Issue #98, opened 2014-12-12
Splitting of netCDF4 files


jhprinz:
Peter @bolhuis suggested that it might be useful to have a splitting feature for the file storage. I thought this should be easy and kind of failed to do quickly implement it. So now I have a draft version that allows us to call a "split" function that then creates a new file and continues storing in the second file, which is itself a valid netCDF4 Storage files. Internally it looks exactly as a normal Storage so that no code outside of the Storage is touched. 

To do so, I needed to subclass the `netcdf.Variable` and especially `netcdf.Dimension` to controle the splitting of a dimension. This means if a file is split the class knows how many items are in each file and if you request no 500 then it find the correct file adjusts the position and reads the object. 

This all works so far, although I need more testing. I think this is especially useful to create milestones in simulations and have files, that are completely separated. And can already be moved and/or analyzed.
Also all of this is independend of the previous implementations. There is just a new subclass of storage that is different. So, we can completely ignore the splitting of we want to.

There are a few questions I have though:

Lets assume that each file can store 100 configurations and if I request number 150 then it will load no50 from file no. 2. Simple. The problem is that all objects can reference each other, which means, that in file 2 might be a trajectory that points to configuration 10, which is in file 1 and even if a trajectory in file two points to object 150, this object is found in the same file at position 50 not 150. This means, that although the file looks intact like a storage, without the other files it is partially meaningless. And I would like to avoid this.

I think, it would be better to have parts that are completely intact (if this makes sense). This will require two additional features:

1. Store some objects multiple times. This is easy since we can easily check, if a file is already in a specific storage file present. If not, add it. This might give some overhead but makes sure that each separate file is complete. In the above example of a trajectory in file no. 2 that points to frame 50 in file 1 we would just store this frame again in file 2

2. Store numbering. Right now each object has a number that is just given by the row it is in and we would have to switch to a numbering that allows each object to store its ID. So we need an additional integer table for each storable object. That could be done automatically, but it would require to change the current file system although to use the arbitrary numbering. We actually have a similar mechanism already, but it uses strings to name objects of certain types and we might add just using numbers to it.
One drawback is, that it will make access a little slower. Maybe not much, but we will always have to search the list of IDs to find the stored position of an object. This has also the nice effect that all frames independent of the stored file have a unique ID. Right now if we store, say, every second frame also in a seperate file they will have different number and we cannot relate them later. We might even think about adding a UUID for the project so that we can later reference all generated object uniquely!

[...]

PR #99, opened 2014-12-12
[WIP] Support for multi-file storage

jhprinz:
This addresses #98 and the current stage is that we can split to multiple files, but it is still highly experimental and I will try to add the things discussed in #98 .

This is definitely not ready to be merged!


jhprinz:
in PR #114 might be a much simpler solution than this.

jhprinz:
With what we have now I will call this closed.

PR #101, opened 2014-12-16
Initial use of SampleSet in PathMovers

### LINKED ISSUE #90: GlobalStateMover (and related)
I’d like to start fixing up the GlobalStateMover and related classes. I picture something that (based on our current naming, which I consider open to changes) looks like this:

* `PathMover` : generates individual `Sample` objects; these are the "atoms" of an overall move: shooting move (on a single ensemble), replica exchange move (on a single pair), etc.
* `GlobalStateMover` : this allows us to combine several of the "atom" moves at once. For example, we may want to do shooting moves on all ensembles at once. This groups the `PathMover` moves together, and statistical sampling is done after each of these.
* `GeneralPathMover` : this may actually be a factory function, but the point is that it should take nested lists of path movers and create an appropriate GlobalStateMover object to handle all moves in the simulation. It's basically a convenience function to create the kind of `GlobalStateMover` we often use in TIS.

A more detailed explanation (based on how these will be used) follows; please let me know what you think of this plan soon, since I’ll start implementing it as soon as I can! Also, as far as I’m concerned, all names are open for changing.

I should also clarify that, at this point, we’ve implemented a few things using `PathMover`s, but we haven’t set up `GlobalStateMover`s (beyond a skeleton) or `GeneralPathMover`s.

### `PathMover`

```python
shooter = ShootMover(UniformSelector(), ensemble)
repexer = ReplicaExchangeMover(ensemble1, ensemble2)
mixedmover = MixedMover([shooter, repexer])
```

`shooter` does is the shooting move for a given ensemble. `repexer` is the replica exchange mover swapping the two given ensembles. The `MixedMover` has a 50% change of doing replica exchange, otherwise it does the shooting.  The probability can be changed by including another parameter, `probability` when setting up the MixedMover.

None of these increment the step on the Monte Carlo sampling: that is, they contain no information about *when* you should take samples from your total distribution for statistics purposes. This way, you can combine `PathMover`s in a way what the subsets of the ordering doesn’t satisfy detailed balance, but the whole sequence (defined in `GlobalStateMover`) still does.

There will also be factory convenience methods for common groups of replica exchange movers:

```python
PathMoverFactory.NearestNeighborReplicaExchange(ensemble_list)
PathMoverFactory.AllReplicaExchange(ensemble_list)
```

`NearestNeighbor` is frequently used (sometimes with a few extra exchanges included); I've only ever used `AllReplica` when testing out some new ideas.

### `GlobalStateMover`

    all_shooting = GlobalStateMover([shooter1, shooter2, shooter3])

`all_shooting.move()` will simultaneously do shooting moves of all 3 types, and will increment the Monte Carlo step, saying that you should sample after each of these moves.

### `GeneralPathMover`

The `GeneralPathMover` is a way of lumping all of this together to simplify setup. You general one huge list of movers: if all the objects in the list are of the `PathMover` type, then it makes a `GlobalStateMover`. Otherwise, it makes all of the objects in the list into `GlobalStateMover`s (you can make a `GlobalStateMover` with a list of one `PathMover`).

    all_moves = GeneralPathMover([[shooter1, shooter2], [repex1], [repex2])

Each list of `PathMover`s is a `GlobalMover`. Each list of `GlobalMover`s is a `GlobalMixedMover`. `GlobalMixedMover`s can be nested. Here we again have to deal with probabilities. My suggestion is that the default be that each `PathMover` have equal probability, and to take the option `probability='equal'` to make all `GlobalStateMover`s within each `GlobalMixedMover` have an equal probability.

### Overall usage

Putting all of this together, the idea is that you'd be able to eventually set up a TIS calculation for a given `ensemble_set` with something like this:

```python
# assume we've already set up ensemble_set, a list of ensembles
from opentis.pathmovers import PathMoverFactory as mf
shooters = mf.OneWayShootingSet(UniformSelector(), ensemble_set)
repexers = mf.NearestNeighborReplicaExchange(ensemble_set, as_global=True)
reversal = mf.PathReversalSet(ensemble_set)
minus = MinusMover(ensemble_set[0], stateA)
general = mf.GeneralPathMover(movers=[shooters, repexers, reversal, minus])

tis = PathSampling(storage=storage, engine=engine, movers=movers)
```

To get the best efficiency, you'd probably have to also include a `probability` argument for the `GeneralPathMover`. A few of the exact calls here might change, but the idea is that it shouldn't really get any more complicated than that.

### LINKED ISSUE #102: Renaming MoveDetails attributes
I think the current names of the `MoveDetails` attributes are a bit confusing. Before we get too far into making more `PathMover`s, I'd like to change some of those names. In particular, I think `final` vs `result` is a little unclear, and `success` vs `accepted` vs `acceptance` is unclear.

Here are my suggestions, in the format `oldname => newname` : description

Things to change:

* `final => trial` : a `Trajectory` representing the trial trajectory (what will go to the output sample if the move is accepted)
* `acceptance => acceptance_probability` : the probability of accepting this step (i.e., Metropolis acceptance probability). Note that for many moves, this is 1 or 0.
* `accepted => trial_is_in_ensemble` : this seems to only be the test of whether the trial path in a shooting move is in the correct ensemble. Unless I'm misunderstanding, it should be moved out of general move details, and only exist for shooting movers.
* `success => accepted` : whether the trial is accepted by the Monte Carlo method

Things to keep the same:

* `inputs => inputs` : list of the `Sample`s selected to be used in the move
* `result => result` : the `Sample` that gets returned by the move
* `mover => mover` : the `PathMover` which generated these details

Note that right now, the things I say are `Sample`s are actually `Trajectory` objects. For now, I'm leaving them as `Trajectory`, but I think we might want to change that to `Sample` eventually. (Should `result` be a `Sample` which is recursively referential, or should it be the `Trajectory` of that sample? Gets weird when thinking about `EnsembleHopMover`s, where the `Trajectory` doesn't change.) 

Any other thoughts on these names? Suggestions for better names to use?

dwhswenson:
Provides new versions of `PathMover`s which use `SampleSet` objects as input, and new version of the bootstrapping calculation which also does the same. My suggestion is to merge this in as soon the tasks below are complete (which should include all aspects in the test suite); other moves/calculation types can be fixed up in future PRs.

This is waiting on PR #97. Work in that is required for this, so we should finish that PR first.

Task List:
- [x] Finish #97
- [x] Modify ShootingMovers to use `SampleSet`
- [x] Modify bootstrapping calculation to use `SampleSet`
- [x] Include storage of the `SampleSet` as part of the calculation
- [x] Rename MoveDetails attributes (closes #102).
- [x] Update ipynbs for new versions of calculations

Extra stuff I included:
* New `EnsembleHopMove` (also to be used in SRTIS)
* New `BootstrapPromotionMove`, which shows how to string several moves together into a more complicated move
* Some tests for `PathMover`s (in progress)

When complete, this closes #90.


jhprinz:
I will check this. All the storing needs also tests. Maybe I can find a general way to test all storage at the same time.

jhprinz:
Found the error... Actually not really an error. I forgot to add the simplest storage to the list of stores so add 
```
self.sampleset = SampleSetStorage(store).register()
```
to storage/netcdf_storage.py at line 60 or so. Should be clear, there are more lines like this.

After that is gets stored correctly. However the ncdump seems to have a bug where not all variable that use variable length are shown correctly. If you look for specific variables they are listed correctly though.
```
ncdump -v sampleset_sample_idx trajectory.nc
```
or
```
ncdump -v trajectory_snapshot_idx trajectory.nc
```

I did not try to fiddle with your branch because this seems faster.

dwhswenson:
Ready for review and merging!

jhprinz:
Looks good to me. I found some notes from myself that seem to be legacy of old stuff. I have to make a sweep of this anyway, so I will merge this now! Good work!

PR #105, opened 2014-12-21
Implement Logging

### LINKED ISSUE #103: Logging
I’m starting to set up logging. This my first time playing with Python’s logging tools, which are quite nice, and I’m doing it now for a few reasons:

* I suspect that as we develop `PathMover`s and `Calculation`s, we’ll be doing a lot of debug-by-print. This is, of course, what logging is really intended for, and including logging statements makes it easier if we have to check the same part of code more than once.
* @bolhuis suggested (a while ago) that when default parameters are used, we should announce that somewhere. This gives a way for users to discover new options (if they won’t RTFM). We can create a logger to do that.
* We can replace our current informative print statements (about how the shooting is going, etc) with logging statements, allowing a user to decide whether to see them.

Anyway, the first of those is the main reason I’d like to do this now. The others are useful bonuses we get by properly setting up logging.

Here’s my suggestion for the logger layout:

1. Each module has its own logger (this seems to be a standard practice)
2. We have a global logger called `opentis.initialization` which logs all initialization parameters (handling the second bullet point above) and logs them at level "info"
3. Configuration of the logger setup is in a `logging.conf` (or `ops_logging.conf`) file. Default configuration is `initialization` logs info messages to a file `initialization.log`, and the module-level initializers log info messages to stdout and ignore debug-level messages. (At least, `Calculation` and `PathMover` info messages should go to stdout, maybe not everything else: all of this is easily configured in the `logging.conf` file)
4. The user will have to explicitly load the `logging.conf` file -- although it puts more work on the user (2 lines of code), everything I’ve read suggests that this is the best practice when writing a framework: the default behavior without the configuration file will be that all log messages are dropped.

Thoughts or suggestions? I've got proof of concept pretty much working on this, it's just a matter of adding messages and finding the right default behavior now.

dwhswenson:
Adding logging to our code. This PR provides the basic setup for our logging system, and will also include usage in at least `calculation.py` and `pathmovers.py`, since we’re about to dive heavily into developing those parts. We can come back to add logging elsewhere later.

When complete, this closes #103.

Task list:
- [x] "Module-level" logger info statements print to stdout
- [x] "Initializations" logger info statements
- [x] Move initializations logger file output
- [x] Without `logging.conf` loaded, no problems (no errors/warnings)
- [x] Finish #101
- [x] Logging for Bootstrapping (initializations and info)
- [x] Logging for ShootingMover stuff (initializations and info)


jhprinz:
Is this ready for merging? I could take care of this?

dwhswenson:
This mainly just needs the fix to #101, which I think you may have given in the comment you posted on that. However, I may not have a chance to settle the last couple parts of #101 until at least Monday (family stuff today and tomorrow).

I think the existing conflicts with master in this branch (preventing automatic merging) will be resolved once #101 is merged in.

dwhswenson:
Logging is ready for review and merge!

jhprinz:
Great. Then I can add my storage logger, too.

Like the `initialization_logging` function. That is very useful. Merging now!

Issue #79, opened 2014-11-22
storage creation not done in engines and make engines storable


jhprinz:
To allow to create several engine objects I would like to move the storage creation out of the engines and let this be handled by the user. So the storage has all the information about the system, n_atoms, topology, an initial configuration which is what an engine might need

```
storage = Storage('file.nc', mode='w/a', topology=..., n_atoms=..., configuration=...)
engine = OpenMM(storage=storage, opts) # creates new engine

# save the options in an engine used to generate it
storage.engines.save(engine)

# create an engine using the stored options and information in the store
old_engine = storage.engines.load(0, OpenMM) 
```

This would save exactly what we save right now, just in a more separate way.

Next, we would still have to pass the engine to the PathMover class. Either by setting the global using
```
PathMover.simulator = engine
```
or per mover
```
mover = ForwardMover(..., engine=engine)
```

Suggestions?

jhprinz:
Is this done? I think I have implemented this already, but there is still the issue that PathMovers require an engine to be set globally. Maybe that is another issue and not present here. Will close this.

dwhswenson:
Regarding the idea that PathMovers have a globally-set engine -- isn't the engine really only relevant for the shooting movers? (And movers that include shooting movers as sub moves). Maybe it makes sense to make it an argument in `__init__` for those.

jhprinz:
Good point. A shooter should know the used engine. Let's just do that. I will make an issue.

PR #110, opened 2015-01-13
Ipythontesting

jhprinz:
Added a python module to run an ipython notebook as a test and integrate it into travis


jhprinz:
The last addition should also close issue #79 which was almost finished.

jhprinz:
Also a little glich in visualization is fixed that could occur when one wants to plot zero samples.

jhprinz:
This is ready for review and merge (if the travis passes). BUT! This also replaces Alanine Example.py by the ipython notebook. Depends, if we want this.

jhprinz:
Last thing, before merging. I want to store the dynamics_engine only as an object and not separate in the storage. This makes is cleaner, but might require a little more testing. But other than that is should work.

dwhswenson:
This is great! I think it is much better to use the ipynb as the test script than to duplicate efforts in the ipynb and the .py. (However, let's keep the .py around until current branches are merged back in. I think I may have made changes in the .py which I have yet to make in the ipynb, so I'd like to be able to quickly diff it.)

Two questions:

1. I noticed that there's a timeout in the most recent Travis test (https://travis-ci.org/choderalab/opentis/builds/46941135, see line 1276). Should the timeout limit be increased? Should we mark a timeout as a failed test? I'm guessing this specific example is the find-first-TIS-path part. However, I'm not sure that the rest counts as a good test if that part fails due to timing out. Besides, with a timeout passing, an infinite loop would look like a passing test, which obviously we don't want.

2. Can you add some of the other ipynbs to the test suite? The `toy_dynamics_tis.ipynb` example should definitely be included. Probably wouldn't hurt to include the others, as well. 

Adding other ipynbs to the Travis tests lets us guarantee that our ipynb-based documentation stays up-to-date (at least runnable). That's really useful!

jchodera:
> I noticed that there's a timeout in the most recent Travis test (https://travis-ci.org/choderalab/opentis/builds/46941135, see line 1276). Should the timeout limit be increased? 

The travis-ci timeout is 10 minutes (600 seconds), so it's best to not increase the timeout over the current 540 seconds unless you switch to a method that prints intermediate output more frequently than once every 10 minutes.

jhprinz:
Yes, the time out issue.The problem is that the example runs a bootstrapping and in some rare cases this might take longer than the 9 minutes. The last run was unfortunately one of these. I would rather simplify the example because this also means that evaluating the cell for a potential user can take more than 10 minutes which is no fun. Or as David said: It could also be the initial run or the bootstrapping which can take so long. Especially since the travis build is done on a CPU. I am not sure if we should increase it over 9 minutes since this will break the whole testing and travis will stop or we output intermediate prints. This is easy since the cell execution is on a seperate thread and non-blocking. It would still defeat the purpose of illustrative examples if cells can run so long.

I also wanted to add the timeout time as an command line argument to make this adaptable to different examples. Third and a really nice option is. That I am checking the first line of a cell, if it starts with `#!` (which is rather arbitrary) then the string behind it is considered a testing command and we could include 1. `timeout:10000` 
2. `strict`: a key word forces the output to match the one in the notebook.
3. `skip` : a key word meaning to not execute the cell
4. `ignore` : a key word meaning to ignore if a cell has an error

Of course other syntax would be acceptable. Similar to decorators in nose tests to skip or...

@dwhswenson I totally agree, let's keep the alanine example around for know. This is often a quick help to check if something is working. I will add the other notebooks too one by one.

jhprinz:
I rewrote the python script and it is much easier to use now. Also two more notebooks are working.

The tool now supports some command-line options as the keywords in the first line.
1. `--no-timeout` set timeout to fail the test
2. `--timeout int` sets the default timeout time in seconds
3. `--strict` activates diff testing, so if the output differs the test fails

dwhswenson:
Those options will be useful additions. But maybe a different name for `--no-timeout`: I would normally assume that name means `--timeout infinity`. Perhaps something like `--fail-if-timeout` ... that specifically doesn't sound great to me, but the idea of using "fail" or "pass" in the option makes it more explicit.

To avoid hitting the timeout, I see three options:

1. Replace most of the AD-based tests with the toy model tests. The `toy_dynamics_tis` notebook should cover most of the same path sampling stuff I put into the alanine dipeptide example; it just has more text explanation and uses a simpler engine. It also doesn't use storage yet; my plan was to only have storage for it with the actual TIS sampling. So maybe a small test to show dynamics and storage with alanine dipeptide (using OpenMM) instead of the whole calculation for the Travis tests?

2. Consider the use of `travis_wait` to push the time limit out past 10 minutes (http://docs.travis-ci.com/user/build-timeouts/)

3. Push some logging output to Travis's stdout (PR #103 includes optional logging to report completion of every N frames; I've been using N=10 with the alanine dipeptide example). Every path move also has logging output.

I think a combination of 1 and 2 would probably be best. We don't want to do any really significant sampling using AD as a test, because it will slow our development cycle if we have to wait hours for the tests to run. However, it would be good to run some basic tests with it.

jhprinz:
Yes, the option will be changed. This makes total sense, I actually thought about this problem when I wrote it, because I was thinking about a way to disable the timeout. I will change that.

I also agree that the testing of all the sampling could be done using the toy_dynamics. If it is meant for testing then the Alanine with all its additional complexity distracts from the real thing to test. I would like to keep most of the Alanine as a testing for the storage, since it creates almost everything that can be stored. Maybe first do integration testing using the toy_dynamics and if this works continue with storage testing using Alanine.

I can also add the travis_wait as an option to some cell so that in case we have something that really might take a long time, we mark it and the notebook tester then tells travis to wait

Number 3 seems not necessary if we do 1 and 2. We could also design the example in smaller steps to be below the 10min threshold. And again agreed we want this to test that the framework works and not run a whole simulation.

jhprinz:
I also realized that the Visualization Example fails for some still unknown reason. My local installation runs fine. Strange the variable `tree` is missing. It get's set one or two cells before. I will check that

jhprinz:
Apparently this has to do with the way the samples are loaded from the TreeBuilder. This can be easily fixed once #114 is done. In principle this case be merged and I will fix the visualization later.

dwhswenson:
Would it work to merge #114 first, so that this one can pass tests before we merge it in?

jhprinz:
I wanted at least this to be separate from the mammut PR for storage. I will remove the Visualization notebook for now so that the test passes. Then we should be able to merge this in. I mean the fail is due to a problem in visualization.py not in the changes files in this PR.

jhprinz:
I will repair the auto merge!

PR #106, opened 2014-12-30
TIS Equilibration & PathMover tools

dwhswenson:
Tools so we can equilibrate within the TIS ensemble, and well as a few general tools for path movers. Also gives us most of the tools for RETIS.

- [x] `PathSampling` : calculation object for path sampling
- [x] `PathReversalMover` (corrections and updates) : PathMover for reversals
- [x] `MixedMover` (adjustments to current version) : PathMover to select a random mover
- [x] `SequentialMover` : PathMover to run several other moves in a row (without worrying about whether previous parts were accepted). Used for, e.g., making one single move that shoots from all replicas.
- [x] `PartialAcceptanceSequentialMover` : PathMover to run several other moves in a row, until one is rejected. All accepted moves are applied.
- [x] `ConditionalSequentialMover` : PathMover to run several other moves in a row. Only if all are accepted, are any accepted.
- [x] Add support for `mover_path` variable (except storage)
- [x] Include TIS equilibration in toy model ipynb
- [x] Finish #105

(Details on various PathMovers listed in issue #108.)

PR #114, opened 2015-01-16
Storage Update and Storage for Tree-like structures

jhprinz:
### todo
- [x] move functions to suitable packages. Especially the JSONifier and the decoraters 
- [x] decouple storage from rest of the code
- [x] minimize necessity of decorators (decorators in objects are now only for JSON not storage!)
- [x] fixed chunksizes to reduce filesize overhead
- [x] fix circular imports (use import opentis as paths)
- [x] rename decorators
- [x] add unit testing for storage
- [x] check documentation
- [x] renamed `ops` to `paths`
- [x] allow easy creation of an empty copy to continue simulation but in a new file. That is a somehow manual but very easy solution to the multi-file problem. Simply once in a while say copy the main parameters (ensembles, movers, enginee, topology, ... ) to a new file, add the last sampleset (which out their heritage) and keep on running with the new file
- [x] merge #106 into this PR

### move to / implement in other PR
- add support for arbitrary metadata such as other topologies (e.g. in only solute files a topology of the solvated system) or the same in files without any coordinates. 
- turn mdtraj.Topology into a storable object for better metadata support

The updates in storage are finalizing. I have made some more simplifications and fixes, and added features.

### changes
- almost all object can now easily be stored. There are two main ways:
    1. An object can be saved only by the contents of its dict and the class name which results in a stub object that has then the same attributes and class name, but no functions. This is useful, if initialization requires additional information that might not be saved and all we need is contained in the dict. These object behave like the original except they have no functions
    2. An object can completely be saved and reconstructed. For this all parameters for the objects init is stored. Since we store only immutable objects this works and will thus created the exact object we had before. Useful e.g. for engines that need initialization
- the object saving is done using two functions that have a standard mechanism depending in the above type, but can be overridden. These two functions are very similar to the construction and deconstruction in `pickle`. As an alternative we could have added more parser to JSON, but I kind of found it easier this way to access the storage directly. These two functions are:
    1. `to_dict` : a generalization of `__dict__` and returns a dict that contains all important attributes. Things necessary for storage and private variables are automatically excluded. Also a private class member `_excluded_attr` contains attributes not to be used for storage
    2. `from_dict` : the counterpart to create the object from a dict. Depending on the storage mode this uses the `__init__` or just calls `setattr`
- the dicts are converted to JSON and stored so we have 
    ```
    Object() <=> dict() <=> JSON
    ```
    and the the object to dict mechanism now supports

    1. native types (int, float, str...) or course. The were built in
    2. slices
    3. dict, list, tuple (only partially, they get converted to lists!)
    4. simtk.unit support
    5. numpy array to base64 (not optimal but might be useful, if we want to store complex metadata)
    6. all objects that have `to_dict`, `from_dict` functions
    7. objects that can be loaded / saved using a Store (there are object that are stored without the to_dict mechanism, which is faster, e.g. atom coordinates)
    8. mdtraj.Topology including non-standard elements!
- I tested compression, but this slows down so much that it becomes the bottleneck before openmm. So, compression does not seem to be useful. If we want to use this for archiving we can still add this very easily. Is just a flag. 
- Chunksizes have been added which remove the storage overhead. Now filesizes are almost optimal. E.g. in the Alanine Example instead of 200MB for an initialized file <1MB and also read/write access seems to have improved. 
- Most objects use the above object storage via JSON. Since this is so flexible also Volume, Ensemble, DynamicsEngine, etc use it. 
- The only special type stored are OrderParameters because these are somehow mutable although only internally there are reserved spots for not yet computed OPs. 
- The storage has native simtk.unit support. Each float array in the netCDF has automatically a JSON readable description of the used unit.
- Snapshots, Configurations and Momenta can be copied while reducing to a subset of coordinates
- There is now an example that show how to reduce the file size by copying all data from one store to another and removing certain atomic coordinates. This can be used to remove all water. Since our order parameters are still based on configurations and not snapshots we cannot completely remove coordinates (momenta is possible) I will also turn this into a command-line tool that can do this with a few options. It might also be useful to also have script to export in some other formats as well. This is of course already possible using the conversion to mdtraj trajectories and saving it. This calls for an example notebook
- some minor bugfixes 
    1. for some reason the `resSeq` in `mdtraj.Topology` is not `int` and caused some problems.
    2. lazy loading of kinetic_energies didn't work
    3. ...

[moved TODOs to a later post]

Comments are welcome.


jhprinz:
### naming of decorators
Right now I am using 4 decorators to mark objects to be storable or be able to turn into dicts in certain ways, but I whink the names are not to good right now, maybe one you has suggestions
- `storable` : decorator for the base class of a certain object that adds a few attributes to a class to make it compatible with the storing mechanism. I store e.g. the reference to the index in a storage once saved and it knows the base_type useful for all derived classes so that we can load by just specifying the base type in the storage. This is the oldest one
- `creatable` : means that this object get automatically `to_dict` and `from_dict` functions that allow to reconstruct the complete object using the `__init__()` function. It also registers the class in a list so that the JSON/dict reconstructor know where to find the class functions if it want to create an object by just a string name. This could be replaced by just searching global namespace for the class, but this seems to be more secure: Only registered class objects can be created. So here we have
    ```
    obj == class.from_dict(obj.to_dict())
    ```
- `dictable` : almost as `creatable` but with slightly different functions. `to_dict` almost returns the `__dict__` and the `from_dict` creates a stub object with the same internal `__dict__` and also changes the class name. So here
    ```
    obj != class.from_dict(obj.to_dict())
    obj.__dict__ == class.from_dict(obj.to_dict())
    ```
    So, it is like the original object except no functions
- `nestable` : this _keyword_ marks, that if an instance of this class appears somwhere as an attribute of some obj of the same base type (e.g. A `InXEnsemble` is pointed to by a `SequentialEnsemble` and they are both of base type Ensemble) then this class will not be stored as a seperate object, but also turned into a dict. Okay, this is confusing. The above example would give
    1. withOUT nestable : 
        ```
        { '_cls' : 'SequentialEnsemble', 'ensembles' : [ 
            { '_stored' : 'Ensemble', '_idx' : 3 },
            { '_stored' : 'Ensemble', '_idx' : 2 }
        ]}
        ``` 
        So, here the ensembles inside the SequentialEnsemble will be stored as Ensemble objects, since an Ensemble can be saved. This is the default behavior. But in this case it might be nicer if we get this
    2. with nestable : 
        ```
        { '_cls' : 'SequentialEnsemble', 'ensembles' : [ 
            { '_cls' : 'InXEnsemble', '_dict' : { 'volume' : { 'stored' : 'Volume', '_idx' : 2 } },
            { '_cls' : 'InXEnsemble', '_dict' : { 'volume' : { 'stored' : 'Volume', '_idx' : 3 } },
        ]}
        ``` 
        Here, we have the nested object as a single JSON block. For Ensembles this makes sense, since we only want a few complex objects. In the case of complex PathMovers we might still use the first one. Because only the first one allows to also reference objects inside of the nested tree. In the first example I could also reference each part of the SequentialEnsemble by `{ '_stored' : 'Ensemble' , '_idx' : ##}` from other object. This is what we want for MovePaths that @dwhswenson proposed.

All right. Much text. I am happy with the names `storable` and `nestable` but `creatable` and `dictable` are confusing. What about `dictable_as_object` and `dictable_as_stub` or so? They should be somehow similar

jhprinz:
Just want to mention that I might port this to use custom JSONEncoders, but for now this does the job. The concept to convert `Object <-> dict <-> JSON` seems to be more common than I first thought. Especially in DB storage. Also this makes all the stuff accessible via the web, etc... But that is not stuff for 1.0

jhprinz:
I have cleaned up some of the package mess with storage which lead to the problem of circular imports. Apparently the better way to do it is to avoid `from package import xyz` which - called several times - will cause an error because two import commands try to set a new command in the global namespace and this is a potential risk. I have changed in most places to `import opentis as ops` and this cleans up the imports in the beginning and removes the battery of import this, import that...

Hope this is okay because it affects practically all packages although to final changes are little.

dwhswenson:
Were we going to use the short form `paths`; i.e., `import opentis as paths`? 

On the `dictable`/`creatable` names... I don't have a suggestion that I consider perfect, but these labels will be the entirety of the line they're on, right? So why not make a longer and very descriptive name? `restores_as_full_object` and `restores_as_stub_object`? (I focus on what happens when restored because I figure that's the user's primary concern.)

Also, this looks like a *lot* of work for one PR. Would it make sense to split it into a couple different PRs so that we can merge it in faster and use parts of the whole?

jhprinz:
I like paths, that looks nice in the code.

Concerning the dict / create I think I will use something descriptive like this. 

About splitting, yes. I started adding stuff and suddenly everything was connected to everything. Especially the import problems with circular imports that I needed to fix. Is there a simple way to split this? Maybe I have to do most of it manually and start by copying once I have merged in your PR.
It has actually gotton worse with changes, but it resulted in less files and much less code.

Anyway, I have cleaned the storage again a lot and I am happy now. No more problems. The main advantage now is that storage is completely non visible in the rest of the code. So if we decide to use another storage idea, we do not need to touch the rest of the code.

jhprinz:
Merged PR #101 and should work now. Luckily only a few minor conflicts.

jhprinz:
### todo
- [x] move functions to suitable packages. Especially the JSONifier and the decoraters 
- [x] decouple storage from rest of the code
- [x] minimize necessity of decorators (decorators in objects are now only for JSON not storage!)
- [x] fixed chunksizes to reduce filesize overhead
- [x] fix circular imports (use import opentis as paths)
- [ ] rename decorators
- [ ] add unit testing for storage
- [ ] check documentation
- [x] allow easy creation of an empty copy to continue simulation but in a new file. That is a somehow manual but very easy solution to the multi-file problem. Simply once in a while say copy the main parameters (ensembles, movers, enginee, topology, ... ) to a new file, add the last sampleset (which out their heritage) and keep on running with the new file

### move to / implement in other PR
- [ ] add support for arbitrary metadata such as other topologies (e.g. in only solute files a topology of the solvated system) or the same in files without any coordinates. 
- [ ] turn mdtraj.Topology into a storable object for better metadata support (optional)

dwhswenson:
On splitting: I suggest getting a version here that works, but doesn't have all the promised features implemented. Then branch independent features off from that (e.g., tests are pretty much completely independent, right?) and we can merge this with partially implemented features (but overall working code).

There are ways with git to really restructure it as if they'd been separate branches all along (assuming your commits were atomic enough), but that's a waste of your time when this approach is much easier.

You can also do what I did with #101, #105, and #106: they needed to be merged in that order, but by putting them on different PRs I reduced number of changes you have to see with each request.

jhprinz:
Almost done. Documentation and lots of tests added and removed much obsolete code.

Last thing is that right now the Visualization requires MovePaths, since it cannot determine from the BootstrapPromotionMove, what actually happened. Thus the pictures look incorrect. Once we have #106 in there we should be able to fix this easily.

Hope this will build nontheless...

jhprinz:
Also lots of internal renamings to make the code clearer to read

jhprinz:
Another good idea: I added the possibility to load using get item. Since the Storage is a subclass of netCDF we can now get the usual variables from the netCDF files using:
```
storage.variables['name'][idx]
# for example
indices = storage.variables['trajectory_idx'][3]
```
and access complex objects using
```
storage.store_name[idx]
# for example
trajectory = storage.trajectory[3]
```
we could also do something like
```
trajectory = storage.objects['trajectory'][3]
```
which is completely analogous to the `variables` attribute. Maybe in another life.

jhprinz:
Ready for merging. Unfortunately there is still a bug which can lead to the visualization to fail when no samples are provided (this can happen in some cases). This was fixed somewhere but seems to have been deleted by some merging :)

The build for jhprinz/treestorage has passed...

PR #115, opened 2015-01-22
Sphinx Documentation

jhprinz:
I took the liberty to copy more stuff from the mdtraj documentation, stripped out most of what we do not need and began filling in the parts from us. 

I also tweaked the ipython support which they (mdtraj) worked out pretty well. It automatically creates an HTML version of the notebooks and includes all of these into the documentation. Looks great. 

I edited the CSS a little to make it look nicer. Is not perfect, but almost :)


jhprinz:
Yes. The idea is that I removed the ipython specific CSS and let sphinx take over to have matching styles, highlighting, etc... And then I added enough CSS to fill the gaps.

jhprinz:
I needed to rename the examples not to use blanks, since I have no idea (which drove me nuts) on how to put blanks into sphinx directives. All that is needed is
```
::notebook [filename]
```
but there seems to be no way to escape a space in the file name. Anyhow, now they are called
`alanine.ipynb` and `visualization.ipynb`

jhprinz:
I would say merge this and we go from there.

jhprinz:
Have a look at the ipython examples and give some comments on the new style

dwhswenson:
Questions:

1. For the main of our docs, do we want to use `.md` (converted to rst using pandoc) or `.rst` directly? Mainly, I know I'm more comfortable with md, but I see that rst has some additional features. (I can always write first drafts in md and use pandoc to convert to rst before making a PR.) Stuff that is heavy on inline math is a lot nicer in a flavored Markdown (e.g., `We set $\alpha=1$ and $\beta=0.5$.` vs. ``We set :math:`\alpha=1` and :math:`\beta=0.5`.``).

2. Is there a way to look at the HTML version of the docs via the PR, much as there's the nbviewer bookmarklet for IPython notebooks? I don't think we should bother putting compiled docs in the repo, but if there's a nice trick that allows us to see the compiled docs, I'd use it!

I'll go ahead and merge here (and then build the docs locally to see how the CSS looks!), but please let me know your thoughts on those points.

dwhswenson:
@jhprinz, can you check whether all the required files were added to the repo? In particular, the `docs/_themes/sphinx_rtd_theme/` subdirectory seems to be missing `theme.conf`, and as a result I get an error when building. Current directory structure at:  https://github.com/choderalab/opentis/tree/2909683a52153173fb6d65a17073dc96cff49eb7/docs/_themes (that commit number is the result of the merge I did)

By changing the theme to `kr`, which does have a `theme.conf` file, I can build the docs. Of course, I can't see the CSS which was customized for the `sphinx_rtd_theme` when I'm using another theme!

Also, if we're not going to use some of the other themes, maybe we can remove them from imports? I had to install `alabaster` and `sphinx_bootstrap_theme` just to build the docs, and it looks like we're not using either of them.

jhprinz:
Yes, this was a bug. Alabaster is not needed as is the sphinx theme, which I attached in a separate folder

jhprinz:
Is addressed by #122.

Issue #116, opened 2015-01-22
Add Storage for Toy Dynamics


jhprinz:
will add storage capabilities for Toy Dynamics and create a Topology object for this as well

jhprinz:
Here a little outline of the changes necessary to make ToyDynamics and OpenMM exchangeable. The goal would be to be able to write
```
topology = ToyTopology(options)
engine = ToyEngine(options, topology)
storage= Storage(filename, topology)
[rest of code]
```
or
```
topology = OpenMMTopology(options)
engine = OpenMMEngine(options, topology)
storage= Storage(filename, topology)
[rest of code]
```
likewise and the rest of the code works the same except the use of Orderparameters! They require knowledge about the structure of Snapshots nothing else is different. And thus development of PathSamplingAlgorithms is detached from the actual problem!

### Proposal for the Topology object
It contains all that is necessary to compute a potential energy surface as well as additional information. Concrete for our two dynamics:

1. Toy Topology
    - Number of Atoms 
    - Spacial dimensions (1 is default)
    - Masses 
    - potential energy

2. OpenMM Topology
    - Number of Atoms, 
    - Spatial dimensions 3, 
    - used OpemMM.Topology(), 
    - (optional) full original OpenMM.Topology()  (in case the storage is cloned with say only solute)
    - ForceField, so that the PES is theoretically available 
    - (optional) additional atomic subsets like: {'water' : range(23,2000), 'solvent' : ..., 'solute' : ..., ions} although mdtraj.Topology can do that, too

The `Storage` and `Engine` needs a Topology to be created

### The EngineObject
The Engine then has only its integration logic and a reference to and (maybe always) an `Integrator`
1. ToyIntegrator
2. OpenMMIntegrator

That should make it flexible yet consistent between engines and Topologies

jhprinz:
This is done in #125.

PR #120, opened 2015-01-23
Storable ToyDynamics

jhprinz:
This adds storage support for ToyDynamics. It can turn
PES, Integrator and Engines into JSON objects and store and restore them. Might come in handy.

I have not yet added Topology support, but I think we do not need it.

Will add some tests and maybe an example to see if and how it works.

This addresses #116 

- [x] adapt ToyEngine for storage
- [x] adapt ToyIntegrator for storage
- [x] adapt PES for storage
- [x] make Topology a separate storable object
- [x] fix OpenMMTopology
- [x] fix ToyTopology
- [x] debug
- [x] correct tests
- [x] update examples (add example for storage)

### Update later
- documentation


jhprinz:
And requires #114 to be merged!

dwhswenson:
@jhprinz, what is the status on this? Is it yet possible to make a `Storage` that will work with the `ToyEngine`?

jhprinz:
Almost ready. It is definitely possible to make this work. Was pretty busy the last two days with robot stuff. Sorry

dwhswenson:
No worries -- I know that you have a lot of loose ends to tie up right now. This is just the one part that I will need soon in order to move forward: #124 is currently waiting on it, although I can move that requirement to later. For now I'm going to finish #137. Once that one is done, though, I want to work on analysis stuff. And to do that efficiently, I need to be able to store things generated by toy dynamics.

Let me clarify: I'm not worried about whether toy dynamics can itself be stored; I'm having the problem that the `template` variable doesn't mean anything in toy dynamics, and therefore I can't even instantiate a `Storage`. We need to figure out how to handle that so I can use toy dynamics to quickly generate data for the analysis routines that I can port over.

jhprinz:
I am getting close with all finishing up :) This will be fixed tomorrow :)

You are right with the template. My idea was to use consequently a Template / Topology also for Toy_Dynamics so that this completly exchangeable with OpenMM run code. Just use another engine

jhprinz:
Here a little outline of the changes necessary to make ToyDynamics and OpenMM exchangeable. The goal would be to be able to write
```
topology = ToyTopology(options)
engine = ToyEngine(options, topology)
storage= Storage(filename, topology)
[rest of code]
```
or
```
topology = OpenMMTopology(options)
engine = OpenMMEngine(options, topology)
storage= Storage(filename, topology)
[rest of code]
```
likewise and the rest of the code works the same except the use of Orderparameters! They require knowledge about the structure of Snapshots nothing else is different. And thus development of PathSamplingAlgorithms is detached from the actual problem!

### Proposal for the Topology object
It contains all that is necessary to compute a potential energy surface as well as additional information. Concrete for our two dynamics:

1. Toy Topology
    - Number of Atoms 
    - Spacial dimensions (1 is default)
    - Masses 
    - potential energy

2. OpenMM Topology
    - Number of Atoms, 
    - Spatial dimensions 3, 
    - used OpemMM.Topology(), 
    - (optional) full original OpenMM.Topology()  (in case the storage is cloned with say only solute)
    - ForceField, so that the PES is theoretically available 
    - (optional) additional atomic subsets like: {'water' : range(23,2000), 'solvent' : ..., 'solute' : ..., ions} although mdtraj.Topology can do that, too

The `Storage` and `Engine` needs a Topology to be created

### The EngineObject
The Engine then has only its integration logic and a reference to and (maybe always) an `Integrator`
1. ToyIntegrator
2. OpenMMIntegrator

That should make it flexible yet consistent between engines and Topologies

jhprinz:
One question: The topology contains all necessary to create the Engine and Storage, but it does not contain a set of coordinates.
So, do we
1. want to only use the topology to create Engine or Storage, or
2. a Snapshot (since they point to a topology) to have amore complete object. i used to call this snapshot a template although it really is only a Snapshot.
3. add coordinates/velocities/box_vectors to the Topology independent of Snapshots
4. Add an initial Snapshot (without a topology) add this to the topology and then add the topology to the snapshot so that they reference each other ?!? (might not be so safe and confusing)

I prefer approach 2 since it forces the user to think about an initial (hopefully feasible) initial structure. It also means the Topology is really topology, describing the structure which is independent of the actual coordinates.

dwhswenson:
Question about the code snippets: are those the same `options` given to `Topology` and `Engine`? Is the idea that `Topology` just ignores the ones it doesn't need?

The idea I keep coming back to in all of this is that there might be non-chemical rare events worth studying with OPS. Forcing the chemistry-centric perspective will make it harder to do that. To me, a `Snapshot` is mainly a point in the space of the dynamical variables of the system. The `Engine`, and everything around it, sets up the differential equations that generate snapshots.

I think it's good to have some chemically-motivated shortcuts (like the way we split the phase space into coordinates and momenta), but I think we should make it so the code can run even without that. The `ToyDynamics` should force some of that. For example, this 2D system should be able to store 2 different masses: the mass is just a parameter of the differential equation; the system doesn't represent any real particle, so `m_x != m_y` is allowed.

But at this point, the main thing is to have something that works, and to clearly document what needs to be done.

jhprinz:
Yes, of course. Sorry to be unclear.

options are totally Engine Dependent and Topology is just an almost whatever object. It is very specific to what you do and is stored with its content. So you could easily add `m_x != m_y`. So the Topology contains all that is necessary to describe the systems PES and whatever might be useful AND number of atoms and spatial dimension, which sole purpose is to create a storage file with the right sized tables. This is almost ready.

jhprinz:
It is all ready and working now. Need to check the notebooks and we are ready to merge this.

This PR needed to change the underlying storage of the Toy_dynamics. Since the storage supports now different no of dimensions I remove the casting of number to make it storable.

jhprinz:
The nose tests all passed in the last travis build and the ipynb is stuck in the bootstrapping call.

Here are the important changes:
- Every storage and engine needs a template which is a Snapshot
- This snapshot is used to determine the units of the coordinates, velocities and energy
- The snapshot also need a Topology, which can be found in topology.py
- For ToyDynamics this is `ToyTopology(n_atoms, n_spatial, masses)`
- Still to not change too much the masses are not totally correct they are mass_x, mass_y independent of the atom. I want also later to add the PES to the topology and not the Engine
- Example to set up the storage would be 
```
topology=paths.ToyTopology(
    n_atoms = 1,
    n_spatial = 2,
    masses =[1.0, 1.0]
)
template = paths.Snapshot(
    coordinates=np.array([-0.5, 0.0]), 
    velocities=np.array([0.0,0.0]),
    potential_energy = 0.0, # Important to set energies dimensionless otherwise kcal/mol is default
    kinetic_energy = 0.0,
    topology=topology
)
pes = (OuterWalls([1.0, 1.0], [0.0, 0.0]) + 
       Gaussian(2.0, [1.0, 4.0], [0.0,1.0]) + 
       Gaussian(2.0, [25.0, 1.0], [0.0, -1.0])
      )
integ = LangevinBAOABIntegrator(dt=0.02, temperature=0.1, gamma=1.0)
sim = ToyEngine(
                options={'pes' : pes, # might go to the Topology
                         'integ' : integ,
                         'n_frames_max' : 5000,
                         'nsteps_per_frame' : 10},
                template=template
               )
```

Be aware that you cannot use mdtraj with ToyDynamics. This requires an mdtraj topology and I have not done that yet. 

Also differentiate between `OP_MD_Function()`and `OP_Function()`. The second is the general, while the first converts the trajectory to an mdtraj.Trajectory and passes this one.

Ready for review. If you want, I can fix the masses and PES before merging.

jhprinz:
I will fix the toy_dynamics notebook after merge. There is a copy that contains most of the new code already

dwhswenson:
This is not mergeable. Not only has `toy_dynamics_tis` lost all the changes of the major restructure from #124, but it doesn't even run basic dynamics. You crash on running a trajectory, not on bootstrapping. From the error, it has to do with the change in spatial dimensionality.

jhprinz:
when I tried to add the latest master it screwed this up. I now copied the file from #124 into by branch, but the notebook is VERY big and takes pretty long lo load. Actually several minutes. I took the liberty to remove all the output from the moves and save it again. In this I will also make it compatible with storage
as I said before

jhprinz:
Sorry for the merge problems. This is working now and I added the toy_dynamics to the notebook testing.

jhprinz:
Also the forcefield is now part of the topology.

jhprinz:
Storage is working now as do all examples. I copy the toy_dynamics Notebook and made a new one out of it that stores the production run at the end and used storageview.py to have a look. There is really lots of metadata besides the few data points... But it works now. 

So, please have another look

dwhswenson:
This works for me!

Question though: for a couple of files, I don't understand the purpose they serve:
* `examples/alanine/toy_example.py` : I thought we were just keeping the IPython notebooks as examples, especially while development is a bit fluid; should we generate this from the ipynb once we have the ipynb in final form?
* `examples/ipython/toy_dynamics_tis-no_vis.ipynb` : Just having glanced at it, this looks a bit like a broken version of the `toy_dynamics_tis` notebook. Do we need to keep it around?

Anyway, neither of those cause any problems.

I assume the plan is that I merge this into #137 and then I'll use the information in `toy_storage` to add support in `toy_dynamics_tis`, right?

jhprinz:
The notebook was indeed just a copy to not mess up the toy dynamics notebook. The other was supposed to be copy of the alanine example but wit ToyDynamics. But I used the notebook then. Both can be deleted safely and contain nothing that is not present somewhere else.

PR #122, opened 2015-01-25
[WIP] Additional Documentation and Utils

jhprinz:
Added a sphinx extension that allows to import a markdown file at any place in the documentation.
Simply use
```
.. markdown: [markdown_file.md]
```
and the file will be converted using pandoc intop rst and then added to the documentation


dwhswenson:
Oooh.. that's a nice way to get the best of both worlds! One question: since pandoc will usually recognize files by their extension, could this be trivially generalized to `.. pandoc:` instead of just `.. markdown:`? That way you could say you'd add any pandoc-compatible file. 

(If I'm guessing correctly on how this works, that really just means change the name of the directive: you could probably already call `.. markdown: [somefile.mediawiki]` and even though the file is not Markdown format, it would work.)

jhprinz:
Hi... Why are you awake? Isn't it like in the middle of the night?
But...

Yes, I just used the nice template from mdtraj for including ipython notebooks nicely. This uses pandoc and just calls it to convert a file and kind of then assumes that this .rst code is independent of the code and include it. You cannot really connect to things in the code yet. I would need to figure out on how to do the linking correctly although I assume it is trivial once I know how to do it.

Concerning your question: The name is right now independent is always assumes that it is markdown, but it can easily extended to use all types of documents that are supported by pandoc, which apparently is a lot. I think there is also a way to use equations. It says it supports mediawiki markdown, etc...

What about changing the syntax to the way you suggested
```
.. pandoc: markdown xyz.md
.. pandoc: mediawiki 
```
That would make it usable very general

dwhswenson:
(Yeah, I'm just about to go to sleep. It's only 3am....)

I looked at your code; I think that if you just remove the "from" part, pandoc will normally guess based on the file extension (possibly based on some file inspection). Having broader pandoc support isn't a big deal, since I don't personally have a desire to work in anything other than md or rst (when I write something for a wiki, I use pandoc to convert *to* mediawiki). However, since the tool is already pretty powerful, it might be nice to use all its power. But I don't think it's worth making the syntax any more complicated. Perhaps you can abstract out the `from` flag for the popen and make it a class property, and use subclasses of a `PandocDirective` to trivially generate specific directives like `Markdown` or `Mediawiki`? Then each of those would be less than 10 lines of code.

(ps: looks like the most recent commit [fec4787] has the problem I was having when I tried to build the docs! See https://github.com/choderalab/opentis/pull/115#issuecomment-71321712)

dwhswenson:
Now that this builds the docs, do you want to merge it at this point (after re-including the standard test suite)? I'd like to have the ability to build the docs in my own stuff as soon as possible, and it looks like this PR includes that.

If so, feel free to merge as soon as tests are re-included and pass, or I'll merge it in the morning if you consider this PR complete. (BTW, `runipy` exists in a version 0.1.1; the `after_success` script calls for 0.0.4 -- must this be such an old version?)

jhprinz:
- ipyrun, yes this was still what is in the mdtraj scripts. I think we can install the latest (hopefully)
- okay, then I will add an optional argument `:from: [type]` for the pandoc extension and name it `pandoc::`, also I add an option to add `[type]::` as a shortcut for all types listed in the options in `conf.py` that makes it rather flexible.

Yes, it builds so far. That means we can now add an amazon s3 and move it up there...

Cool... Then have a good night :)

PR #123, opened 2015-01-26
[WIP] Fix Path_Mover Saving and Visualization

jhprinz:
This repairs the storage of path_moving and everything that is related to it.

- [x] repair delayed loading for Configuration, Momentum and Sample
- [x] change attributes for OptionalEnsemble and SingleFrameEnsemble for easier storing
- [x] repair storageview.py
- [ ] restructuring PathMover creation of move_path
- [ ] move move_path out of details and be a separate property of a `Sample`

Optional
- [ ] turn legal_sample_set into a decorator or some kind of filter on the input SampleSet

All this stuff is storage related and the main problem is that MoveDetails should be created immutable as all the other objects. To do this we have to revert the order in which the path is created. I would also think that MoveDetails be present only for *leave* moves the other are moves are for picking a move and not making it.

So `MoveDetails()` are strictly information about the generation of a sample. While the `MovePath()` (which is only a list of DecisionMoves and a final ConstructionMove) explain how it came to the selection of the move that constructed the sample.

Okay, lets call the move tree leaves `Constructions` and the branches `Decisions` just for talking about it.

My main question is, if we will loose any information when no details about the Decisions are stored. We will also loose information about Sequential generated samples, although samples will be stored in the correct order. 

Another question is, if we store in-between samples or just the final `SampleSet`? Consider a SequentialMove of 10 ForwardShooters for a single Replica. After the full move we have one new SampleSet, but in-between 10 Samples that we could throw through away, but since Movedetails store the input trajectory these will point sequentially back to all trajectories in the samples. 

Somehow I am not satisfied. It looks too complicated and I think it should be possible to make this simpler. But that could just be me. I will make some changes to play around and maybe we can come up with some simplifications.


dwhswenson:
`MoveDetails` *have* to be mutable after the leaf builds them. See `ConditionalSequentialMover`. Samples which were accepted must be turned to rejected if later moves are rejected.

Example: simplified version of minus move.

1. Try a replica exchange.
2. If that's accepted, try a shooting move.
3. If the shooting move is rejected, then you have to go back and reject the replica exchange as well. The WHOLE move gets rejected.

One approach is to only allow the samples to be saved from the `Calculation` level. At that point, the samples don't change. Within composite movers, they do.

jhprinz:
You are right... Hmmm, that is a problem...

jhprinz:
So, I think (which is not a too bad sign) I have a way to get out of my dilemma. So I try to explain how I would run the pathmovers and then please look carefully if I haven't neglected or forgotten anything. Especially if we are still able to do everything (moves) we want and can save all information we want. And, of course, if you think it makes sense at all or is too complicated to explain, too far away from the way this is usually done, etc... Probably this is only to convince me that we have to leave it the way it is :) Alright, this might get a little confusing and long, so please bear with me a little.

## Objects we need and what they contain
### `Sample()`
- pointer to trajectory
- pointer to used leave mover (for convenience)
- move_details : the information how the sample was generated using the *leave* mover
- move_path : list of  moves run that lead to this leave
- replica_id
- ensemble
- maybe information if the sample was an accepted one

### `MoveDetails()`
- arbitrary information about the particular move. This is totally Move dependent and does not necessarily share and attributes. Mainly just a `dict()`

### `SampleSet()` 
the objects that are moved by pathmovers 

What we basically want is moves in SampleSet space. There is one SampleSet (the generalization of a set of replicas) and these are moved to a new SampleSet. (I hope this is correct and the most general way to describe what we want to do). 

And this moves have a few steps
1. From the initial `SampleSet` determine which samples are relevant for this particular move (I would call this the *filterstep*. Potentially also some random picking of the resulting samples
2. Use some generation (swapping, shooting, etc) to generate new samples including acceptance
3. Replace some samples in the initial `SampleSet` to return this as the new one

This actually is very similar in nature to the shooting moves. 
1. Get an initial set of frames
2. Generate new frames
3. Replace some of the old frames with new ones.

Now, important is that a for later analysis we ONLY look at the samples that are in the `SampleSet` because only these are part of the real generated *ReplicaSpaceSamples*. So we can have intermediate SampleSets with intermediate Samples that are generated but are not used for later final analysis. I would prefer to only store the SampleSets are are real wanted and not all potential intermediates, although we could do that. We only might need a mechanism to mark these are *real*

## Immutable Objects

If we want to have immutable savable objects while they are created we need the intermediates, otherwise there is no way to do this correctly. What we need to set of a complete Move in the SampleSetSpace we need ways to implement a little programming language with things like `if` or `block`. The interesting for this is that the only two things which a move can depend on is
1. the original `SampleSet`
2. the bool of being accepted or not.

We can even classify moves by their requirement:
1. `SampleSet` : Shooters, ReplicaExchange, etc.
2. `accepted` : ConditionalMovers, PartialSequential
3. `None` : SequentialMovers

So let us assume that each move returns these two things, then we could write a (simplified) MinusMove (and I am sorry for not really knowing too much about MinusMoves) like:
```
mover = Conditional(Conditional(ReplicaExchangeMove, ShootingMove, FalseMove)):
```
1. This applies the Conditional mover which returns the trial if the inner Move has been accepted and the original otherwise
2. Applies Conditional again, but this time will return the inner result either applied to a Shooting Move if true or return the original, but with a False as accepted. Thus the FalseMove Mover.

Instead of Conditional with FalseMove we could just write a new ConditionalFalse Mover
This mimics the PartialConditional

In contrast, if we want to mimic the ConditionalSequential we would write

```
mover = Conditional(ReplicaExchangeMove, ShootingMove):
```

The inner part would attempt a shooting move only on an accepted ReplicaExMove. If the Shooting fails its return an ACCEPTED ReplicaMove and an accepted Shooting otherwise.

Let's do this more general

a PartialAcceptanceMover looks like

```
partial_mover = Conditional(move1, Conditional(move2, Conditional(move3, move4)))
```

and a ConditionalSequenceMover looks like

```
conditional_mover = Conditional(move1, Conditional(move2, move3, FalseMove), FalseMove)
```
or simpler rename this to
```
conditional_mover = ConditionalSeq(move1, ConditionalSeq(move2, move3))
```

ConditonalSequential and PartialAcceptance could then be factories to build these nested lists.

Maybe be might use different names for the two types.

## How to handle application of samples

Since each mover returns the already updated SampleSet, at every point in the nested chain we have the correct object at hand and need not to worry about this.

### Handling of MovePath
This is a little difficult and not so elegant. Each Mover need to be passed the previous move_path as an option to be able to add oneself to all coming path movers. We might be able to do this logic using wrappers or include in the Class Logic so that a use doesn't see this. It results in some kind of recursive calls.
The other option to handle this is to add the current Path_move to the SampleSet, this is very easy and actually very consistent. I would prefer this second way. We can also store the move_path that lead to the final SampleSet (not only to the last samples) this might actually be what we want in the first place: All steps that lead to the next SampleSet.
Theoretically the move path contains the last accepted (as well as all previous) boolean so we do not need the last accepted to store separately. This might still be convenient. It also allows to create moves that work on the complete past of a move which is more than we can do now. (Is this used somewhere)

## other high-level movers
### SequentialMover
This could be written as
```
seq_mover = Sequential(move1, move2, ...)
```
or using a list
```
seq_mover = Sequential([move1, move2])
```

### Parallel Moves (future)
This is already a step into the future and not yet necessar and makes sense for Sequential move (as you mentioned in a previous post)
```
parallel = ParallelSequential([move1, move2])
```
The only difficulty here is to determine the resulting SampleSet from all the moves. This would require some analysis between the original and all the resulting SampleSets. But if it could really be run in parallel all changes must be independent and so that should be possible. 

### Moving SampleSet creating to Mover
It would require to do this within a mover to be consistent, but this should not be a problem.

## Naming of SampleSet
I like the idea of `Bucket` or `SampleBucket`, since it is such unordered bunch of samples, like a bucket full of samples where some get replaced added removed, etc...

## Final words
#### Pros
- So, if this works I think we can implement this with almost no interruption in the calculation class. This would only be internal changes in pathmover.py. 
- This will also keep your idea of the high-level movers intact, which I think is great.
- SampleSet contains also the move_path
- Storage is easier
#### Cons
- add move_path (and accepted) to SampleSet. Costs more storage
- It might be complicated to think about it and explain how to use it?

Okay, that was long. Any suggestions?

dwhswenson:
I still need to think more about what you've written there (and will write a longer post later), but why can't it be this easy:

The reason things need to be immutable is that they can't be changed after they've been stored. So why not just enforce a rule that they can't change after they're stored?

Doesn't this basically amount to overriding `__setattr__` to give an error if the object has already been stored, and behave normally otherwise? (Maybe with some other work for lists, although if lists are cast to tuples on storage that makes them immutable.)

It seems to me that raising an error if the code tries to change a saved object would solve all the problems without having to make lots of other major changes. (Although I know overriding `__setattr__` is a bit  of a minefield.)

dwhswenson:
Before I write a longer response, let me start out with a few questions, because I don’t fully understand your proposal.

First, I like the idea of switching `ConditionalSequential` and `PartialAcceptanceSequential` to being based on a `ConditionalMover`, but it isn’t clear from what you wrote how that works. (See second clarification question below.) Also, my main question is whether that should be a refactoring for version 2.0. I think we’ve hit the point where we need to get a working version out the door ASAP, even if it means leaving a few hacks in there that should be cleaned up later.

Questions for clarification:

### Type signature of `PathMover.move()`

Even though Python doesn’t require type signatures, I think it is extremely important that `PathMover.move()` have a single, well-defined type signature. In the current implementation, this is:
* takes a `SampleSet` as its only argument
* returns an (ordered) list of `Sample`s

Do you mean to change this? In particular, are you now saying that you want `PathMover.move()` to return a `SampleSet` instead of an ordered list of `Sample`s? I think `SampleSet` *does* retain the order that the `Sample`s were added, but only implicitly (the idea is that the user-facing parts of `SampleSet` are more like a dictionary than like a list, since that’s usually what we want.)

Retaining the order is essential to later updating a global state correctly. Retaining all the samples (with associated `mover_path`) is essential to do analysis of move acceptance (e.g., relative acceptance of forward and backward shooting paths).

I don’t think we should make the type signature more complicated unless we really must.

### Clarifications on `ConditionalSequential`

You propose a `ConditionalMover`, which, if I understood correctly, has the following:
```python
class ConditionalMover(PathMover):
    def __init__(self, condition_mover, if_true_mover, if_false_mover=None):
         #stuff
```
(well, obviously you fill in the `#stuff`.) If the `condition_mover` is accepted, you call `if_true_mover`; if not, you call `if_false_mover` (if it exists).

To use this to mimic a `ConditionalSequentialMover`, you propose something you call `FalseMove`. How does that move work?

If I understand the scheme you have proposed, it’s just like the frames generating when calling a function (in fact, it *is* doing exactly that). So let’s consider the example of the conditional sequence:
```python
mover = Conditional(
    condition_mover=move1,
    if_true_mover=Conditional(
        condition_mover=move2,
        if_true_mover=move3,
        if_false_mover=FalseMove_B
    ),
    if_false_mover=FalseMove_A
)
````

I’ve distinguished between `FalseMove_A` and `FalseMove_B` for discussion; in principle it could be the same object. Other than that, this is exactly what you wrote, just formatted differently.

Let’s say that `move1` is accepted, but `move2` is not. But that means that somehow `FalseMove_B`, which is called with an argument `SampleSet` for which the results of `move1` have been applied, has to somehow know to undo `move1`. I don’t see how it has that information in scope.

Is it more correctly something like:
```python
mover = Conditional(
    condition_mover=Conditional(
        condition_mover=Conditional(
            condition_mover=Conditional(
                conditional_mover=move1,
                if_true_mover=None
            ),
            if_true_mover=move2
        ),
        if_true_mover=move3
    ),
    if_true_mover=None
)
```
You still need to be careful to ensure that, if `move1` and `move2` are accepted but `move3` is not, the previous moves are undone. The idea of returning a list of `Sample`s instead of the active global state in my version made it trivial to identify which `Sample`s come from submoves. I think figuring that out in yours would be messy, but not impossible.

### Including `mover_path` in `SampleSet`

What does this mean for `SampleSet` in the "global state" context? As we have implemented it now, those are the only `SampleSet`s that are saved, and the reason for that is it gives us the correct statistics. However, there is *not* a unique leaf per global state (see `SequentialMover` or any of its subclasses). So the `mover_path` variable is nonsense in the one case where we save these. There is a unique leaf (and thus a defined `mover_path`) per **sample**, not per `SampleSet`.

***

I think answers to those would explain enough of your proposal that I can list a few other pros and cons. Anyway, I hope you're staying comfortable during the snowpocalypse! Looks like the worst of it went east of you, at least.

jhprinz:
Yes, it has been very quiet here and I liked it a lot. I have never seen NYC to quiet. No cars after 11pm on the streets. Just like 10cm of snow and sunshine today. Kind of nice day actually

Okay, I totally see you point and have been working on this yesterday and I think I managed to do get it running and surprisingly it simplified the code - not much - but a little. 

The main reason I came up with this is because I read a good article about functional programming and have a friend that explained a few things about it. And it seemed to me that this would be a good place to use it to avoid some problems with backtracking and fixing objects that have not been created immutable. This can be difficult to keep working correctly. It can also makes testing and debugging more difficult. And the main reason I think is beacuse I think its cool and I wanted to see, if I can get it to work.

But lets first answer your questions:

### Type signature of PathMover.move()

Correct. I propose to use is strictly `SampleSet` -> `SampleSet` so that it can be chained and combined, like we did with `Ensemble` or `Volume`, just that we have combinations that are not based on sets, but on functions. While for Ensemble we combine the results in a correct way to create a new function `Ensemble` -> `bool` here we just chain different possible functions depending on results.

To make this work we need to add everything that is necessary to make these decisions to the SampleSet, so a sample set contains if this SampleSet has been produced by an accepted move or by a rejected one. We also need to keep track of the previous move_path so that we add this to the SampleSet. 

Finally we need to think only locally about what a move does, what it returns and how the move wants to change the move_path that is passes on to inner moves and what it returns to outer ones.

I think that this would work a lot like recursive/nested functions and that python takes then (automatically) care of all the backpassing, etc...

It would also mean that we do not have to retain the order since the new samples are applied automatically to get a new SampleSet or the new samples get discarded by a rejection.
Since we now apply all samples in the given order, we could also do this while we are walking the tree and if there is a chance that we reject like in a ConditionalMover we would keep a copy, but only the ConditionalMover needs to know about this.

### ConditionalSequential

Let me maybe answer this with two example on how I implemented the Movers now

```
class RandomChoiceMover(PathMover):
    def __init__(self, movers, ensembles=None, replicas='all', weights = None):
        ... # as usual

    def move(self, sample_set):
        # figure out idx
        mover = self.movers[idx]

        new_set = sample_set
        # Add self to mover_path if that is desired
        new_set = SampleSet(new_set, accepted=True, move_path=sample_set.move_path + [self])

        # Run the chosen mover
        new_set = mover.move(new_set)

        return new_set
```

the main point here is that you just apply the mover that you have picked. Everything else does the mover itself:
- adding oneself to the move_path and 
- return the new sample_set where the changes have already been applied to

Now for the ConditionalSequentialMover
```
class ConditionalSequentialMover(SequentialMover):
    def move(self, globalstate):
        subglobal = SampleSet(
            self.legal_sample_set(globalstate),
            accepted=True,
            move_path=globalstate.move_path + [self]
        )
        false_state = SampleSet(
            self.legal_sample_set(globalstate),
            accepted=False,
            move_path=globalstate.move_path + [self]
        )

        for mover in self.movers:
            subglobal = mover.move(subglobal)
            if not subglobal.accepted:
                return false_state

        return subglobal
```

it really as simple as that. Just apply movers in sequence as long as the return `accepted==True` and this is what we want. For the PartialSequential this changes only
`return false_state` to `return subglobal`

### Including `mover_path` in `SampleSet`
Yes, here you are right and this question solves a little a question I had myself. Let me first ask what the correct move_path for a Sample would be and then try to get back to the question what move_path makes sense for ´SampleSet´.

First question is, if it contains the result of the movers, too or just which ones have been called. I prefer the second one, but I guess the first is also enough. 

Assume the following: Three movers in a ConditionalSequencetialMover 1,2,3 and the 3 one fails, what is the move path: It could be
1. `[]` nothing since the new sample is the old one.
2. `[ContidionalSequentialMover]` since this one has been done but not been accepted
3. `[ConditionalMover([move1, move2, move3])]` all moves that have been done (look more like a tree then
4. `{ConditionalMover({move1: True, move2 : True, move3 : False})] : False}` like a nested tree

Now to SampleSets: Right now I assume (which I think we are not doing yet) is that each mover just changes the SampleSet and all we can do is go back to an earlier stage and discard what has been done already (like in a rejection, no using the trial). If this allows to do all the moves we want then the move_path for a SamplePath is just a concatenation of all the little moves involved in getting to the final SampleSet which we then call GlobalState and this one is saved for statistics. So in the writing of 4. If we use two of these movers to create the final global state then the move_path for the SampleSet is

```
{ SequentialMover({
    ConditionalMover({move1: True, move2 : True, move3 : False}) : False,
    ConditionalMover({move1: True, move2 : False}) : False
}) : True }
``` 
where I assumed an OrderedDict() and the last False is irrelevant, because there is no meaning to True or False. 

Now one last strange thing: I said before that each Mover changes the SampleSet and at the end we will have the same set as if we would have added all samples at the end. Okay, but how do then move_paths look like for any but the first sample. In theory the first sample will have a simple path, the next sample could in principle depend on the last move_path. And the next sample will even be longer. This is similar in thinking that if you consider 3 global states then the 3rd on depends on both what happened 1->2 and 2->3. So we would have
```
sample1.move_path = { SequentialMover({
    ConditionalMover({move1: True, move2 : True, move3 : False}) : False
}) : True }
sample2.move_path = { SequentialMover({
    ConditionalMover({move1: True, move2 : True, move3 : False}) : False,
    ConditionalMover({move1: True, move2 : False}) : False
}) : True }
``` 
which makes sense under all of the assumptions. And finally let's get back to the question what move_path makes sense. I think for a sample we want the actual shortest path that leads to a node, but SequentialMoves introduce lots of nodes before that potentially are not relevant for that particular sample. So we would like to have something that allows to simplify this. Example:
```
mover = Sequential([ForwardShooter(ensemble1)] * 100)
```
Each subsequent sample will have a longer move_path because it depends on an even longer path of moves from the initial SampleSet, that was given. 

### Solution to the move_path problem
I think I just (while I am writing this) got the solution. So (oh man, this posts are getting too long :) )...

If we actually call it a move_path then it is in fact the path that goes from an initial sample_set to another one and just contains the set of all possible ways to move from initial to final the actual chosen one. While the actual used Mover contains all possibilities with their probabilities, the move_path contains the route taken. Then, as you correctly say "A move_path for a SampleSet does not make sense", unless we specify the initial SampleSet. More generally between any SampleSet that are generated using Movers we can write down the exact path of movers that was taken.

We can also write down the path from a SampleSet to a Sample exactly the same. Although it is very redundant. Alternatively we can add the samples to the move_path
```
{ SequentialMover({
    ConditionalMover({move1: [sample1], move2 : [sample2], move3 : [sample3])] : [sample3],
    ConditionalMover({move1: [sample4], move2 : [sample4]}) : []
}) : [sample3] }
```
This tree contains all generated samples and at what point which are still relevant.

Okay. Stop here. I am getting lost in new ideas...

Last comment. All of this bears some similarity to versioning systems with commits that change the code :)

dwhswenson:
Again, just writing a few comments rather than a really long thing.

My biggest concern here is about time and timing. At this point, I think we need to be getting a functioning product out the door, not rewriting code that already basically works. I’ve been working on `PathMover`s in the current version since mid-December, with more stuff coming in the next week. Keep in mind that you’ll have to redo a lot of that work, and, in particular, you’ll have to redo large sections of the `PathMover` tests. I just don’t see benefits that make this worth delaying version 1.0.

So I think the question is whether we can find an way to get the key features in less time. **What is wrong with just saying that saving samples only happens at the calculation level?** It seems to me that this solves all the problems with essentially no work.

It does require that a different input vs. return values for `move()`. You seem to be under the impression that doing that breaks something. You say:
 
> To make this work we need to add everything that is necessary to make these decisions to the SampleSet, so a sample set contains if this SampleSet has been produced by an accepted move or by a rejected one

I do not understand that sentence. We already have working `Sequential`, `PartialAcceptance` and `ConditionalSequential` movers. What can we not do?

Also, I don't think you're correct when you say:

> Each subsequent sample will have a longer move_path because it depends on an even longer path of moves from the initial SampleSet, that was given.

That is not `mover_path` that I implemented. The path is ALWAYS the height of the tree: one entry for each decision point. The only downside to this is that you need to have different instances of any move repeated in a `SequentialMover`. I have a fix for this based on the tree you suggested above. It wouldn’t take nearly as much work as changing the return type to `SampleSet`, but I still think it is more important to get something that works at all rather than rewriting code for cases that we won’t see between now and version 2.0.

With your approach:

1. Do you save all the intermediate `SampleSet`s? 

  * If so, then you're saving approximately one `SampleSet` for every `Sample`. That starts to waste storage (compared to one `SampleSet` per MC move in the existing approach). It's also harder to track the changes between two `SampleSet`s: a list of `Sample`s is a list of diffs; the `SampleSet`s are the two files. Finally, it means that hunting for the `SampleSet`s that actually contribute to the averages takes more effort: in my approach, every saved `SampleSet` consists of `Sample`s for the average.
  * If not, then how do I get rejection rates for submoves? That's the major advantage of this approach.

2. I think your tree structure still requires that you only save at the calculation level. Otherwise, you don't know when the tree is complete. So what do you gain over just saving samples whether they are returned from the root mover, as I do now?

***

(A couple comments on a more conversational side):

> The main reason I came up with this is because I read a good article about functional programming and have a friend that explained a few things about it.

Ha! Yesterday one of the thoughts I had was "this looks entirely too much like LISP: powerful and elegant, but much harder for a beginner to wrap their head around it." So I see how that was the case! But using recursion through stack frames is actually already what I'm doing. Stricter functional programming is useful when you don't care about the call stack, but in this case, the call stack is *exactly* what we care about.

> Last comment. All of this bears some similarity to versioning systems with commits that change the code :)

Yup. Again, I had a similar thought yesterday. But my analogy is that `Sample`s are like commits, and `SampleSet`s are like PRs. A `Sample` logs a change that occurred, but `SampleSet` is saved so that you have a set of related changes that tell some sort of story (the title of each PR being "Monte Carlo move 1", "Monte Carlo move 2", etc.)

jhprinz:
Okay, I thought a lot about this. (Somehow I could't let it go) and I agree mostly with your arguments, so I kind of rethought my idea and it is much better now and already got it to work and fixed the examples. The goal was mostly to keep the complete move path and allow storage, which is possible now and quite elegant I think. Okay, the solution was quite subtle. I left the SampleSet and instead of outputting Samples from a Move it now outputs a `MovePath()`. This is almost a list of samples, but can be nested and extended by other things than samples. Then A SequentialMove does not output all the samples from it contained movers, but it stores a Sequence of `MovePath()`. So I copy exactly in structure the part of the tree, which was traversed, including all Samples and whatever happened.

### Example

```
mover = Sequential([
    moveA1,
    Progressive([moveB1, moveB2]),
    Exclusive([moveC1, moveC2]),
    moveA2
])
path_a_to_b = mover.move(sampleset_a)
sampleset_b = sampleset_a + path_a_to_b
print path_a_to_b
```
which outputs e.g.
```
SequentialMove : True : 28 samples
 +- SampleMove : ForwardShootMover : True : 1 samples [<openpathsampling.sample.Sample object at 0x10e784710>]
 +- PartialMove : True : 2 samples
 |   +- SampleMove : ForwardShootMover : True : 1 samples [<openpathsampling.sample.Sample object at 0x10e622ed0>]
 |   +- SampleMove : ForwardShootMover : True : 1 samples [<openpathsampling.sample.Sample object at 0x10d5efcd0>]
 +- ExclusiveMove : True : 6 samples
 |   +- PartialMove : True : 3 samples
 |   |   +- SampleMove : ForwardShootMover : True : 1 samples [<openpathsampling.sample.Sample object at 0x10d5dce10>]
 |   |   +- SampleMove : ForwardShootMover : True : 1 samples [<openpathsampling.sample.Sample object at 0x10d5c4e90>]
 |   |   +- SampleMove : ForwardShootMover : True : 1 samples [<openpathsampling.sample.Sample object at 0x10d5f8c90>]
 |   +- PartialMove : True : 3 samples
 |   |   +- SampleMove : ForwardShootMover : True : 1 samples [<openpathsampling.sample.Sample object at 0x10d5daad0>]
 |   |   +- SampleMove : ForwardShootMover : True : 1 samples [<openpathsampling.sample.Sample object at 0x10e747f10>]
 |   |   +- SampleMove : ForwardShootMover : True : 1 samples [<openpathsampling.sample.Sample object at 0x10dc23f90>]
 +- SampleMove : ForwardShootMover : True : 1 samples [<openpathsampling.sample.Sample object at 0x10dbf5e90>]
 +- SampleMove : ForwardShootMover : True : 1 samples [<openpathsampling.sample.Sample object at 0x10dbfe490>]
 +- PartialMove : True : 2 samples
 |   +- SampleMove : ForwardShootMover : True : 1 samples [<openpathsampling.sample.Sample object at 0x10dbfea90>]
 |   +- SampleMove : ForwardShootMover : True : 1 samples [<openpathsampling.sample.Sample object at 0x10dbfedd0>]
 +- ExclusiveMove : True : 5 samples
 |   +- PartialMove : True : 3 samples
 |   |   +- SampleMove : ForwardShootMover : True : 1 samples [<openpathsampling.sample.Sample object at 0x10dc06ed0>]
 |   |   +- SampleMove : ForwardShootMover : True : 1 samples [<openpathsampling.sample.Sample object at 0x10db79bd0>]
 |   |   +- SampleMove : ForwardShootMover : True : 1 samples [<openpathsampling.sample.Sample object at 0x10db81f10>]
 |   +- PartialMove : True : 2 samples
 |   |   +- SampleMove : ForwardShootMover : True : 1 samples [<openpathsampling.sample.Sample object at 0x10db92c50>]
 |   |   +- SampleMove : ForwardShootMover : True : 1 samples [<openpathsampling.sample.Sample object at 0x10db9bad0>]
 |   |   +- SampleMove : ForwardShootMover : False :[]
 +- SampleMove : ForwardShootMover : True : 1 samples [<openpathsampling.sample.Sample object at 0x10e726fd0>]
 +- SampleMove : ForwardShootMover : True : 1 samples [<openpathsampling.sample.Sample object at 0x10e4eaf10>]
 +- PartialMove : True : 2 samples
 |   +- SampleMove : ForwardShootMover : True : 1 samples [<openpathsampling.sample.Sample object at 0x10e4ebfd0>]
 |   +- SampleMove : ForwardShootMover : True : 1 samples [<openpathsampling.sample.Sample object at 0x10e4da890>]
 +- ExclusiveMove : True : 6 samples
 |   +- PartialMove : True : 3 samples
 |   |   +- SampleMove : ForwardShootMover : True : 1 samples [<openpathsampling.sample.Sample object at 0x10e4dae50>]
 |   |   +- SampleMove : ForwardShootMover : True : 1 samples [<openpathsampling.sample.Sample object at 0x10e4e1cd0>]
 |   |   +- SampleMove : ForwardShootMover : True : 1 samples [<openpathsampling.sample.Sample object at 0x10e4b6ad0>]
 |   +- PartialMove : True : 3 samples
 |   |   +- SampleMove : ForwardShootMover : True : 1 samples [<openpathsampling.sample.Sample object at 0x10e4b6e10>]
 |   |   +- SampleMove : ForwardShootMover : True : 1 samples [<openpathsampling.sample.Sample object at 0x10e4c0950>]
 |   |   +- SampleMove : ForwardShootMover : True : 1 samples [<openpathsampling.sample.Sample object at 0x10e718d50>]
 +- SampleMove : ForwardShootMover : False :[]
```
to get all samples that are actually used to make the new set by applying them to the old set use
```
path_a_to_b.changes # only the necessary (accepted) samples
path_a_to_b.all_samples # all generated samples on the way
path_a_to_b.samples # all samples that only belong to this Movepath not children
```

I need to leave now. Just wanted to put this up for discussion, since I think it does exactly what we want and you can now also search for specific samples.

Storage works fine  and is shown in the storageview.py as well as for SampleSets and a SampleSet knows the path that lead to its existance.

What do you think? :)

dwhswenson:
After a zeroth glance at this (i.e., a minute reading on a Friday night at 3am right before going to bed), I'd say the following: 

You know the part where I said

>  I have a fix for this based on the tree you suggested above.

?

Well, this is almost exactly what I was thinking of. You returning a tree of the whole move, right? Basically with tuples of `(mover, [samples])`? I'll have to look more, but if you're doing that I was thinking, the sequence of applying the samples is trivial because you just apply each accepted sample in the order it appears in the tree structure (depth-first traversal). This allows you to distinguish between the first and second time a repex is called in the same `SequentialMover`, which my original approach didn't, while saving all the other advantages of the original methods (especially not having to rewrite everything).

It is still the case that you aren't returning the same type of object you input (although I suppose you could redefine the input object so that you did, but I think that would actually make analysis harder). I think that is okay, although I understand the feeling that it would be prettier if you could.

Among the advantages of this approach is that the work to implement it isn't too bad. It is (1) relatively easy to modify current path movers to do this (just need to make wrap the `Samples` currently returned in this tree; mainly that's a task for the 4 "multi-movers" [`RandomChoice`, `Sequential`, `ConditionalSequential`, `PartialAcceptance`]) and it is (2) quite easy to fix up the tests (a flattened list of the `Sample`s in the right order satisfies the tests, and that's an easy function to write). My big concerns were changing these aspects. As you can see, even beyond the changes needed for code already in master, I've been adding a lot of move-related stuff in my two open PRs. I didn't want us to spend a week or more redoing that; I hope that this approach won't take more than an afternoon.

What you wrote looks like actual output: how close is this to a mergeable PR? I think that the best way to do it might be for me to first finish #124 and #137 (I suspect I can finish those both over the weekend, if I delay the issue of storage with toy dynamics to a later PR: I think I cracked the MinusEnsemble tonight, and the rest is just convenience classes, docs, and tests.)

I'll email you to ask about your schedule in the near future. I really think we're no more than a week or two from production-ready code, if the contributions from each of us can line up well enough with our schedules!

jhprinz:
YES, YES and YES... You pretty much nailed it

With the tree (structure) you suggested I meant your idea to use SequentialMovers, etc. So that a Move looks like a tree. The output you saw is actually the output of the `print str(movepath)` code.

Yes, the actual changes were not so much. it was merely adding the MovePath and changing the output.

The main reason to have a different output is that otherwise in the recursing I will need to point to all the previous intermediate SampleSets and these get stored.

The main thing missing is good naming and how we finally want to write the application of a Mover. Some possibilities:

- `mover(sampleset)`: add a `__call__` to a mover to make it create a move path starting from a given `SampleSet()`
- `new_sset = old_sset.move_by(mover)` : Short-cut for writing `new_sset = old_sset + mover(old_sset)`.
- `new_sset = mover(old_sset)` : This is probably the most elegant way to write it. The `__call__` method just replaces a `new_sset = old_sset + mover.move(old_sset)` where we should rename the `.move` because it create a move path and not a 

My proposal would be
```
mover = <some PathMover>
old_sset = <some SampleSet>

new_sset = mover(old_sset)
# properties of new_set
new_sset.move_path # the move_path used
new_sset.origin # the initial sset (here old_sset)
new_sset.step # the step which we could increase in the move
```

Also rename some classes
```
PathMover()  -> Mover()          # the general mover
MoverPath()  -> MovePath()       # the path used by a particular mover
                Sample()         # as before, building blocks of SampleSets
MoveDetail() -> SampleDetail()   # details about the move relevant for a `Sample` might 
SampleSet()  -> ???              # better name? 
```

All of this is practically mergeable. Little cleanup to do, but I fixed all tests already to the new scheme to test the actual accepted samples that are applied and all generated samples in the process to be consistent.

One thing about SampleSet. Would be possible to move the random.choice logic out of the SampleSet or is that required? If possible it would be good to have the SampleSet to behave reproducible when the same functions are called. We might use your `FilterSetMove()` idea that just limits the view on all used samples to a random single choice that the underlying mover will use. This way we can even keep track of which sample (if there are many possible ones) is chosen. 

One more comment about the new MovePath implementation. It allows to make other changes to SampleSets in principle not only based on Samples. Examples could be:
- Remove a `Sample` from `EnsembleXY`
- Change ReplicaIDs...
- Replace an existing Sample with a new one (this is what we are doing all the time!)
I am not using this explicitly since we still assume that all can be expressed using Samples that we `apply` to a SampleSet. But in the MovePath idea it is just a special case of other possible ones.

dwhswenson:
### Function behavior

I would suggest the following:

* `mover.move(old_sset)` : generates the `MovePath` (see below on name?) object
* `movepath.samples()` : returns the list of samples in the order they would be applied

Combining those two means that `mover.move(old_sset).samples()` will give us exactly the output of the current `mover.move(old_sset)`.

Are the properties you're suggesting a replacement for the current `SampleSet`, or additional properties? Currently, `SampleSet` has a record of **ALL** active trajectories, and the assumption is that the `SampleSet` passed to the `PathMover` contains **ONLY** active trajectories. This needs to remain true. As additional properties, those are fine.

A path sampling calculation would then call something like:

```python
move_record = root_mover.move(globalstate)
move_record.save(storage)
globalstate = globalstate.apply_samples(move_record.samples())
globalstate.save(storage)
```

With possible simplification down to

```python
move_record = root_mover.move(globalstate)
globalstate = globalstate.apply_move_record(move_record)
globalstate.save(storage)
```

I think the normal expectation is that any function from a path mover which returns a `SampleSet` will have one sample per replica. You seem to be suggesting versions that will have the old sample as well as new samples. Since we disagree on what is expected of a convenience function, we probably shouldn't do anything either way to make `__call__` a convenience at this point.

### Renaming Classes

I'm not so fond of some of that renaming. I know that namespaces will help this, but I can't help but think that `Mover` is too generic. I assume you wanted to change it because it leads to potential confusion with `MovePath`. I think it would be better to change that. That name was based on the idea that it would be the recursive list of parents in the move tree. What you have is more complete. What about naming that to `MoveRecord` (or `MoveHistory`, `MoveLog`, or something else) to avoid confusion?

So I'd go with:

* `PathMover`
* `MoveRecord` (or other)
* `Sample`
* `SampleDetails` (I like that one, although I'd keep it plural)
* `SampleSet` (I have no better ideas)

jhprinz:
Allright. I think we are practically there:
With 
```
.move() -> MovePath
move.samples() # right now this is called .changes() and .samples() are the samples only in this MovePath step and not all below in the tree
```
I kind of liked the idea to detach the movepath from only applying samples, but this is a good compromise. RIght now it is exactly as you wanted (except for some name changes).

```
.origin
.movepath
```
are just convenience additions to the current SampleSet. It stays exactly the same otherwise
`.origin` is set when a move is apply to know where the movepath started for this particular sample. So one could traceback all steps and remove `.step`
`.movepath` is added when a move is made so that a SampleSet knows how it was created. Like a Sample has SampleDetails. `apply_move_record` is right now called `move_by` but I like the longer version better.

> I think the normal expectation is that any function from a path mover which returns a SampleSet will have one sample per replica. You seem to be suggesting versions that will have the old sample as well as new samples. Since we disagree on what is expected of a convenience function, we probably shouldn't do anything either way to make __call__ a convenience at this point.

Actually, no. I was just not sure if this could be the case? Especially for special moves like AMS. I guess I still know too little about this. My idea was to just use
```
class PathMover():
    def __call__(self, sample_set):
        return sample_set.apply_move_record(self)
```
so that it can be used as
```
new_sset = mover(old_sset)
```

Will finish this soon.

dwhswenson:
> right now this is called .changes() and .samples() are the samples only in this MovePath step and not all below in the tree

Ah, I think I see the point of that. Meaning only leaf moves return non-empty sets, right? I'd call that `.local_samples()`, and have `.samples()` (or maybe some clearer name?) return the subtree.

Regarding the properties, I just wanted to make sure you weren't switching to saving *only* changes (meaning that you need the entire history to figure out which samples are active). Hypothetically, the properties you proposed make that possible; doing that is frustrating in practice. But is sounds like that's not your goal.

Regarding `__call__`, are we then saying the same thing? My thought is that path sampling boils down to this:

```python
for step in range(n_monte_carlo_steps):
    globalstate = root_mover(globalstate)
    # whatever stuff for saving and labelling with the step
```
The point being that if the return type is the same as the input type, then the return result should be a valid input for the next step.


For AMS, you would probably WANT only the new sample. It would be something like:

```python
while (ntransitions < ntransitions_required):
    globalstate = ams_mover(globalstate)
    new_path = globalstate.move_record.samples()[0]
    if (is_transition(new_path.trajectory)):
        ntransitions += 1
```

(`is_transition` checks whether the path crosses into the other state. I think AMS uses `ntransitions_required=1`, but I need to check the math to see if that's necessary. The funny thing is AMS is that we really need to know what the *previous* version of the trajectory was, so your `origin` might prove really useful.)

So let's we have 10 replicas in `old_sset`, and `mover` does a shooting move on each one of them.

```
new_sset = mover(old_sset)
```

Does `len(new_sset)==10` or `len(new_sset)==20`? I vote for 10.

jhprinz:
Just quick. Yes, 10 Samples. But the new set has also a reference to the old set and a reference to the move_path. But of course it represents only the new state and nothing else. Like A sample has only the trajectory, but the SampleDetails that have more information.

Cool. I see we agree. Great

PR #124, opened 2015-01-27
RETIS

dwhswenson:
Support for minimal RETIS calculation.

To be able to do a RETIS calculation, we need:

- [x] ReplicaExchangeMover (new version) : path mover for replica exchange
- [x] Tests
- [x] `NeighborEnsembleReplicaExchange` factory function
- [x] `PathReversalSet` factory function
- [x] Add RETIS support to `toy_dynamics_tis.ipynb`


jhprinz:
Good point, we need to get a new version of the ReplicaExchangeMove!

dwhswenson:
Actually, it's less a "new version" than a "switch so the `.move()` function works like other `PathMover`s." Your original RepEx part of the code remains mostly intact. :-)

dwhswenson:
Problem with the AWS key?

My recent commits in this PR (e1373b5, and previously 99941cf) were to bring the the changes @jchodera made to do the push to S3. I tried another (ea77f7b) in the hopes of fixing the errors (it didn't). Checking the Travis log, we stop with:

```
+python devtools/ci/push-docs-to-s3.py
  File "devtools/ci/push-docs-to-s3.py", line 22, in <module>
    '''.format(**os.environ))
KeyError: 'AWS_ACCESS_KEY_ID'
The command "bash -x devtools/ci/after_sucess.sh" exited with 1.
```

(See Travis log at https://travis-ci.org/choderalab/openpathsampling/builds/48699188)

Obviously something is missing, probably from the environment, when my PR tries to push to Amazon. But John's PRs were working just fine with Travis. Just glancing at recent diffs, I notice that 99941cf disabled tests (which probably need to be re-enabled anyway). `nosetests` pass locally, as does `make html` in the docs directory. Nothing else has been changed since the last Travis pass in this PR.

Fixing this is outside my knowledge. Help! (Please?)

dwhswenson:
Move discussion of Travis problems to #135.

jchodera:
Not sure if mentioned elsewhere but encrypted keys are not substituted in
PRs except from branches of the main repo for security reasons.

Supposedly the after_success script has logic to skip the S3 and binstar
pushes so this doesn't cause travis to fail, but maybe we have to tweak
that.

dwhswenson:
I fixed it in #135, which has been merged to master. `after_success` was looking for `TRAVIS_PULL_REQUEST == "true"`. That variable holds the PR number if true; "false" otherwise.

dwhswenson:
Moved adding storage support into #137 (otherwise I'll have merge conflicts in ipynbs, which is a real pain to deal with).

That means this is ready to merge.

jhprinz:
Looks good. Will merge...

PR #125, opened 2015-01-28
Changes in path mover implementation

jhprinz:
This is more for testing and review. Not clear that this will be merged ever!


jhprinz:
see #123 for discussion

dwhswenson:
Could this PR include examples of reading the move records from storage, too? A couple simple examples will make it easy for me to get information out of our storage and into my existing analysis routines.

jhprinz:
Help... I get hundreds of pages of debugging commands and can't turn the logging off. :)

jhprinz:
Question about the `FirstSubtrajectorySelectMover` this reports always `True`, but is tested for `False` in testpathmover. Is this wrong or did you still assume that the ConditionSequentialMover will set this to false. I will assume the latter and change the test. Otherwise let me know.

jhprinz:
So, fixed the nose tests. Now to the notebooks

jhprinz:
Help please: Bootstrapping is broken and I need to adapt the BootstrapPromotionMove. If I understand correctly, then this should replace what we have now
```
def move(self, globalstate):
    top_rep = max(globalstate.replica_list())

    shooter = self.shooters[top_rep]
    hopper = self._hoppers[top_rep]
    replicaID = paths.ReplicaIDChange()

    return paths.ConditionalSequentialMover([
        shooter,
        hopper,
        replicaID
    ])
```

where I am not sure if the replicaIDChange Step is implemented correctly.

dwhswenson:
Oh, the joys of working in different timezones! :-) I'll respond to everything in this, although you may have already figured out some of these.

### Logger verbosity

Right now we have module-level loggers. This is what everything I've read on logging says is standard behavior, although I thought class-level made more sense (and the more I work with it, the more I think we'll eventually implement class-level loggers; but inheritance is kind of a pain there.)

Anyway, for now, you're probably getting overwhelmed with debug statements from ensembles while you're actually trying to debug path movers. The fix for that is to (either at the module level or at the level of the individual test) include the following:

```python
logging.getLogger('opentis.ensemble').setLevel(logging.CRITICAL)
```

Obviously, this assumes that you've already done `import logging`, but it you're working in the path mover tests, you have (and you'll see I used the same trick to silence the initialization logging.)

If we had class-level loggers, you could use similar tricks to only get the classes you care about. This is why I'm surprised that everyone does module-level.

### `FirstSubtrajectoryMover`

Yes, you're correct that the `Conditional` flipped it to false. If you're changing this, please also add some `assert`s to the tests to check that after `apply`ing the move, we have the expected `SampleSet`. (The old approach guaranteed that.)

### `BootstrapPromotionMove`

What you have would work, although instead of `Sequential` of one move followed by `Conditional`, this is a good place for a `PartialAcceptance` mover. Reimplementing `BootstrapPromotion` was on my list of "this could be made nicer, but as long as the current version works, other tasks are higher priority".

And you are correct that the `ReplicaIDChange` was not complete. (Also why I hadn't finished reimplementation.)

### `ReplicaIDChange`

I like the idea of keeping the generation of samples separate from the application of them. It makes the samples more atomic and a true log of the steps taken in the move. How's this for a reasonable approach:

* we use null samples as you described above (creating a clear history of how we got from sample set A to sample set B)
* `SampleSet.apply_samples` is changed to include logic that null samples don't get applied (otherwise we end up carrying around a null sample for every dead replica). This might be set in an optional argument to `.apply_samples` with default `remove_null_samples=True` or something. (I'm trying to imagine when you might want to re-use the previously dead replica ID; not sure whether that ever makes sense. But for previously dead ensembles, I could imagine something analogous to pseudoparticle birth/death as done in diffusion quantum MC.)

Note that in this sense, the `ReplicaIDChange` you propose is *not* what we want for bootstrapping: this would remove the lower replica from the list of active samples. At the end, we want `bootstrap.globalstate` to include samples for all the replicas, not just the last one we worked on. Even with `remove_null_samples=False`, this is have you carrying around null samples for the replicas, not the last ones that had a trajectory.

For bootstrapping, you can just implement a `ReplicaIDPromote`, which is initialized with list of replica IDs (or a list of pairs of IDs), and in its move, finds the replica ID of the chosen input sample in that list, and returns two samples: one with the same replica ID, and one with the replica ID after that. In bootstrapping, you don't want to return a null sample in the sense above, because you want to keep the old sample around as the last version of that replica.

jhprinz:
Sure, I will add a simple example. It mainly works, but we need to fix the bootstrapping.

jhprinz:
Almost there... Just spend 1h fixing a problem in the Backwardstep in Calvin Dynamics. I was sooo blind. Should work now. Hopefully...

jhprinz:
AHHHHHHH! There was a bug that drove me nuts:
The bootstrapping and everything works, so I just wanted to run the visualization example which should work now and it got stuck in a loop. Turns out that the EnsembleHopMove were saved twice and one referenced itself. And if you try to start at an ensemble and try to traverse through the file and find all previous used samples at this sample it gets stuck in an infinite loop.

Now, this of course should not happen and all PathMover stuff worked totally fine and kind of before saving it seemed fine because I could get the path tree and after loading a stored file it did not work.

After 2h of headache I found this as the error:
The EnsembleHop uses discarded samples as intermediates so if a hop is successfull there are two samples and when these get saved it first stored the second one which itself references the first. And the automatic recursion saving does this automatically. Now since the second one is saved first it requests an IDX to be stored in the db, then calls save of the first one, which itself requests an IDX to be stored.
Sadly, since the second object has not been written yet, the method that looks for free IDX returns the same IDX and stores the first object, then the second is stored under THE SAME IDX and this causes a loop. 

i have fixed that now, but this was soooo difficult to find... Sorry, needed to post this...

jhprinz:
There is a general ambiguity in the use of Bootstrapping() and PathSampling() which is not compatible.
Bootstrapping uses only a single replica and walk through different ensembles. This is in accordance with the tests. PathSampling however uses a set of movers for different ensembles which does not work with a single initial SampleSet of only one ensemble in Inferface6. I deactivated this to make the notebook run. But I don't know, how this is supposed to be.

jhprinz:
We really need to make sure that logging is TURNED OFF in ipynb. It takes about 10min to open a notebook on my MacBook Pro. And editing these files by hand is no fun :)

jhprinz:
Alright. Except for the out commented PathSampling I would call this readz to merge, so that we can go from there. Once this is merged we can go on and implement the rest of the PathSamplingalgorithm...

dwhswenson:
I'm a little confused by some of those last statements about PathSampling. In particular this comment:

> There is a general ambiguity in the use of Bootstrapping() and PathSampling() which is not compatible.
Bootstrapping uses only a single replica and walk through different ensembles. This is in accordance with the tests. PathSampling however uses a set of movers for different ensembles which does not work with a single initial SampleSet of only one ensemble in Inferface6.

Here's how it should work:

```python
>>> bootstrap_calc = Bootstrapping(parameters)
>>> len(bootstrap_calc.globalstate)
1
>>> bootstrap_calc.run(nfails_max)
>>> len(bootstrap_calc.globalstate)
ninterfaces


>>> tis_calc = PathSampling(parameters, globalstate=bootstrap_calc.globalstate)
>>> len(tis_calc.globalstate)
ninterfaces

>>> tis_calc.run(10)
>>> len(tis_calc.globalstate)
ninterfaces
```
where `ninterfaces` is the number of interfaces, 6 in our case (and obviously other parts of the code are a bit simplified into `parameters`).

You still only have ONE active sample in the final `globalstate` of bootstrapping. You should have 6.

The `globalstate` variable is so-named because it contains the active samples for ALL replicas. This is the necessary input for a TIS calculation. The only reason we have the bootstrapping is to have a quick-and-dirty way to create that input.

Is this just a problem with the new version of the bootstrapping mover, or is this a fundamental issue with the way you're treating moves in this PR? I'll try to look at it a bit to see if I can find a fix. In the old version, existing samples stayed active unless the move changed them (this is what we want: many moves make rather small changes to the global state, e.g. one pair repex'd).

Let's hold off on merge until this is fixed. I've made of branch off of this; if I find a fix, I'll send a PR to your fork.

dwhswenson:
Regarding logging in the ipynb:

1. You don't want to huge load times. Understandable.
2. I think it is really important that users see output in examples. That's important pedagogically, and also gives the user a bit of the thrill of "I got it to work!" (This is the whole point of doing this in ipynbs!)

I think I have a happy medium that manages the best of both worlds, although it requires installation of a third party tool: https://github.com/ipython-contrib/IPython-notebook-extensions/wiki/limit-output

Also, the problem isn't that we have too much output. The problem is that the way notebooks render that output in the browser creates a ridiculous number of objects. This is a known issue on ipython (see ipython's issue 6771, which is also where I found the notebook extension.)

dwhswenson:
Aha! You never implemented the replica ID change in the bootstrap mover. An accepted step should go from `(old_traj, ensemble0, replica0)` to `(new_traj, ensemble1, replica1)`. You were going to `(new_traj, ensemble1, replica0)`. Since there can only be one active copy of any given replica, you were overwriting the old sample in `globalstate`.

I'm working on implementing `ReplicaIDChangeMover`. That should give us the ability to fix this. (Assuming everything after that works.)

jhprinz:
#### replicaID change...
Yes, that was what I guessed, but when I added the replica change step, then all the tests stopped working. There seemed to be designed to work all with replica zero (they use sample_set[0]) and so I assumed that this was the new idea to have one trajectory ID that advances through the ensembles.
So, then we both agree and it makes more sense to use the bootstrapping to generate all initial moves.
One thing that I am not sure about the way this works now: Since we have only samples to be applied to the simpleset, we need to create a special combined shooting and hop move and cannot use the partial, because the partial will apply always the shooting and if it has crossed then add a sample in the new ensemble using new replica ID, but this will leave the result from the shooting move untouched, meaning, that the old ensemble will have a trajectory that is also in the new ensemble (does this make sense). This is not wrong because they are nested, but it is a rather poor choice for an initial trajectory in each ensemble. We could add the old sample again though as a cheap trick!

Let me know, when this is working and I will add it. And then we can finally merge. 

With the ipython notebooks that was more that we should find a way that we do not store the large files. Working with the output is totally fine, but the loading is annoying because of the way it is stored I guess. It might just be a good idea to crop some of the output automatically before we create the package.

Actually, right now the example are only available from github and nowhere else. Is there a way to bundle these with the package? Or have a downloadable zip file that contains the documentation, examples, scripts and a conda install script? That might be nice.

dwhswenson:
OK, I think I've got my head around most of it now, and I think I have a way to solve it (but I need a little help on the MovePath side).

Yes, the problem is that we're accepting all 3 of the samples generated, when we only want to accept the last one. That's part of why the whole thing into a single specialized move before, which I'd forgotten. (You'll notice the references to a combined ensemble hop/replica promotion step still in the comments.)

First off, I think that this is a bit of an unusual edge case: I don't think we'll often alter the replica ID. However, what is not uncommon is that we might create a sample which is just being made as a step toward its final status (for example, we also have such intermediate samples in the minus move). So my suggestion is to add an attribute `intermediate` to the `Sample` object. Then the `apply` function(s) just don't use any samples where `intermediate==True`.

Two challenges:

1. How many places need to have this option added in order to pass it along? I suggest having as an optional argument to `PathMover.move` and also as an attribute for `PathMover`. I think that covers everything.

2. What does this variable look like? I suggest that it can either be a list or a single value. 

  * If a single value, it obviously applies to all `Sample`s generated by that move. 
  * If a list and given to a leaf mover, the values in `intermediate` should be mapped to the `Sample`s in the order they are created.
  * If a list and the path mover is a multimover, it applies as if it were a single value given to each submove. We test that `len(intermediate)==len(movers)` for sanity. 

  If the user has nested movers and a more complicated "intermediate" definition than that, then the user can place the attribute on the appropriate submoves. (Allowing nested lists is tempting, but I think it could be a source of both programming error and user error.)

This also makes me feel better about the intermediate samples generated in `MinusMover`. It's very weird to generate samples that are actually invalid. Any sample that gets generated but that can never be accepted in the `globalstate` is marked as an `intermediate.`

I'm starting to implement this idea by adding `intermediate` to `Sample` and to the `PathMovers`.

@jhprinz , if this approach seems reasonable to you, could you modify the various return functions of `MoverPath` so that we have convenient options to get sets of samples with/without intermediates?

dwhswenson:
... and that still doesn't fix everything, but now I think I have an idea that (combined with the above) does. It only involves adding one more class, and its a class we've talked about adding before: It just requires the if-then-else `ConditionalMover` (which should not be confused with `ConditionalSequential` ... we need to work on some names before we release this).

Have a draft of it now; we'll see if it works!

jhprinz:
Before you implement something with `intermediate`. I think there is a simpler way to mark samples as not to be used (actually this idea is great, always thought only of ways to remove samples we don't want anymore instead of not adding them in the first place). The simplest way to do that would be to not use `SampleMovePath` (which add a sample) but another class like `VirtualSampleMovePath` which references the Sample, but will not return its samples to the total samples that should be used. For this we need to implement the ´samples´ function of the Mover. In the same way we can also make a ´Mover´ that only returns the last accepted sample instead of all of them.

I will have a look at your changes tomorrow and maybe implement my idea to show what I mean. But that should make it almost trivial. You can also imagine that instead of an additional boolean variable for an instance we add a subclass and create the one or the other class depending on the case. So it is in principle your case. Since we have already the mechanism in place that each Mover knows its samples we can simply alter the behavior. 

I guess I should sleep. :)

dwhswenson:
(PR was issued to JHP's fork [the branch associated with this PR] yesterday: see https://github.com/jhprinz/msm-tis/pull/3.)

Most of the features associated with `intermediate` are changes to `PathMovers`, and will be needed in either case (for example, the ability to either mark a `PathMover` instance as always producing intermediates, or to override the way its `move` function is called so that it produces intermediates, and the function to tell the mover whether the sample it is generating `is_intermediate`).

With that in mind, I'm not sure that "simpler" is the right word: aside from those changes, there are tiny modifications to `SampleSet` so that there's one function that applies intermediate samples (used in multimovers) and one function that does not (used at the calculation level). From there, the only difference is that when you build a `Sample`, you also assign whether it is `intermediate` (defaulting to `False`).

On the other hand, depending on whether the user is interacting more with `Sample`s or with `MovePath`s, it may be more intuitive/make better logical sense to use one or the other. I won't have a sense for that until I'm actually writing analysis code.

jhprinz:
With simpler I meant, less code to write, or easier for me to implement not simpler in usage. 

Most changes are required for both appraoches. 
It would also only require to add some code in the HopMover and write the new MoverPathType. I guess it is mostly the question again, which part nows about applying samples, is it the Sample itself or the MovePath. I think the MovePath, but it is kind of philosophical. I would like to make it, so that
1. Samples contain the trajectory and information about the creation of the trajectory, but not about the "replica" process. The only information we should store is the input samples, the new one and (which in theory could also be into the MovePath) the acceptance probability. The SampleMovePath only return the sample if the acceptance is True and we could of course create `AcceptedSampleMovePath` and a `RejectedSampleMovePath` instead
2. SampleSet includes only a list of samples and are the statistical correct step in the whole scheme that we want to generate
3. MovePath contains all other information to move between two SampleSets and the way this is done is exclusively by applying samples to sample sets. This way an ordered list of samples is a reduced way to express a move path with alladditional information stripped. And we have a property `.samples` that allows to cast a MovePath to the relevant list of samples. 
4. A SampleSet can be "moved" by applying a list of samples (not a sample set, which is an unordered set of samples). Alternatively - again a philosophical question - we can delegate the knowledge about how to apply a move path to a sample set (right now replacing/adding samples) from a function of the SampleSet to a function in the MovePath. I think the second is the move clear separation to say a move path knows how to move something. But either way is fine.

So, instead of adding the intermediate property I would like to keep this information in the MovePath. The last thing to think about is, that it might come in handy to mark each sample as virtual or not. But this is again different from being in a sample set or not, so be might have to add that, too. Not sure, still think all of this information is part of the MovePath, not of the sample.

Does that sound reasonable?

jhprinz:
After more thinking I agree that the idea to use only a list of samples that is applied is very flexible and I would like to keep this and discard the idea of other ways to change samplesets for now. As I wrote in your bootstrapping PR we can do everything. So, I would keep the current way we do things, but replace intermediate sample flag, by a `KeepLastSampleMovePath`. So we have the following way to go from one globalstate to the next:

1. PathMover (SampleSet/globalstate) -> MovePath

    ```
    movepath = mover.move(globalstate)
    movepath = mover(globalstate) # shortcut ?!? or
    new_globalstate = mover(old_globalstate) # as shortcut for getting .move and applying it in one step
    ```

2. MovePath -> list of samples

    ```
    samples = movepath.samples
    sample_iterator = movepath # shortcut
    ```

3. SampleSet (list of samples) -> SampleSet / new globalstate

    ```
    new_globalstate = old_globalstate.apply_samples(movepath.samples)
    new_globalstate = old_globalstate.apply_samples(movepath) # shortcut by maybe just
    new_globalstate = old_globalstate.apply(movepath) # because it reads better
    new_globalstate = old_globalstate(movepath) # even shorter but confusing with globalstate[xyz]
    new_globalstate = old_globalstate.apply_samples(mover(old_globalstate))
    new_globalstate = movepath(old_globalstate) # alternative ??? but would mean
    new_globalstate = mover(log_globalstate)(old_globalstate) # which is totally confusing
    ```

Which of these do we want? I really like
```
movepath = mover.move(globalstate)
new_globalstate = old_globalstate.apply(movepath)
```
as the long way, or if movepath are not required we do not confuse the used and simple allow
```
new_globalstate = mover(old_globalstate)
```

dwhswenson:
I don't think that `KeepLastSample` is as flexible as the `intermediate` flag, and therefore I don't think it should replace `intermediate`.

***

Aside: if we sequentially apply a list of samples, replacing whatever previously existed in the replica of the given sample, then `intermediate` is only **necessary** when we change replica ID, which is an extremely rare thing to do. In that case, careful wrapping with `KeepLastSample` will do the same thing, and while it may require an ugly combination to do it correctly, it's such a rare case that it doesn't matter. 

But in general, without `intermediate`, the list of samples can include samples aren't internally consistent. That's weird, and I liked the idea of being able to immediately identify samples that only exist as a step toward building a real sample. In a lot of real moves, there are samples which never have a chance to be considered real: they truly are built as intermediates. It seems weird to me to save them on the same footing as valid samples.

***

Example: the Minus Move. I'll use the notation `(r, t, e)` for a `Sample`'s replica, trajectory, and ensemble attributes.

In the case of the minus move, there are three ensembles: the innermost `e_i`, the minus `e_m`, and the "minus segment" `e_s`, which is almost the same as `e_i` (for the vast majority of trajectories, they are equivalent). There are two replicas (trajectory IDs); let's call them `a` and `b`. I'll use an asterisk `*` to indicate a sample that would be marked `intermediate` in my scheme (all non-marked samples should be included in a `SampleSet`.

Let's look at a successful minus move. It generates following samples:

1. `(a, t_a, e_m), (b, t_b, e_i)` : Initial condition
2. `(a, t_a', e_s)*` : Get segment of `t_a` (call it `t_a'`) which satisfies `e_s`
3. `(b, t_b, e_s)*, (a, t_a', e_i)` : repex trajectories in `e_s` and `e_i` (near 100% acceptance)
4. `(b, t_b, e_m)*` : force-hop into the minus ensemble
5. `(b, t_b+, e_m)` : Extend trajectory `t_b` into trajectory `t_b+`, which satisfies the minus ensemble.

Note the following:

* There are TWO samples which must be returned from this. How would `KeepLastSample` handle that? 
* One of the samples which should be returned (on success) in generated in move 3 -- it isn't even generated in the last submove of the `ConditionalSequentialMover`.
* The sample from step 4 is actually nonsense: `e_m(t_b) == False`. It is truly used as an intermediate. Saving it like other samples seems weird.

We could solve the last problem by making a more complicated `ShootingMove` (and I think that might be a good plan for the future) but the other problems remain. And technically, you certainly *can* use `KeepLastSample`, but it involves making the `MinusMover` into a `ConditionalSequential` of two `KeepLastSample`s, each of which wraps a `ConditionalSequential`. I don't think that's simpler than just assigning the proper value to the `intermediate` attributes of each submove.

None of this actually causes problems in this specific case, because, as I said, each move overwrites the last version of that replica. So the last thing to happen to replica `a` was in move 3, and the last thing to happen to replica `b` was in move 5: we treat everything as possible. But if you want to talk about what will work generically, `intermediate` is more flexible than `KeepLastSample`, and it has the advantages that it lets us label some samples as not even potentially real.

jhprinz:
Woohoo... I just ran the toy_example completely and it worked. I saved the simulation and could read everything using storageview.py
Pretty cool. For this I merged your changes and mixed our ideas a little bit. I took out the intermediate related code and added the KeepLastSampleMover. This worked as expected and made the code in BootstrappingPromotionMover a little shorter.
After that I also needed to fix the storage of a few new objects. This mainly involved removed too much information that kept the loading from correct restoration.

Still checking why it is not building. But I would say that we finally achieved our goal to make a production and store it. YESSSS! So, cool to see that is finally works!

Alright, we should check, that the stored data is really what we want, but so far it looks fine!

dwhswenson:
@jhprinz : `n_crossings` is not quite correct. `n_l` is used because that's also the expression used in the paper. `n_crossing` is ambiguous, as it could refer either to exits of the state or to exits of the interface. It actually refers to the number of segments satisfying the minus segment ensemble (so, exits from the state which are then followed by an exit of the interface and an entrance into the state). You can call it `n_segments`, but I'd still prefer to follow the notation of the cited paper.

jhprinz:
I see. That makes sense. I was just confused when reading the code and it took some time to figure out, what this actually is. Maybe we should just name it back and add a little more documentation.

What do you think about the changes. It that okay? Sorry, that I just removed this. It seemed like the quickest fix from my perspective. Let me know what you think.

From my perspective we should merge this and go from there. This PR has grown so big that it became practically impossible to see the relevant changes.

dwhswenson:
Is there any way to get back the `intermediate` stuff? I think it added a decent bit of flexibility which is no longer available. Did it break something else? The PR I sent you passed nosetests and did everything except the `toy_tis` storage (because of the JSON bug.)

It's easy enough to switch the `n_l` name back. And there's already documentation on `n_l`: the docstring for the minus ensemble includes both a "parameters" section and a reference to the article.

Let's deal with those last two issues before merging this.

jhprinz:
n_l will take two minutes, will do that quickly.

Do you have an example for more flexibility? I think both approaches allow for the same possibilities. Just storing the information not in the sample itself and the way to write it is different.

I will just add this back. But it will take a little while...

dwhswenson:
Is it not clear from the example in [this comment above](https://github.com/choderalab/openpathsampling/pull/125#issuecomment-73385426)? 

(Actually, I'm guessing that you missed that comment because you were writing your comment below it at the same time as I posted mine)

In that specific example, neither approach is actually needed. But imagine if we threw in some `ReplicaIDChange`s in there....

dwhswenson:
Not something to worry about before closing this PR, but where you said

>  Alright, we should check, that the stored data is really what we want

I would second that. Strongly. In the ipynb, you run 10 production steps, and have 201 `Sample`s in storage. So 20 samples per step. That is not right. The minus move generates up to 5 samples per step, replica exchange generates 2, and path reversal and shooting each generate 1.

jhprinz:
Hi, yes, missed it :) Sorry, that was a good example

jhprinz:
So, I totally see the point and you are right, but I was kind of arguing for something else. My only point was, that I want to implement the intermediate idea differently and used the BootstrapMove as an example. And there is became simpler (I think). I loved your comment about Texas :)

So, we definitely need a way to incorporate the MinusMove naturally into the rest. This is clear and it needs to be feasible and clear to use. I would like to try to come up with something that can do everything that intermediate does, but stores it a little different and does not require to change every Mover to do so. I am not 100% sure that this is possible, but I would like to try til tomorrow.

Would that be okay? I think we are arguing on a very high level and potentially noone will understand what we talk about :)

dwhswenson:
Well, as it stands now, the actual `MinusMover` isn't broken. It's just that most of the `Sample`s it generates are between useless and nonsense. However, the two that are correct are the only two applied to calculation's `globalstate` variable. To be clear, the issue is about whether it makes sense to mark that nonsense as nonsense in storage, and whether some hypothetical replica-ID-changing move might be possible.

I think that the `intermediate` attribute of a `Sample` can be moved to the `MovePath`, although it amounts to the same thing. The part that I don't think you can avoid are the changes to the PathMovers, although in the end all that requires from the developer is that you assign the value of `mover.is_intermediate(samplenumber, intermediates_list)` to  `sample.intermediate` (or `movepath.intermediate`). 

This is just from an information flow perspective: whether a step is `intermediate` needs to be user-determined. `MovePath`s (and `Sample`s) are not made by the user. How would you get the information into one? The only answer I see is through the object which creates the `MovePath` or `Sample`: i.e., the `PathMover`. Conveniently, that's also the thing whose output we want to "mask" by calling it `intermediate`. This seems the most straightforward approach.

But feel free to look for another way -- I'm headed to bed now anyway!

I do think your `Bootstrapping` is easier. Actually, I think I could have undone the inverted algorithm as soon as I developed replica filters, and that would have worked, too. But then I would have had to remove the line about Texas. ;-)

jhprinz:
Okay, I admit I had a wrong view on the intermediate, it really does what we want. Your last post pretty much got it to the point and this clarified the rest to me.

To be honest, I think I wasted too much time for the both of us. Sorry, but I had this hunch that we can do it more elegant. Here are my thoughts on how we should implement the *virtualsample* idea. I should provide almost the same functionality, but has some advantages when it comes to saving with the trade off that it does not allow to exactly tweak every subsample.

### hiding intermediate samples

The main point is that I did not want the user to have to pass on information about the underlying movers, since it would mean to know how these work and what they generate. So, if we change the implementation of say SequentialMover, we have to change the intermediate property of everyone that calls it. I would like for each mover to be as effective as possible on its own and that it can be used without knowledge of unterlying functionality. This is indeed not so flexible as allowing total control and could be less efficient, but I think it might not be too bad.

If we want to reduce the number of unnecessary saved samples then this approach does not help in the current implementation. Although a sampleset only saves its new samples it also saves the move path and this has references to all intermediate samples not only the real ones, because their moves might depend on them. We can of course somehow prevent saving intermediate-marked samples from being saved but that might create problems with correct references in the movepath.

The only way to fix this is to replace Parts of the MoveTree by a new simplified structure. Effectively, cutting out the part that contains the virtual samples and replace it by something else. So, here my proposal and if this does not work out, I will add the intermediate and stop bothering you.

Okay, if one constructs a MoveTree using the Movers, we allow to add a new Mover that is called `ContractMover` and all that it does is to create a Fake Wrapper mover that calls everything that is under it, and removes all samples that are not finally used. This is easy because samples are applied in row and is kind of the ingenious advantage of using a list of samples. Just use our sampleset.apply_samples to check if a sample would be replaced and return only the samples, that are really relevant. So, if you now store the MovePath, it will not save everything below the `ContractMover` and mimic a new mover that just magically created some new samples. Its name can be given arbitrarily. 

This way we will remove non used samples at chosen stages in the movetree, which I assume is more at the fundamental level of mover than higher-level combinations. Like the MinusMover should be saved as a MinusMove that generates 2 samples and not as some complicated combination of moves. And this would be easily possible with this approach.

If we decide to store the whole move path, then even that would be possible since the MovePath still has the reference to the underlying hidden parts and we could request the full tree and then just save it if we wanted to. Or add it to logging, etc. Since all subtrees are valid move paths we can also only store the minusmove interior itself separately for whatever reason.

I realize that this is different from marking samples as intermediate, but so the user has not to worry about this. It also means that we cannot just open the storage and request all real samples, but I fear that this choice is somewhat arbitrary, since effectively all samples play a role at some point. Even when they get rejected, it is just a question what to keep and what to compress into a single move.

The picture I have in mind is a file directory tree with little button to open and close nodes. This is how I like to think about it. The ContractMover (or adding the ContractedMovePath to the output) just means a node is closed and save is only, what you can see.

To make it even close to the `intermediate` idea we could add a flag to each mover to say if it should return collapsed or full.

For comparison, right now we have 
- `movepath.samples` to get all accepted samples / depends on open/close status
- `movepath.all_samples` to get all samples
- `movepath.local_samples` to get only the ones created in this step
we add
- `movepath.closed` creates a closed copy
- `movepath.opened` creates an open copy

Some example for the minus move
```
minusmover = MinusMover(
    minus_ensemble=...,
    innermost_ensemble=...,
#    collapsed=False # override default behaviour (True for minusmover)
)
minusmover.collapsed = False # Alternative override after creation
movepath = minusmover(old_state)
print movepath # will print the movepath in default setting
print movepath.closed # print the 
print movepath.opened # 
# the view or storage form does not change the way a movepath is applied
new_state = old_state.apply_samples(movepath.collapsed)
storage.save(movepath.opened)
```

### replicaID change treatment

For this problem I would use a FilterMove or KeepLastSample or something similar to get rid of the samples that have been created and should not be applied. Then apply the same idea above to mask everything as one move.

Alright. Last attempt. Let me know what you think :)

P.S. Maybe I just try to add the Texas line again

jhprinz:
So, done... The changes are in a PR in https://github.com/jhprinz/msm-tis/pull/5
to keep this one clean.

We can now even easier collapse a complicated MovePath into one and store the collapsed. Just use
`.opened` and `.closed` to switch. It is also completely consistent with what we had before. So, if we don't use it, that code is the same as before. 

The only change I made was to set standard output of MinusMover to be collapsed.

So, for the tests instead of getting the samples using
```
movepath.samples
```
you will now only get the two relevant samples and
```
movepath.opened.samples
```
gives the old result. Hope you like it. I find it pretty convenient, though it is not perfect.

dwhswenson:
Merging this part (to by followed by JHP issuing a PR and/or merging in jhprinz#5).

We've got stuff that works, but I think that our lack of consensus on the best way to implement this stuff shows in the code, and will be confusing for future contributors. We need to clean that up before release, but that cleanup will require a time when we can talk long enough to come to consensus. I assume @jhprinz doesn't have that kind of time in the next few days!

In the meantime, merging this will allow us to move forward on developing tools for analysis.

PR #126, opened 2015-01-28
[WIP] Upload to binstar

jhprinz:
Checl if we can put the conda package on binstar


.travis.yml
```diff
@@ -46,3 +46,7 @@ env:
     - python=2.7  CONDA_PY=27  CONDA_NPY=19
 # - python=3.3  CONDA_PY=33  CONDA_NPY=17
 # - python=3.4  CONDA_PY=34  CONDA_NPY=18
+
+  global:
+    # encrypted BINSTAR_TOKEN for push of dev package to binstar
+    - secure: "Ou9KukGxCOdNWvEpr+9YrWTlB0xezu1mNV1gXtFjJ0RDIvzPRGUkpU7AUAdKwn6fUcbzq5E1y6rU9aG/QQe9YZkRHhPo585ZzieTNa0UxsyINk7XuwxL7MpRZijHCHFBvRVNrXkIh3CtmXvZS2ksAa8+FuYiO/lBbOMZ0nNAJac="
```
jhprinz: Here is the token

devtools/ci/after_sucess.sh
```diff
@@ -2,20 +2,25 @@ coveralls
...
+#    echo "No deployment on BRANCH='$TRAVIS_BRANCH'"; exit 0
+#fi
 
+echo $BINSTAR_TOKEN
 
-if [[ "$TRAVIS_BRANCH" != "master" ]]; then
-    echo "No deployment on BRANCH='$TRAVIS_BRANCH'"; exit 0
+if [[ "2.7" =~ "$python" ]]; then
+    conda install --yes binstar jinja2
+    binstar -t ${BINSTAR_TOKEN}  upload  --force --u omnia -p openpathsampling-dev $HOME/miniconda/conda-bld/linux-64/opentis-dev-*
```
jhprinz: This throws an error from banister saying `omnia` is not a valid directive, should be something like `upload`, etc... I assume that the token is not working and messes up the command somehow. Anyone seen this problem before?

jhprinz:
Help here please. I assume that the token is not correct and messes up the binstar command somehow.

kyleabeauchamp:
Have you confirmed that things work manually from the command line?

jchodera:
Note that encrypted environment variables are not substituted in PRs
because that would be a security issue.  Instead they are just blank, and
you get an error like this. You probably just have to commit this and debug
from there.

PR #127, opened 2015-01-28
Test secure keys for binstar and Amazon s3

jhprinz:
Apparently secure key do not work on forked repos so I clone also the choderalab/opentis and test it here.


jhprinz:
IT WORKS!!!! So cool...

Actually the conda package already assumes that the project is called openpathsampling, so I assume we need to rename this. But the documentation is online! Still VERY simple, but it's coming...

http://openpathsampling.s3-website-us-east-1.amazonaws.com/html/

jchodera:
Looking good!

jchodera:
I'm going to merge this so that I can start a PR to tweak things to work with http://openpathsampling.org

PR #132, opened 2015-01-28
Push to S3 now pushes to latest/ folder

jchodera:
This should fix the automatic redirection from http://openpathsampling.org


jchodera:
Works now!  Merging.

PR #133, opened 2015-01-28
Added link to documentation website to README.md

jchodera:
This just adds a link to http://openpathsampling.org to the `README.md`.


jchodera:
I've also replaced the base `index.html` with a less problematic version, since we can handle redirects on the S3 side.

PR #135, opened 2015-01-29
Trivial PR

dwhswenson:
A tiny update to README.md to see if Travis likes this or not. See problems in #124.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,9 @@
 [![Build Status](https://travis-ci.org/jhprinz/msm-tis.svg?branch=visualization)](https://travis-ci.org/jhprinz/msm-tis)
 
-# opentis
+# OpenPathSampling
 
-An open source Python framework for transition interface and path sampling calculations.
+An open source Python framework for transition interface and path sampling
+calculations.
 
 See documentation here:
 
--- a/devtools/ci/after_sucess.sh
+++ b/devtools/ci/after_sucess.sh
@@ -2,14 +2,14 @@ coveralls
 
 echo $TRAVIS_PULL_REQUEST $TRAVIS_BRANCH
 
-#if [[ "$TRAVIS_PULL_REQUEST" == "true" ]]; then
-#    echo "This is a pull request. No deployment will be done."; exit 0
-#fi
+if [[ "$TRAVIS_PULL_REQUEST" != "false" ]]; then
+    echo "This is a pull request. No deployment will be done."; exit 0
+fi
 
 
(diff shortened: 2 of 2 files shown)


dwhswenson:
This is a fresh checkout of `master`. The only difference is an minor change to the README.md. It doesn't pass Travis. This is the thread where we should fix that.

jhprinz:
This is actually not a problem, but a security feature. Patrick actually helped me with the same issue. The script only works, if the build is a PR from the choderalab/ops and not from a fork. This is to ensure that the forked script do not spill out the secure passwords and only the original git is trusted to use secure variables. So all environment variables in .travis.yml that are set using the secure part are only run in these real PR, hence it cannot find AWS_... in PR from our forks and fails. In the upload to s3 is some code that checks if this is the case and stop the script otherwise. I will repair this or you just do this.

```
coveralls

echo $TRAVIS_PULL_REQUEST $TRAVIS_BRANCH

if [[ "$TRAVIS_PULL_REQUEST" == "true" ]]; then
    echo "This is a pull request. No deployment will be done."; exit 0
fi


if [[ "$TRAVIS_BRANCH" != "master" ]]; then
    echo "No deployment on BRANCH='$TRAVIS_BRANCH'"; exit 0
fi

if [[ "2.7" =~ "$python" ]]; then
    conda install --yes binstar jinja2
    binstar -t ${BINSTAR_TOKEN}  upload  --force --u omnia -p openpathsampling-dev $HOME/miniconda/conda-bld/linux-64/openpathsampling-dev-*
fi

if [[ "$python" != "2.7" ]]; then
    echo "No deploy on PYTHON_VERSION=${python}"; exit 0
fi


# Create the docs and push them to S3
# -----------------------------------
conda install --yes pip
conda config --add channels http://conda.binstar.org/omnia
conda install --yes `conda build devtools/conda-recipe --output`
pip install numpydoc s3cmd
conda install --yes `cat docs/requirements.txt | xargs`

conda list -e

(cd docs && make html && cd -)
ls -lt docs/_build
pwd
python devtools/ci/push-docs-to-s3.py
```

dwhswenson:
`TRAVIS_PULL_REQUEST` return the number of the PR if it is a PR; "false" if not. The code tested for the value "true", which it never returns. http://docs.travis-ci.com/user/ci-environment/

I'll go ahead and merge this, since the changes are trivial and this in should fix the build problems on our current PRs. (Although we should double check that a non-PR Travis run does successfully push to S3 now; but that can't be done from within a PR!)

PR #136, opened 2015-01-29
Newpackagename openpathsampling

jhprinz:
This renames everything to use openpathsampling. Hope I did not forget anything...

PR #137, opened 2015-01-30
[WIP] Minus Interface: Minus Ensemble, Minus Movers

dwhswenson:
Support for the minus ensemble/move. Technically, this is designed to implement the [multiple interface set minus move](http://scitation.aip.org/content/aip/journal/jcp/141/4/10.1063/1.4890037), including the possibility of multiple re-entries ($n_l>2$).

The `MinusMover` involves two ensembles: the innermost interface ensemble, and the minus ensemble, which consists trajectories with `n_l` distinct segments which satisfy the innermost interface ensemble. (Usually `n_l=2`). 

Technically, there's also another ensemble at work here: we'll call it the minus interface segment ensemble. This is almost the same as the innermost ensemble, except that paths must end in the initial state. (This ensemble is saved as an attribute of the minus ensemble.)

Let's say we start with path A in the innermost interface, and path B in the minus interface. Path B includes two segments, B1 and B2, which satisfy the innermost ensemble.

The `MinusMover` is a `ConditionalSequentialMover`. It takes the following steps, only going to the next one if the previous was successful. Unless the whole process is accepted, no moves are accepted.

* Select either the first or last subtrajectory (B1 or B2) satisfying `innermost`. (Pretend we got B1)
* Perform of replica exchange of that subtrajectory with the trajectory actually in the innermost interface. (This should have nearly 100% acceptance.)

At this point, we have trajectory B1 in the innermost ensemble, and trajectory A in the minus ensemble. There a little subtle point here: technically, we put B1 into the minus interface segment ensemble, and now A is in that. Otherwise, the replica exchange wouldn't have worked. Of course, A doesn't satisfy the minus ensemble: it contains only one of the $n_l$ segments. To fix this, we need to extend the trajectory. But first, we need to force it to think it is in a different ensemble with:

* Use a `ForceEnsembleChangeMover` to move the sample with path A from the segment ensemble to the minus ensemble.
* Extend A until it satisfies the minus ensemble. This is done by randomly selecting from a `ForwardShootingMover` which selects the final frame of A (extending forward) or from a `BackwardShootingMover` which selects the first frame of A (extending backward).

At this point, the path A has been extended and satisfies (and is in) the minus interface, and the path B is in the innermost interface.

Note that these last two steps could be replaced with a `ForwardExtension`/`BackwardExtension` pair of moves, and that would mean that we don't have the weirdness of `ForceEnsembleChange` leaving us with a sample that isn't technically valid (`sample.ensemble(sample.trajectory)!=True`). That may be a better way to re-implement this in the future.

To-Do:
- [x] FirstFrameSelector (ShootingPointSelector)
- [x] FinalFrameSelector (ShootingPointSelector)
- [x] RandomSubtrajectorySelectMover
- [x] FirstSubtrajectorySelectMover
- [x] FinalSubtrajectorySelectMover
- [x] ForceEnsembleChangeMover
- [x] MinusEnsemble
- [x] MinusMover
- [x] Docstrings for all the above
- [x] Tests for all the above
- [x] Include storage usage (for RETIS part) in `toy_dynamics_tis.ipynb`


dwhswenson:
I'm going to move MultipleSetMinus to a later PR, that way (1) @jhprinz can pull this into the renaming PR #136 and merge it (following the plan in #139), and (2) he can start modifying the *many* added moves to match the new move structure he's preparing in #125.

jhprinz:
Cool. Just fixed the renaming and will add this PR now

jhprinz:
Merge almost done for renaming. Checking notebooks now and rename packages in there.

jhprinz:
waiting for the travis build to pass

jhprinz:
Merging

PR #140, opened 2015-02-01
Docs: No small caps in ipynb inline code

dwhswenson:
Previous CSS used smallcaps for anything marked `code` in the Markdown cells of IPython notebooks. This made it (1) look different from the actual code cells; (2) look different from inline code in the rest of the sphinx docs; (2) look really ugly when there were many inline references to things from main code blocks (see the "Example trajectories for these ensembles" in http://openpathsampling.org/latest/examples/sliced.html for an example).

This fixes that. Also replaces one header MDTraj->OpenPathSampling.

### DIFF
--- a/docs/_themes/sphinx_rtd_theme/static/css/ipynb.css
+++ b/docs/_themes/sphinx_rtd_theme/static/css/ipynb.css
@@ -69,7 +69,7 @@ margin: 2px;
 
 /* font-family: Courier; */
 font-weight: normal;
-font-variant: small-caps;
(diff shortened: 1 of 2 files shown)


jhprinz:
Yes, I played around with that ;) Wasn't so sure, if it was a good idea either. It partially looked cool, but I didn't like the large brackets and numbers. This looked awkward.

The general idea was to remove the automatic ipython notebook css and let the one from RTD take over to that it looks consistent. 

Merging.

PR #145, opened 2015-02-04
Fix tests

dwhswenson:
**Problem**: The refactor of tests makes it so that Travis doesn't see that our tests fail, even when they do!

**Solution**: this PR

**Irony**: We currently have a failing test, so this PR should be merged if it FAILS Travis, not if it passes!

See, e.g., [build 351](https://travis-ci.org/choderalab/openpathsampling/builds/49457319#L1430) around line 1430. The `toy_dynamics_tis.ipynb` fails, but the build passes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,8 +24,8 @@ script:
   - conda list -e
 
   # Run tests
-  #- source devtools/ci/nosetests.sh
-  #- source devtools/ci/ipythontests.sh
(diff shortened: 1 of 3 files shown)


dwhswenson:
Failure is success!

(And war is peace, freedom is slavery, etc. It may be double plus ungood that failing tests got into master, but at least it won't have to happen again.)

This is ready to merge as soon as we want tests re-enabled.

PR #148, opened 2015-02-06
Should allow all tests in subscript to run

dwhswenson:
I think this will make a small improvement as an addendum to #145 : previously the `ipythontests` stopped after the first failure. Better to run all the notebooks and see if anything else fails.


jhprinz:
Merging

PR #149, opened 2015-02-07
Move repo/scripts to repo/package/scripts.

### LINKED ISSUE #147: `setup.py` and scripts
(Splitting this into a separate issue. Something @jchodera said reminded me of a problem we should be aware of.)

In #112, @jchodera wrote:
> For scripts, use the entry_points feature of setup.py to handle these, since they will then be installed correctly by conda as well: https://github.com/choderalab/yank/blob/master/setup.py#L145

One side warning on `setup.py` and scripts: there is a bug in setuptools that makes it a bit of a problem to keep scripts outside the main package directory. For details see [this bug report in setuptools](https://bitbucket.org/pypa/setuptools/issue/230) and [this very long explanation in distribute](https://bitbucket.org/tarek/distribute/issue/177/setuppy-develop-doesnt-support-package_dir).

In short, the issue is that if the user wants to install an editable version a live and editable version of the package (i.e., installed using `setup.py develop` or `pip install -e`) the scripts don't install, because the effective directory rewriting in `package_dir` never happens. The entry point is at `package.scripts.scriptname`, but no such module exists.

Two solutions that will work:

* Move `repo/scripts/` to `repo/package/scripts`
* Warn such users that they have to add the symlink (I suspect most such users will be pretty advanced)

There may be some other `setup.py` trick that works; but I can't find any evidence of it.

Of those options, I prefer the former, since it makes the user's life easier (and I'm one of the users who works this way!)

(Based on the `setup.py` John linked to from yank, I don't think this is a problem there; however, I encountered it while playing with MDTraj, which uses the same directory structure as us.)

dwhswenson:
Resolves #147.


jhprinz:
Merging.

PR #157, opened 2015-02-10
Add Query support for storages

jhprinz:
This allows to easily define a Query for the storage and load all objects that match this.

### Example
Load all samples that are samples for a specific ensemble
```
ensemble = ... # ensemble to be searched for
query = paths.storage.StorageQuery(lambda self : self.ensemble is ensemble)

# Query the storage and return an iterator over all matchin samples
for sample in query(storage.samples):
    print sample.ensemble.idx[storage] # check, if the idx is the same for all found samples
```


jhprinz:
Few things I realized that we need to change in Sample and SampleSet for the analysis:

### SampleSet
To search for only relevant samples we need to search sampleset. The only problem is, that we might have different calculations and it is not clear, which to pick from. We might have stored all these results and need to know, which sampleset is from which calculation. So, I propose to add either a name or a reference to a Calculation to know which samplesets we want. We could in principle just store all calculations with it.

- option 1
    
    Storable calculation and refrence to it, the `.run` of the calculation set the property automatically, so we know, what it was run by. 
    ```
    my_calculation = BootStrapping(...)
    sampleset.calculation = my_calculation
    ```
- option 2

    Just a string to mark it
    ```
    sampleset.calculation = 'Production'
    ```
- option 3
    
    Nothing. Just use the MovePath as a reference what made the simpleset
    Still, a calculation is somehow a path mover factory with some convenience to apply it several times and save the results along the way. So, if the Calculation would wrap the full mover in a named mover then the move path would contain which Calculation was done.

### Sample
I realized that due to using the same sample in several samplesets the .step property is not meaningful and should only be stored in the sampleset. Only in the context of the sampleset the step has a meaning.

jhprinz:
### Much better using iterators and list comprehensions
A storage is effectively a list. One that is stored and can only be appended and not changed, but still a list.
So I implemented `__iter__` and `__len__` and now we do not need the query we can use simple python list comprehensions

```
# get all samples in all sample sets that are in ensemble ens
samples = [
    sample for sampleset in storage.sampleset for sample in sampleset 
    if ee.ensemble is ens
]
```
or if we have a list of ensembles use
```
samples_per_ensemble = {
    ens : [
        sample for sampleset in storage.sampleset for sample in sampleset 
        if ee.ensemble is ens
    ]
    for ens in ensembles
}
```
of course this list can contain several samples multiple times!
The above examples create lists which can take a while. If you want an iterator instead you can use generator expression and replace the `[]` with `()`
```
# get all samples in all sample sets that are in ensemble ens
sample_iterator = (
    sample for sampleset in storage.sampleset for sample in sampleset 
    if ee.ensemble is ens
)
for sample in sample_iterator:
    print sample
```

The `Query` function I wrote only adds caching that would be the only benefit, so objects we know are false from a previous run will no longer be loaded and returned a second time.

jhprinz:
So, i had to import the other PR, I hope this was okay. It mainly fixes a few glitches and all the other functionality is not used yet.

On my machine all tests passed.

jhprinz:
So, I made a longer example about how to work with the storage. In there is shown how to get the data that you want. I did not use the new Query function that I mentioned before. Just using generator expressions and list comprehensions is so python like and the storage now really behaves like immutable lists. Have a look at the examples. i show how to get all samples from sample set with a specific ensemble and compute the average pathlength in basically 2 (long) lines.

jhprinz:
Forgot to mention: the example is an i python notebook.

dwhswenson:
This looks great: what did you end up doing with `SampleSet`? Of the three options, I don't really like the second one, but the others seem fine to me.

How close is this to ready?

jhprinz:
It is ready to merge besides the sampleset. I would pick something between 1 und 3 actually. Save the calculations because we can, but actually wrap them in a CalculationMovePath that references the additional information so that we know that a particular move was done by a specific calculation.

dwhswenson:
So for a PathSampling calculation, a CalculationMover would essentially be the `root_mover`, but instead of being a RandomChoice, it would be a subclass of random choice that also includes a pointer to the calculation? That works by me.

jhprinz:
So, I think this is ready for merging. I am not completely finished with cleaning up the Calculation storage, but this is independent of the other storage stuff.

here an example what is now contained in the movepath (aka move record)
```
        4] 7 sample(s)               : [0, 1, 3, 5, 6, 8, 11]        
                CalculationStep : PathSampling : Step # 3 with 1 samples
                 +- RandomChoice :
                 |   +- RandomChoice :
                 |   |   +- RandomChoice :
                 |   |   |   +- SampleMove : ForwardShootMover : True : 1 samples [<Sample @ 0x10c26a190>]
```

It will also save the calculation (not yet active for Bootstrapping I think) and its contained properties. It will not however store the initial globalstate, since the global state contains the movepaths and thus we get a circular dependency which causes still some problems in the storage. And I would like to avoid checking for loading circles. Will look into that. It usually means we store redundant information.

if you wanne use it, wait for the build to pass (hopefully) and merge.

Let me know if you need something else quickly.

dwhswenson:
All looks reasonable, although I can't tell from the above whether `CalculationStep` is the `root_mover` or not.

However, this can't be auto-merged. There's some conflict with master. (You might also consider merging #159 first; that way if the storage test causes problems again, we'll get a hint what's going on.)

Anyway, if you can make this merge-ready tonight, I think it is (in principle) good to be merged.

jhprinz:
I think this is good to go. Merged #159 and found a last error in the correct storage of PartialSeqMovePath. All notebook run now. So if it passes, we are good to merge.

jhprinz:
> So for a PathSampling calculation, a CalculationMover would essentially be the root_mover, but instead of being a RandomChoice, it would be a subclass of random choice that also includes a pointer to the calculation? That works by me.

Yes, kind of. although this is done internally and the calculation stores its root_mover again.
I am not completely convinced that this is the best way, but it works for now and contains all information.

As an alternative we could have made a special PathSamplingMover that just applies the root mover and knows the the calculation details. That's actually what it kind of is, but a general class for all calculations.

For Bootstrapping I did not to that because we have already a "unique" BootstrappingMove that has a descriptive name.

jhprinz:
Let's hope this builds :)

jhprinz:
Fixed the case, that an interface had no frames and a division by zero occurred in the example.

jhprinz:
Ready to merge. The last errors were mainly due to sloppy examples, but it should work now.
Let me know, if you need anything else to continue.

dwhswenson:
Oh good! I figured the only problems were in the examples, so I'd already merged it into my analysis branch to work on.

jhprinz:
Perfect...

PR #158, opened 2015-02-11
Add Pandoc Markdown support

jhprinz:
Checking what is wrong with markdown conversion to enable adding Davids Examples!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,8 +24,8 @@ script:
   - conda list -e
 
   # Run tests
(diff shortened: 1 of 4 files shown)


jhprinz:
This works now. The fix was to apt-get install pandoc was missing. Now it actually compiles markdown using the
```
.. markdown:: [filename].md
```
directive

Am example is in readme.rst in the main docs folder

jhprinz:
Ready to merge, although we can really test it only if it merge with master!

Issue #138, opened 2015-01-31
Problem with `test_stored_topology`?


dwhswenson:
I've had 2 travis builds timeout while running `test_stored_topology`:

* https://travis-ci.org/choderalab/openpathsampling/builds/48935600
* https://travis-ci.org/choderalab/openpathsampling/builds/49001096

They're not using the same worker. And other commits pass this test without problems. Any idea what's going on?

jhprinz:
That is indeed strange and I believe I have seen this too, but it was not reproducible. Will have a look. Maybe problem in the way I implemented the test because it uses several storages...

dwhswenson:
Had another failure:

* https://travis-ci.org/choderalab/openpathsampling/builds/49083367

This is with the same worker as the build 48935600 from the first example above (7a054b10). Not sure if that is significant.

jhprinz:
Me just, too. I made some modifications to the teardown and all of the tests with Storage so that there will be no open files, etc. It is strange though since there is no error and what this does is basically save a file with a topology and read it and compare. I assume that there might be some file handle that block access to a file or so. Let me know if it still occurs.

jhprinz:
Has this ever happened on your machine? It never happened on mine so I assume it has to do with memory or storage. Any other ideas?

dwhswenson:
I've never had it happen in a local test (either desktop or laptop), but it seems to be happening with nearly half the Travis tests. I assume there must be *something* different in that environment. The only thing I can think to debug it is that maybe we can `travis_retry` the nosetests, and manually check whether we ever see it fail once and then work on the retry (this would give us some idea whether it is a problem with some specific servers or whether it's something that hits randomly).

jhprinz:
It has never happend on my local machine, but you are right every 2nd or 3rd run it gets stuck at various tests. This could really be a netCDF bug.

jhprinz:
This is happening more and more. No idea what that is. We might want to disable testing for this. I have no idea why that is happening or how we can investigate what the problem is. Any ideas?

dwhswenson:
The only idea I've had so far is to add `travis_retry` to see if the problem is with a specific machines (we we could identify based on the fact that `travis_retry` won't help).

Another possibility is to enable more output on the offending tests. The `-s` option for nosetests should mean that you see `stdout`. Adding some printf-debugging to the tests that are causing problems might help you identify exactly *when* things freeze up, at least.

jhprinz:
[...]

PR #159, opened 2015-02-11
Storage test debug

dwhswenson:
Added output to test that has been giving us trouble lately, per https://github.com/choderalab/openpathsampling/issues/138#issuecomment-73869458


dwhswenson:
Okay, how is it that when I WANT this thing to crash, it won't?

I've added some print-debug statements to the test that was causing us problems. Rather than add a bunch of empty commits, I suggest merging this into master and into our working branches. When the error happens again, we can post to #138.

(This also includes a one-character fix of a bug in the ipynb test script that prevented it from registering errors in there.)

jhprinz:
:) That is a really nasty bug. Will merge

PR #162, opened 2015-03-02
Add support for openmmtools

jhprinz:
Adding a topology and engine that allows to simply use an openmmtest system.

There are still some question on how to best store information about an integrator. Right now our openmm_engine only supports VVVR.


jhprinz:
Considering that openmmtools is getting so much more features I will add a possibility to just setup your own open simulation using whatever and then plug it into OPS. This will then use the pickling from open and store the engine with it. Problem is that it is not so easy to get back to the original OpenMM simulation and see its parameters.

jhprinz:
All of this could be part of the refactor engine branch I started working on. See #197.

jhprinz:
Added some folding to travis to make the output more compact.

This is ready for review.

dwhswenson:
If this can only be used in the test systems that are included in `openmmtools`, I don't get the point of it. Wasn't the goal of this PR supposed to be that we could use the integrators from `openmmtools` with any system of our choosing (and we could remove the VVVR integrator from OPS, since we could get it from `openmmtools` instead)? That does not appear to be what it does.

If we want to use some particular model system without showing how to set it up, I think it makes more sense to reload it from a stored version. That shows one of our (coolest) features, and one that a user is likely to make use of, without introducing new things that the user won't need later. Building a system from `openmmtools.testsystems` isn't going to be relevant for the average user.

Is there some other purpose to this that I'm not seeing?

jhprinz:
> If this can only be used in the test systems that are included in openmmtools, I don't get the point of it. Wasn't the goal of this PR supposed to be that we could use the integrators from openmmtools with any system of our choosing (and we could remove the VVVR integrator from OPS, since we could get it from openmmtools instead)? That does not appear to be what it does.

No, this RP enables OPS to do two things:
1. Use ANY openmm.system and openmm.integrator with OPS. There is a new Engine that just takes your openmm system object, an integrator and as usual a template (although the template is in this case only used so that the engine has a template. All information necessary to run (topology) are already encoded in the system object -- which does NOT allow us to get back the topology unfortunately. This is completely independent of `openmmtools` which as the name suggests are just some tools to get test systems and integrators for openmm
2. Provide a tool to create a template directly from an `openmmtools.testsystem` this simplifies the case where you use openmmtools to generate a openmm.system object that you can then use with (1). To get the complete engine you use the convenience function to extract a template from the openmmtools.testsystem and pick ANY integrator either from `openmm.integrators` or from `openmmtools.integrators` which has a lot of custom integrator implementations.

To show how this works I added a notebook that shows that instead of all this complicated setting of alanine parameters you can now to the following which I would include in the alanine example should we choose the include openmmtools.
```
# imports
import openmmtools as omt
import simtk.openmm as omm
import smirk.unit as u

# pick a testsystem
testsystem = omt.testsystems.AlanineDipeptideVacuum()
# get template, system and integrator
template = paths.tools.snapshot_from_testsystem(testsystem)
system = testsystem.system
integrator = omm.VerletIntegrator(0.002 * u.picoseconds)
# create engine
engine = paths.OpenMMEngine(template, system, integrator)
```

This is as easy as it gets to set up a system like this. Only downside is that all objects are only stored as XML serialization strings which are hard to read.

That said, `openmmtools` seems to be a useful and well maintained addition when you use open with python.

> If we want to use some particular model system without showing how to set it up, I think it makes more sense to reload it from a stored version. That shows one of our (coolest) features, and one that a user is likely to make use of, without introducing new things that the user won't need later. Building a system from openmmtools.testsystems isn't going to be relevant for the average user.

Yes, I agree that loading from a stored version is better if you are interested in understanding how OPS works, but allowing to use openmmtools is probably useful for the openmm community. There seem to be several people involved and it demonstrates how well OPS is useable with openmm.

jhprinz:
I think this is ready to review, but since we will merge #300 first. I might merge 300 already in here so that we can merge faster once 300 is done.

jhprinz:
It should not load pyemma now.

dwhswenson:
I like the ability to use the integrators, but I still really don't like the extra engines. Is there a way to make `classmethod`s out of some of these, instead of completely new engines? Or maybe just get rid of them?

Should someone use `OpenMMEngine` directly, or `SimpleOpenMMEngine`, or `OpenMMToolsEngine`? It seems like we have 3 things that all do the same thing. This actually makes things ***harder*** for the user: how do you decide which to use unless you understand all of them? It violates the principle that "there should be one -- and preferably only one -- obvious way to do it." Paradox of choice. 

As I've said before, I particularly don't see the point of having a separate engine for `openmmtools.testsystems`. That looks like unnecessary complexity in our list of classes. What users does this benefit? You say that it would benefit the OpenMM community. How? The systems available in `testsystems` are limited, and most of them aren't relevant for rare events. And it looks like all that class does is wrap two functions that anyone who knows OpenMMTools well enough to want to use it should be able to figure out on their own.

Would it make sense to get rid of the old version and say "set up your simulation as you would for OpenMM" and get rid of the so-called `SimpleOpenMMEngine`? Can you set up alanine without using the `openmmtools.system` for it and without using `SimpleOpenMMEngine`? If so, I think makes more sense. Let people who know OpenMM just use OpenMM. Trying to wrap it was already weird in the way we forced a specific OpenMM integrator in our code.

Basically, my suggestions would be:
1. Definitely get rid of `OpenMMToolsEngine`. Any `openmmtools` user knows how to get `testsystem.system`. Leave `tools.snapshot_from_testsystem` (but correct the docstring, which doesn't give the correct parameter names). The `openmmtools` user can find that and use it. Then the extra class is unnecessary.
2. Probably get rid of `SimpleOpenMMEngine`. It isn't that simple, and it isn't flexible at all. Setting up a basic OpenMM system isn't that hard once you've see an example, and unless we're going to support *all* possible options, this object ends up with arbitrary and unpredictable limitations.

As for making it easier to set up OpenMM simulations for absolute beginners: I think any efforts we do will be pretty weak compared to the main OpenMM team. I think it would be better for us to point users to resources like builder.openmm.org if they're just starting with OpenMM.

jhprinz:
> Should someone use OpenMMEngine directly, or SimpleOpenMMEngine, or OpenMMToolsEngine? It seems like we have 3 things that all do the same thing. This actually makes things harder for the user: how do you decide which to use unless you understand all of them? It violates the principle that "there should be one -- and preferably only one -- obvious way to do it." Paradox of choice.

I think I understand your concern, but I think these are 3 engines for 3 cases and maybe my idea of an engine just as something that produces frames is too general.

1. `OpenMMEngine` is the engine that takes open ingredients, this is the one for experiences openmm users
2. `SimpleOpenMNEngine` is the engine we used before for illustration purposes. It also uses understandable parameters and hence the storing is way smaller and does not use the XMLSerialization. Just using a factory and 1. would remove that advantage, but it can be done without the user realizing the difference. It is not flexible, correct, but could help people get started with no knowledge about openmm.
3. `OpenMMTestEngine` is just a convenience function and does

    ```
    return OpenMMEngine(
        template=paths.tools.snapshot_from_testsystem(testsystem),
        system=testsystem.system,
        integrator=integrator,
        options=options
    )
    ```
    and nothing else. 

If you want to remove the specialties I am fine with that. If we want we can later add some class methods for convenience. For openmmtools we might leave an example notebook in the docs.

> Would it make sense to get rid of the old version and say "set up your simulation as you would for OpenMM" and get rid of the so-called SimpleOpenMMEngine? Can you set up alanine without using the openmmtools.system for it and without using SimpleOpenMMEngine? If so, I think makes more sense. Let people who know OpenMM just use OpenMM. Trying to wrap it was already weird in the way we forced a specific OpenMM integrator in our code.

I think so.

dwhswenson:
I think the code in the most recent commit looks a lot easier to understand. Assuming it passes tests, I think this is the way to go. Two small suggestions:

1. It looks like the `.create()` method can be combined with `__init__` now. Is there a reason not to do that?

2. `RandomMDEngine` is only used for testing, right? It might make more sense to move it to `test_helpers.py`, with other mocks and stubs.

jhprinz:
> It looks like the .create() method can be combined with __init__ now. Is there a reason not to do that?

Yes, there is. If you analyze you might not want to create the engine and this just saves the trouble. It could be that the engine must run on CUDA and you analyze on a different machine. It is only for efficiency after loading.

> RandomMDEngine is only used for testing, right? It might make more sense to move it to test_helpers.py, with other mocks and stubs.

Agreed.

jhprinz:
Should be ready. I would leave the create for now.

dwhswenson:
> Yes, there is. If you analyze you might not want to create the engine and this just saves the trouble. It could be that the engine must run on CUDA and you analyze on a different machine. It is only for efficiency after loading.

Hmm.. I see the point, but I wonder if there's a way around it. From a user perspective, it's weird that `OpenMMEngine` would require the `create` method while most engines wouldn't.

Is there a way to make it so that `create` doesn't get called when restoring from storage, but does during normal object creation?

jhprinz:
> Is there a way to make it so that create doesn't get called when restoring from storage, but does during normal object creation?

Yes, but I would not do that. That might be more confusing why in some cases you have to create and in some cases you don't. 

I think I had also things like lammps support on my mind and I think it makes sense to have this two step procedure in general. For ToyDynamics it is useless, but for lammps there is this step to create the actual instance. I understand your point and before we had the `initialzed` and this is the replacement.

It also reminds a user that an engine is something very complex.

I don't know. Guess this is philosophical. Engines have different parameters also, and some require a create to be used, but can be stored and loaded without it.

We can take it out for now and put it back in later. If you want I can add a `self.create()` to the init.

jhprinz:
I found a nice way. An option to turn of auto calling of `create()`. So, if you have an engine that is slow or expensive to create you can (when creating it) use the option `auto_create: False` deault it True and it will not be created automatically and you need `create()`. Default will be that this is not needed.

dwhswenson:
The `auto_create` idea sounds like a good one -- at least, it sounds like that's a way to hide the need for a `create` unless the user really wants it.

Is that intended to be a part of this PR? (Just noticing that no more commits have been pushed yet).

Another thought: would a different word than `create` be better? It seems strange to say "this engine need to be `create`d, but this other engine doesn't." Maybe `prepare` would make more sense? Or something else along those lines.

jhprinz:
I thought I had pushed it, but apparently only committed.

I have another proposal, that is even easier: If an engine provides lazy initialization. Then it will be automatically initialized the first time you call generate or run. Otherwise it will not. I remember that some tests failed because we created too many Openmm instances at a time and this way should aboid it. Do you can look in storage for engines and get their parameters, but they will only occupy your memory or GPU once you actually use them.

The auto_create flag is kind of clumsy and forces you to decide this at creation which is not really flexible and too restrictive. 

If your engine takes a long time to initialize (which should never be the case) you can alsways run `initialize` (the new name) manually before.

jhprinz:
Alright. Removed the `initialzed` property since it is not used anymore and added auto_initialize on first engine usage. In addition I added a convenience method to create a new engine out of an old one by only changing options and/or integrator.

This way you can reuse most of the engine and only change the integrator, hence the temperature or the integration steps, max_length.

jhprinz:
Ready for review

dwhswenson:
Just starting to review, but one quick and easy point: Could you correct the docstring for the OpenMMEngine class? It currently reads "This is only to allow to use the examples from openmmtools.testsystems within this framework".

dwhswenson:
This looks very good. Only two things:

1. As mentioned above, please correct the docstring in the OpenMMEngine class.
2. Is there a test to ensure that we can save and reload OpenMMEngine, and then use the reloaded version? I couldn't find one, and since this is involving the external APIs, I think it would be good to have a test like that.

jhprinz:
> Is there a test to ensure that we can save and reload OpenMMEngine, and then use the reloaded version? I couldn't find one, and since this is involving the external APIs, I think it would be good to have a test like that.

Only as a notebook, it creates an engine (using system and integrator from openmmtools), runs it. Closes the store. Loads the engine from store and runs more steps from it.

> docstring

Has been updated.

dwhswenson:
That looks good. Will merge now, with the following suggestions for near future:

* make some of the tests in that notebook into nosetests
* do we need both the openmm integration and openmmtools integration notebooks? they look very similar.

jhprinz:
> nosetests

yes, we should do that

> two notebooks...

That is a mistake. I thought I deleted the ..._intergation one. I renamed it and apparently did not correctly removed the other from the repo. Can be cleaned up anytime

PR #165, opened 2015-03-02
Fix IPython Notebook test to work with IPython 3.0.0

jhprinz:
This should make the tests work again. But we might wanna try to adapt the script and make it work with iPython 3.0.0 and eventually put it into a seperate package, since it might be useful for other projects as well.


jhprinz:
This is ready to be merged. 

### The following changes have been made:
- port to IPython 3.0.0+, does not work if a smaller version is installed
- it can load notebooks created with earlier versions. I have only tested nbversion 3 and 4 which correspond to IPython 2 and IPython 3. I guess it works also with earlier version, but I had non to test it. This means it does not matter which version you (we) have installed and is used to create the notebooks.
- Note that once IPython 3 is installed and notebooks are converted when opened and saved. They are not backwards compatible except IPython 2.4.1 should support nb.version 4.
- I renamed the file to `ipynbtest.py`
- Removed old code
- Added documentation for classes

dwhswenson:
Looks ready to merge!

PR #166, opened 2015-03-03
fix storage

jhprinz:
this tries to solve the problems with storage. See #138


jhprinz:
This should be ready to merge and stop the freezes in the `teststorage.py` nosetest by only creating one openmm instance.

PR #168, opened 2015-03-10
OpenMM tutorial

jhprinz:
removed the openmm folder and created a simple example notebook to show openmm.


jhprinz:
This can be merged since it passed. It does not change code. It only adds the ipython notebook.
We might think about adding a test for this as well.

PR #169, opened 2015-03-11
TIS Transitions and Analysis

dwhswenson:
I’ve been working a little on some tools to handle common TIS analysis steps. It turns out that the analysis is easiest if I also make a tool that makes creating a TIS calculation much easier. Now, instead of having to build all the ensembles, path movers, and so forth, you should be able to build a (default) calculation by defining a an order parameter, volumes for `stateA` and `stateB`, and a set of volumes for `interfaces`. Once you've also set up an `initial_globalstate`, a `storage_object`, and an `md_engine`, then the whole process turns into this:

```python
tis_trans = paths.TISTransition(stateA, stateB, orderparameter, interfaces)
tis_calc = paths.PathSampling(
    storage=storage_object,
    engine=md_engine,
    root_mover=tis_trans.default_movers(md_engine),
    globalstate=initial_globalstate
)
tis_calc.run(n_steps)
tis_trans.storage = storage_object
tcp = tis_traj.total_crossing_probability(method="wham")
```

Of course, you can always change the default behavior to be something fancier, but this makes it dead-simple for new users. The pseudo-script above gets us from defining the states and interfaces through having a calculated total crossing probability, and looks pretty easy.

This PR is only aiming to have things which are used for all versions of TIS. RETIS-specific features will come in a later PR.

- [x] `TISTransition` class builds ensembles and path movers
- [x] `TISTransition.default_movers(engine)`
- [x] `Histogram` class: wraps numpy histogram with some features we’ll find useful
- [x] Pathlength histograms per ensemble
- [x] Crossing probability per ensemble
- [x] Update `wham.py` from dynamiq to work in OpenPathSampling
- [x] Total crossing probability


jchodera:
Cool!

Not sure if you guys have seen [seaborn](http://stanford.edu/~mwaskom/software/seaborn/) for plotting, but it is totally awesome.

jhprinz:
You are kidding. How many commits did you do? Wow

dwhswenson:
Well, note that I've been working on this for nearly 2 months... first commit was on Jan 30.

BTW, the most recent fail is the same one that I've been seeing on my laptop when running newer versions of netCDF. Any idea what's going on there?

jhprinz:
Let me just paste here what I came up with on the train as to the discussion with Peter and David about a typical way setting up and running a simulation:

```
# Setup the system and engine
template = snapshot_from_pdb(pdbfile)
engine = Engine(template, [options])

# Create transition
states = [ define states here using Volume ]
# I think this might be optional, since the orderparameter is 
# only used to determine the interfaces, or do we need it for lambda_max
orderparameters = [ define the orderparmeter used]
# for interfaces we could also allow to use TISTransition and give the order parameter an lambda_values
interfaces = [ these are also done using Volumes]
tis transition = TISTransition(states, order parameters, interfaces, [options])

# Setup a storage
storage = Storage(filename, template=template)

# use a function from TISTransition that knows how to setup bootstrapping and run it
bootstrapping = Bootstrapping([options for bootstrapping])
initial_sampleset = tistransition.create_initial_using_bootstrapping(bootstrapping)
# we could name a simplest (not yet activated, but might be useful !?!)
# also, a sampleset knows its creating PathMover, so this sampleset 
# knows that it has been generated using a Bootstrapping PathMover
initial_sampleset.name = ‘from bootstrapping’

# Store the initial sampleset (if we wanted to, don't have to)
storage.save(initial_sampleset)

# Do some equilibration
equilibrator = tis.equilibration_scheme()
equilibrated_sampleset = equilibrator.run(initial_sampleset, steps=tis.equilibration_default_steps)
# same here as bootstrapping, although I like the naming idea
equilibrated_sampleset.name = ‘from equilibrated’

# Store the equilibrated sampleset (also optional)
storage.save(equilibrated_sampleset)

# Get the actual simulation/calculation
sampler = tis.sampling_scheme()
sampler.storage = storage

# Run 
sampler.run(equilibrated_sampleset, steps=3000)
# here the sampler would also return the last (active) sampleset and
# we could use it to run again. But here we do not need it

# [optional] computation of all missing orderparameters
[ op(storage.snapshot) for op in storage.op ]
# store the new data (if any)
storage.cv.sync()
# never wrong (but I think this is automatically done before closing)
storage.sync()
# [end optional]

# Cleanup
storage.close()
```

jhprinz:
Hmmm, no, but I will have a look. I think I know what is going on and it would only require to create a numpy int array from a list. So the fix would be easy. Still wondering why that is.

BTW. I have not yet upgraded. Maybe that is why it still works...

dwhswenson:
Problems with this proposal:

* `TISTransition` only tells about a single transition, which is not a full calculation. It can not generate a calculation. I'll be working on a `TransitionNetwork` object to handle that. 
* I'm generally against this "one-pot-cooking" approach. The steps are (1) get first trajectories, (2) equilibrate within the path sampling ensemble, (3) production run. It is not logical to put the results from all of that into the same file, in the same way that when doing MD you wouldn't make one giant trajectory with energy minimization steps, NVT equilibration, and NPT production run. It makes no sense to put all of that here -- instead, run 3 separate calculations (especially since only the last 2 are necessarily the job of OPS)
* Putting it this way assumes that bootstrapping is the approach which will be used to get the first paths. Bootstrapping works for toy models, but I wouldn't recommend it for real systems. There are many ways to get a first transition trajectory. (The approach I'm currently favoring is the one Jocelyne Vreede and I are using on a DNA conformational change: metadynamics followed by equilibration with TPS. But arguments could be made for runs based on higher temperature, or you could even make an approached based on starting with nudged elastic band. If we're lucky, we'll solve this problem once and for all with MSM-TIS!).

Basically, I would say that the starting point for path sampling is that you have some path that satisfies each ensemble. I would do the equilibration in another file (and often I just do shooting for equilibration, no replica exchange or anything fancy like that).

Anyway, this whole approach is already illustrated in the ipynb examples in this PR.

So I would put stuff into two files. In `equilibration.py`, you would input initial trajectories, and run some subset of the `tistransition.movers`. Note that this is essentially what is done in [`presentation_simple.ipynb`](http://nbviewer.ipython.org/github/dwhswenson/openpathsampling/blob/ffa273f323c4b3ee23e1a68fd1c8764dc24c52c6/examples/ipython/presentation_simple.ipynb) in this PR.

In `production.py`, you would load the final `SampleSet` from `equilibration.nc` and run using a more complete set of a movers. In an example where you're only doing A->B transitions, this would mean that you set the `PathSampling.root_mover` to `tistransition.default_movers(engine)` (probably during initialization of the `PathSampling` object). But the point is that `TISTransition` gives you the tools to handle ONE transition; more than one transition requires either manual intervention or another object (the `TransitionNetwork` object I'm proposing.)

Another note: When presenting this stuff, I **strongly** recommend skipping the bootstrapping. Explaining it takes time, and people mistakenly think that bootstrapping is important and then they miss the real point, which is the path sampling part.

jhprinz:
Totally agree except a little on the storage, which I think should be the recommended case but I personally would rather store everything in a file to make sure I can 6 months later still check what happened. This is mostly due to me not being so good of taking care of old files. Still, I think recommended and safer would be to split files as you do in the example. This is also how you should setup a normal MD simulation. Prepare everything and then run only the production.

Bootstrapping was in there just as an example how the use of bootstrapping could look like. It was not meant to be the way it has to be. Since it was the script we always used I just _converted_ it into this example. Of course the are other ways to get a first trajectory. 

I assume the use of _typical way to setup_ was misleading :)

Okay, so Bootstrapping is not that common, agreed. Loading is much simpler although you probably need to know how to get the first trajectory then. This can be tricky... 

I really like the `TransitionNetwork` as the more general term. We might think about separating the general network of states, like the 3 states for alanine and all the other stuff need for a specific type of sampling Technique. That would mean that one general object is `TransitionNetwork` and `ABTransition` is a subclass. Then `TISTransition` is not a subclass of transition, but a factory where you give a TransitionNetwork with 2 states. That means that TISTransition and MSTISTransition is the same object and just depends on the number of states.

Does that make sense? So you could have the same `Transition` object but different ways to sample it. And then apply a FFSFactory and get an FFSSampling (or the multistate version if that even makes sense ?!?)

An example what I imagine something like this (on the fly) for TIS

```
a2b = Transition(stateA, stateB)
# one possibility
tis_scheme = TISSampling.create_using_cv(
    transition=a2b,
    orderparameter=my_op,
    interface_lambdas=[0.20,0.25,0.30,0.35,0.40]
)
# other more flexible
tis_scheme = TISSampling.create_using_interfaces(
    transition=a2b,
    interfaces=[intfc_1, intfc_2, intfc_3, intfc_4, intfc_5]
)

# finally run 
# idea #1
# in this case tis_scheme is of type TISSampling which is a subclass of PathSampling
tis_scheme.storage = ...
tis_scheme.engine = ...
tis_scheme.run(active_state, steps) # where steps has a default so it can be omitted

# idea #2
# use PathSampling as the runner
sampling = PathSampling(
    storage=...,
    engine=...,
    root_mover=tis_scheme
)
sampling.run(active_state, steps)

# idea #3
# use a Sampler which can be reused
sampler = PathSampler(
    storage=...,
    engine=...
)
sampler(scheme=tis_scheme).run(active_state=initial_sampleset, steps=5000)
# here sampler() would create the PathSampling object
```

There are more possibilities and it is again a philosophical question of how to best reflect the simple things in the code.

Perhaps this is too much new objects, but #3 reflects how I think about doing this.
1. Transition : Only the states
2. Scheme : Add Interfaces etc and how to sample it (PathMover) which is usually coupled (is it?)
3. Sampler : Knows Storage, Engine, other details independent of Sampling Scheme
4. PathSampler : Thing that can run MovePaths using engine and Store results

This is only my typical `Jan need to think loudly about this in public`. So nothing is meant as : We have to do it this way. Just seom food for thought. We are so close to the finish line that I can almost smell it :)

jchodera:
> Totally agree except a little on the storage, which I think should be the recommended case but I personally would rather store everything in a file to make sure I can 6 months later still check what happened. This is mostly due to me not being so good of taking care of old files. Still, I think recommended and safer would be to split files as you do in the example. This is also how you should setup a normal MD simulation. Prepare everything and then run only the production.

I just wanted to chime in to say that it's important to store critical information about the *provenance* of the data along with the data!  Even if you need to store things twice this way, it's important to have all of the input data needed to generate the output data *stored along with the output data* so that it's possible to understand where the data came from (for exactly the reason @jhprinz described).

dwhswenson:
This now calculates a total crossing probability, using WHAM. However, the resulting TCP is obviously not right. But, the individual crossing probabilities also don't look right, so I think the problem isn't with WHAM but with garbage-in-garbage-out. Testing this modified WHAM script against calculations from previous work show the same results as before, so I think the WHAM part works.

@jhprinz : if you want access to `Transition` stuff quickly (like, for showing people in Vienna), feel free to merge (if Travis okays it). Otherwise, I'll probably add path mover acceptance ratios and maybe run a longer test of this toy model to see if more data will clean up the messy crossing probability.

jhprinz:
Strange. If you did not merge the fix I would know what causes this, but so far I am a little puzzled. The generated trajectories look fine and the computed order parameters are too. So why does it not look right? Weird, I will wait for you getting more results and then see, what is wrong.

dwhswenson:
With a little more data, it looks a bit better. Even this isn't fully sampled. However, it is using TIS, not RETIS: RETIS would probably fix the annoying little bump in this version of the total crossing probability.

I'll call it ready for review and merge, and I'll complete mover acceptance ratio calculations in a later PR.

jhprinz:
Looks good. Will merge this now. All the stuff below is not urgent and just some ideas and/or suggestions.

A few questions I have though:
```
def sample_generator(samples):
    i=0
    while i<len(samples):
        yield samples[i]
        i = i+1
```
I think
```
sample_generator(samples) = iter(samples)
``` 
or if you want to have a another name 
```
sample_generator = iter
```

What does `get_n_samples` do? It seems not to be used? Can it be deleted?

`pathlength` could be added to `Sample` as `__len__`. As could be done for other `Trajectory` properties, e.g. `__iter__, __slice__, __getitem__`. We could also delegate all functions to trajectory that cannot be found. This way we would only have to implement features like `max_lambdas` only for trajectories.

```
    JHP, please don't remove this! You created a loop where we need
    MoveDetails to make a sample, and we need a sample to make
    MoveDetails. It makes more sense to have THIS be underlying, and use
    initialization_from_sample when appropriate.

    THIS IS USED BY Sample.initial_sample
```

??? I think I do not understand the reason for this comment... I think originally I used this trick to keep some code that was there before and not change too much... Also this is part of legacy when we used to use trajectories and not samples... Anyway, I would remove the comment...

dwhswenson:
Both `sample_generator` and `get_n_samples` can be removed: they were put in place before I knew how I'd be able to get samples out of the storage. Hypothetically, `get_n_samples` was designed to take the generator and create a buffer of a certain length, because `numpy.histogram` doesn't take generators. However, I think the better solution is to replace `numpy.histogram` with something like the arbitrary-dimensional histogram code I wrote (in C++) a few years ago for a side project. That algorithm will allow generators. Instead of defining the histogram bounds by the number of bins, it defines it by the bin width. In my experience, this makes more sense: I can more often guess what resolution I need on the data than what the range of data will be (e.g., pathlength: resolution of 5 steps per bin is probably sufficient under most cases, but I have no idea what the max pathlength will be, and therefore I have no idea how many bins I need to have that resolution).

> pathlength could be added to Sample as `__len__`. As could be done for other Trajectory properties, e.g. `__iter__`, `__slice__`, `__getitem__`. We could also delegate all functions to trajectory that cannot be found. This way we would only have to implement features like max_lambdas only for trajectories.

No, don't do this. It would only make things more confusing. Let the `Sample` and `Trajectory` be distinct. If a user is developing new analysis methods, they need to know that a `Sample` consists of `trajectory`, `ensemble`, etc. Abstracting that out will only confuse people and creates too many paths to the same function.

In any case, *some* function needs to be used by the histogrammer (see `TISTransition.ensemble_histogram_info` in its `__init__`). If the user wants to histogram some other property of the trajectory, they now see two examples: `pathlength` and `max_lambda`, and both examples are done with the same steps: (1) write a function to do it per sample; (2) create the appropriate histogrammer. The example with `pathlength` requires no additional arguments to the function, whereas the example with `max_lambda` does.


> ??? I think I do not understand the reason for this comment... I think originally I used this trick to keep some code that was there before and not change too much... Also this is part of legacy when we used to use trajectories and not samples...

You had removed this. Or more accurately, you'd changed `MoveDetails.initialization` to ONLY work with a `Sample` as input, instead of offering the ability to create from a `trajectory` and an `ensemble`. Consider how we really start a calculation: we get a transition trajectory from somewhere, and use that as the initial condition for all the ensembles. This means that we do not start with any `Sample` objects. So the problem is this:

```python
first_sample = Sample(
    trajectory=traj,
    ensemble=ens,
    replica=rep,
    details=MoveDetails.initialization(WHAT_DO_I_PUT_HERE)
)
```

`WHAT_DO_I_PUT_HERE` has to be a `Sample`, but no `Sample` exists. This is the loop I’m referring to. I re-added the old function so that line could become

```
details=MoveDetails.initialization_from_scratch(trajectory, ensemble)
```

I put the comment there so you wouldn’t change it, thinking it redundant. When I’d put the original function in there, I specifically intended it to take `trajectory` and `ensemble` because I knew we’d need that functionality later. Since you changed it before, I wanted to be sure you’d know there was a reason for it to be that way. But now that you do, I can remove the comment!

Being able to create move details from non-sample sources like this is particularly important if we're going to use the details object to handle John's concerns about provenance for these initial trajectories, which I believe we had decided was the best way to do that.

jhprinz:
Just  quick reply because it was very late yesterday when I wrote the comments after trying to fix the problems with the number of trajectories :)

Agreed, Samples should be Samples and not Trajectories. At some point though we could even subclass Trajectory and call it a Sample, but keeping it separate made it also easier to explain it. 

`sample_generator` and `get_n_samples` : Good, that was my impression. I just wasn`t sure if it is necessary for other branches, etc...

Sorry for the short comment. You are totally right that this was bad style. I remember that I need to do some tricks to get this to work. Not even sure why I did this in the first place. It is quite a while ago.
I was mostly confused about the long comment and thought you should have just fixed it and mention it here in github. Anyway, good work!

PR #170, opened 2015-03-12
Fastquery

jhprinz:
This PR adds delayed loading of frames in a trajectory. Now if you load a trajectory it only loads the snapshot IDs which is fast and only if you access a single frame it loads the frame from storage and create the full snapshot object.

Internally this creates a tuple `(store, idx)` instead of a `Snapshot()` instance. Which will be interpreted correctly by all list operations applied to trajectory.
Not that of course the iterator also starts loading all frames one by one unless the snapshots have been cached.

One particular problem right now:
We are still using orderparameters which are based on `Configuration`. And to get the configuration of a snapshot, it needs to be loaded so in the current use of orderparameters you will not see any benefit (which means that the fall back works). To access all values in an orderparameter fast, you have to use an `Orderparameter` that is based on `Snapshot` (which on the other hand means, we might have little redundancy, but there is no elegant way around it.

Thus, in this PR I will change the default behaviour of Orderparameters to `Snapshot`.


dwhswenson:
Looking forward to merging this into #169: currently, making the path length histogram there (for a 10k MC step calculation) takes about 5 minutes. From our testing, this will mean it takes a few seconds.

I think we want order parameters to be based on snapshots anyway: we've used state definitions which included velocities before (see the first paper on TIS, where a velocity aspect of the state definition is used to avoid recrossing in an NVE simulation). I don't see how this creates unnecessary redundancy: if all order parameters are defined as mapping a snapshot (a point in the total phase space) to a resulting number, this covers all cases. The fact that it makes analysis faster makes it even better!

jhprinz:
Almost ready. Just doing some more testing.
The only redundancy is, that if our OP depends only on the configuration then we can have several snapshots that have the same OP but different velocities and it needs to be computed again.

We could however cheat and add the configuration per snapshot as an additional "hidden" order parameter which would solve this, but I don't see the point.

Bottomline is, the new framework is MUCH faster, but you have to rerun your simulation since the OP is stored referenceing the configuration not the snapshot ID.

dwhswenson:
Ah, I see what you're saying. Yes, we do take a little performance hit on the analysis side, but only when doing two-way shooting, and even there probably only something like a maximum of 1-5% (1 / avg traj length), and again, only on the analysis side. Barring really hard order parameters, this won't cost much for real dynamics.

So theoretically wasteful, but practically, who cares, right?

jhprinz:
Ready to merge and passed all tests...

dwhswenson:
Unclear to me: is usage such that we still use the old usage, and this implementation just avoids loading unnecessary things? (In particular, this looks like it hasn't implemented the query function we discussed yesterday; is that to be a separate PR?) This is a drop-in replacement so that when I call `len(storage.trajectory[n])` it doesn't load all the configurations, right? I don't have to do anything else?

Also, please correct docstrings. I notice in particular the `ConfigurationVariable`, `SnapshotVariable`, and `OrderParameter` have the same description. Is there even a difference between those last two? Check that any other sections where you copy-pasted don't have misleading docstrings.

I think there should be a better way -- or a least a better name -- than all the `._is_instance` type stuff, where you make a hidden method of a standard function. Try this:

```python
def item_type(item):
    """Gets type for orderparameter: if tuple, assumes element 0 holds content_class info"""
    if type(item) is tuple:
        return item[0].content_class
    else:
        return type(item)
```
(Not sure that `type(item)` is exactly right for the second return, but there should be something along those lines.)


Then you start `__call__` with:

```python
i_type = item_type(items)
if i_type is paths.Snapshot:
    # etc
```

This way you can remove all these `_is_instance`, etc, all of which are methods that should be static but aren't. Just create a stand-alone function that removes the need.

You should be able to similarly simplify `_compatible` with this. The approach is to regularize the inputs to a single format, instead of enumerating all possible formats.

jhprinz:
Sorry about the confusion.

Yes, everything stays the same. I have not changed any code for the tests, etc. Except for some clearing up. That means we should not need the function we talked about, although it might still be helpful to have it. So I will still implement it, but the rest should work.

I also tried not to change the functionality in `trajectory.py` so this stays the same. Only the trajectory store imposes some changes on the list functions like `__getitem__` and `__getslice__`. I also needed to add a new `__iter__` to be compatible. All the rest stays the same.

Next step was to adapt `order parameter.py` to be fast and use the changed trajectory. The only changes are in the StorableObjectDict, this is the part that knows about the _fake_ snapshots. The rest again stays the same. 

The little clumsy writing with isinstance is because I decided at some point in the past to allow for subclassed objects. Snapshots and Configurations  are not subclasses (so far) but other objects can be (e.g. Ensembles) and here type fails, so I wrote a replacement for isinstance which works also for subclassed items. Unfortunately it needs an instance, that is why the `_isinstance` looks to strange. 

You are definitely right that it could look simpler. Maybe we should not allow subclassed objects and just keep it brief. What do you think?

I have removed the `SnapshotVariable` which was indeed just a copy.

Docstrings will be adapted. Sorry for that.

dwhswenson:
The confusion was mine: this PR does exactly what you described it as doing. Still, no harm in checking before I start using it!

The difference with the query function we talked about is that it should only return samples which were saved by a SampleSet. The current stuff just looks through all samples in storage, right?

On all this `_isinstance` stuff: does the fastquery version work with subclasses? (sorry I can't browse code very well: this is coming from my phone.)

If that does handle subclasses, then is there a way to change the part that I suggested should return `type(item)` so that it returns the same thing as the `item[0].content_class` returns? This would regularize it.

jhprinz:
This should be ready to merge (when it hopefully passes soon -- just did).
I simplifies the item_type and used the base_cls property.
Also the `storage.cv.sync()` should work now

PR #174, opened 2015-03-13
Visualization for MovePath

jhprinz:
Adaption of the PathTree Layout to show what happens in the simulation.


jhprinz:
Here is an example picture for the `toy_tis.nc` from the toy dynamics example notebook
movepath

jhprinz:
This should be ready to merge. It contains an example notebook to show how to get networkx graph from a mover, do some simple D3 visualization and the old PathTree and MoveTree stuff. 

It adds a `submovers` property to PathMovers that contains a list of all submovers used by a pathmover. 
in addition it uses the same tree functions as PathMoveChange. In the larger PMC update #251 I will try to put the tree functionality into a mixin to keep the code better maintainable.

if it passes it should be ready to merge.

dwhswenson:
Very much in support of moving the tree functionality into a mix-in (with the understanding that you're planning that as part of #251, I'll merge this once it passes tests). As it stands, the tree functionality in `PathMover` makes the basic `PathMover` object pretty hard to understand, which I fear will scare off contributors. Something like 1/4 of the `pathmover.py` file is currently devoted to maintaining the new additions of tree-visualization-related information. Is that all really necessary? I would think you could construct the `PathMover` decision tree with a few `hasattr` checks.

jhprinz:
Exactly. This will be part of #251. The tree mix-in will just allow to loop over the whole tree and these sort of things. This is more useful for `PathMoveChange` which is really a tree, while the `PathMover` itself is more a way to describe an ensemble of possible `PathMoveChanges`... At least kind of. So iterating over a `PathMover` is mostly helpful for visualization.
The different branching in the tree for RandomChoice, Sequential, PartialAcceptance, etc, leads to different sets of possible subtrees. If we add some functionality we might extract some information about probabilities that a certain mover is called, etc... But that is not yet so important.

dwhswenson:
Now that it passes tests, merging!

jhprinz:
yes, took a while for me to figure out that networkx was missing from the installed packages :)

dwhswenson:
Yeah, I had also added it as part of #250. But since it was imported from a module, not from a notebook, it was pretty obvious what was wrong.

Fortunately, it looks like we both added it to the same point on the list, so we shouldn't have any weird stuff (like trying to install it twice) when they merge!

PR #175, opened 2015-03-17
Switch conda to https following JHP's fix

### LINKED ISSUE #173: Anaconda MD5 uses now https://
Haven't gotten around to fix that in all PRs, but Anaconda require now https:// for the MD5 and (I think) also to download Miniconda.

just change the two line in `install.sh` and it works. See #170 for an example.

dwhswenson:
Resolves #173 

Quick fix here to be merged in so it can be merged into all branches.


jhprinz:
Independent of other error and fixes what it should, so I merge to make the other PR run

PR #177, opened 2015-03-17
fixed delayed loading from multiple storages

jhprinz:
this updates cloning to all the new things to clone and
fixes a bug when using delayed loading with two storages


jhprinz:
Finally. Found a last stupid bug that made it slow again. It had to do with automatic decision when to safe which in some cases will run a safe all the time although the snapshots are not safed yet. ...

Long story. I hope this will build now and everything is now very flexible and buffering, intelligent caching etc can be added. FOr the user everythings looks as before. 

Let's cross fingers...

PR #178, opened 2015-03-20
Fastquery

jhprinz:
Much more speedup. I found some things that made the storage much faster.
The default behaviour is now that we only keep orderparameters cached after a storage to disk for all snapshot that are in the storage. 

That means we assume that before syncing we store all snapshots we think are worth to be saved.
After that the order parameter cache knows that only the saved ones should be kept in memory for caching. Especially minus moves generate lots of temporary order parameters and snapshots that we will ignore. As an example, I was running the old toy_dynamics example with minus_mover equally likely as the other movers. This resulted in 200,000 cached order parameters while only 20,000 were saved!
Also this means that the garbage collection will now remove the samples, since they are not referenced anywhere. The minus mover should cut these out.

The result is, that I ran 300 steps for the TIS production in about 30s(10/second) compared to almost 900 (1 every 3 seconds) before. This is of course only so obvious for the toy system, while for real systems the shooting will be dominant.

As a last thing, I think the new bottleneck is the can_append for the SequentialEnsemble, but we should investiage that a little more. But this is less a problem that the storage was!

If it passes it should be ready to merge. Enjoy the blazingly fast store now!


jhprinz:
Oh no... The build fails because of the new netCDF version...

They have changed the way you can select indices/slices. First simple lists of integers stopped working and now it has to be a numpy 1D integer array that is sorted and does not contain any duplicate values.
This is stupid since these operations are much slower done in python itself... Too bad...

I can fix this, but potentially not before Sunday... So long...

dwhswenson:
Do we want to force conda to use the old version of netcdf for now, so this can get merged in, and then you can fix this later with a little less rush?

jhprinz:
Yes that might do it for now. It is not really a big change, but it requires some more testing since the documentation seems to not reflect the changes.

jhprinz:
This is ready to merge and adds the some more features and bug fixes. Check, if it passes. Just added 2 more nice features.

- Loading and saving of trajectories is now consistently fast
- Use of order parameters is much faster. Storing and Loading. and uses `(snapshot_store, idx)` instead of real `Snapshot()` objects 
- `storage.snapshot.all()` returns a shallow trajectory that refers to all snapshots that means to get all order parameter one can use either  `op(storage.snapshot.all())` which is really fast or `op(storage.snapshot)` which will also load all snapshots. The second variant will also load all 
- a trajectory will use cached snapshots even in shallow load, so a loaded Trajectory can consist of mixed Snapshot types
- this also fixes problems with netCDF 1.1.6 which requires that no index int lists are empty, unordered or not unique.

### Still possible improvements, but skipped for now
1. If you load a shallow copy (only tuple of store and idx) and use it with an order parameter it will be loaded and cached. If you now load the snapshot with the idx and use that one with the OP the value is theoretically already cached, but right now will be loaded again. I think we can live with the little overhead for now.
2. The other way round can also happen - first cached a snapshot object and then the shallow version is required. To reuse this information would require to scan the cache for an object that has this `idx` which is costly. So I do not recommend it.

PR #181, opened 2015-03-23
Important FIX: Wrong order or orderparameters

jhprinz:
Due to the new netCDF sorted indices requirement I had this fixed, but apparently my test was not able to see, that the back-ordering after requesting from the netCDF file to the original index order was permuting in the wrong direction. This one fixes it.


jhprinz:
query store is also adapted to the new netCDF stuff

PR #183, opened 2015-03-25
[WIP] RETISTransition

dwhswenson:
Adds more features needed to handle RETIS. Extends `TISTransition`.

Tasks:
- [x] Create minus ensemble, minus move
- [ ] Storage for `RETISTransition`
- [x] `RETISTransition.default_movers()`
- [x] Clean up replica exchange
- [ ] `RETISTransition.minus_move_flux()`
- [ ] `RETISTransition.rate()`
- [ ] `RETISTransition.replica_flow()`
- [ ] Add `PathMover` acceptance ratio analysis

This still needs to be followed by two significant improvements to the analysis side: (1) need to add support for networks; (2) need to add block averaging to put errors on the calculations.


dwhswenson:
@jhprinz : The most recent commit of the `toy_tis.ipynb` (c856768) raises an error in storage when I run it. My guess is that this might be because the order parameter which was previously used in the equilibration stage (and stored in the equilibration file) is also used in the production stage, which uses a different file for storage.

Suggestions on how to fix?

jhprinz:
@dwhswenson Looking at this now...

jhprinz:
Hmmm, took a little while but I fixed and simplified the new OP stuff and it should work fine with the PR now. It will fix the linear increase and handle some other OP related loading and saving more efficiently.

jhprinz:
We might merge this before all of the points are done. In particular I can take care of the Transition storage!

Issue #182, opened 2015-03-25
Storage has linearly increasing time with number of trajectories


dwhswenson:
See ipynb in gist https://gist.github.com/dwhswenson/0264e84fccc1414c313e for details. Note that the ipynb there uses the `TISTransition` class, so you'll need a branch with that in it (since that hasn't been merged to master yet).

jhprinz:
This is no bug, just not the most efficient caching. I will add a PR to improve this.

dwhswenson:
This has been fixed by #184, which was merged as a part of #201, and therefore can be closed, right?

jhprinz:
Yes! Juhuu (german for Hooray!)

PR #184, opened 2015-03-26
[WIP] Improve inefficient saving of orderparameters

jhprinz:
Storing of orderparameters with lots of cached values needed to filter for potential not storable cached values. This is now done when adding new computed values and only once. This removes the linear increase in Issue #182.

Still adding some more tests.


jhprinz:
No need to merge this. Will be fixed when #189 (on @dwhswenson s fork) is merged!

PR #188, opened 2015-03-28
Toy TIS separation; TISTransition cleanup

### LINKED ISSUE #186: Separate `toy_dynamics_tis`
In a real calculation, this would be split into (at least) 3 stages: (1) get initial transition trajectory; (2) equilibration; (3) production run.

I'm suggesting that we split this into at least two files: `toy_bootstrapping` and `toy_tis`. The first covers the first part, the second covers parts 2 and 3. A few advantages to this:

1. Saving the bootstrapping result instead of recalculating each time simplifies debugging in the notebooks -- indeed, that part of the calculation should be pretty much done.

2. Explaining bootstrapping is an unnecessary complication when introducing this to novices. You get the first transition trajectory by hook or by crook; bootstrapping is just one of many options. Splitting this lets us focus on the path sampling aspects.

I'm planning to make this split part of #183 (because I want this for development there), unless there are objections. My initial thought is that it would look like this:

#### `toy_bootstrapping.ipynb`

* Get first trajectory without saving it
* Use storage for the entirety of the bootstrapping procedure (`toy_bootstrap.nc`)
* Add in some of the move history visualization stuff (hoping @jhprinz can do that part later; I won't try to add it)

#### `toy_tis.ipynb`

* Load last trajectory of last ensemble from `toy_bootstrap.nc`, use that as initial condition for all ensembles (also mention how you could load the entire sample set, and how that might be better in this case, but use the single-trajectory approach because that's more general)
* Run the equilibration (shooting and path reversal only), saving to `toy_equil.nc`
* Run the production calculation (includes replica exchange), saving to `toy_tis.nc`

#### `toy_analysis.ipynb`

* Load everything from `toy_tis.nc` and do the analysis: path histograms, crossing probabilities, and replica flow within the interface set.

This will make testing and development a lot easier. Note that we'll still need more notebooks to describe how to do things at the network level: probably one for the simulation and one for the analysis. In practice, I think the recommended thing will be to start with networks (the 2-state version MSTIS network is probably best even if only doing a 2-state problem, at least if you want to get free energies, etc.)

dwhswenson:
This PR includes stuff that was originally going to be part of #183, but I thought @jhprinz might prefer to have this merged in sooner.

Includes some cleanup after #169, and separates `toy_dynamics_tis.ipynb` into two notebooks, which closes #186.

Should be ready to merge; no real new features here, just some clean-up.


jhprinz:
Looks good to me... Merging now...

PR #192, opened 2015-03-30
Rename n [rename merge no. #1]

jhprinz:
- [x] Rename everything that as todo with numbers.
I have also changed that mostly
- [x]  `len()` is used instead of `n_snapshots`, etc... which is more pythonic and also mdtraj allows this `len(traj) == traj.n_frames`
- [x] AllIn, AllOut, etc...
- [x] Union, Intersection, etc...


jhprinz:
Should be ready to merge. The only thing is that loading by string only works with the .uid property and not the name anymore. That is the downside. We might think about this but still, the name is not unique and not so suitable for loading.

jhprinz:
fixes conflicts in notebooks. Can be merged now!

dwhswenson:
One question: didn't we want to keep the `X` indicating a volume? I think that is clearer. `AllInEnsemble` isn't clearly a volume ensemble; it could be all the trajectories that are in some ensemble. So I think `AllInXEnsemble` to be brief or the more long-winded `AllInVolumeEnsemble` would be clearer.

jhprinz:
Sure. I like both so I changed it. Should pass now.

dwhswenson:
Merging, although `ensemble.rst` needs to be changed to `InX` from `InXX` now :-)

PR #193, opened 2015-03-30
Rename calculation [rename merge no. #2]

jhprinz:
Renames `Calculation` to `PathSimulator`

Also renames all other usage of a `calculation` variable


jhprinz:
THis should work except for wrong uid setting in the notebooks. This happened because notebook overwrite themselves when they are still open. So switching between branches can screw these up (and this cost me at least 2 hours of additional testing today)

jhprinz:
updated here the nb. Let's hope this works now :)

jhprinz:
If it passes, this is also ready to merge

jhprinz:
damn, this was not meant for this commit ...

dwhswenson:
Put it on the wrong branch? Use `git cherry-pick` to apply it to the right place.

jhprinz:
This is also ready to be merged. Merge after rename n.

jhprinz:
Ready to be inspected.

dwhswenson:
Please do a read-through of the changes to see cases where "calculation" was used in a context other than the object which is now `PathSimulator`. In a lot of cases, you might replace it with "simulation" instead of "calculation", but "pathsimulator" (especially lowercase) doesn't make sense. This shows up in text (including in ipython notebooks) and in docstrings.

jhprinz:
This should be better now. I fear we need a final pass through after the code is ready. Especially all the ipython notebooks change a lot due to the new simpler `Transition` class and splitting. There are also a lot a notebooks that can be deleted...

Anyway. I think the last commit should have fixed the wrong renames.

dwhswenson:
A few more to fix:

#### `toy_tis.ipynb`

* l. 92 (formerly l. 98): This is not about `PathSimulator` objects; this is about the general concept of path sampling as a simulation approach
* l. 178 (formerly l. 190): again, this is about the simulation being set up. There exists no `PathSimulator` yet -- the concept hasn't been introduced. Using the word there is wrong.

#### `histogram.py`

* l. 14 : The `histogram.py` script is completely independent of OpenPathSampling. It should have no reference to `PathSimulator`s.

#### `tis_analysis.py`

* l. 173: flux and rate are results of analysis calculations, not results from pathsimulators.
* l. 236: nonsensical with "pathsimulators"
* l. 305: nonsensical with "pathsimulator"

A few other problems in the attic; we should just clean the attic.

Anyway, better to fix these now, when they're easy to find in the differences.

jhprinz:
Actually, I have not started to read through all files. The whole renaming and refactoring including merge issues and fixing notebooks took about 8h. And since my english might require another pass-through in general ;) , I thought to be more efficient not to spend fixing everything. 

In this case I think I used the wrong find meaning I did not catch all `pathsimulators`. Anyway, you don't need to explain why something is wrong. I think there is better use for your time :) If you see something just tell me the line. I just realized that there is a little glitch in PyCharm and you cannot find a word when searching for it. It displays the rendered notebook and so you cannot search for line numbers...

You are right. We should remove the attic at some point. Thanks for the help!

dwhswenson:
As soon as those changes are made, and this can be automatically merged, I think we're ready for this one. I'm hoping we can get through the renaming PRs today.

jhprinz:
Oh no, I think the merge of your PR caused this. Will fix this now.

dwhswenson:
Looks good now! Merging.

PR #195, opened 2015-03-30
Rename Orderparameter [rename merge no. #3]

jhprinz:
Orderparameter is changed to CollectiveVariable


jhprinz:
if it passes this is ready to be merged after rename calculation.

jhprinz:
Ready to be inspected.

jhprinz:
This has now the corrections in it from wrong calculation changes.

dwhswenson:
A few small changes where we really do mean specifically `orderparameter` and not general `collectivevariable`:

* `tis_analysis.py` everything except `OP_Function`->`CV_Function` on l. 24 should be reverted (the order parameter in here is the real TIS order parameter)

The rest are just a result of changing the above:

* `toy_analysis.ipynb` l. 133
* `toy_bootstrapping.ipynb` l. 286
* `toy_tis.ipynb` l.192, 261

jhprinz:
Removed the toy_storage.ipynb which is indeed redundant and was only used for testing at some point.

jhprinz:
This should be ready to merge. Just added the latest PR.

PR #196, opened 2015-03-30
Rename movepath [rename merge no. #4]

jhprinz:
### Renaming
- Rename MovePath to `PathMoveChange`
- Rename stores to use plural `trajectory` -> `trajectories`

### Visualization
- fixes visualization of PathTrees. These should work again and the new sample attributes help here a lot. 

### ChainDict
- fixes a bug in `chaindict` that could lead to wrong results when several storages were used and only some of these stored a CV
- makes `chaindict` a little faster for writing and loading
- there should be now lots of obsolete code in `chaindict` but I leave that for now

### PathMoveChange and PathMovers
- These have been reorganised as was discussed in PR #210.
- There are now things attached to PathMovers and Samples. Both are accessed using details
- Samples have more information, like `.accepted`, `.mover`, `.parent`, etc...
- PathMovers can now be queried for their tree structure using `__contains__`
- Samples load faster and use delayed loading of rarely accessed attributes
- `__iter__` / `__reversed__` will traverse the PathMoveTree in order.
- MoveDetails will be attached to MovePathChanges and not Samples!
- change all tests for PathMovers and Shooters to check PathMoveChanges and not the Samples.
- removes lots of unnecessary initialization of Samples with empty Details
- PathMoveChanges are now trees and will be treated as such.
- `__getitem__(self, n)` will select the n_th subchange
- `__len__` will give the number of all underlying nodes (not the number of subcanges), this will be given by len(pm.subchanges)
- `.subchange` is a property that returns the first sub change for conveniene and backwards portability


jhprinz:
This should be ready, but the tis_ notebooks need to be adapted and had been corrupted a little from other PRs. Battery is empty now and will leave it this way. The other rename PR can be merged!

jhprinz:
Ready to be inspected.

jhprinz:
Adding the fixes changed a notebook back and I had to reemerge some code at the end. This is mainly the examples on how to extract some data using the only `sample.detail` which does not work anymore.

jhprinz:
Ready to be inspected. Will keep on working on the `movepath.key` idea...

jhprinz:
still adding more functionality to query easily in PathMoveChange trees.

jhprinz:
So, enough for today. I think this should be ready. I might move some more stuff into `.details` to speed up loading. Right now loading of `Samples` is not as fast as I would it like to be. Let's see if if passes.

jhprinz:
So, finally. This should work now. Fixed a bug in chaindict. Was not so severe after all... (Just took too much time to find it)

jhprinz:
So, finally. All tests passed on my computer and so far travis looks good. I also checked the storage view.py to make sure that all data is as expected. In particular CVs.

jhprinz:
Ready to be merged.

dwhswenson:
Woohoo! Looks good from the quick review I could give (52 files changed is a *lot*). Merging.

PR #197, opened 2015-03-30
[WIP] Refactor engine

jhprinz:
Moves all engine stuff into separate package


jhprinz:
I think this is totally outdated and needs to be started from scratch. There is no way to fix all merge conflicts. But I can still post some ideas. 

In this I had an engine package structure like this:
```
├── dynamics
│   ├── __init__.py
│   ├── dynamics_engine.py
│   ├── external
│   ├── openmm
│   │   ├── __init__.py
│   │   ├── openmm_engine.py
│   │   ├── topology.py
│   │   └── util.py
│   ├── topology.py
│   └── toy
│       ├── __init__.py
│       ├── engine.py
│       ├── integrators.py
│       ├── pes.py
│       ├── topology.py
│       └── util.py
```

It's called dynamicsengine and has some shared files and then engine specific subfolders to clearly separate which engine you want to use.

dwhswenson:
I like the idea of that basic structure (and it should be pretty easy to redo the structure in a new branch, right? Much easier than trying to merge changed files into this). Even just the file structure could help contributors who want to contribute new engines, and wouldn't the use of subpackages also speed up our main import? (My suggestion would be that `DynamicsEngine` is `paths.DynamicsEngine`, same with `Topology`, but all engine-specific stuff requires a separate import of that engine's subpackage.)

I'm just trying to figure out how this should look in practice.

I know we've talked about whether to change the `OpenMMEngine` class to something just separated by the namespace: `openpathsampling.dynamics.openmm.Engine` ... I like the way that reads, but that's outweighed by my concern that there's the risk of someone doing too many `from package import *` commands and mixing the namespaces (of course, this is *exactly* why you shouldn't use that form of import, but you know that some novice user will).

I can't think of a way around that, and I'm not sure whether there even *is* a solution. 

I also would prefer to keep the *real* names of the classes a little clearer. This helps with readability, especially so new developers can easily see which kind of `Engine` this is. One trick we could use is something like `Engine = OpenMMEngine` in the `openmm/__init__.py`: that way `openmm.Engine` can be the recommended way to get the `OpenMMEngine` class.

jhprinz:
Sorry for the brevity. I just pasted this yesterday so I don't forget about it.

> I like the idea of that basic structure (and it should be pretty easy to redo the structure in a new branch, right?

Yes, that is just shifting files and classes and fixing imports :)

> Even just the file structure could help contributors who want to contribute new engines, and wouldn't the use of subpackages also speed up our main import? (My suggestion would be that DynamicsEngine is paths.DynamicsEngine, same with Topology, but all engine-specific stuff requires a separate import of that engine's subpackage.)

Yes, again. That would be great. My idea was initially to just write
```
import openpathsampling.engine.openmm as engine
```

To prevent the problem of `from package import *` we could exclude engines from `openpathsampling/__init__.py`.

The problem we have is with storage, because we would need to add information about the package the engine instance is from and not just a class name. As long it would be okay that you can only load classes from packages that have been imported. So you can only load a Storage that contains OpenMMEngines if you have imported using `import ops.engine.openmm as omm` or similar. Importing the class will add the special engines and since they all derive from `OPSObject` you should be able to load these too.

> I can't think of a way around that, and I'm not sure whether there even is a solution.

Besides forcing the user to explicitly importing an engine I don't think so. We should make this clear in the examples and that Python best practice is to avoid importing with `*`.

jhprinz:
I will close this also since we had so many changes that this little renaming of files is not worth fixing.

PR #200, opened 2015-03-30
[WIP] Traverse

jhprinz:
This add some functionality that is part of movepath visualization and allows to easily
traverse movepath trees in various (bfs / dfs) order and apply a function. This way you can filter for specific properties, count moves. etc... basically in one line! This works like
```
movepath.traverse_dfs(lambda x : x.__class__.__name__)
>>> ['ForwardShoot', 'RandomChoice']
```
and would return a list of names of all classnames in order where now the innermost "leaves" are called first. Useful, if you want to get the order in which samples are generated

`traverse_bfs` will call it in the order how the movers are called
```
movepath.traverse_bfs(lambda x : x.__class__.__name__)
>>> ['RandomChoice', 'ForwardShoot']
```

The is also a version of both to also get the depth of the leave, so you get a list of the type
```
# dfs
>>> [(1, 'ForwardShoot'), (0, 'RandomChoice')]
```
```
# bfs
>>> [(0, 'RandomChoice'), (1,'ForwardShoot')]
```

I think this is ready to be merged, if considered useful and with good names, etc...


jhprinz:
This is / will be part of #196 and is obsolete...

PR #203, opened 2015-04-01
Add storage of RETISTransition, Transition and TPSTransition

jhprinz:


jhprinz:
This adds storage for Transitions. It also improve on some other stuff of the storage:
- all object will now be fully restored
- `@restores_as_full_object` is renamed to `@ops_object` because it actually prepares a class for all the pickling, storing and recreation which is used by the storage. This is important also for the conversion to a JSON string
- orderparameters can now be loaded even faster. Before, if an order parameter was not present in the storage the snapshot would be loaded and then fed to the OP function (although this is not present in loaded OPs). It just made the loading slow. Now this is removed and using partial stored orderparameters is as fast as full ones
- all classes have been checked that they load and save correctly now. Still more tests for that are required.
- the `.name` property is now treated as every other variable and will be stored, even if not explicitly in the parameters for `__init__`
- the default behavior for storage is now that all member variables (the ones in `.__dict__`) are stored unless they are private (starting with a `_`) when restoring an object these variables are filtered to the ones necessary to run `.__init__` of that class and used to recreate the object. All other missing variables are set afterwards.
- every class has the `._excluded_attr` variable that contains a list of excluded member variables
- every class has the `._exclude_private_attr` variable (`True` is default) . If set to `False` also private members are stored
- every class has the `._restore_non_initial_attr` (`True` is default) . If set to `False` then the member not used to initialize are not set to recreate a class. This is not used yet, but can make the recreation code simpler

If it passes this is ready to be merged and should be merged before all the renamings!

jhprinz:
Alright, this passed so go ahead and store RETISTransitions. 

Docstrings still need to be updated, but I will do that after easter :)

dwhswenson:
Changes look good to me: this seems like a much simpler interface to storage than we had before. The distinction between `@restores_as_stub` and `@restores_as_full` was going to be hard to explain to a user; folding them into the a single decorator makes it a lot clearer!

We might still think about whether there's a clean way to add storage support through inheritance instead of through decorators. Right now, every subclass that inherits from certain classes (`PathMover`, `Ensemble`, etc) needs to use the same decorator. That looks to me like a case where the clearer approach would be to use inheritance. Is it possible to decorate the base class such that inherited objects retain the decoration? (This way you'd decorate `PathMover`, but not all of its myriad subclasses.)

Also, I assume the changes to files in attic/ were due to automated mass changes? We can probably clean out the attic sometime soon.

jhprinz:
Yes, you are totally right with the clumsy _each subclass needs the decorator_ approach. The problem is that this decorator fills a list with all classes so that the JSON can easily reconstruct a class just by name. This seemed clearer than searching in the global namespace for the class name since we can really say, these classes should be storable and not this ones. Now for subclasses it would be nice to do this automatically. In a way the storage does this, but the pickling (the serialization) still cannot. There is some code that need to run per (sub)class and the only way to do that would be to use metaclasses. This is a very powerful and cool concept, but is not often used, so I tried to avoid it. We could try to hide it from the user and just say: 

If you want an object to be storable/jsonifyable subclass it from `OPSObject` and python will do the rest. If OPSObject has a metaclass set, then this will be used in all subclasses as well and then you can e.g. register all subclasses when they are created. I know that these things are used often for database stuff, since it can hide a lot of complexity from the user. I will keep this in mind and might come up with a solution that could work. In this case I would really like your opinion first to make sure that this does not make things more complicated!

Yes, I did not care too much to fix problems in attic since lots of this code (from me) is obsolete and can or should be removed.

PR #204, opened 2015-04-03
[WIP] Faster SequentialEnsemble

dwhswenson:
The current version of `SequentialEnsemble` checks the entire trajectory every time can-append is called. This is redundant, since the earlier frames were already assigned to an ensemble in the sequence the previous time can-append was called. The current implementation has $O(N^2)$ scaling with the number of frames in a trajectory.

For normal TIS trajectories (length ~100 frames), that doesn't hurt too much. However, this can be a serious problem for minus interface trajectories (length >1000 frames).

Looks like it should be pretty easy to fix: either I was thinking about doing something like this when I initially implemented `SequentialEnsemble`, or I'm really lucky. The code even pretty much points out how to start with the cached values; just need to figure out when to cache the values and when to reset the cache.

Tasks

- [x] Implement and test basic cache structure
- [x] Implement and test caching in `can_append`
- [x] Implement and test caching in `can_prepend`


dwhswenson:
This was a bit more complicated than I thought (I should have known -- anything dealing with `SequentialEnsemble` is a guaranteed headache). In any case, what I have now is still not perfectly optimized (there are more checks of the volume than I would have expected) but it is much faster than before. I'm now able to do 1000 RETIS steps in the toy model in about 15 minutes -- I suspect we'll be able to further speed that up in the future, but that's a lot better than the several hours it used to take.

I suggest we merge the progress so far, although we should make more improvements along these lines later.

jhprinz:
Looks good. Looking forward to faster Ensembles!!! Yes

PR #206, opened 2015-04-05
"Trusted" trajectories in Sequential Ensembles

dwhswenson:
Based on the idea sketched in https://github.com/choderalab/openpathsampling/issues/205#issuecomment-89774884. Hoping this gets rid of of the $O(N^2)$ scaling discussed also in #204.


dwhswenson:
This is ready for review and merge. It won't fix the scaling problem until combined with #204 (still in progress), but the aspects related specifically to the title of this PR (and not #204) are all done. Primarily, this does the following:

* renames the `__call__(..., lazy)` to `__call__(..., trusted)`
* adds the `trusted` argument to `can_append` and `can_prepend`
* adds `check_reverse`, which is analogous to `__call__`, but assumes you're starting from the end of the trajectory. Only implemented so far in order to use its `trusted=true` to select the first frame instead of the final frame in `AllInX`.
* implements `trusted` for `AllInX` (I believe that `AllOutX` inherits that, and those are the only places where we really need it so far)
* uses `trusted=true` in calls from `SequentialEnsemble` where we know it can be trusted (and which cause our bottleneck)
* Independent of the above, it also changes `toy_bootstrapping.ipynb` so that it actually fills all ensembles with a trajectory (unless it times out). This should fix some annoyances in testing, since we count on the bootstrapping result for later tests.

At some point, it would be good to implement a version of `SequentialEnsemble.check_reverse` that properly checks the backward-growing trajectory, since it is possible for a sequential ensemble to accept a trajectory grown in one direction while rejecting the same trajectory grown in the other direction. But I'm not including that in this PR, because there are too many higher priorities. (This is more for sanity checks and illustrations than practical use in the code.)

jhprinz:
Looks good. Merging now.

PR #209, opened 2015-04-07
Speed up TIS equil

dwhswenson:
Tired of tests taking so long: this PR trades temporary speed-up in the ipynb tests for some ugliness in the `toy_tis` file. We can remove the ugliness later.

(Ugliness is that instead of equilibrating properly, we do shorter equilibration runs until the innermost interface trajectory can be used to generate a minus interface trajectory.)


dwhswenson:
This is ready to merge, if we want the test speed-up.

jhprinz:
Good idea!!!!! Merging...

PR #214, opened 2015-04-21
TISEnsemble as a separate class

dwhswenson:
Makes TISEnsemble into a full-fledged class instead of factory function. This allows us to store some extra info in it, which will make analysis easier (and will make it easier to create modified versions of the ensembles if we want to change things, like remove a given state from the allowed final states).


dwhswenson:
Passes tests, so I'd say it is ready for review and merge!

jhprinz:
Looks good… Merging...

PR #216, opened 2015-04-22
fixed wrong loading of extended sample data

jhprinz:
a small bug fix that I overlooked and is necessary for visualization. Storage was fine but some properties of `Sample` did not get reconstructed to the right type. 

This also has an updated little sample for making pathtrees


dwhswenson:
Merging.

Issue #223, opened 2015-04-24
Problem in Ensemble.split


dwhswenson:
Ensemble.split isn’t behaving as I’d expect it to. I’m writing a test for this first, so that @jhprinz might also be able to look for a fix if he has time (so I can work on networks).

- [x] Write test (skipped) to illustrate problem (#224)
- [x] Fix split behavior

Quick summary of the problem (this is what the test will check):

1. Trajectory `traj` is in-out-in-in.
  * `len(inA.split(traj))` : expected 2, got 2 (correct)
  * `len((inA.split(traj))[0])` : expected 1, got 2 
  * `len((inA.split(traj))[1])` : expected 2, got 1 
  * `len(outA.split(traj))` : expected 1, got 0
2. Trajectory `traj` is in-out-in-in-out-in.
  * `len(inA.split(traj))` : expected 3, got 1
  * `len((inA.split(traj))[0])` : expected 1, got 1 (correct)
  * `len((inA.split(traj))[1])` : expected 2, got error
  * `len((inA.split(traj))[2])` : expected 1, got error
  * `len(outA.split(traj))` : expected 2, got 0
  * `len((outA.split(traj))[0])` : expected 1, got error
  * `len((outA.split(traj))[1])` : expected 1, got error

jhprinz:
Found the error and have a fix, will explain tomorrow. Basically two things...
1. overlap might lead to additional trajectories when using can_append
2. when stopping we need to check if the trajectory up to now is in the ensemble and if not also check if the one before (one frame less) is in. This is missing. Has to do with your math stuff...

Can probabily fix this on Sunday...

dwhswenson:
Great! Once this works, it should be extremely easy to get the flux from the results of the minus move, which makes it easy to calculate the rate.

What do you mean by "overlap might lead to additional trajectories"? I would suggest that the trajectories we return from `split` should have no overlap. Otherwise, consider `AllInXEnsemble.split(traj_all_in_vol)`: you might have to return every subtrajectory of the trajectory. That scales as $n!$. We don't want an algorithm that, in the worst case, scales as $n!$.

dwhswenson:
This looks fixed by #252.

PR #224, opened 2015-04-24
Test for Ensemble.split

dwhswenson:
Adds test mentioned in #223. Right now I `raise SkipTest` so it can pass tests for merging, but either my expectations of `Ensemble.split` are way off, or something is wrong in the code.


jhprinz:
Could you check for merging? After merging the other PRs this will not automerge and I assume you can fix this in a blink!

dwhswenson:
Yep ... only problem was that I'd added different tests at the end of `testensemble` in different branches, and git saw that as a conflict. Newest commit includes both tests (the one for this PR, and one I added in `retis_transition`, I think -- I was afraid this might happen). This commit should fix.

PR #229, opened 2015-04-26
Fix for SeqEns([A & B]) != SeqEns([B & A]) bug

dwhswenson:
This fixes a weird bug where the order of a combination mattered when used with a SeqEns. Example of the problem, assuming you've set up a reasonable `bigvol` and length==1 `traj`:

```python
vol_ens = paths.AllInXEnsemble(bigvol)
len_ens = paths.LengthEnsemble(5)

combo1 = vol_ens & len_ens
combo2 = len_ens & vol_ens

seq1 = paths.SequentialEnsemble([combo1])
seq2 = paths.SequentialEnsemble([combo2])

print vol_ens.can_append(traj)
print len_ens.can_append(traj)
print combo1.can_append(traj)
print combo2.can_append(traj)
print seq1.can_append(traj)
print seq2.can_append(traj)
```

All of these should print `True`. However, the last one was printing `False`.

The bug appears to be that our short-circuit logic used `is True` instead of `== True`. In particular, the line in `EnsembleCombination.can_append` needed to read `if b == True` (even when `b` evaluated to `True`, it was following the `else` statement).

I'm not sure why this bug only manifested under the last set of circumstances, but this PR will fix it. Should be ready to merge as soon as tests pass.

### DIFF
--- a/examples/ipython/debug_logging.conf
+++ b/examples/ipython/debug_logging.conf
@@ -1,0 +1,45 @@
+# Default configuration file for OpenPathSampling loggers
+[loggers]
+keys=root,initialization
+
+[handlers]
+keys=stdout,initf,stdf
+
+[formatters]
+keys=standardFormatter,msgOnly
+
+[handler_stdout]
+class=StreamHandler
+level=NOTSET
+#formatter=standardFormatter
+formatter=msgOnly
+args=(sys.stdout,)
+
+[handler_initf]
+class=FileHandler
+level=INFO
+formatter=standardFormatter
+args=('initialization.log', 'w')
+
+[handler_stdf]
+class=FileHandler
+level=DEBUG
+formatter=standardFormatter
+args=('debug.log', 'w')
+
+[logger_root]
+level=INFO
+handlers=stdf
+
+[logger_initialization]
+level=INFO
+handlers=initf
+qualname=openpathsampling.initialization
+propagate=0
+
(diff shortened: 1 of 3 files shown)


jhprinz:
weird fix with the `== True` instead of `is True`. This should only make sense if your variable is not always a boolean in which case it will not be wrapped in a `bool()`. So I believe the following is true:

```
>>> True is True
True
>>> 1 is True
False
>>> 1 == True
True
```

in the last case you can also write:
```
var = 1
if var:
# True
if var == True
# True
if var is True:
# False
```
because if wraps variables in a `bool()`

Also I recently learned that the following works
```
>>> a = set()
>>> b = set([1])
>>> a == True
False
>>> bool(a)
False
>>> bool(b)
True
```

so you can write
```
if a:
# tests if set is empty
```

Anyway. Will merge this...

dwhswenson:
I agree that it is really weird. What's even weirder is that with the sequence (simplified here, obviously):

```python
logger.debug("b is " + str(b))
if b is True:
    logger.debug("doing thing if b is True")
else:
    logger.debug("doing thing if b is False")
```

I was getting output in the log like:

```
b is True
doing thing if b is False
```

Changing to `==` fixed that.

jhprinz:
Could you add a `type(b)` to the logger? To check that maybe b is a numpy bool or so and maybe the `is` will report `False` as it should also that does not make sense in that context?

Issue #228, opened 2015-04-25
CollectiveVariable equality is incorrect


dwhswenson:
(Assigning this to @jhprinz because it needs to work with his fancy storage tricks for CVs.)

Easiest to explain with code:

```python
import openpathsampling as paths

def circle_sq(snapshot, center):
    return (snapshot.xyz[0][0]-center[0])**2 + (snapshot.xyz[0][1]-center[1])**2
    
opA = paths.CV_Function(name="opA", fcn=circle_sq, center=[-0.5, -0.5])
opB = paths.CV_Function(name="opB", fcn=circle_sq, center=[0.5, -0.5])

def xval(snapshot):
        return snapshot[0][0]

opX = paths.CV_Function(name="xval", fcn=xval)

print opA == opB
print opA == opX
```

Both of these should print `False`; instead they print `True`. I think this is because it is using the `dict` definition of equality (and the dictionaries are both empty, so far). It should use something based on the `fcn` and `kwargs` given -- or even the (required?) `name`.

This is needed for #227 (MSTIS networks), so I'm tagging this as "ASAP".

jhprinz:
Should have been corrected by PR #230. Will close this for now.

PR #230, opened 2015-04-28
Fix CV equality and full storage of CVs

jhprinz:
This fixes the bug in issue #228 and will compare by checking if either the bytecode of the function is equal or - for mdtraj, etc. - if the same function is called with the same arguments.

This also adds the possibility to store and completely restore order parameters. This also covers arbitrary functions, also it causes some security issues.

The function must be independent on other functions unless the are reconstructed manually. For example
```
def func(snapshot):
    return snapshot
```
works, but
```
def func(snapshot):
    return other_func(snapshot)
```
does not, but
```
def func(snapshot):
    return mdtraj.some_func(snapshot)
```
works, if we manually run
```
import mdtraj
```

Also CV_MD_Function and CV_Featurizer now work the same. You just pass the arguments and the class or function

```
CV_MD_Function('name', mdtraj.function, kwargs)
CV_MD_Function('name', msmbuilder.Featurizer, kwargs)
```


dwhswenson:
Excellent! I think having any support for rebuilding a function (rather than just using the dictionary of saved computations) is pretty powerful.

Some questions for clarification: I suspect a common case might be that you'd use some functions from an analysis package (MDTraj being a likely example) to build your own function. So, for example:

```python
import external_analysis as ext

def my_func(snapshot):
    return ext.min_dist(first_list_of_atoms, ext.center_of_mass(second_list_of_atoms))

CV_Function(name="cv1", fcn=my_func)
```

That's a reasonable CV (it's actually an example we're using in the group now). Under what circumstances would this be reconstructed? Do I need to do:

```python
import external_analysis as ext
storage.cvs.load("cv1")
```

or can I get away with

```python
import external_analysis
storage.cvs.load("cv1")
```

or can I not use `import as` in the code?

What if, instead of having the import *outside* the function, I instead do:

```python
def my_func(snapshot):
    import external_analysis as ext
    return ext.min_dist(first_list_of_atoms, ext.center_of_mass(second_list_of_atoms))
```

Would that work?

Any tricks to raise a warning on save to say the the function can't be restored? I imagine you might just try to load it into a dummy storage and try to use it once, and catch whatever error pops up.

This is a great feature, but we should make sure the conditions that support it are well-documented!

jhprinz:
Good questions. Just some quick answer, later more...

In general a function works if all other global variables needs (modules, etc...) are the same as when the function has been saved. So the first should work (I whink I tested a similar example). The second probably will not, but I am not sure if the internals of marshal will take care of this. The reason is that you might assign something else as ext and since the function does not save this there is no way of knowing.

Your third approach should work again. That seems like a very smart workaround.

Right now you cannot use CV_MD_Function with arbitrary function, but I will add that later. This is easy to do. The only difference to function is that the trajectory will be casted to an MDtraj trajectory.

jhprinz:
Let's see if that passes. I realized some more intricacies of python, which are concerned with namespaces. I tried to make a general and clean way to store functions, but this is not as easy. The main problem is the following:

When you define a function in ipython (or whatever code you use to run openpathsampling) and call it this function has access to the imported packages as well as your global namespace. If you want to save this and the saving function is part of a package then this function has only access to the globalname space of the package and not the one of your code. Means that the byte code of your function know only to access the global namespace to get a package or variable but not which one. This means that the saving function has no access to the content of the variable in the namespace that you want. 
However you can still save the code and reconstruct it, you just need to make sure that then you have the same variables in the global namespace present. And without really badly hacking into the python interpreter there is no way to get this information. A simple example:
```
atoms = [4,6,10,18]
def my_md_cv(traj):
    return mdtraj.compute_dihedrals(traj, indices=[atoms])

my_cv = paths.CL_MD_Function(my_md_cv)
storage.save(my_cv)
```
This might be a typical case and the function `my_md_cv` stored references to global vars `atoms` and `mdtraj`. While saving we have no idea, what these globals are in the context where they are used. This is a very powerful feature of python and at the same time the problem. 

This is a better and works independent of the global namespace:
```
def my_md_cv(traj):
    atoms = [4,6,10,18]
    import mdtraj
    return mdtraj.compute_dihedrals(traj, indices=[atoms])

my_cv = paths.CL_MD_Function(my_md_cv)
storage.save(my_cv)
```

All right. We cannot figure out what the content is, BUT we can figure out what things a function would like to import ad which global variables are requested. This helps and the current implementation uses that. It checks the bytecode for `LOAD_GLOBAL` and `IMPORT_NAME` commands and retrieves the associates variables. I then check, if load_globals is empty (because that is safe ot output a warning) and check i f the imported modules are in a list of `safe ones` or ouput a warning.

There is also a possibility to disable loading of custom functions using
```
CV_Function.allow_marshal = False
```

I have also simplified the way the featurizer works. There are now two main ways to use external things. One uses functions, the other uses callable classes (like the MSM3 featurizers).

The class `CV_MD_Function` is actually just a `CV_Function` with the casting to mdtraj.Trajetory. Saving, etc is the same.

This means of your 3 examples:
1. should work but might throw a warning (depending on the package. Right now you might only use `math`, `numpy`, `mdtraj` and `msmbuilder`
2. Does not work since your function will look for `ext`
3. Works fine and is the preferred way to do it. The interpreter will figure out not to load the package everytime itself. So it will not be too slow

dwhswenson:
Passes: can this be merged?

jhprinz:
Just fixing the docstrings and then I think so...

jhprinz:
This should do it. Please merge when ready. I might improve this over time, but for now it works.

dwhswenson:
Only thing I would say differently is that we might set `allow_marshal` False by default (security). Merging!

PR #231, opened 2015-04-29
Faster AppendedTrajectoryEnsembles

### LINKED ISSUE #222: AppendedTrajectoryEnsembles are slow
I still see unexpectedly (and unacceptably) slow behavior in `SequentialEnsemble`, especially in its `.can_append`. This is an issue that should be dealt with before release.

dwhswenson:
Once this is done, trajectory generation should be blazing fast.

- [x] Caching for `ForwardAppendedTrajectoryEnsemble`
- [x] Linear scaling for `ForwardAppendedTrajectoryEnsemble`
- [x] Caching for `BackwardPrependedTrajectoryEnsemble`
- [x] Linear scaling for `BackwardPrependedTrajectoryEnsemble`
- [x] Remove/refactor debugging statements to reduce prefactor

I'm also overriding the unused of `can_append`/`can_prepend` for these to raise an error if called. `ForwardAppendedTrajectoryEnsemble.can_prepend` is nonsense: I'm not even sure what the expected behavior would be, so I'm calling it an error.

Fixes #222.


dwhswenson:
As mentioned in #222, the prepend cost is actually only "near-linear." The nonlinearity is visible (see `shooting_speed_test.ipynb` in my [gist](https://gist.github.com/dwhswenson/27de391da5450fc249b1) for details), but the effect is small enough that I don't think we need to worry about it.

I've done what I can to speed things up here. Hypothetically, a different implementation of the sequential ensemble could speed it up a bit (by replacing the current approach with an approach which assigns frames to the trajectories, and then uses that assignment as a cache).

I've attached a recent call graph (from a 200-cycle RETIS calculation). You'll have to load it to read anything, of course, but a few main points:

* storing SampleSets: 27% of total time (done after each MC cycle).
* sanity check: 14% of total time (done every 10 MC cycles, although that is easily adjusted)
* `chaindict`-related stuff: 46% of the total time

Part of the problem with `chaindict` is, of course, that it gets called a lot. But it still makes a lot more calls within itself. So anything that could reduce that would probably speed things up considerably.

Anyway, this is ready to merge now. I'm getting reasonably faster (and much more consistent) calculation speeds.

retis_200

dwhswenson:
Application of other PRs made this unmergeable, and #242 made a step in the tests here redundant (and thus they would throw errors). All of that should now be fixed, and (assuming it passes tests, which it should) this should be ready for review and merge.

dwhswenson:
Fixed so that this is ready to be merged again (at least, if we merge it before merging anything else this time!)

jhprinz:
lets go...

PR #232, opened 2015-04-29
Load samples fast

jhprinz:
This adds a function to cache all samples at once instead of loading them one by one.

In particular for samples this is verz effective since samples contain not too many complex information.
Just to load Samples the new function is 2-3 orders faster than the old method.

See example NB for comparison.


jhprinz:
If is passes this should be ready to merge...

dwhswenson:
Nice... merging!

PR #234, opened 2015-04-30
[WIP] Better Tree Handling for PathMoveChanges

jhprinz:
This will mostly cleanup the tree handling of PathMoveChanges. This includes

- [x] nested lists based tree description
- [x] implementation of `__contains__`
- [x] searching using subtrees
- [x] searching using wildcats in trees
- [x] searching using PathMover instances, PathMover class and PathMoveChange instances
- [ ] create list of all possible movetrees from a PathMover
- [ ] create list of all possible keys from a PathMover
- [ ] allow to check if a MovePathChange could have been generated by a PathMover

#### The new tree structure is as follows

1. A tree consists of nodes
2. Each node can have zero, one or more children
3. Each child is a node itself

#### The tree structure in openpathsampling is expressed as

1. The tree structure is given as a nested list.
2. The first element in the list is the head of the node
3. Element 2 to N are the children. This way we can access a child using __getitem__[#child] while [0] get the element of the node
4. Children are always wrapped in brackets
5. The head can be a PathMover instance or PathMoveChange instance or a wildcat

node = [head, [child1], [child2], ... ]

This means that the brackets naturally indicate going down `[` or going up `]` the tree

A tree can be a subtree if the subtree (always starting from the top)
fits on top of the tree to match. Here child nodes are ignored as long
as the mask of the subtree fits.

In searching wildcats are allowed. This works

1. slice(start, end) means an a number of arbitrary children between
    start and end-1
2. '*' means an arbitrary number of arbitrary children. Equal to slice(0, None)
3. None or '.' means ONE arbitrary child. Equal to slice(1,2)
4. '?' means ONE or NONE arbitrary child. Equal to slice(0,2)
5. '.', '?' and '*' on the head 

#### Examples

```
>>> tree1 = [mover1, [mover2]]
>>> tree2 = [mover1, [mover2], [mover3]]
>>> tree3 = [mover1, [mover2], [mover4, [mover5]]]
>>> tree4 = [] 
```

for wildcats we can write
```
>>> search = [mover1, [mover3]] # mover1 with first submover mover3.
>>> search = [mover1, '.', [mover3]] # mover1 with second submover mover3.
>>> search = [mover1, '*', [mover3]] # mover1 with submover mover3 at any point.
>>> search = [mover1, ['?', '.'], [mover3]] # mover1 with arbitrary first child hat must have one child (or more and second child of mover1 is mover3
```

Don't know how much we might use this, but it could be useful.

### DIFF
--- a/openpathsampling/__init__.py
+++ b/openpathsampling/__init__.py
@@ -25,7 +25,7 @@ from pathmover import (
     ReplicaExchangeMover, ConditionalSequentialMover, EnsembleHopMover,
     PartialAcceptanceSequentialMover, ReplicaIDChangeMover, SequentialMover,
     ConditionalMover, FilterByReplica, RestrictToLastSampleMover,
-    CollapseMove, PathSimulatorMover, PathReversalSet,
+    PathSimulatorMover, PathReversalSet,
     NeighborEnsembleReplicaExchange
 )
 
--- a/openpathsampling/pathmovechange.py
+++ b/openpathsampling/pathmovechange.py
@@ -105,6 +105,21 @@ class PathMoveChange(object):
         if type(item) is int:
             return self.subchanges[item]
 
+        if type(item) is list:
+            # this is assumed to be a tree
+            if item[0] is self.mover:
+                if len(item) > 1:
+                    for ch in self.subchanges:
+                        r = ch[item[1]]
+                        if r is not None:
+                            return r
+                    return None
+                else:
+                    return self
+            else:
+                return None
+
+
     def __reversed__(self):
         for subchange in self.subchanges:
             for change in reversed(subchange):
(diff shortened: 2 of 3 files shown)


dwhswenson:
Just for the record, I hope you're searching using wildCARDS in trees. I feel a little unsafe wandering the PathMoveChange forest if there are actually wildCATS in trees. :-)

jhprinz:
Well, hmmm, maybe :) I think so ...

jhprinz:
One Idea I came across: If we adopt networkx for the graph analysis anyway we might use the same way to express trees as they do. I think that could be useful.

dwhswenson:
That could be useful. I'm definitely planning to use NetworkX for analysis of replica exchange networks -- even if we only directly implement some basic visualization, it gives users immediate access to more powerful graph theoretic routines if they want them, with no extra effort on our part.

Up to you if you want to use it here as well.

jhprinz:
Will be part of #251.

PR #235, opened 2015-04-30
[WIP] add paired Snapshots and trajectory improvements

jhprinz:
This will add that Snapshots are created in pairs. Each Snapshot and it's reversed one. 
This allows us to handle multiple reverses of snapshots to be identified as the same.

```
snapshot.reversed.reversed = snapshot
```

This requires to use
```
snapshot.reversed
```
to get the reversed snapshot. Not sure if we really want to use this. But that fixes the problem @dwhswenson had.

Last,
```
reversed(traj)
```
will return an iterator that runs in reversed order over reversed snapshots. So to get a reversed trajectory we can write
```
Trajectory(list(reversed(traj)))
```
or just
```
for reversed_snap in reversed(traj):
    # Do something
```

### DIFF
--- a/examples/ipython/alanine.ipynb
+++ b/examples/ipython/alanine.ipynb
@@ -112,7 +112,7 @@
    "cell_type": "code",
    "execution_count": 4,
    "metadata": {
-    "collapsed": true
+    "collapsed": false
    },
    "outputs": [],
    "source": [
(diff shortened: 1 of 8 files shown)


dwhswenson:
Excellent! With any luck, this will help me fix the last bit of caching for `BackwardPrependedTrajectoryEnsemble.can_prepend`. Thanks!

Merging.

PR #236, opened 2015-05-01
Adds MC cycle counter to log file.

dwhswenson:
Exactly what it says on the tin.

@bolhuis suggested that it would be useful to have an easy way to check how many MC cycles have been done by the calculation while it is in progress. This is a one-line addition to `PathSampling` to do that.

It replaces what @jhprinz and I have been doing with `ncdump -h $FILE | grep "sampleset \="` (or similar).

Should be immediately ready to merge.


dwhswenson:
Hold on, don't merge this yet; going to add something else to it.

dwhswenson:
Okay, the new edition has a little hack that tracks the MC step number in the cell output. By clearing the output after each step, we don't accumulate a bunch of wasteful stuff.

On non-ipynb output, it'll act like a print statement each time; so we might want to come up with a different way to handle that. But for interactive ipynbs, it looks pretty cool in action.

Now this will be ready to merge.

jhprinz:
Neat trick with clearing the IPython cell

jhprinz:
merging

PR #242, opened 2015-05-03
Fix ensemble testing

dwhswenson:
We'd recently been getting occasional errors in `test_ensemble`. This PR fixes that.

In a test before this fix, I got errors 12/100 runs of `nosetests test_ensemble.py`. After this fix, 0/100.

The problem was in the way the test trajectories were built. For simplicity, early on I made test trajectories as lists of numbers instead of lists of snapshots. However, this created problems in certain parts of the code, because two "snapshots" in this fake trajectory list could have the same value, and the caching behavior assumes that snapshots never repeat in the trajectory.

The change makes the tests run a little slower (getting an order parameter from a snapshot is more expensive than reading a number from a list of them), but at least it's correct and tests pass consistently!

Anyway, this should fix the problem, and can be merged as soon as it passes tests.

### DIFF
--- a/openpathsampling/tests/testensemble.py
+++ b/openpathsampling/tests/testensemble.py
@@ -1,6 +1,6 @@
 from nose.tools import assert_equal, assert_not_equal, raises
 from nose.plugins.skip import SkipTest
-from test_helpers import CallIdentity, prepend_exception_message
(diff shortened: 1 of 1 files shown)


jhprinz:
great. Will merge

Issue #247, opened 2015-05-06
Summaries of each move


dwhswenson:
Moving @bolhuis's comments to a new issue:

> The most important about these summaries is that they also give the properties of the path.

> I really think that we should focus on providing the user with a summary of the simulation such as:
> step = 1 : move type = forward shot ; interface = 3: intlamb=0.15: initial state A: final state A: shooting point =10 ;trial length = 100 ; original path length = 50; new path len = 200 ; lambda_max = 0.2 ensemble = true : metropolis = 0.9: accepted 
> step = 2 : move type =- backward shot: interface = 4: intlamb=0.20: initial state A: final state B: shooting point =43 ;trial length = 33 ; original path length = 88; new path len = 133 ; lambda_max = 1.5 ensemble = true : metropolis = 0.1: rejected 
> step = 3 : move type = revesal: interface 5: intlamb=0.25: initial state A: final state B: new path len = 133 ; lambda_max = 0.25 ensemble = true : metropolis = 1: accepted 
> step = 4 : move type = replica exchange : interface 2: intlamb=0.10: initial state A: final state A: lambda_max = 0.2 interface 3 : intlamb=0.15: initial state A: final state A: lambda_max = 0.3 : exchange= >  allowed: accepted
> etc etc

> Are you looking into this?

> Especially the type of the path is important. E.g. a path can be typed as A1A A2A, A3A, AB, AC AD etc
> Are there any tools to identify such path types?

dwhswenson:
This needs to be considered in two parts:

1. Summaries of how the ensemble sees the trajectory
2. Summaries of what the move does.

The log will NOT show this as a single line, because the main point of the log is to output as we go along so we can see where things went wrong if a bug shows up. But we can write an awk script to join it into a single line after the fact, if you think that is really important. We can create a python function to create these one-line summaries in any format you want, but as I've said, outputting all of them in an IPython notebook will crash your web browser, and besides, if you want them lines like this it is just as easy to get them from the log. If you want to analyze them, that's what python is for, and we've been adding more analysis tools.

I've already added the part 1 above. I'm checking with @jhprinz about a few details before issuing the PR, because I might change a few small things to make it easier to do other stuff later. Current log output for that looks likes this:

```
Trial trajectory: initial_state=-inf<xval<-0.3 final_state=-inf<xval<-0.3 min_lambda=-0.313912 max_lambda=-0.168661
```

[...]

PR #248, opened 2015-05-06
Ensemble trajectory summary

dwhswenson:
(just going to put this PR up; I think my statements on TrajectoryCVs are accurate; if not, @jhprinz can correct me)

New function called `trajectory_summary` as a part of each ensemble. The main purpose of this is to satisfy @bolhuis’s request for the following information on every TIS shooting move:

* initial state
* final state
* max value of lambda for the trajectory

To do this as part of the shooting process, we have to make this a function for ALL ensembles, because shooting is not specific to TIS ensembles. However, those three statements make no sense in the context of a general ensemble. So this has to be a function attached to each ensemble individually.

The way I’m setting this up, there are two functions:

* `trajectory_summary` : returns a dictionary of `{ name : value }` pairs
* `trajector_summary_str`: calls `trajectory_summary`, and returns a printable string

If the values in `trajectory_summary` are numeric; we can save these as trajectory CVs (@jhprinz tells me we’re very close to being able to implement that).

In this PR, I’m only implementing for the `TISEnsemble`. In that case, the following would be a reasonable return value for `tisensemble.trajectory_summary(trajectory)`:

```
{ 'initial_state' : 0,
  'final_state' : 2,
  'max_lambda' : 0.35,
  'min_lambda' : -0.01
}
```

The `TISEnsemble` object has attributes `initial_states` and `final_states`, each a list of volumes. The summary function returns the index of that state. This should make it easier save those results as `TrajectoryCV`s.

The associated output for `tisensemble.trajectory_summary_str(trajectory)` would be something like:

```
initial_state=stateA final_state=stateC max_lambda=0.35 min_lambda=-0.01
```

Note that this also requires that `TISEnsemble`s have an `orderparameter` associated with them. That defaults to `None`, and in that case, `max_lambda` and `min_lambda` return `None`. If the user specifies the `orderparameter` for a `TISTransition`, this is set automatically for the generated `TISEnsemble`s. An `MSTISNetwork` requires `orderparameters` for each state-based interface set, so this gets set automatically there as well.

Although this currently only does anything with the `TISEnsemble`, I think it might also be useful on a general `SequentialEnsemble` to know which frames the ensemble is associating with which subensemble. I’d been thinking about finding a way to report that anyway; this seems like the right tool for the job. (But that will have to be a future PR).

Task list:
- [x] Implement `TISEnsemble` summary functions
- [x] Tests for `TISEnsemble` summary functions
- [x] Add `ensemble.trajectory_summary_str(trial.trajectory)` logging to Shooting move

This should be ready to merge. Handles part of #247.


jhprinz:
lgtm. Merging

Issue #244, opened 2015-05-04
New `MCStep` object


dwhswenson:
@jhprinz and I discussed the fact that analysis is a bit confusing because some things are studied using the `SampleSet` and some are studied using the `PathMoveChange`. Furthermore, it sounds like some of his changes to speed up loading of `PathMoveChange`s may make it harder to loop over them in a one-per-MC-step way.

We realized that we can solve all of this by saving an `MCStep` object which has attributes for both the newly-determined `SampleSet` and the most recent `PathMoveChange` (from the top of the tree). This should make analysis a good bit easier from the user's point of view.

dwhswenson:
Related to this: if we have such an object, it might be possible to create something like `slice`s for `storage` based on subsets of this. A couple examples where I would want something like that:

* while developing analysis methods, it can be nice to work on a small subset of the data for speed purposes
* block averaging can be easily implemented by using slices

If each `step` contains both the `SampleSet` and `PathMoveChange`, slicing the `storage` would just be slicing the `steps` list.

jhprinz:
Will be part of #251.

jhprinz:
Alright. I have this implemented in #251. The current MCStep has the following attributes
```
class MCStep(OPSObject):
    """
    A monte-carlo step in the main PathSimulation loop

    It references all objects created and used in a MC step: 
    The used mover, and simulator as well as the initial and final sampleset, 
    the step number and the generated pathmovechange.

    Attributes
    ----------
    simulation : PathSimulation
        the running pathsimulation responsible for generating the step
    step : int
        the step number counting from the root sampleset
    pre : SampleSet
        the initial (pre) sampleset
    post : SampleSet
        the final (post) sampleset
    change : PathMoveChange
        the pathmovechange describing the transition from pre to post
    root : SampleSet
        the first sampleset at step 0.
    """
```

Questions:

- Is the naming good?
- Do we need both `.pre` and `.post`? I thought this makes it very clear that this is a step which goes from a pre-state to a post-state and not a state itself.
- Do we need the root? This is somehow redundant and in most cases the same as `steps[0].pre`
- Do we want an additional reference to the used mover? This is accessible by `step.change.mover`. We could add this as a property. Actually I think I just do that.

@dwhswenson suggested to allow to splice a store. Do you mean just internally and do something like:
```
splice_store = storage.splice_steps(slice(0,None,10))
```
[...]

PR #249, opened 2015-05-07
[WIP] Fast PathMoveChange loading...

jhprinz:
This PR will change the way `PathMoveChange`s are stored. This will allow much faster loading.

The store will now save a PMC by directly storing attributes in table and not use JSON except for details.
Each PMC now has the following attributes and we should think about good names for these:
```
.mover # this is a reference to the used mover
.subchanges # used by all node changes to store the subPMCs
.generated # list of all samples generated in this move. Only for SamplePMC
.details # all the rest that is specific to a move
```

more details coming


dwhswenson:
Regarding names, I think most of those are good. The only one I would definitely suggest changing would be `generated`, which isn't clear on its own. `generated_samples`, maybe?

If we change the name of `PathMoveChange` (it is a bit unwieldy) then obviously `subchange` might also need adjustment, but it works for now. I think `mover` and `details` are quite clear.

Is the whole of `details` stored as JSON, or just parts? In principle, there are a few details which *all* PMCs will have (`accepted` is the first that comes to mind; `trials` (pointing to a list of `Sample`s) is another one). It might be nice to make these load faster than JSON requests would.

Definitely looking forward to the faster loading; but looping over all moves: what will be the way to do that? I assume it is no longer `for pmc in storage.pathmovechanges` (since I assume these are now each sub-pmc saved individually). We talked about adding some sort of `MCStep` (or `MCCycle` if you prefer) object (see #244) -- I think that's no more than a container for `SampleSet` and `PathMoveChange`, and each new step is made in the `PathSimulator` object. Should that be part of this PR?

dwhswenson:
Oh, and please make appropriate modifications to the analysis stuff. For a couple reasons (mainly that my WHAM script fails if input data is insufficient) we haven't added `toy_analysis.ipynb` to the test suite. I suspect this may cause problems for a few parts of that -- let me know if you need help fixing those problems.

jhprinz:
Will be resolved also by #251

jhprinz:
Should have been closed by #251 and is already solved.

PR #250, opened 2015-05-07
Replica networks

dwhswenson:
This will include various analysis tools to look at replica exchange networks.

Tasks:
- [x] RepEx acceptance matrix
- [x] RepEx diagram (based on NetworkX)
- [x] Flow (for given `top` and `bottom`)
- [x] RoundTrip times (for given `top` and `bottom`)
- [x] `trace_interfaces_for_replica`
- [x] `trace_replicas_for_interface`


dwhswenson:
I'm calling this ready for review and merge. A nice overview of the included feature set is in the [`repex_networks.ipynb`](https://github.com/dwhswenson/openpathsampling/blob/replica_networks/examples/ipython/repex_networks.ipynb) notebook.

There's still plenty of room for improvement. A few things to keep in mind:

* better layout for images: `pygraphviz` is better, but not yet in conda; a custom implementation may be better still
* clean up use of labels: still kind of messy, and in particular the flow plot required too much extra work to get appropriate labels

Still, this is a nice start, and should be merged so this can be adjusted to work with #251.

jhprinz:
I will merge this. Looks good. Cool, there is a transition_matrix function!!!!!!!!!!!!!!

One question though: Why did you add ReplicaExchangeMover to MinusMover as a baseclass.
Is this to check for subclasses or does it actually require the __init__ of RepExMover to be called for initialization. If so, this require some careful calling of __init__. To be honest, I still have not completely understood how this is handeled in Python.

dwhswenson:
The reason for adding `ReplicaExchangeMover` as base class of `MinusMover` is that I needed a way to identify the `MinusMover` as a kind of `ReplicaExchange` move. Since I couldn't see inside the `MinusMover` (because it is collapsed) I tried to come up with a hack to check that

Actually, as I think about it more, I think the current implementation is a bit dangerous. Let me think a bit about finding a better implementation. Some of this will have to depend on how things are implemented in #251.

PR #252, opened 2015-05-12
Splittest

jhprinz:
This should fix the split problems in Issue #223 .

There are two very subtle problems:
1. @dwhswenson had figured out that some can_append will stop, when the trajectory is in the ensemble and some one frame later. So that the actual subtraj is in some cases `traj[start:end-1]` and `traj[start:end]`. It will now check the longer one first and then the one reduced by one frame.
2. The other issue was that due to restarting with overlap it could happen that if the last subtraj ended at the last frame, it would still restart and end with another shorter traj of length overlap that is already contained in the last subtraj. This is also fixed now.

As soon as this passes travis it can be merged.


dwhswenson:
Great! This should allow us to get the flux from minus move trajectories, which will make it possible to get rates in basically a one-step analysis.

PR #253, opened 2015-05-12
Fix incorrect fast caching of samples

jhprinz:
This should do the trick. Tested the toy_analysis.ipynb and it works now and is fast.

Let's see if the tests pass.


dwhswenson:
Merging!

PR #254, opened 2015-05-13
Flux and rate for transitions

### LINKED ISSUE #225: Remaining tasks for complete RETIS analysis
Listing a few remaining tasks, so the current status of #202 can get merged in. Each of these may be separate PRs.

- [x] `RETISTransition.minus_move_flux()` (included in #254; was waiting on #223, fixed by #252)
- [x] `RETISTransition.rate()` (included in #254; requires `minus_move_flux`)
- [x] `RETISTransition.replica_flow()` (included in #250)

dwhswenson:
This PR adds rate calculations to the TISTransition object and adds automatic calculation of the flux using trajectories from the minus move.

#### Rate

The rate is the product: 

$k_{AB}$ = \phi_{A0} P(\lambda_m | \lambda_0) P(B | \lambda_m)$

where the terms are:

* $\phi_{A0}$ : flux through the innermost interface
* $P(\lambda_m | \lambda_0)$ : total crossing probability at $\lambda=\lambda_m$, which should be the value of the order parameter for the outermost normal TIS interface
* $P(B | \lambda_m)$ : conditional transition probability for the outermost normal TIS interface (defined as the fraction of paths in that interface which end in state B)

```python
retis.rate(outer_interface_lambda, flux=None, outer_ensemble=None, errors=None)
```

This will automatically calculate and cache the flux (see below), the total crossing probability and `outer_ensemble`’s conditional transition probability. It will assume that the outermost ensemble is `retis.ensembles[-1]`.

If the user provides a triplet `errors=(err_flux, err_tot_cross_prob, err_cond_trans_prob)`, the function will return the tuple `(rate, err_rate)`.

#### Flux

For normal TIS, the flux must be calculated separately. In RETIS (assuming usage of the minus move), the flux can be calculated by splitting the minus trajectory into parts that go from the exit of the interface to the entrance of the state (we’ll call those "outside" segments) and parts that go from the entrance of the state to the exit of the interface (we’ll call those "inside" segments). 

In the case that the state definition is exactly the first interface, inside segments correspond to frames inside the state, and outside segments correspond to frames outside the state. When there's an interstitial between the two, things are a bit more complicated.

The flux is given by (see [van Erp, PRL **98**, 268301 (1997)](http://dx.doi.org/10.1103/PhysRevLett.98.268301)):

$\phi_{A0} = ( <\tau_{in}> + <\tau_{out}> )^{-1}$

For minus interface with `n_l` excursions through the interface, there will be `n_l` outside segments and `n_l + 1` inside segments. However, the first and last inside segments are not complete and should not count in the statistics (in the case with no interstitial, these are the single frame in the state at the start and end of the trajectory: obviously, including these would not give you good statistics on $<\tau_{in}>$). So we ignore those two, and use the other `n_l - 1` inside segments.

#### Tasks

- [x] Flux from minus ensemble
- [x] Rate calculations

This resolves #225.


dwhswenson:
Slight updates and corrections to the original post:

### Rate

Easiest usage is now:

```
retis.rate(storage)
```

Most complicated usage is:

```
retis.rate(storage, flux, outer_ensemble, outer_lambda, error, force)
```

where `error` hasn't been implemented yet. For each of the other items, if not given, they are guessed from:

* `flux` : flux from the minus ensemble
* `outer_ensemble` : `self.ensembles[-1]`
* `outer_lambda` : last lambda for which the conditional crossing probability for ensemble `outer_ensemble` is 1
* `force` : assumed to be `False`; if user sets it `True`, then cached values are recalculated

Note that it could happen that `outer_lambda` and `outer_ensemble` contradict each other. I couldn't think of any way to avoid this. I think there are cases where each may need to be user-specified. But for most simple cases, the defaults will do the job.

Everything is the same for `TISTransition`s, except that the `flux` must be provided by the user.

To make it perfectly clear, the following code be all you need (assuming you have enough data in the file to avoid crashes in the analysis):

```python
storage = paths.storage.Storage("toy_retis.nc", mode="r")
retis = storage.transitions.load(0)
retis.ensemble_histogram_info['max_lambda'].hist_args = { 'bin_width' : 0.02, 'bin_range' : (-0.35, 0.35) }
retis.rate(storage)
```

I don't think it gets much easier than that!

### Flux

Minor correction to my explanation above: we can't just split the trajectories. We need to actually measure "out" from the first exit of the interface to the first entrance of the state, and "in" from the first entrance of the state to the following exit of the interface. In the case that `state` and `innermost` are the same volume, this is a distinction without a difference. However, it really does matter if we have an interstitial between the state and interface.

Anyway, the code as implemented gets it right with or without interstitials, and furthermore, it gets it right for an arbitrary `n_l`. (For `n_l` larger than 2, we *do* use the extra crossing to obtain more statistics ... there are always `n_l-1` contributions to both the "in" and "out" parts).

I also save the distribution of "in" and "out" lengths, just in case there's any interest in analyzing that.

dwhswenson:
Note that also, the flux (and therefore the rate) are in units of per-timestep instead of per-time, pending #255.

Despite that lack (and the lack of error bar support) I think is ready for review and merge.

jhprinz:
Looks good.

Nice to add .nc to the gitignore!

Also you have already something with simple names for Volumes in place. Great. I had a little idea to allow for naming wrappers (perhaps for more than volumes) that allow to name volumes and give them a 'Letter' and interface number. This could be used by the analysis when it generates ensembles.

Merging now

Issue #233, opened 2015-04-30
Better printing for objects


dwhswenson:
I've been making a few things pretty-printable, but I think we should make some decisions about what functions return what kind of string for our objects.

In particular, I'm thinking about things like `PathMover`s and `Ensemble`s, where it would be nice to have a clean string representation. If I want to check that my move scheme from a network is correct, seeing `<openpathsampling.pathmover.ReplicaExchangeMover at 0x114025650>` doesn't help very much.

The common functions here would be `__str__` and `__repr__`. As I understand it, `__str__` should be easily human-readable, and `__repr__` should be as close to code that reproduces the object as possible, and most importantly, should be unambiguous. I think we may want to create a third function to print something which is not quite code and not quite human-readable. Maybe call that `.description()` or something like that?

Consider `Ensemble` as an example. Right now, `Ensemble.__str__` prints a set-theoretic description of the ensemble. This is unambiguous, but not really human readable (at least, not for complicated ensembles). I think this should be moved to the `.description` (or whatever we call it). 

In many cases, we can assign a name to an ensemble (`Interface 1 from State B`, `Minus Interface State A`), and I've been working on having my network code create those names by default. I think this is what the `__str__` object should print, with a fallback to calculate something like the `.description` (but using `name` for subobjects) if `self.name` is not available.

Also, `__repr__` (which we have rarely implemented) should be made more like its standard behavior, which is unambiguous and often code-like.

Basically, my suggestion is this:
* `__str__` : human readable, simple, not necessarily unambiguous -- usually `self.name` if available
* `description` : sort-of human readable, unambiguous
* `__repr__` : code-like, unambiguous

jchodera:
This all sounds great!

jhprinz:
Sounds good to me. Do you have anything available for this already? Or working on a branch?

dwhswenson:
Not really working on it yet... just wanted to put the idea out there. I've been using `__str__` in this way for `Transition` and `Network` objects (`Transition.__str__` prints the transition's name, the initial state, the final states, and a list of the names of the ensembles, which are set to something based on the interface values during setup). Example:

```python
mstis = paths.MSTISNetwork([
    (stateA, interfacesA, "A", opA),
    (stateB, interfacesB, "B", opB),
    (stateC, interfacesC, "C", opC)
])
print mstis.from_state[stateA]
```

returns:

```
RETISTransition: Out A
A -> A or All states except A
Interface: 0.0<opA<0.04
[...]

PR #261, opened 2015-05-15
Change .name storage

jhprinz:
This PR simplifies storage a little and add better support for the `.name` property.

All objects have now a special name property which can be set once before it is saved (we could actually change this).

There is also for all nameable objects a `.named(..)` function that creates a shallow copy with a new name. 

Instead of the decorator `@ops_object` there are new two base-classes `OPSObject` and `OPSNamed`.
Subclassing from these does the same as before in a easier to understand way.

The `.name` property is available for all elements as a string. Even if not set.

To be compatible with the only once setting of the name I had to remove the automatic setting of a name for `LambdaVolume`. I think the nice `min<op<max` output could be moved to either `__str__` or `.description`


dwhswenson:
What is the reason for doing a ***shallow*** copy with a new name? I would assume that one might want a deep copy, if anything.

I think that the `min<op<max` should be set as the `name` by the `VolumeSet` thing that makes interfaces (that's where we really need this). This might also be an option of making that `description("short")` or something like that; I think that `description("long")` should be the current `__str__`, and `__str__` should be something like:

```python
if name is not None:
    return name
else:
    return self.description()
```

where I'm not sure if that should be `"short"` or `"long"`. But those changes are to be done much later in #233.

jhprinz:
Important: Saving an unnamed object sets the name to an empty string. After that the name cannot be set anymore.

All objects that can be saved with name are
```
Volume()
Ensemble()
PathSimulation()
PathMover()
CollectiveVariables()
DynamicsEngine()
ShootingPointSelector()
Transition()
```

After tests pass, this should be ready to merge. 

Only question is if we want to keep the name only once policy. Not sure if this makes sense. It forces the user not to rename at will.

jhprinz:
A deep copy would also make a copy of all contained stuff like ensembles, volumes and double the objects. That would be bad. A shallow copy only replicates the object and fills it with the same objects. There is one thing though and I need to check, if it correctly handles indices. See last commit.

I have not yet used the .named feature, we can also just use it to set the name. This still allows us to write it in a chain way
```
ens = Ensemble(...).named('myensemble')
```

dwhswenson:
1. I think "objects-immutable-after-saving" is a fair rule to play by. Mutable until saving is what matters to me.

2. Now I see the chain approach, and that's not bad. But without that, I think `object.name=myname` is better than `object.named(myname)`, esp. since the latter creates a copy (which means references to it won't be the same, so it is NOT the same as setting the attribute, right?). I could imagine someone wanting to create a deep copy with a new name and modify it; that's why I'm concerned this might be confusing.

2. I think it might be nice to have a `default_name` function, or at the least have it default to the class name, rather than the empty string. I suppose some of that can be achieved in most cases by doing `if self.name == "":` in the `__str__`, which is, I think, where we'll usually want to access the name. But a non-trivial default might be good to have.

jhprinz:
1. Mutable until saving might require an additional flag to carry around. Setting a name once does not, which makes the code a little clearer, but I agree that it is not so common in python

2. Exactly, using `.named(myname)` will be different from `.name = myname`. The first will clone the object which is equal to call the constructor with the same parameters, so both objects represent the same information/are identical, but are not one and the same. This is the idea of functional programming and I agree that might be too much. The upside is flexibility (which is what we want), the downside is that a user can do something that is not wanted or reasonable at an arbitrary point in the code and sometime later your code crashes and you do not know why. This is of course the extreme case. 

An example: One of your tests is testpathmover sets the name of an object to 'stateA' and will check for/search for it later. That code is hidden and noch accessible to the user. At a later stage you change the name for convenience and all references change with it. When you do the testing you cannot find your object. This is - I agree - very constructed - but the point is that if you create an object with properties and can depend on that object being immutable you know where to find an error, because there is only one place where the variable could be set.

This should only be an explanation why I like this approach so much. The little overhead for constructing a shallow copy is negligible is most cases (if the compiler or interpreter is smart). Anyway, I think this is in this particular case not really necessary, also because the name is more for convenience and will not destroy a calculation if fiddled around with.

I also agree that this approach is not so well known and might be confusing. Especially because of this:
```
x = MyObject()
x.named('test')
print x.name
>>> None # x did not have a name and it is immutable
y = x.named('test')
print y.name
>>> 'test' # this works
```

This concept of _chaining_ is often used in javascript and requires each setter to return the object itself.
This confusion would be my reason not to use it.

3. Good idea. I would call it a `default_name` property and return the result, if the `._name` is empty automatically like
```
@property
def name(self):
    if self._has_name or self._name is '':
        return self._name
    else:
        return self.default_name
```

4. I will change the code so that you can change it as long as you want. Empty string means use default and we save the empty string so the default name will adopt to changes in `.default_name`

jhprinz:
I added a flag to make sure you cannot change the name after storage.

Concerning the searching by name: This will only look for really manually set names, not for the default_name which I think makes sense.

jhprinz:
Finally, searching by name works again, with the following caveats.

1. You can only load by a set name not by a default name
2. If there are more than one object with the same name you get the first one (no errors thrown, but it is mentioned in the logger)
3. `storage.[object].find(name)` searches for all objects of a name
4. `storage.[object].find_first(name)` searches for the first object of a name
5. `storage.[object].find_indices(name)` returns indices for all objects of a name

Usually there should be only one object of a given name, but who knows. With this PR you can now use

```
# save
my_retis = RETISTransition().named('myRetis') #give it a name
my_retis.name = 'myRetisTransition' # change name afterwards

storage.save(my_retis)

# and load
storage.transitions.save(my_retis)
retis = storage.transitions['myRetisTransition']
```

jhprinz:
fixed the storage tutorial to work with name and not uid.

Should be ready for review and merge!

dwhswenson:
When I say "mutable until save", I really mean "mutable until production; save at production". Saving the setup is the first step in production.

Here's a realistic usage that should be allowed:

```python
mstis = MSTISNetwork([
    (stateA, interfacesA, "A", opA),
    (stateB, interfacesB, "B", opB)
])
outermostA = mstis.from_state[stateA].ensembles[-1]
# MSTISNetwork gives reasonable standard names, so that, for some min and max:
# "min<opA<max" == outermostA.name

outermostA.name = "outermostA"
# changing that standard name to a custom name
```

(Terminology: I'm using "standard name" to refer to a name set computationally by another part of the program without user interference; this is to distinguish from "default name" which is set by the object itself or "custom name" which is set by the user.)

Three points that I think need to be handled:

* We absolutely do NOT want a copy here, because the ensemble is in a context of `PathMover`s etc which would also all have to be changed. 
* We do not want the user to have to lose all the conveniences that come from the `MSTISNetwork` object, just because s/he wants to give a custom name to one part of it.
* We want the `MSTISNetwork` object to be able to assign standard names based on the context that it understands, since those standard names will be good enough for most users.

As the code stands now, is all of that supported?

jhprinz:
Quick answer for your 3 points. 
- agreed, no copies. The idea of making a renamed copy is only meant to be used when creating the object, this is why you should the `.named` ONLY in chain. And that is too dangerous. So I removed all of the copying and it does what you proposed
- also agreed, the MSTISNetwork, etc is the center of the calculation this must be one object. And it is, no changes here, too.
- Yes, it can assign standard names. If it does not the `.default_name` property will be used to generate a name. This `.default_name` is always computed and will not be stored. This means you need to be aware of what you get when using a name of one object to build the name of another. You might have to assign names in the correct order! The user can now always change the names as he wants unless the objects are stored, which is done by starting the simulation or manually by the user. You will get an error message in that case.

So, I think it should be what we wanted, right? Sorry, for my immutable excurse... 

Will fix the merge issue now.

dwhswenson:
You changed the `toy_tis` notebook to outputting `toy_retis.nc` instead of `_toy_retis.nc`. You'll need to change that in the `repex_networks` notebook too.

jhprinz:
Ups, yes I thought that was a mistake of mine. Sorry, which one do you need? With or without the leading  underscore?

jhprinz:
Actually, I thought I had fixed that. I guess the last merge messed that up :)

dwhswenson:
Doesn't matter: I switched it at some point to keep some 10k MC step results around. The main thing is that `toy_tis` has to generate something, and `toy_analysis` and `repex_networks` need to use the same thing after ... `toy_analysis` still needs to be added to tests (but not yet -- WHAM chokes if the data is insufficient)

jhprinz:
okay, so I move all of them (also visualization) to `_toy_retis.nc`. Just a second

dwhswenson:
I think you made the changes in the wrong branch.

jhprinz:
Yup. I did...

jhprinz:
Woohoo... Travis passes. That is a good sign...

dwhswenson:
Merging!

PR #251, opened 2015-05-10
PathMoveChange improvements

jhprinz:
See #246 for detailed discussion and ideas.

- [x] refactor sample generating movers into Generators
- [x] allow generators to be used to generate samples without SampleSets
- [x] remove `.accepted` from samples
- [x] `.valid` to property
- [x] move treelogic to separate file
- [x] fix tests to work with changes
- [x] add new PMC_store for faster loading and saving : Closes #237 
- [x] add `.trials`, `.results` to PMC
- [x] add `.in_ensembles` and `out_ensembles` computation
- [x] fix visualization
- [x] better `__contains__` for trees
- [x] add `.canonical` property : Closes #243 
- [x] remove ForceEnsembleChange from MinusMover
- [x] add `MCStep` : Closes #244
- [x] add bias for `EnsembleHop` : Closes #142 
- [x] move `PathMover.engine` to `ShootingMover.engine` : Closes #150 
- [x] fix analysis
- [x] merge MSTIS
- [x] little renames in MCStep
- [x] extreme speed-up for analysis
- [x] make `canonical` a property
- [x] `cache_all` for orderparameters
- [x] add `StateSwapMover`
- [x] fast loading of MCStep
- [x] check docstrings

I kept the old names `.trials` for all changes and `.results` for the actual final changes.

Will move some minor cleanup to a separate PR:
- add simpler `ensembles=[...]` parameter for PathMovers : Means to use a more unified way to express the ensembles attribute for PathMovers.
- cleanup for ensemble selection. This is related to better EnsembleHop support with bias. This might require little more discussion
- change tree description to match networkx use of dicts of dicts of dicts (optional). This is not necessary right now and might take longer than I thought
- add specific unit tests for treelogic, chaindict, etc...

PR #263, opened 2015-05-16
Replica networks cleanup

### LINKED ISSUE #262: Normalization in replica exchange matrices
We have a decent skeleton of the functions we want from #250, but I have a few questions to check details of how some things should be implemented. In particular, I'm hoping @jchodera and/or @bolhuis might be able to give some guidance on what the normalization should be for the matrices we create in the replica exchange analysis. 

Right now, we report two matrices: the mixing matrix and the transition matrix. Let me explain what I have implemented so far -- the sparsity structure of these is definitely what we want, but I'm not sure about what the relative values should be.

The way it is implemented now, what I'm calling the transitions matrix $T$ is such that 

$T_{ij} = P_{acc}( i<->j )$
$T_{ji} = 0$, 

where $P_{acc}( i<->j )$ is the acceptance probability of the replica exchange swap between $i$ and $j$, and $i>j$ (selecting a lower-triangular matrix).

What I'm calling the mixing matrix $M$ is then just $M = 0.5 (T + T^T)$ (where superscript T indicates transpose).

I see several options for normalization:

1. per trial of the exchange for this matrix element (which is what I'm doing now, but I don't think it is generalizable to all replica exchange schemes)
2. per MC step in the calculation (which gives much lower values, since many MC steps are not replica exchange moves -- matters if we includes self-transitions)
3. per MC step which includes a replica exchange (may be the hardest to implement)

One thing to keep in mind: some replica exchanges (e.g., minus move) are done less frequently than others, so the number of trials can be very different from cell to cell. 

Let's do a quick hypothetical to show the differences that we'd see with different normalizations. Assume only 2 swaps possible: minus-0 and 0-1. The minus move typically has ~100% acceptance. Say that the 0-1 swap has 25% acceptance, but is 4 times more likely to be called. Let's say we tried 200 total moves, with 50 shooting, 50 path reversal, 80 0-1 swaps, and 20 minus-0 swaps.

* In normalization 1, we would get `T[0,minus] = 1.0`; `T[1,0] = 0.25`
* In normalization 2, we would get `T[0,minus] = 0.1`; `T[1,0] = 0.1`
* In normalization 3, we would get `T[0,minus] = 0.2`; `T[1,0] = 0.2`

If we include self-transitions, this gets more complicated; in normalization 2, `T[0,0] = 0.8` (which is 8 times `T[1,0]`) whereas in normalization 3 `T[0,0]=0.6` (3 times `T[1,0]`).

Note that only first highlights the difference in acceptance ratios. But the whole point of modifying the frequency of selecting the move is to balance the number of *accepted* moves, so maybe the equality in number of accepted moves is what we want to see.

dwhswenson:
This will involve a little more work on the replica exchange network analysis. In particular, it supports:

- [x] Counting transitions from `SampleSet` instead of identifying repex moves
- [x] Clean support for labels (numeric or string)
- [x] Clean up matrix/dataframe generation to clearly separate building the sparse data from building the output objects (this will make it easy to create analysis based on other measures)
- [x] Docstrings
- [x] Add `is_ensemble_change_mover` property (waiting on #251)
- [x] Replace current matrix calculation by something which normalizes to all RepEx moves (waiting on #251)
- [x] Add tests for `is_ensemble_change_mover`

This resolves #262.


dwhswenson:
Passes Travis! This is now ready for review and merge.

I still find the specific question asked in #262 a little problematic (see my most recent https://github.com/choderalab/openpathsampling/issues/262#issuecomment-107016235), however, I've completed everything I set out to do in this PR.

jhprinz:
Looks good.

Still thinking that we might make `is_ensemble_change_mover` computable and not having to specify it directly.

dwhswenson:
How would we make it computable? And does that make sense?

Right now, you *only* have to do anything if you're creating a SampleGeneratingMover that also involves changing ensembles. Any combination movers already compute it from their submovers.

And consider this minus-like sequential move:

1. Take trajectory from innermost interface, extend it forward until it satisfies the minus ensemble
2. Select the new subtrajectory that satisfies the innermost interface.
3. Replace the old innermost interface trajectory with that new one.

For this, `is_ensemble_change_mover` should be `False`. Currently, the user would have to set that manually, but my approach allows the user to set that manually. (This is why it is extremely important that the default be `None`, not `False`). A computational approach would need to figure that out as well.

This approach also allows a user to exclude certain ensemble-change movers from analysis; for example, maybe you want to look at *only* the normal replica exchange, and none of the minus moves. You can set the flag to "False" on the minus moves before doing your analysis.

jhprinz:
> How would we make it computable? 

I don't know, maybe that is not possible :) Just thought it would be cool, if we could avoid setting it and derive it from some properties.

> And does that make sense?

Why not, if it works without fiddling? Then we could remove the code and a potential user that wants to write a new Mover does not need to worry about it.

> This approach also allows a user to exclude certain ensemble-change movers from analysis; for example, maybe you want to look at only the normal replica exchange, and none of the minus moves. You can set the flag to "False" on the minus moves before doing your analysis.

I think this is the important point. it is for analysis to manually set, what you are interested in. A computed approach does not have this flexibility. Although it also changes the meaning a little.

So maybe move that idea to distant future :)

Issue #259, opened 2015-05-15
Add support for external CV computation (pyña)


jhprinz:
[add more discussion here...]

dwhswenson:
Some start to the discussion: in our Skype call today, I mentioned to @jhprinz that I've been working on a package that summarizes the output from another program that analyzes the structure of DNA. We were discussion how to make an OPS-ready collective variable out of this. Although we discussed it in the context of this specific set of tools, I think it might represent a more general paradigm for generating order parameters which we should support.

My little project is [pyña](http://github.com/dwhswenson/pyna). It parses the output from the [Curves+](http://www.ncbi.nlm.nih.gov/pmc/articles/PMC2761274/) program into `pandas.Panels`, and then has a few conveniences for analysis.

Basically, the issue is this: Curves+ will analyze a snapshot, but it will gives *all* its analysis at once. You can't ask for just part of the analysis. Given that (1) it gives some of the best results for the groove widths, and (2) it would be a bit of a pain to reimplement its groove width algorithm, we'd rather just re-use what they do.

So how do we do that in OPS? Consider the following:

Assume that maybe you wanted 2 numbers out of the whole Curves dataset. Each of these is an OPS CV. In principle, the easiest thing is to run Curves once for each CV. That will work, but that will obviously be computationally wasteful, and probably pretty slow.

So the question becomes whether we can use the fact that I've already generated a `pandas.Panel` (well, technically several panels) and cache that result for use in later order parameters. Even taking the string output from Curves and caching that might be useful, but the point is to avoid re-running Curves for every CV.

I think this general approach of using an external program which calculates several collective variables (or at least extracts parts of what are needed for them) might be desired in a lot of specialized situations. It would be nice to have a clean way of handling that.

jhprinz:
I think this is now supported by the new storage...

dwhswenson:
Self-assigned as a reminder to implement a check. In Skype call with @jhprinz, we *think* this has been fully implemented.

jhprinz:
This is already implemented. Will close this now. You can define arbitrary value CVs and store these without caching. You can also design cached CVs that are dependent on other CVs.

jhprinz:
Can I close this?

dwhswenson:
Per https://github.com/choderalab/openpathsampling/issues/259#issuecomment-173409947, I'd rather not close it until there's a test in place for exactly what I was thinking to do.

dwhswenson:
[...]

PR #264, opened 2015-05-17
[WIP] Allow CVs that do not store the cached values

jhprinz:
This will solve PR #259 by allowing a special CV that can be stored with its definition (the way values are computed), but the concrete values will not be stored. 

The way pyña would be used is then create a pyña CV that knows how to handle the external computation and will not store its pandas frames. A second CV can access single values and will know which value to pick and can also be stored.

There is a little caveat to make this work because of python scopes, but it is not as hard as I thought. 

Will leave it like this and finish once the bigger refactoring is done.


jhprinz:
This should be done when travis passes. Must be merged after #251 and solves part of #259.

jhprinz:
Cool. The PR adds a way to store a CV without storing its cached values. Very useful for CVs that do not return floats, but are useful. Consider the following example

create a CV that creates a tuple of the two first x coordinates (of atom 0 and 1). It can be saved, but it will not store its cached values
```
cv = paths.CV_Function('myfnc', lambda x : (x.xyz[0][0], x.xyz[1][0] ), store_cache=False)
```
create two dependent CVs
```
cv2 = paths.CV_Function('my1', lambda x, cv : cv(x)[0], cv=cv, store_cache=False)
cv3 = paths.CV_Function('my1', lambda x, cv : cv(x)[1], cv=cv, store_cache=True)
```
one with storage and the other without. Note, that cv2 and cv3 use `cv` as an external argument. This has the effect that the function stays a pure function (no use of external variables.
Arguments will be stored the way they were set at creation, so if you store cv2 or cv3 also the dependent `cv` will be stored!
```
storage.save([cv2,cv3])
```
Let's compute some values and store these for cv3
```
cv3(engine.template)
cv3.sync(storage)
storage.sync()
```

If you want to implement pyna now this would have to look like this:
```
def pyna.from_snapshots(list_of_snapshots):
    """
    Compute the pandas df from a list of snapshots
    Parameters
    ----------
    list_of_snapshots : list of openpathsampling.Snapshot

    Returns
    -------
    list of pandas.Dataframe (or similar)
        Important is for this to work that it outputs a list that is of same length as list_of_snapshots. 
        What you return per snapshot is arbitrary!

    Notes
    -----
    This function should be part of a package that can be loaded and should be locally defined!
    """

    return [ ... ]

pyna_cv = paths.CV_Function('pyna', lambda x : pyna(x), store_cache=False)
single_cv = paths.CV_Function('pyna1', lambda x, pyna_cv : pyna_cv(x)[0], pyna_cv=pyna_cv, store_cache=True)
```

jhprinz:
I thought to add two more feature to CVs.
- a CV requires a testing snapshot that is used to determine if it is storable, the potential unit, the dimension, etc and,
- a flag (could autodetermined from the testing snapshot) that states whether the function uses lists of objects of single objects. That makes it faster and MD_Function already uses that.

The second is trivial to implement and the first requires some test adding which would be very nice.

jhprinz:
Okay, this works, but unit support is still missing, which I will postpone to future. This is not really needed right now. Means, a function needs to output a number or a list of numbers to work, not a Quantity!

Issue #257, opened 2015-05-15
Delegate trajectory properties to sample


jhprinz:
I realized that we often use a sample as a trajectory since it basically is a trajectory with additional
information (although samples can have empty trajectories or even None).

This would mostly allow to write
```
len(sample) # length of the trajectory
sample[slice] # same as sample.trajectory[slice]
```
in addition, use samples in order parameters and ensembles
```
ensemble(sample) # is ensemble(sample.trajectory)
# note that 
sample.valid = sample.ensemble(sample.trajectory)
cv1 = CollectiveVariable()
cv1(sample) = cv1(sample.trajectory)
```

I think this makes scripts more readable and shorter. In the long run, we might be able to use trajectories only internally with no need to expose these. Like we do with configuration or momentum and also with MCStep

dwhswenson:
I think this is a very good idea. I think the main usefulness would be that we could, in many cases, directly use:

```python
function_that_takes_trajectory_argument(sample)
```

I can't count the times I've tried to use a sample as a trajectory when showing live code to Peter. If *I'm* doing it (given that I should know better), then that's probably the behavior a user would expect.

Note that, in one interpretation, this means that `Sample` becomes a `Trajectory` with extra metadata. I don't think that version should be reflected in the implementation, but it might be worth trying write code that supports that interpretation.

dwhswenson:
Closed by #265.

PR #265, opened 2015-05-17
allow simple trajectory access for samples

jhprinz:
#### New features to use samples as trajectories
- [x] use samples in CVs 
- [x] access the trajectory using slicing
- [x] len(sample) return length of trajectory
- [x] iter and reversed works on the trajectory


jhprinz:
This should be ready for review and merge.
I basically call the code from Trajectory

jhprinz:
Closes PR #257 .

dwhswenson:
Merging, but with the following caveat:

I'm concerned about the amount of explicit type-checking that is growing in our code. This is generally viewed as unpythonic. Python expects us to use duck-typing to our advantage. The standard Python approach is to use try-except clauses whenever possible, not if-else type checking. EAFP over LBYL ("easier to ask forgiveness than permission", "look before you leap"). Google "python type checking" and you'll see nearly every result says "don't do this" before showing how you *can* do it.

Not something that necessarily needs to be changed now, but implementing it the more pythonic way now will prevent rewrites in the future.

jhprinz:
> I'm concerned about the amount of explicit type-checking that is growing in our code. This is generally viewed as unpythonic.

Maybe, I prefer to explicitely write checks for things that I know can go wrong. If you want to use the try except approach then you need to start adding all exceptions to the API definitions. And should actually write tests for what can go wrong.

Consider a function that that throws an exception for some reason and you catch it. Then you need to exactly know the possible reasons for that to happen. Otherwise your action might be based on a false assumption and your code continues and noone knows what happens.

An example from your code
```
try:
    hist = self.histograms['max_lambda'][ens]
except KeyError:
    run_ensembles = True
```

If I read the code I wonder what did he want to do:
1. check if `end in self.histograms[''max_lambda]`
2. check if `'max_lambda' in self.histograms`
3. both

Obviously, 3. is what the code does, but first thing I though when reading the code was 1.

I think you get what I want to say. But we should be careful when to use it. For example at some parts I am using isinstance

After reading a little more about this I understand that this is a typical thing to do in python. Performance wise in some cases `try` if faster in other cases `if`. So to do it right we can use at some places, but not necessarily always, even though it is considered pythonic. In some cases (as I learned from two excellent programmers recently) these pythonic things are not, what is used in more modern languages. 

Personally I like reading in code that calls function with valid parameters and check for that then hope that the function will report correctly what will go wrong afterwards. 

Maybe I should use it more. I guess it is just a habit. Last comment from the programmers about Duck-Typing "You don't wanna do this" ;)

Although I like it! I will give this little more thought in the future.

dwhswenson:
> So to do it right we can use at some places, but not necessarily always, even though it is considered pythonic.

I would argue that general rule should be to use approaches that are considered "best practices" for the language you're using, unless there's no way to accomplish that. Try-except constructs are not *always* a best practice in Python. However, explicitly checking type at runtime is *absolutely not* a best practice in any programming language I'm familiar with. Either you trust that you have the right type and count on errors to arise if you don't (dynamic typing, as with Python) or you check the type at compile time (static typing, as with C). You can argue that static typing is better than dynamic, but once we chose Python, static typing wasn't an option. (Until 3.5, when I guess they'll be implementing best-of-both-worlds gradual typing.)

There are cases where type checking can't be avoided (serialization is one example). But in general, we should not need type checking. If type checking would identify an error, then the code should raise an error anyway. If we're avoiding errors in special cases, it is better to use exception catching for both speed and clarity (the `try` section usually implies "this is what I expect to happen frequently").

> In some cases (as I learned from two excellent programmers recently) these pythonic things are not, what is used in more modern languages.

But they **are** what is used in Python. And our code is in Python. That's the point. Scala, for example, is a statically-typed language. Because of that, the best practices can be very different from they are in Python. Scala is also fundamentally a functional language with object-oriented features, whereas Python is fundamentally an object-oriented language with functional features. This again leads to different best practices.

> Last comment from the programmers about Duck-Typing "You don't wanna do this" ;)

That's a bit ironic, given the subject of this PR. What's the actual complaint? Is it about static vs dynamic typing (in which case the only solution is "don't use Python") or is it about method name collisions (which can be a problem, but is part of the reason that I'm always arguing for clear usage of terms and usage.) Those are the only arguments against duck typing that I'm familiar with.

This PR is an example of a good use of duck-typing. You made it so that samples quack like trajectories. This means that, if I now write `ensemble(sample)`, I get the same result as `ensemble(sample.trajectory)`. And you did that by changing `Sample`, not by changing code in every single one of our ensembles to check whether the input is a `Sample` or is a `Trajectory`. You also got immediate support for this behavior in plenty of other functions: I assume that `trajectory.shared_configurations(sample)` will work now. You got most of the support for `CollectiveVariable`s for free as well -- my point is that you *shouldn't* have had to change anything in `CollectiveVariable` either. Duck-typing (possibly combined with some type conversion) should reduce the amount of code in exactly that kind of case.

jhprinz:
> However, explicitly checking type at runtime is absolutely not a best practice in any programming language I'm familiar with.

I think that depends, especially if you look at the argument I had with that try except might fail because of the wrong reasons. All I wanted to say is that we should not try to replace all if decisions, which might have a good reason with try except. Exceptions should be exceptions.

I think you got me a little wrong. I said that in general dynamically typed languages have the problem of not doing what you expect and this means if you fiddle with your code you need to be careful. Try except might lead to unstable code while being more readable or shorter. 

In some cases type checking seemed the right way to fix some duck typing issues, like using samples in cvs. There are some caveats that require testing the type because just trying will be bad practice and might fail very late in a computation, while a simple check would prevent the whole cascade rom failing. 

> Last comment from the programmers about Duck-Typing "You don't wanna do this" ;)

This was a bit ironic, because I like using Duck-Typing and they are actually scala programmers :) and of course there are still problems with type-safe languages. 

All I wanted was not to always use try-except because it looks ugly to me. In theory, if your code it very good you should need neither, but this requires a few more iterations. 

> Those are the only arguments against duck typing that I'm familiar with.

The reason I mean is that you use a function that is supposed to do the something and replace it by something that is supposed act the same way. But I am using the function wrong with potential useless or false parameters and it is even hard to tell what the original function does and that is fails properly. Even more difficult to assume that the duck-typed will behave wrongly in the same way!

I also agree that we should be pythonic as possible, but still in some cases I want to not do this at any cost because in these cases it will lead to the wrong results. Let me explain the reason I used it here. Because that was your initial concern.

```
elif item_type is paths.Sample:
    if len(items) == 0:
        return []
    elif len(items) == 1:
        return [list.__getitem__(items.trajectory, 0)]
    else:
        return list(list.__iter__(items.trajectory))
```

Here I need to explicitly check because in some cases (only Sample and Trajectory) I need to treat it differently. Problem is that we allow to use a single Snapshot, a list of snapshots and the reduced form of snapshots which are a tuple. Now there is right now no other way to check, if it is just in iterable of snapshots or a tuple which is treated as a single snapshot. I agree that there are better ways of internally handling this. All of this is to make this fast loading with need of coordinates work.

That is the only reason it is used here. Otherwise I still would not have used try because my implementations usually do not take care of throwing everywhere correct exceptions. 

Do you get the point? You last comment turned into a lecture about duck-typing is a good thing because we use Python and all I wanted to state in the end is that we should be careful and our code is (maybe only my part) is not written to be used safely with try-except.

Anyway, as I said last, I will give this issue more thought.

dwhswenson:
Oh... no wonder you didn't reply to my comment. I forgot to click "comment" to post it. At least, it isn't here!

A few things I remember saying in it:

> they are actually scala programmers :) 

Well, they didn't like duck typing, and they referred to programming languages "more modern" than Python -- it was pretty easy to guess. ;-)

> You last comment turned into a lecture about duck-typing is a good thing

Sorry about that. I misread what you'd written as saying you didn't think we should use duck-typing, which was pretty weird on a PR where you'd implemented some duck-typing!

To the substance:

> try except might fail because of the wrong reasons

If this happens, you're writing bad `try` clauses. The `try` clause should be very short. The whole thing is try-except-else: if you have stuff to do if it doesn't fail (but not if it does), then you use the `else`. This allows you to keep `try` to a couple lines at most. And you catch each exception specifically, acting differently if you get a `TypeError` than an `AttributeError`. At worst, you end up with the same number of lines as in an if-else block, but you naturally follow the most likely path. I don't see any way it is less safe than LBYL. 

 > But I am using the function wrong with potential useless or false parameters and it is even hard to tell what the original function does and that is fails properly

I'm not sure what you mean here. I mean, a user can *always* put unacceptable parameters into a function in Python. At some point, that should raise an exception. That's what we expect. Then you trace up the stack and see where it went wrong.

> Here I need to explicitly check because in some cases (only Sample and Trajectory) I need to treat it differently. Problem is that we allow to use a single Snapshot, a list of snapshots and the reduced form of snapshots which are a tuple. Now there is right now no other way to check, if it is just in iterable of snapshots or a tuple which is treated as a single snapshot.

This is where I had a question: I'm not clear exactly what this function is doing. Why does the output differ? Is there a reason not to always make it return a list?

(Since I failed to ask that question, you never had a chance to answer it -- here's some of what I've played with since.)

Even aside from the question of try-except and type checking, I keep trying to point out that there are often ways to reuse code. This function is a great example. Instead of writing a separate branch of an if-statement for every possibility, try reducing the number of problems by making them equivalent. For example, instead of adding a whole new block for the case of `Sample`, you could have started by saying, "If I have a `Sample`, let's turn in into a `Trajectory`." I tried this, and it leads to less code and a speed boost for everything:

```python
def mid_pre_item(cv, items):
    item_type = cv.store_dict._basetype(items)

    if item_type is paths.Sample:
        items = items.trajectory
        item_type = paths.Trajectory
    if item_type is paths.Trajectory:
        if len(items) == 0:
            return []
        elif len(items) == 1:
            return [list.__getitem__(items, 0)]
        else:
            return list(list.__iter__(items))
    elif item_type is paths.Snapshot:
        return items

    elif hasattr(items, '__iter__'):
        return list(items)
    else:
        return items
```
Timings:
```
Empty trajectory:
Old version: 0.260112285614
Mid version: 0.225152254105

Empty sample:
Old version: 0.329460859299
Mid version: 0.251954078674

Snapshot:
Old version: 0.203765392303
Mid version: 0.197712421417

Single frame trajectory:
Old version: 0.340366363525
Mid version: 0.294761896133

Trajectory:
Old version: 0.905314207077
Mid version: 0.877851724625

Sample:
Old version: 1.1665391922
Mid version: 1.0147023201
```

I'm also not sure about the special cases within the `paths.Trajectory` case. I replaced that whole block with `return list(list.__iter__(items))` and got the same output, with some performance cost to single frame cases and empty trajectory cases, but a (smaller) performance boost to trajectories and samples (plus simpler, reduced code). I'm not sure which case we need to optimize for, though.

All of that is in [an ipynb](http://nbviewer.ipython.org/urls/gist.githubusercontent.com/dwhswenson/a89fdb01801fb97a63d7/raw/9007cdec919bed321eb4c0d295ca66e16fd7dc97/speedtest_pre_item.ipynb) that I just put up in a [gist](https://gist.github.com/dwhswenson/a89fdb01801fb97a63d7). The pure try-except approach was only much faster for Samples (where it doubled the speed), but Samples won't be the most common case.

jhprinz:
Okay, let's make this short. :) I think we talk too much about this.

All I am saying is that there are cases where try, except are not save to use. I just had another place where try would not have thrown any error but given the wrong result. Which means of course that I am using Duck Typing not correctly. But that was the point. Your code need to be safe for these type of errors and since I did not use enough or appropriate raise xyz enough and not even sure if I used the right Error I wanted to point out that we need to be careful. 

So my argument is : Because I never really used try, except we need to be careful. Iam not saying it is bad or better. It can lead to cleaner and shorter code, but you need to know what you are doing (which I did not necessarily in the past!)

The trick at the end is nice. I think we should add that. Actually I think we can even improve on that. I will try to keep that in mind and fix at some point.

Was a nice discussion though!

PR #266, opened 2015-05-18
MSTIS Network Analysis

dwhswenson:
Analysis for MSTIS networks. Also includes improvements to `TISTransition` analysis tools.

- [x] `MSTISNetwork.rate_matrix`
- [x] Add `toy_analysis.ipynb` to testing suite
- [x] Add `mstis_analysis.ipynb` to testing suite
- [x] Move transition copying functionality over to `Transition` objects
- [x] Bring `network` and `transition` analysis function calls and set-up into agreement
- [x] Move `summarize_trajectory_volumes` to `Trajectory.summarize_by_volume`
- [x] Add docstrings for analysis objects
- [x] Separate `move_summary` from `TISTransition` objects (requires #251)


dwhswenson:
The last step of this is to move the `move_summary` calculation out of the `TISTransition` object. Since it involves a few subfunctions, it seemed reasonable to move that into a class. And I've been thinking about making a `MoveScheme` class (to make it easier to modify the move scheme; issue on the subject coming soon). Because of that, I decided to name the class which holds these functions as `MoveScheme` -- after all, this is the analysis of the how many times each move gets accepted, which is basically an analysis of the move scheme, right?

Anyway, that name can change if you don't like it, but I thought the choice needed some explanation.

dwhswenson:
Passed Travis tests. Ready for review and merge.

jhprinz:
From what I can tell this looks fine. You moved some stuff to separate files which seems to be a good idea.

Will merge now...

PR #268, opened 2015-05-21
MISTIS setup

dwhswenson:
Multiple interface set TIS.

- [x] Upgrade minus ensemble to multiple interface set minus
- [x] Upgrade minus mover to multiple interface set minus (waiting on #251)
- [x] MISTISNetwork builds movers (except minus/outer)
- [x] MISTISNetwork builds minus
- [x] MISTISNetwork builds multiple state outer
- [x] Storage support for MISTIS
- [x] Include MIS Minus ensemble and move as part of MISTISNetwork
- [x] Check whether any minus calculations need to be adjusted for multiple interface set minus
- [x] Test MISTIS on simple system
- [x] Add mistis_setup.ipynb to tests
- [x] Fix edge cases on input that cause tests to fail
- [x] Speed up creation of initial trajectories


dwhswenson:
After discussion with @bolhuis today, we decided that the correct way to implement the process of getting the flux from the multiple interface set minus move is a bit worse than we'd originally thought. So we'll instead require that MISTIS calculations use a separate flux calculation.

I might someday add the correct flux calculation, but it requires something much more complicated than the current multiple interface set minus ensemble. (The existing version can still be used to assist in decorrelation and to facility replica exchange across states: this is also all that we used it for in our paper introducing it.)

dwhswenson:
The main problem I'm having right now is with setting up trajectories to use as input to the MISTIS calculation. MISTIS itself seems to run fine, but I can't always get it started. There are two reasons for this:

#### Too slow

I'm having problems with getting these tests to complete: it seems that some of the ensembles are scaling badly again (in particular, things get really slow when I have `SequentialEnsemble([stuff_for_two_states]) & AllOutXEnsemble(other_state)`, which is the easiest way to describe some of this MISTIS stuff). 

To fix it, I'm just saying that all functions which are allowed as "continue functions" in `dynamics_engine.generate` need to have the signature `func_name(trajectory, trusted)`, instead of just `func_name(trajectory)`. Once we start running the dynamics, the trajectory is always trusted (the very first trial in `generate` is not trusted, of course). 

In general, this is probably a better place to manage identifying "trusted" trajectories that by using the caches in the ensemble code. I don't want to remove those yet, however: among other things, I think that the `SequentialEnsemble` cache can be really useful in debugging.

There still seems to be something slowing it down, but my guess is that before this fix, it was $O(n^3)$ and now it is $O(n^2)$. Right now, this only appears in the process of finding a first trajectory during bootstrapping, so the additional cost is manageable. But that needs to be hunted down at some point.

#### Incorrect input trajectory

I still need to work on ways to guarantee that input trajectories here are always acceptable. It is a lot harder to automatically bootstrap correct trajectories in this MISTIS example than in other cases. I'm working on finding the cases that can create bad initial trajectories and avoiding them.

dwhswenson:
Ready for review and merge.

The speed fix was just the removal of 5 characters (`trusted`, not `self.trusted`). I think I've got most of the special edge cases handled correctly for getting an initial trajectory.

Also, we're about to hit 1000 Travis builds. :-)

openpathsampling/pathmover.py
```diff
@@ -1616,17 +1616,14 @@ class MinusMover(WrappedMover):
...
-    classes, the calls to `super` will use the version in
-    ConditionalSequentalMover. However, analysis routines will see
-    `isinstance(minus, ReplicaExchangeMover)` as True.
     """
-    def __init__(self, minus_ensemble, innermost_ensemble, 
-                 ensembles=None):
+    def __init__(self, minus_ensemble, innermost_ensembles, ensembles=None):
         segment = minus_ensemble._segment_ensemble
+        try:
+            innermost_ensembles = list(innermost_ensembles)
```
jhprinz: This is a use of try and except. I like that!

openpathsampling/volume.py
```diff
@@ -10,12 +10,24 @@
...
+        empty
+    """
+    volume = EmptyVolume()
+    # EmptyVolume is smart and knows its OR just takes the other
     for vol in volume_list:
-        if volume is None:
-            volume = vol
-        else:
-            volume = volume | vol
+        volume = volume | vol
```
jhprinz: I just discovered a great website about python that covers lots of features and great packages.

http://people.duke.edu/~ccc14/sta-663/

There the function 'reduce' was mentioned what interatively applies a function reusing the result. While `map` applies each element to the function, reduce takes two elements and uses the result and the following element again. 
So you could write your function as
```
return reduce(lambda x,y : x | y, volume_list)
```
or 
```
return reduce(lambda x,y : x | y, [EmptyVolume()] + volume_list)
```
if you want to allow for empty lists. The lambda could be replaced using the `operator` module
```
import operator as op

return reduce(op.or_, [EmptyVolume()] + volume_list)
```

jhprinz:
Phantastic. Don't forget to look at 

http://people.duke.edu/~ccc14/sta-663/

Merging!

openpathsampling/pathmover.py
dwhswenson: :-) I've been trying to find a good way to "clean" the input so that it is a list. This is the version I like the most. As I come across others, I think I'll change them so that they are like this.

openpathsampling/volume.py
dwhswenson: That would be better. I've been using `reduce` in a lot of other circumstances, but I don't usually think in terms of it. I find it harder to read (I almost always add a comment to explain what I'm doing with it), but in this specific case, the function is then just a clearly-named wrapper for a commonly used one-liner.

I won't worry about changing it in any immediate PR (just not a high priority), but I am in favor of the change.

jhprinz: That's why I merged it. Just wanted to make sure you know about reduce (which I didn't until recently).

Issue #150, opened 2015-02-07
Remove global PathMover.engine


jhprinz:
Only Shooters need an engine (so far), so we add the engine to the shooter as an option or
add it as a class variable to ShootMover

The first approach would allow to store the engine, but store the engine with each mover separately.
The second approach is simpler because it only need to set the engine once, but we would need to store the used engine somehow separately

dwhswenson:
Note that this also needs to be added to classes which use the ShootMovers: that makes it a little more complicated to use initialization arguments, since it would require changing `OneWayShootingMover`, `OneWayShootingMoverSet`, `MinusMover`, and possibly others.

On the other hand, I feel weird having it be a class variable. Is it possible to have two Engines at the same time? You could imagine a situation where you want to combine a path sampling approach with temperature replica exchange. In that case, you need a different engine for each T (at least, the way we have engines implemented now).

jhprinz:
Hmmm, it would be kind of cool to allow for multiple engines, although I don't know if our current setting would allow for multiple openmm instances. But in principle we could just program a regular Temperature RepEx algorithm. Interesting. No, wait, we would do Temperature RepEx in Trajectory space. Which could turned into regular RepEx if we only shoot and continue at the old positions.

Just hypothetical. WOuld it make sense to add the temperature and combine TIS trajectories at various temperatures? I don't see why that should not work.

Interesting and easily doable once we have released the first version.

dwhswenson:
When she was here for MolSim, Toni and I talked about the idea of using her TRAM (now xTRAM, if I remember what Frank told me?) in combination with RETIS. If I remember correctly, it was basically a way of handling replica exchange along two axes (by creating sort of "blocks" of one axis within the other? is that right?). So the thought was that there might be some way to do that where one of the axes was RETIS. Not sure if temperature is exactly the right thing to do, which is why I suggest keeping in generic.

In any case, I think that for parallelization we'll eventually take an embarrassingly parallel approach, and farm trajectories out to different nodes. This will probably require a different `DynamicsEngine`, and *that* engine can do the work of making running different instances of OpenMM. So when we really need it, there will probably be a way around it. Just a thing to think about.

dwhswenson:
Related note: whether the `engine` is associated with instances of `ShootMover` or the class `ShootMover` or the class `PathMover`, it seems to me that we don't actually need it (in general) for `Calculation`s.

Currently it is used in initialization of a calculation, but I don't think anywhere else. Should I remove that?

[...]

PR #269, opened 2015-05-23
Engine.snapshot_timestep

### LINKED ISSUE #255: Universal access to timestep as `engine.snapshot_dt` or `engine.snapshot_timestep`
Some analysis routines (flux, in particular) need to know the timescale. That information is currently stored in the engine, but in an engine-dependent manner. I think we should make that accessible as `engine.snapshot_dt` or `engine.snapshot_timestep`, instead of `toy_engine.dt` vs `openmm_engine.options['timestep']`.

Also, this needs to be not the internal timestep, but the time per frame, so `dt*engine.nsteps_per_frame`.

The only catch is for dynamics that report frames at variable timesteps. I *think* path sampling can still make sense there, though the algorithms to calculate the flux would be different (you'd need actually total times instead of number of frames). But I don't think that's a scenario that needs a lot of worry at this point.

I would implement this as a get-only `@property` in each engine. It would be part of the required API for any new engine (since common analysis routines won't work without it).

Any opposition to this approach?

dwhswenson:
Resolves #255.

- [x] Add `snapshot_timestep` property to OpenMMEngine and ToyEngine
- [x] Add tests for `snapshot_timestep`
- [x] Use `snapshot_timestep` in RETIS flux calculations

Requires ability to extract the `engine` used by the `MinusMove`, which should be in the solution of #150 (which I guess means this is waiting on #251).


dwhswenson:
Passes Travis; ready for review and merge.

Side note: I noticed a potential problem in the automatic identification of the outermost value of lambda. The way it currently works, it can pick the wrong value if there isn't enough data. This probably isn't an issue in real situations: if you don't have enough data, you get the wrong rates for SO many reasons, and this won't be a problem once there's sufficient sampling. Besides, you can always manually assign the value of `outer_lambda` while doing the rate calculation.

jhprinz:
Looks good. We should keep track of this max_lambda value though. Do you know what the reason is?

jhprinz:
Merging...

dwhswenson:
> We should keep track of this max_lambda value though. Do you know what the reason is?

Sorry, meant to write a more full description later. It really is just a matter of insufficient sampling, and should never cause problems in the real world.

The `outer_lambda` is the value of the order parameter at the outermost interface (i.e., $\lambda_i$ for that interface). This is determined automatically based on the fact that, by definition, $P(\lambda | \lambda_i) = 1$ for $\lambda \le \lambda_i$. So, given results for $P(\lambda | \lambda_i)$, it is assumed that $\lambda_i$ is the largest value where $P(\lambda | \lambda_i)=1$.

The problem is that if you are massively under-sampled, then maybe $\lambda_i = -0.05$ but all values of $P(\lambda | \lambda_i)$ are 1 until $\lambda = -0.01$. The `outer_lambda` will be assumed to be $\lambda_i = -0.01$, but the total crossing probability `tcp(-0.01) != tcp(-0.05)`. Since this is one of the factors that goes into the rate, the rate will not be what you would expect it to be.

So the obvious question to ask is, what if $P(\lambda | \lambda_i)$ really *is* 1 beyond the $\lambda_i$ defined by the interface? This absolutely can happen (to within sampling, at least): if every single path that reaches $\lambda=-0.05$ also reaches $\lambda=-0.01$, then this will be true. In fact, this is exactly why the crossing probabilities histograms level off: once you're far enough past the barrier, (nearly) every trajectory continues into the other state.

But note what this means: not only does $P(\lambda=-0.01 | \lambda_i) = P(\lambda=-0.05 | \lambda_i)$, but the same would be true for *any* interface $i$, meaning that the total crossing probability would be the same at these points (again, see the leveling off of crossing probability reflected in the total crossing probability). So it wouldn't matter in the calculation.

The reason it messes up the rates when you don't have enough data is not just that $P(-0.05 | \lambda_i) = P(-0.01 | \lambda_i)$, it is that you also have $P(-0.05 | \lambda_{i-1}) \ne P(-0.01 | \lambda_{i-1})$: in some other interface, this ratio is different. This means that the non-equality shows up in the WHAM-combined total crossing probability.

The ratio of crossing probabilities at these two location should always remain the same -- that's the basic principle behind the splitting approach in TIS. So this is just an artifact of (very) insufficient sampling.

PR #270, opened 2015-05-24
Trajectory summary

### LINKED ISSUE #247: Summaries of each move

dwhswenson:
Resolves #247.

Most of that is already dealt with in other branches. All that this PR does is to add `Trajectory.summarize_by_volumes_str`, a string version of the summarize-by-volumes list.

In https://github.com/choderalab/openpathsampling/issues/247#issuecomment-102423790, I suggested we might also identify volumes of interest in the TIS transitions. I think this is already done: the user has access to the list of `interfaces`. If the user wants `transition.interfaces[i] & ~transition.interfaces[i-1]`, that's pretty easy to generate manually; I'm not sure that an extra function will make it any clearer (especially since I'm not sure what to call that function).

Tasks

- [x] Add `Trajectory.summarize_by_volumes_str` and tests
- [ ] Wait for #263 and #266 to be merged (requires parts of both of those)


dwhswenson:
This should be reviewed and merged immediately after #263. It requires some of that PR, and I've made it so that it should be able to automatically merge after that PR is merged.

Only involves a few small changes beyond #263, so after that's merged, review for this should be easy.

jhprinz:
merging...

PR #271, opened 2015-05-25
Block averaging preparations

dwhswenson:
Some initial work on the numerics which I think will make it easier to implement block averaging of analysis in the near future.

Tasks:
- [x] LookupFunction plotting in matplotlib
- [x] LookupFunctions support numpy ufuncs
- [x] Modify histogram functions (.normalized(), .cumulative(), etc) to return LookupFunctions
- [x] Add LookupFunctionGroup class, with ability to determine mean and std
- [x] Tests for LookupFunctionGroup
- [x] Docstrings in `lookup_function.py`


dwhswenson:
This is ready for review and merge. Main improvements are as follows:

1. `LookupFunction` is easier to plot in matplotlib (e.g., `plt.plot(wham.x, np.log(wham))` instead of `plt.plot(wham.keys(), np.log(wham.values()))`, and supports numpy ufuncs, and output to `pandas.Series`.
2. Histogram functions (`.cumulative`, `.normalized`, etc) now return a `LookupFunction` instead of just a numpy array. This makes them easier to plot, and will make them easier to average as part of block averaging. They can set their x-values to the left bin edge (`"l"`), the right bin edge (`"r"`), or the midpoint (`"m"`). Cumulative and reverse cumulative default to "r" and "l", respectively, and others default to "m" -- but the user can change this with the `bin_edge` argument.
3. New class `LookupFunctionGroup`, which takes several `LookupFunction`s and gives convenient ways to calculate the mean and std dev. Individual constituent functions can be accessed using `__getitem__` notation: `func0 = group[0]`. Calling notation acts as a call to the mean, so in principle the group looks just like a `LookupFunction` of the mean, with the additional property of `std` to calculate the standard deviation.
4. More tests and docstrings.

dwhswenson:
Assuming this still passes tests (it should) this is back to being ready to merge.

dwhswenson:
... and once again, after cleaning up from the last PR merged in, this one is ready for review and merge.

jhprinz:
Nice. Quick question about the implementation of the LoopupFunction. Do you plan to subclass this from numpy.array?

dwhswenson:
> Do you plan to subclass this from numpy.array

No. `LookupFunction` is, to the user's eyes, a function -- specifically, a 1D function (for the foreseeable future). So it should act like `sin` or `exp`, not like an array. Many of the `numpy.ndarray` behaviors are completely irrelevant here (how do you `sort` the sine function? what is the `std` of $x^2$?)

The main thing we want is support for numpy's ufuncs, so that we can write things like `np.log(func)` (or even, and I should check that this works, `np.log(func(xvals))` for arbitrary `xvals`). This is done by implementing `__array__`, `__array_wrap__`, and `__array_prepare__` (and I think `__iter__` as well).

If I subclassed `numpy.ndarray`, I'd still have to rewrite (most of) those for my subclass (plus a few others; `__array_finalize__` at least). And I'd inherit methods that are nonsensical in this context. That's why I wrote it as I did.

Some of the statistical functions make sense for `LookupFunctionGroup`, but again, the main point is that the group should mostly look like a `LookupFunction` (for the mean of the group). So it inherits from `LookupFunction`, and adds some extra features. Assume that `group = LookupFunctionGroup([func])`, so there's only one function in the group. So, for any `x`, `group(x)==func(x)`. It would be weird for `func.std()` to return a number which is different from `group.std()`. (Here, `group.std()` should return `0`, and `func.std()` should error.)

Note that the behavior of `group[item]` to select a specific subfunction of the group is the reason not to implement `LookupFunction.__getitem__`: it would be weird for `func[item]` to return a number while `group[item]` returns a function. The desired behavior is that either `func(x)` or `group(x)` returns a value, while `group[num]` return a function and `func[num]` errors.

To summarize, I considered subclassing `numpy.ndarray`, and explicitly chose ***against*** it.

Issue #272, opened 2015-05-28
Mover need to be ensemble specific


jhprinz:
#### Question:

Would it be okay or desirable if all movers need to be told which ensembles to use explicitly on creation?

#### Example
```
randomrepex = ReplicaExchangeMover([[ens1, ens2] , [ens2, ens3], [ens3, ens4]])
```
would have to be written explicitly
```
randomrepex = RandomChoiceMover([
    ReplicaExhangeMover(ens1, ens2),
    ReplicaExhangeMover(ens2, ens3),
    ReplicaExhangeMover(ens3, ens4)
])
```
so that each mover knows exactly which ensembles it acts upon and creates. Very useful for cleaner implementation using input and output ensembles

dwhswenson:
If I understand, you're asking whether we should say that each instance of a mover is tied to the most atomic level of trial; i.e., the instance can only represent a move for a single, specific ensemble for shooting/path reversal, or a specific pair for replica exchange, or a specific to-from pair for hopping. Is that correct?

I'm very much in favor of this. When we were first implementing this, I thought that others didn't like that approach, so I tried to implement things in such a way that it could work either way (but then made the default setups so that they would use this approach, because it adds a lot of flexibility).

This also simplifies the code significantly. I wrote some fancy stuff to select the correct samples. It was a pain to wrap my head around that code when I wrote it, and I really hoped it was correct, because trying to understand it again to fix it later would be a nightmare.

The only caveat is that, in that sample selection process, you still need to keep in mind the possibility that there is more than one sample in a given ensemble. It should get a list of all active samples in that ensemble, and pick one of them at random. (Think AMS, for example, where you need pick a random trajectory in the ensemble to start your next trajectory from.) But I think that's what SampleSet gives you anyway, isn't it?

jhprinz:
Precisely, that was my idea. And the problem I had is, that the code gets longer and simple things create long lists. On the other hand it is much clearer and will remove the kind of hidden selection of ensemble pairs and put it explicitely into the move scheme. 

As you said, writing a simple to use selection is very diffiult and in the last PR I ran into the same problem (with no satisfying answer). 

The main reason I had to require this is that it make the input output ensemble treatment much simpler.

[...]

PR #273, opened 2015-05-28
[WIP] Shoot and Extend

jhprinz:
Solves making a general ExtendByShooting Mover still missing from PR #251.

Forward and Backward, Shooting and Extending is basically now one mover.


jhprinz:
I would wait with this one after merge of PR #272 .

jhprinz:
This is a VERy old PR. Maybe we will close this and redo it at a later stage. All it does is doing some restructuring to turn the Shooting Moves into one move and after all the changes we have made I fear that trying to fix this will cause more trouble than just redoing it after we have done the major changes.

It is only to make the code more reusable but will not add features or change the API.

dwhswenson:
Given that this PR is sort of dead, how about I do this as part of the pathmover overhaul I'm about to start? The point is that this would give a superclass to contain all the detailed balance stuff I'll be adding, so that code doesn't have to be repeated in the convenience subclasses.

dwhswenson:
(In other words, unless you disagree, let's close this without merging)

jhprinz:
Yes, please do so. I was not so happy with my implementation anyway and fixing this will take longer than doing it from scratch!

Issue #153, opened 2015-02-09
Renaming


dwhswenson:
In a way, this is a continuation of #24, but we'll try to do it in a more methodical way. Feel free to edit this initial post to add more things to the task list. I'm treating this as a list of things I'm concerned might be confusing or might take a better name; some of them we might agree not to change.

I'll put discussion points for each of these in h4 (`####`) in comments below.

- [x] `PathMover` and `MovePath` confusion
- [x] `Momentum` doesn't contain momenta -- agreed that it doesn't matter at this point
- [x] [agreed] `Snapshot.atoms` and `Trajectory.atoms` vs. `Topology.n_atoms` #160
- [x] [agreed] `Trajectory.frames` vs `Trajectory.n_frames` (or `Trajectory.n_snapshots`) 
- [x] [agreed] `OrderParameter` vs. `CollectiveVariable`
- [x] [agreed] `LeaveX` `InX` `OutX` `HitX` rename to `PartOutX` `AllInX` `AllOutX` `PartOutX`
- [x] [agreed] `AndVolume`/`AndEnsemble`, `Or`, `Sub`, `Xor` rename to `Intersection`, `Union`, `RelativeComplement`, `SymmetricDifference`
- [x] [agreed] `Calculation` rename to `PathSimulator`

#### fix in PR #274 
- [x] [agreed] `LambdaVolume` to `CVRangeVolume`
- [x] [agreed] `ForwardAppended` to `Prefix`; `BackwardPrepended` to `Suffix`
- [x] [agreed] `root_mover` to `move_scheme`; and `default_movers` to `default_scheme`

dwhswenson:
#### `PathMover` and `MovePath`

Obviously, these two names are too similar. I think we should leave `PathMover` the same and change `MovePath`, for the following reasons:

1. `MovePath` could be confused for `PathMover`, but not vice versa.
2. One might have a script with movers of things other than paths, so using just `Mover` is ambiguous.
3. "Path" is an overloaded term in this project. `MovePath` does not refer to trajectories, but to the path a move took.
4. `MovePath` is a record of the what happened in a move. So I think `MoveRecord` or `MoveLog` or something similar would be a clearer term to describe it. (I personally like `MoveRecord`, to avoid possible confusion with `logging`-related logs.)

#### `Snapshot.atoms` and `Trajectory.atoms` vs. `Topology.n_atoms`

[...]

PR #275, opened 2015-05-30
Rename

jhprinz:
Tried again


jhprinz:
This should resolve Issue #153 for now until new things pop up :)

Ready for review and merge when it passes. Be careful though. This changes the LambdaVolume which is used in a lot of notebooks. It might be better to just not merge potentially changed notebook and fix this by hand.

jhprinz:
Ready for review :) Merge whenever you think it makes sense. I will try to keep it mergeable when we add more PRs.

dwhswenson:
My plan will be to merge this right after #268. I'd like to get the notebooks from that in before the major name change.

jhprinz:
Sounds good. Let me know, I will probably have to fix it after merging #268 .

jhprinz:
I will merge #281 first and then fix this and then adapt my other PRs.

jhprinz:
alright. This is ready to be merged once it passes and #281 is merged.

jhprinz:
Interesting. There was an API change in mdtraj topology. I need to fix that. Hmmpf.

dwhswenson:
Any way to update it so that it is also compatible with older versions of MDTraj?

jhprinz:
the problem is that the Element object which we store for the ToyDynamics needs now more parameters. Not sure if I can fix that. At least version 1.4.0 breaks.

dwhswenson:
this would be a good place for a try-except ;-)

(Temporarily, until we can be sure that all our users are on more recent versions of mdtraj -- I'm personally running an older version, as is Peter, and I'd rather not deal with upgrading quite yet)

jhprinz:
exactly. ;) That is what I am doing right now. try and otherwise try to use Elements with a radius.

dwhswenson:
Now that this passes tests again, is it fully ready for review/merge?

jhprinz:
This should be ready to merge. It is only renamings of some functions. It can be merged now or later but will need fixing again (I assume)

dwhswenson:
Merging. Only comment is that we should consider the `move_scheme` name as it relates to the `MoveScheme` object in PR #285. One possibility is to rename the `move_decision_tree` in #285 as `move_scheme`. At first it seemed weird to me to say `MoveScheme.move_scheme`, but looking at examples (see new notebook in #285) it doesn't seem so weird.

Issue #258, opened 2015-05-15
"Forget" caching


dwhswenson:
Right now we cache everything, and the cache never forgets. Obviously, this won't work once we're dealing with hundreds of GB of data. (In fact, it's already getting into dangerous territory on my old laptop.)

I think the discussion of caching needs to split into two sides: sampling and analysis. Optimal solutions might be different for each of these.

One question to consider is how much we really need to cache positions and momenta. We want to keep the currently active trajectories while doing sampling, but we certainly don't need anything else. For analysis, I think we can do about the same: most analysis routines (especially any that will go to the samples) will loop over MC steps, and so will know the active samples for any step.

I'm a little concerned about scaling for this: it would be easy to fall into quadratic scaling. This should be linear. Looking up some of the algorithms used for garbage collection would be useful.

jhprinz:
I will look into a smart way for better caching once #251 is done.

jhprinz:
Should have been closed by #300 . Closing.

PR #276, opened 2015-05-30
[WIP] Least Recently Used (LRU) Cache

jhprinz:
Adding standard limited caching to storage and collective variables.

Will resolve Issue #258 .

I have implemented a simple LRU Caching.

- [x] add LRU to object cache
- [x] add LRU to collective variable caching
- [x] add LRU to collective variable store caching
- [x] add storage option to quickly change caching options on the fly
- [x] add special storages for production / analysis that will set the caching appropriately

Caching can already be changed during runtime. I have something like
```
storage.set_cache_mode('production')  # almost no caching
# run pathsampling
storage.set_cache_mode('analysis')  # maximum caching for cv, sample, pmc and sampleset
# run analysis
```
and  with special storages I mean
```
storage = AnalyticStorage("my_simulation.nc")
```
will open the store as usual, then set the caching to maximal for the usual stuff and run cache_all() to fill the cache


dwhswenson:
Nice! -- I would change `AnalyticStorage` to `AnalysisStorage`, though. "Analytic" gives me nightmares about my complex analysis course.

jhprinz:
Seems this is almost done. Bootstrapping works fine and fast with `AnalysisStorage(filename)` now.

dwhswenson:
@jhprinz : all the tasks are checked off, and there's no [WIP] tag, but you haven't said that this is ready for merge or assigned it to me -- should I be reviewing this one?

dwhswenson:
Oh, and glancing through this now: could you add (or maybe you already have) logging for cache misses? I don't think we have a sense of what those default sizes for the cache really *should* be: at least, I can only make guesses, and I assume your values are also best guesses. As we start to work with real systems, we can check logs of cache misses to get better default values.

I really wish there was a way to set this in terms of GB, but I don't think that would be easy. Still, maybe not too hard: we'd just have to implement `__sizeof__` for our classes. Even just doing that for the major offenders (basically, the stuff you're already putting limits on) might allow us to do something.

jhprinz:
Very good point. All values were just guesses and potentially bad ones. The GB idea is great. Just trying to implement this. 

Actually this was ready, but I would rather add some mode tests.

jhprinz:
Hmmm, I realized the problem to figure out how big a dictionary with all objects actually is requires to keep a list of all objects, really all somehoe referenced objects with their sizes to not overcount and always updating this list. Kind of like adding a second garbage collector. I think the overhead will kill us. 

Too bad, because I liked the idea so much. 

It is very hard to determine the actual size of the cache. It could be that the user has made another cache and so the cache itself if practically for free or removing items from the cache will not remove them from memory since there is still the other cache. 

I was thinking about using weak referencing instead of caching. Would that we an option? This means: That caching works only on objects that are still referenced by some other part of your code. If not then you can get it from the cache, otherwise only until the gc has not deleted them.

It could also be useful to hook into slicing and the `__iter__` implementation instead of having a cache that we have to set. 

We can also implement `__sizeof__` for all our objects and return a realistic memory approximation we want to use for determining how many objects we want given some memory.

Do we maybe wanna just set a global cache size. And let it be used on a first come first served basis?

I need to think about this more. But it will be complicated... :)

dwhswenson:
I think we could use a GB based estimator without worrying about it being exact, and it would still be really useful. There's a difference between saying "our whole memory space won't be larger than 8GB" and saying "no single cache will hold more than 6GB data". The former would be really hard to accomplish. But is the latter really that bad? The user has to know to leave padding for other stuff. But I think that's still more user-friendly that telling them to guess how many of each object they'll need before they fill up their cache.

A couple questions:

1. How much does the size of each object in a cache dictionary vary? I mean, a `Snapshot` may or may not have the associated `Configuration` loaded, but the memory taken by the `Snapshot` is the same no matter what -- as is the memory taken by every `Configuration`. Those will probably be the main culprits for memory usage, so even putting a limit on them would be hugely powerful (say I typically have 4GB RAM free on my laptop: maybe I set the total cache for those two to be 3GB and then we use some number limits as already implemented on the others). The size of `Configuration`s varies widely by calculation, but not at all within one calculation. So there exists no good default number for all systems, whereas any given system *does* have a good default.

2. Would it be possible for each cache to track its own memory footprint as it grows/shrinks? Just call `__sizeof__` on every object that gets added/removed (after, of course, implementing `__sizeof__` on our objects). This adds some overhead, so I'm not sure how painful that would be. (I think that despite the normal recommendation of using `sys.getsizeof`, this might be a case where we'd use `__sizeof__` both for speed and because we don't really need an exact measurement including garbage collection overhead: just a rough idea of how full it is relative to a user's preferred max). When it gets near full, start emptying it.

> It could be that the user has made another cache and so the cache itself if practically for free or removing items from the cache will not remove them from memory since there is still the other cache.

I don't understand this sentence. Or more, I think I understand it in theory, but it isn't clear to me why it is a realistic concern. I don't think users will be making secondary caches. If they do, then they are serious power users and should know enough to turn this automatic cache management off (or upgrade it to handle their special case). What example situation are you thinking of here?

> I was thinking about using weak referencing instead of caching. Would that we an option? This means: That caching works only on objects that are still referenced by some other part of your code. If not then you can get it from the cache, otherwise only until the gc has not deleted them.

I don't understand what you mean here. If the cache is only weak referencing, isn't Python already handling it with its own garbage collection? I thought the point of caching was to keep extra references around so we could more quickly access them, even if other parts of the code weren't using them before.

jhprinz:
That was me dumping ideas to not forget about it. Alright. After some more thought I think the memory based idea is practical if not optimal. 

Quick to your questions and ideas I had.

1. The size for the _normal_ objects is practically fixed, like configurations and momentum. For these a `__sizeof__` should be easy to be implemented and could result in some speed up.

2. That is what I have implemented right now. But it does not count redundancy. E.g. If all PMC point to the same Simulator then this will be counted n times. No way around this unless implementing our own garbage collector. 

3. Second cache was just an example that assuming that a user has strong references to all objects, the cache takes almost no additional memory and hence we could leave it larger. This is related to the idea of weak references. Let me explain. The cache would be somehow optimal, if we keep the last N objects until our memory limit is reached and in addition all objects, that are around anyway because the user has still some strong reference to it. And the garbage collector would take care of that. 

    This could be implemented as follows. Keep the memory based LRU cache and if an object is removed put it into a second cache with weak references. So we have still quick access to elements that occupy space anyway. Of course we need to limit this in size, too. But since references are very small and almost independent of the object just limit this in number. Maybe 10,000 or 100,000. The the cache works as follows. 

    1. Get : If an object is checked if being in the cache. First check the normal cache. If present return, else check the weak cache. If present check if reference is pointing to an existing object. if present return else raise exception (please not my use of exception now :) Learned my lesson) and remove from weak cache. If not present in weak cache then raise exception.

    2. Set : Add object to OrderedDict. Update size. Then check size. If max_length is set check if too many objects. If max_size is set, check if total _approximate_ size is too large, which is an overestimate due to overcounting. If either is true, remove least recently used object. And correct total size.
I would also add some statistics to keep track, of the caching and see if it has potential problems. And add debugging output.

What do you think?

dwhswenson:
> That is what I have implemented right now. But it does not count redundancy. E.g. If all PMC point to the same Simulator then this will be counted n times. No way around this unless implementing our own garbage collector.

At least it would be reference counting. But this isn't what I was thinking. I was thinking of a more quick-and-dirty approach that didn't involve deep/recursive "free"ing. So my thought was that when a given dictionary got too large, it would just empty some of its own values. Your example is exactly where we *don't* want to remove objects (there usually only be 1 `PathSimulator` in the storage?).

In theory, the approach I was suggesting could lead to weird situations where you have configurations that aren't referenced by any other objects in the cache. In practice this won't be as much as an issue: we typically define data structures in order to abstract lots of data into less. So configurations take up a lot more space than snapshots, and trajectories take up a lot more space than samples.

Of course, the relative memory cost of configurations to snapshots and trajectories to samples will depend on the system. For configurations size/snapshot size, it is a constant that can easily be determined at runtime. On the other hand, trajectory size/sample size is a lot harder, since it really only exists as an average value.

> What do you think?

I still don't fully understand the approach you're proposing here. How are you doing the weak references if not reference counting? I don't understand what the secondary cache gains us. 

> If max_length is set check if too many objects. If max_size is set, check if total approximate size is too large, which is an overestimate due to overcounting.

I definitely agree with this as the overall user-facing behavior. And I would say that, if both `max_size` and `max_length` are set, use an `or` statement: if at least one is violated, you clean up.

However, I'm not sure about actually doing the cleanup at every addition to the dictionary. That requires a lot of overhead. Instead, it might be better to batch it and require some `storage.cleanup()` to run from the `PathSimulator` (with some frequency which we can set adaptively or make user-determined), and then reduce it by more than just keeping at the maximum value. Maybe reduce it to some 75% `max_val`. Again, some of this can be either user determined or possibly optimized on-the-fly.

The on-the-fly optimization stuff should probably be held for the future, but as long as you can track the number of cache misses that were found on disk, it should be feasible to implement.

But the most important thing is to look at how anything you do here affects performance. Set some small limits and test it with the toy model.

jhprinz:
> At least it would be reference counting. But this isn't what I was thinking. I was thinking of a more quick-and-dirty approach that didn't involve deep/recursive "free"ing. So my thought was that when a given dictionary got too large, it would just empty some of its own values. Your example is exactly where we don't want to remove objects (there usually only be 1 PathSimulator in the storage?).

Okay, wrong explanation. The problem is, that if we want fast caching without expensive memory counting than we will overcount the used memory. Some objects like PMC reference the same object. So if you have one object, full memory, several of these objects less than N times the memory. This is potentially only the case for smaller objects rather than Configutations or Momenta. So we either don't care about that overcounting when determining when to reduce the cache or we use a fixed object size determined somehow, which then corresponds to just specifying the number.

> The weakref cache...

This is just a little trick to fast load stuff that is in memory anyway. Consider the thing that we do not use any cache for PathSimulators. Then the weakref cache would still work and fake caching because the PathSimulator object will be alive as long as it is referenced somewhere. This effectively means we could remove any caching for some object types. I think this could be a very elegant way to make it easy for the user to specify cache sizes.

Since we get this praktically for free I would add this feature. It could also be useful for objects that we loop over instead of referencing them randomly like snapshots for example.

I will make a little notebook and play around with that.

dwhswenson:
> The problem is, that if we want fast caching without expensive memory counting than we will overcount the used memory.

Not if you don't calculate subobjects.

The cost of a configuration shows up in the configuration dictionary's count, not in the snapshot dictionary's count.

The cost of a snapshot shows up in the snapshot dictionary's count, not in the trajectory dictionary's count. It doesn't matter if that snapshot is the same in 10 trajectories.

Any object that doesn't have its own cache dictionary is, by definition, not in the cache. We don't need to claim that we get an exact count of the total program memory footprint, just that we have a rough idea of how big the cache is.

This is what I was trying to say above.

> Then the weakref cache would still work and fake caching because the PathSimulator object will be alive as long as it is referenced somewhere.

I'm still not entirely clear why this isn't just reproducing what Python does already.

dwhswenson:
Another thought on this thread (at least, if you're going to be continuing some updates in here, it might be related enough):

Would it be possible to initialize the storage in such a way that we *don't* use the chain dict caching, and instead always calculate the CV involved?

My concern here is with production, where the most common case is for chain dict to go to its last resort. Of course, it tries everything else first.

It's clear that for a sufficiently complicated order parameter (e.g., something using pyña, which requires sending a frame to an external program, running the program, and reporting back on the results), the overhead will still be worth it. Chain dict is also *always* worth it for analysis, because the most common case gets hit first.

However, for the cheapest order parameters (as used in our toy model), a very large percentage of total production time is spent in the links of the chain dict. So it would be nice to avoid that.

Common and realistic cases (e.g., Ramachandran angle in Ala-dipeptide, mindist in dissociation/association of protein-ligand complex) will fall somewhere between these two extremes, and I think we should experiment with both approaches to see which version is faster during production.

(There might also be a way to do a hybrid where production parts of the code know that they're probably creating new values, so they know to skip to the "run the function" part ... but I suspect that would be harder to implement and maintain.)

jhprinz:
Okay, a few numbers:
- fill a regular dict cache with 200,000 object (including creation of these) : 185ms
- fill an OrderedDict with 200,000 : 613ms
- fill an LRU Cache based on ordered dict with length_limit > 200,000 : 893ms
- fill an LRU Cache based on ordered dict with length_limit = 1 : 1430ms
- fill an LRU Cache with nested memory counting (using empty objects!), no_length limit: 3030ms
- fill an LRU Cache with nested memory counting and wearef, weak_cache_size=1 : 4320ms
- fill an LRU Cache with nested memory counting and wearef, weak_cache_size=1000000 : 4170ms

So, the ordinary number based cache add 5x setting, reading is always fast of course. More later :)

jhprinz:
> Would it be possible to initialize the storage in such a way that we don't use the chain dict caching, and instead always calculate the CV involved?

I think what you suggest will be supported with the pyña support PR. The idea is to make chain dict more flexible. The question would be if it is enough to just differentiate between CVs which store their cache and those who don't. If it is really that cheap to compute it would only matter to save it, if the frames are expensive to load!

The slow part in chain dict is actually the support for storing the cache in multiple files. And it also requires an additional cache for each file to be fast (enough). 

So besides fast or slow computation of the order parameter the other problem to be considered is: If we don't cache a simple value, we need to load whole frames and that will be the bottle neck. Especially in analysis. Now what makes it slow is the possibility to save stuff in multiple files. Say you compute some CV then sync with the first file. Then compute more CVs and store the same CV in a second file and only sync that one. Now you compute more CVs and want to sync again with the first store. For this to work you need to keep track of all computed values in the cache since the last sync. This needs to be done per file to really keep track of what you saved and what not. 

Handling all of this makes it slow. Maybe the idea of having multiple files open at the same time and allowing the same object being stored in multiple files is not a very efficient concept.

jhprinz:
> Not if you don't calculate subobjects.

> The cost of a configuration shows up in the configuration dictionary's count, not in the snapshot dictionary's count.

> The cost of a snapshot shows up in the snapshot dictionary's count, not in the trajectory dictionary's count. It doesn't matter if that snapshot is the same in 10 trajectories.

> Any object that doesn't have its own cache dictionary is, by definition, not in the cache. We don't need to claim that we get an exact count of the total program memory footprint, just that we have a rough idea of how big the cache is.

I think you are wrong here. Let me try to give an example: 

I was first thinking that this will work. Every cache takes care of its own type and has a quick and dirty approximate size of its objects. Possible just `single_size * count` could work. Problem is, that the caches are not coupled. For example.

```
a = st.trajectory.load(0) # a super long trajectory e.g. 1000000 frames
```
this trajectory will have to be in memory and loads all snapshots. Always a max of 100000 snapshots can be in the cache given some memory limit. Still `a` has references to all frames so the whole thing is in memory although both caches are fine and we get an error.

If you think one trajectory cannot be that long consider 100 very long ones. The cache for trajectory is fine, but the memory used by the contained object might fill the whole memory. This can only be checked by recursive memory counting. And this usually overcounts, if you want not to check for shared objects inside.

In short: Only counting the object without its referenced objects might lead to memory problems, although all caches are fine.

jhprinz:
Another idea that might be helpful:

We could detach the iterator from caching. Usually if we iterate over lots of objects we do that once. So caching is pointless. If we do it several times we might force the user to fill the cache before, like we do in  analysis.

Actually, I think, that we do not need caching at all during production. Also from a user point, is there any need for caching? This is only useful for analysis and interactive usage. And the user can always cache something himself by e.g.
```
all_trajs = list(st.trajectories)
```
and then do what needs to be done. 

When does it really happen that we run
```
for i in range(1000):
    for snap in st.snapshots:
        # do something with different with snapshot depending on i
```
Either cache everything (or it gets slow, because we will always throw away elements we will need next) or loop over snapshots first. 

For iterators we should add automatic chunk loading.So if you iterate over 1000000 object. It loads 10000 and then iterators over these, loads again etc...

The idea with weak referencing could be nice, since then caching would simply work for all objects that the user uses at some other place automatically. Back to make the last example work you would write:
```
all_snaps = list(st.snapshots)
for i in range(1000):
    for snap in st.snapshots:
        # do something with different with snapshot depending on i
del all_snaps
```

We could keep say 100 elements in a fast LRU cache for interactive things, but mostly use a weaker cache.

jhprinz:
I have merge #278 into this because fixing tests became very cumbersome and since #278 is almost ready it seemed like a good idea.

jhprinz:
So, I have implemented now a LRU Cache with weak referencing. Turns out this is so easy, because it is already build-in. The weakref package allows to keep a dictionary of still weakly referenced objects and they get removed, when the garbage collector removes them. 

This way we have best of both worlds: 
1. Everything that is stillin memory because of other variables can be accessed. 
2. Everything else will only be kept because we assume that the user might want to reuse recently accessed items.

I think this is a good solution and will keept memory as small as possible, the code is extremely short and there should not be too much performance issues, since most of the work is done by the garbage collector.

This also contains the new `AnalysisStorage` which will load things MUCH faster for analysis. Very nice :)

dwhswenson:
Since this is mixed with #278, I'm going to hold off on proper code review until that is merged first. But one thing: could you do some profiling to show off how much this improves things? I'm curious about both speed performance and memory usage during sampling.

I'd especially like to know whether this means we now how truly constant memory usage or whether there may still be some slight increase over MC steps. (Previously, we had a *large* increase over MC steps, if the system was large.) After all, the fundamental problem in issue #258 is about memory usage, so it would be good to check that this PR does fix that.

jhprinz:
Good point. I am not too sure on how to proof this works now. In principle this is ready for review and merge, once I found a test to check for it to work :)

dwhswenson:
The `%mprun` and `%memit` magics in IPython are supposed to track memory usage. But you might find it easier to use a large-ish system (Ala-dipeptide even?) and track memory usage over some number of shooting moves just by watching in `top`.

jhprinz:
thanks, will check that out...

dwhswenson:
Or even one of the toy systems, if you run long enough: I saw GB-range in memory usage with a few tens of thousands of MC steps with the toy samples; if this works as expected, usage should stay constant and well below GB! -- and speed checks will be obvious.

jhprinz:
I found a problem :) Actually the storage seems to be fine. The point is that `sample.parent` points to the previous sample so all previous samples will be kept alive. Not sure about the best way to fix that.

Strangely, if I start 500 from the same initial sampleset and move I also get a memory leak. There must be something that keeps references to old objects and thus increasing memory usage...

More tomorrow.

jhprinz:
Okay, I found a way to fix the problem of a recursive pointing to all ancestors. When you save it, the referene will be removed and if you request it, it will be reloaded automatically. This is already implemented and used to speed up loading. 

Only problem with this is, that after you save a sample and want to access the `.parent` property then it will be loaded and it requires to load from the storage. Downside is that this is slower, but seems to be the only way to remove these unnecessary dependencies.

Loading for analysis can have the same problem. The cascade of pointing to the parent will keep the parents alive.

There is a generic way to handle these things:
For trajectories we have implemented that samples inside of a trajectory can be fake objects like
`( store, index )`. So, we only reference an object by its _key_ but do not load the object.

The question is, if we find a way to allow `OPSObject` to in general point to another `OPSObject` by key. In you request the object, it will be loaded. In general this would be the best way, since you can decide during runtime to remove an object (like `.parent`), but keep the reference. We might also use this special key to point to other files, etc... But this _key_ needs to be very unique.

Due to limitations in python you cannot replace an existing object directly, like you would in C by changing the pointer to another object.

But what you can do is use an ObjectProxy. This automatically delegates all access to another object given by the `__subject__` attribute. The only difference between the ProxyObject and a real object is that `type` will be different and the the Proxy has a `__subject__` attribute. So, I would have to stop using type ;) but that shold be doable.


#### Example

```
mover = LazyStorageProxy(
    store = storage.pathmover,
    idx = 10
)
mover.name # will trigger loading the object and then it is accessible as usual
```

We should have functions to switch between 
1. a memory representation (meaning it is readily loaded in memory and can be used)
2. a storage representation (meaning it only knows where it is stored and how to load it)

```
mover.load() # will switch to memory
mover.unload() # will switch to storage (if possible)
```

I hope there is a way to make this proxy behavior part of `OPSObject` so that if an object is created you automatically get a proxy.

Will keep you posted.

dwhswenson:
> The question is, if we find a way to allow OPSObject to in general point to another OPSObject by key. In you request the object, it will be loaded. In general this would be the best way, since you can decide during runtime to remove an object (like .parent), but keep the reference.

Questions:

1. Will this have a negative impact on speed during analysis? I think that's important to check.
2. Does this change anything in the user-facing stuff? I think the same behavior we have now needs to still work.
3. Does it make more sense to only apply this proxy object indirection to specific attributes? Especially if there's a speed cost to it, I would think most attributes don't need this.

> But this key needs to be very unique.

Isn't that completely trivial, though? Each file needs a unique key. That's easy to make. Then `(filekey, store, index)` is a unique identifier for any object ever created by OPS.


> Due to limitations in python you cannot replace an existing object directly, like you would in C by changing the pointer to another object.

I don't understand this sentence. I certainly can do this:

```python
a = ObjectA()
b = ObjectB()
myobject = a
myobject = b
```

Isn't that exactly the same as changing the pointer to another object? In fact, because C is strict on types, this is something I *couldn't* do in C.

jhprinz:
Answers:
1. No, no impact no (noticeable) impact on speed for analysis. Overhead should be really minimal this time. It is only a delegation to another object, when certain attributes are called
2. No, no changes for the user. It just gives us the flexibility to represent an OPSObject as either a pointer to a file or a pointer to an object in memory. That's all. I just thought that this is impossible, but apparently it is not.
3. It is not meant to replace the fast loading. So the delayed loading of single attributes is not affected, although in some cases the implementation of delayed loading would be trivial. It just extends the idea of having tuples instead of samples in a trajectory (for speed up) to other objects as well. This means, that right now Trajectory is something like a list Proxy for Samples. Now I would make this generic and for other objects as well. It will also remove some clumsy code in chaindict to handle the tuple stuff.

4. It will simply some of the code

> Isn't that completely trivial, though? Each file needs a unique key. That's easy to make. Then (filekey, store, index) is a unique identifier for any object ever created by OPS.

Yes and no. This works if we can deduce the way the object is stored from the key somehow. If the store is a class that implements load and save then this should work. We might have to add the file type to the key tuple.

> I don't understand this sentence. I certainly can do this:

Sure but you cannot do this (which is what I need)
```
a = ObjectA() # tuple indicating store and id
b = ObjectB() # actual object
myobject = a
yourobject = a
print myobject
>>> ObjectA
a = b # redefine what a is
print myobject
>>> ObjectA # still ObjectA
```
I want to replace all references to the store_key by the actual object. Everywhere someone points to it.
With a proxyobject this works! Thanks to python.

dwhswenson:
Thanks for the clarifications: both the problem and the proposed solution make more sense to me now.

Asking for more clarification on a couple sentences:

> If the store is a class that implements load and save then this should work. We might have to add the file type to the key tuple.

* Don't all stores for `OPSObject`s support load and save in your current implementation? That is, in what case is the first part a problem?
* What do you mean by the "file type"? Can't that difference be managed by having difference classes handle accessing data from the file, and just providing an abstract interface that they all implement?

jhprinz:
An update on the remote Snapshot feature. What, if the engine will exclusively take care of this. Meaning an engine generates Snapshots that can be proxies, but only the engine knows how to reach the actual frame data. Since the engine is responsible for generating the data it would make sense that it returns a SnapshotProxy that knows the engine along with additional (engine-specific) data that allows the engine to find the data. For Gromacs this could be a filelocation and a frame number.

dwhswenson:
I'm not entirely clear on what you mean by "remote snapshot." First, about scope: Is it necessary to make it part of this PR? This PR is about a memory usage problem. We can not run any real calculations until that problem is fixed. Anything beyond fixing that problem should be done in a separate PR, because we *really* need that problem fixed ASAP.

Other questions, more to the specifics:

* How do we handle translation of trajectories between different engines? (i.e., changing different versions of these `SnapshotProxy` objects?) This needs to be handled both in cases where the translation is obvious (Gromacs to OpenMM: data maps 1-to-1), and translation between engines where the information is not 1-to-1 (e.g., OpenMM and an ab initio trajectory that needs to store wavefunction information: in one direction, we drop the wavefunction info; but in the other, we need to somehow create this). This is a problem we would face anyway, but if you're restructuring snapshots, let's face it now.
* Why would this need to be a `SnapshotProxy` instead of simply a `Snapshot`? I previously proposed adding an extra field to `Snapshot` to handle the additional information; another approach we discussed would be subclasses of `Snapshot`. I think having engines create subclasses of `Snapshot` (possibly made abstract by moving some functions into `NotImplementedError` in the `Snapshot` superclass) might be a cleaner way than creating a `SnapshotProxy` hierarchy.
* Is it necessary to implement it now? If not, let's leave it for later. Fixing the caching so we can work with realistic memory expectations is necessary before we release 1.0. Increasing the variety of engines we can support is important, but not before 1.0.

jhprinz:
fixed changes from adding unique ensembles

PR #277, opened 2015-05-30
[WIP] Add reporter to PathSimulators

jhprinz:
I tried to cleanup PatkSimulators and make BootStrapping and PathSampling use the same basic function since they looked so similar. This resulted in a little experiment to have 2 hooks in the PathSimulation:
1. start : before first step
2. pre : before every step
3. post : after every step
4. final : after last steps

You can define an MCObserver and give it up to 4 functions that are triggered at the time. This way you can easily specify what to output where.

Example
```
bootstrapping = Bootstrapping(...)
bootstrapping + IPythonReporter(
    post=lambda this : "The step is %d" % this.step
)
```
The `+` add the observer. Each Simulator has default observers for logging etc, but you can add whatever you want.

We could use this also to cleanup PathSimulators in general (including docstrings)

- [x] add reporters
- [ ] cleanup
- [ ] check docstrings

This should be merged after a #276.


jhprinz:
There were no comments on this one. Should I continue in this one or can I delete it. I assume the would require now so many changes that I should better start from scratch.

jhprinz:
Decided to abandon this one. The PathSimulators surely could use some better outputs, but the way the simulators work now is very different from before. Maybe I redo this at some point.

dwhswenson:
I don't know that this is really necessary, anyway. It looks like you're trying create a completely generic framework for any `PathSimulator`. I think that it will often be easier just to write a different subclass of `PathSimulator` instead of trying to make a single `PathSimulator` that works for every kind of simulation one might run.

This approach would work in principle, but I think anyone who wants to implement a different function will find it more intuitive to subclass `PathSimulator` than to figure out how exactly to shoehorn their method into a generalized `PathSimulator`.

As for output, we might do that, but: (1) I don't think we need four hooks for it -- wouldn't one hook be enough? (2) I think for most things it would probably be better to get whatever analysis possible from the storage, instead of changing the production. The only exception to this that I can think of is if you only wanted to save, e.g., the first frame of each trajectory, instead of the whole thing (which I've done before). But that would require significant other changes to storage anyway: right now we store everything (including all snapshots) or we store nothing at all.

jhprinz:
I think you are right. It is probably better to have a set of nice tools to output status messages to ipython like you did. If we improve on that somehow like using a widget it would be enough.

PR #278, opened 2015-05-31
Use Unique ensembles per mover

jhprinz:
Attempt to implement #272 .


dwhswenson:
I think you were holding back on this until MISTIS (#268) was merged (which it now is). This PR would be useful before I get too far into the `MoveScheme`/`MoveStrategy` stuff -- so just a nudge to put this higher on your to-do list. :-)

jhprinz:
This should be ready and my nose tests passed already. 

It now requires that all Mover have exactly one single set of input ensembles. This meant also adding the needed input ensemble in some way to all Movers (some were missing this).

So far all movers need exactly one except for RepExMovers.

With Davids idea to replace this with EnsembleHop we will have that all basic movers only need a single input, although higher level movers can of course have more.

This is one step closer to be able to build all Movers from little building blocks in a very concise way.

openpathsampling/pathmover.py
```diff
@@ -1624,33 +1855,36 @@ class MinusMover(WrappedMover):
...
+        sub_trajectory_selector.name = "MinusSubtrajectoryChooser"
 
-        repexs = [ReplicaExchangeMover(ensembles=[[segment, inner]])
-                  for inner in innermost_ensembles]
-        repex_chooser = RandomChoiceMover(repexs)
-        repex_chooser.name = "InterfaceSetChooser"
+        repex = ReplicaExchangeMover(
+            ensemble1=segment,
+            ensemble2=innermost_ensemble
+        )
```
dwhswenson: breaks MISTIS

openpathsampling/pathmover.py
```diff
@@ -1602,18 +1778,73 @@ class OneWayShootingMover(RandomChoiceMover):
...
+            ensemble=dct['movers'][0].ensemble,
+            selector=dct['movers'][0].selector
+        )
+
+        # override with stored movers
+        mover.movers = dct['movers']
+
+        return mover
+
+class OneWayExtendMover(RandomChoiceMover):
```
dwhswenson: Do we need this?

openpathsampling/pathmover.py
```diff
@@ -1244,6 +1286,146 @@ def move(self, globalstate):
...
+                    if ens not in present_ensembles:
+                        # ens might be required but is not present
+                        weights[idx] = 0.0
+
+                if weights[idx] > 0.0:
+                    found = True
+            else:
+                weights[idx] = 0.0
+
+        return weights
```
dwhswenson: Please explain why these are needed. I'm pretty sure the `FirstAllowed` and `LastAllowed` don't satisfy detailed balance.

openpathsampling/analysis/tis_analysis.py
```diff
@@ -65,27 +60,22 @@ def all_movers(self):
...
+    # def _assign_sample(self, sample):
+    #     if sample.ensemble in self.all_ensembles:
+    #         try:
+    #             self._samples_by_ensemble[sample.ensemble].append(sample)
+    #         except KeyError:
+    #             self._samples_by_ensemble[sample.ensemble] = [sample]
+    #         try:
+    #             self._samples_by_id[sample.replica].append(sample)
+    #         except KeyError:
+    #             self._samples_by_id[sample.replica] = [sample]
```
dwhswenson: not sure on this one

openpathsampling/analysis/tis_analysis.py
```diff
@@ -65,27 +60,22 @@ def all_movers(self):
...
-            all_movers += self.movers[movetype]
-        return all_movers
-
-    @property
-    def all_ensembles(self):
-        return self.ensembles
+    # TODO: can this be removed? Only used in _assign_sample
+    # @property
+    # def all_ensembles(self):
+    #     return self.ensembles
```
dwhswenson: No. This is overridden by other transitions, but that gives the default behavior.

openpathsampling/analysis/tis_analysis.py
```diff
@@ -353,6 +343,9 @@ def total_crossing_probability(self, storage=None, method="wham", force=False):
         force : bool (False)
             if true, cached results are overwritten
         """
+
+        tcp = dict()
```
dwhswenson: Don't do this. If WHAM fails to return a correct value, we want to give an error, not to return an empty LookupFunction. In general, I think it is better to let python raise an error if something goes wrong, instead of giving a number which will then be wrong (and possibly published by an inattentive user!)

openpathsampling/pathmover.py
dwhswenson: To clarify: you should change everything in this section back to plurals, and leave the list of repexs. Just change the `ensemble1=, ensemble2=` part. Everything else is a part of the changes I made so that this implements the multiple interface set minus move -- it was that way for a reason.

openpathsampling/pathmover.py
jhprinz: In principle yes. But I wrote it a little shorter. The current way first creates a new object with the correct ensemble and selector and then fixes it to use the stored movers. The new way does not instatiate using OneWayShooter, but creates an empty object using __new__ and then calls the super __init__ function to create a 'SelectionMover'. This is shorter and reuses more code.

openpathsampling/pathmover.py
jhprinz: Only LastAllowed is used so far and only in BootStrapping. This way we can remove using the kind of clumsy check in which ensemble we are and run the appropriate mover. It does not satisfy detailed balance!

openpathsampling/pathmover.py
jhprinz: this is a merge error. Will fix this.

jhprinz: Just saw you comment. Exactly. Sorry about the confusion.

openpathsampling/pathmover.py
dwhswenson: Sorry, the selection didn't make very clear what I was asking about. I'd actually only selected the line

```python
class OneWayExtendMover(RandomChoiceMover):
```

That's what I didn't know whether was necessary!

openpathsampling/analysis/tis_analysis.py
jhprinz: My IDE says the function is not used anywhere. So I was wondering if this is legacy that can safely be removed or is it something that you would like to keep. I commented it out to see, if the code breaks.

openpathsampling/pathmover.py
dwhswenson: Okay, I see what you're doing, but I think we should eventually find a better solution to this. This is still a pretty nonintuitive way to approach the problem, and feels a bit kludgy. Honestly, the better solution is probably just to keep state in the simulator, and rather than doing these contortions to avoid keeping state. There's nothing equilibrium about the bootstrapping sampling, so these hacks to make it act like it is are just leaving us with a bunch of path movers that are *really* bad ideas for most situations. (These, and the partial acceptance sequential, and I think we've done a few other silly things over time.) It is cruft that will confuse contributors.

openpathsampling/analysis/tis_analysis.py
jhprinz: I understand. It is usually bad practice not to initialize variables at all and hope for raised errors. I think I had some problems with this not being initialized somewhere and added it. At least my IDE was complaining that this could fail, since is has never been initialized and does not even exist. Meaning it could fail, but for the wrong reasons. (I guess we are getting back to the old debate about duck-typing :). Anyway, I can remove it.

openpathsampling/analysis/tis_analysis.py
dwhswenson: It might be in an ipython notebook somewhere. I couldn't find it either -- and I'm not entirely sure what it was supposed to do -- I think it predates your fast-loading of samples, so it's probably not needed. I'd say leave as is: commented out, but not deleted, in case it comes back to bite us somewhere!

openpathsampling/analysis/tis_analysis.py
dwhswenson: Well, it doesn't exist if you choose "mbar", because that isn't implemented. If a user tried to use mbar, I would rather they get errors than be told their rates are 0!

jhprinz: The other disadvantage I remember is that autocomplete in ipynbks will not work until .tcp is created.
`analysis_object.<TAB>` will not list `tcp` as an option, which can be confusing.

dwhswenson: You're mixing two things. Very intentionally, there's a distinction between the local variable `tcp` and the instance variable `self.tcp`. The first is what this part is about, and is a dictionary. The second is a `LookupFunction`. We could initialize the second to `None`, although I think I use its existence (`hasattr`) as a test for whether analysis has already been performed, so you'd have to hunt down those cases and replace that with testing for `is None`. I'm fine with that approach if you think it's worthwhile.

jhprinz: Okay, my mistake. Indeed the autocomplete is for the self.tcp which would be good to exists from the start. I misunderstood. The tcp = dict() was only meant to make sure that the last line does not break. I was not aware that this is what you wanted.

openpathsampling/pathmover.py
jhprinz: Not sure if I agree on this one. The idea was not to make it look like okeying detailed balance. The idea was to have a clean way to write bootstrapping moves the same way as all the other moves. 
The way I think about moves is that there are some that obey detailed balane for a very good reason while others do not. Also for a good reason. This way the code looked cleaner to me. And we might reuse the `FirstAllowed` and `LastAllowed`.

These two movers fall into the same category as the `SequentialAllowed` mover that is new and was meant for single replica TIS. From a set of possible ones it selects the one that is possible (according to input/output ensembles). Instead of having to have this selection process to program in the PathSimulator.

I agree that this introduces a few more Movers that could be confusing to a user, but I assume that these functions will usually just be ignored and are only for advances users relevant.

jhprinz:
all recent stuff merged and it has passed all tests on my machine. Let's see what happens on travis.

openpathsampling/pathmover.py
dwhswenson: > The idea was to have a clean way to write bootstrapping moves the same way as all the other moves.

This is exactly the idea I disagree with. They're not the same. We're shoehorning them in, and everything we've done like that is a kludge. Furthermore, by listing these moves alongside the others, we hide the from the user the fact that they are fundamentally incompatible with other methods. Bootstrapping is really simple: it isn't a move, it is a simulation approach. You have a separate list of moves for each step; you run them until all moves are accepted.

I think the original approach was the best. Trying to make it into something more general with `PartialAcceptance` was a bad idea. We've rewritten bootstrapping 3 or 4 times, each time adding *more* code that "might" be reused, but when we don't have any actual examples where we reuse it. We keep trying to make it more like a normal move, which hides the fact that it isn't.

If you leave it as is, please find a way to mark moves that can not be expected to have a detailed balance partner. I'm in favor of (shortly after 1.0) rewriting bootstrapping and removing partial acceptance.

> And we might reuse the `FirstAllowed` and `LastAllowed`.

When? One of my concerns is that we keep exploding with more and more classes because we "might" reuse them. We're falling into the trap that is the reason programmers use the expression YAGNI: "ya ain't gonna need it." By making too many overly generic classes, I'm concerned that we risk making it easy to do things that aren't physically correct. These moves will never satisfy detailed balance. I think it is a bad idea to include them alongside the moves that *do* satisfy detailed balance.

Adding in something like this later would be easy if we find an actual use for it. But I think we should aim to keep code slim when possible.

Note that the distinction here is that, like PartialAcceptance, you can not *create* a move that satisfies detailed balance. Other first-or-last (like choose first/last subtrajectory) can be made to satisfy detailed balance by wrapping them in a RandomChoice.  

> These two movers fall into the same category as the `SequentialAllowed` mover that is new and was meant for single replica TIS.

So one move in the category does satisfy detailed balance, and the others don't? Isn't that another potential source of confusion?

> I agree that this introduces a few more Movers that could be confusing to a user, but I assume that these functions will usually just be ignored and are only for advances users relevant.

I agree, but I think we shouldn't let that be an excuse for adding code that isn't needed. We have over 200 classes, around 50 in `pathmover.py` alone (according to `grep "^class"`). The larger that number, the higher the barrier to a user contributing. I think we need to work hard at reducing our classes, and turning anything possible into convenience classes of more general treatments.

jhprinz: Okay, then I tried to satisfy your original idea and apparently you don't want that. This is how I understood the original idea which is the reason why I thought it makes sense to have both move with and without detailed balance and they should be possible to be used as moves.

> This is exactly the idea I disagree with. They're not the same.

Exactly, they are not the same, but they can be used the same way. Like using plus and times with numbers they are very different but are applied the same way.

You said that bootstrapping is a way to generate initial trajectories. It is not meant to generate samples, since be will not sample correctly from the given distribution. This is the way we used it before and for this reason why not allow to use moves we had before to drive the system toward a target ensemble or something else. For sampling we only use detailed balance.

I think you came up with the way to use bootstrapping. So, do I understand that you think this is not a good idea now and we should have a clear distinction between sampling and other trajectory generation? That might be a good idea. I still think that it could make sense to have the same mechanism (using movers) for both approaches. 

> around 50 in pathmover.py

This is due to some restructuring and only about 15 of them are to be used. Especially the Generator Idea is still in there which I need to remove and doubles the number of classes.
Also some more I think are legacy and can be removed.

I hope you did not think that I wanted it to be this way from the start. It was my attempt to combine your bootstrapping with the movers in a consistent way.

Anyway. It seems that PathMovers need more cleaning. This I totally agree upon...

dwhswenson: (starting with a quote from the end, because the answer to that answers a lot)

> I hope you did not think that I wanted it to be this way from the start. It was my attempt to combine your bootstrapping with the movers in a consistent way.

No, I didn't think that at all. What I meant to say is that, if I recall correctly, It was ***me*** who started this direction. ***I*** started us down this road when I introduced the `PartialAcceptanceMover`. (As much emphasis as possible, to say "I am to blame".) But the more I look at it, the more that looks like it was a bad idea. We've both come up with some really clever tricks along that road, but the problem is that a user needs to be clever to understand and/or correctly use those tricks. And that's what I'm concerned about. So unless you have ideas for really useful cases where we need to keep doing this (they might exist, I just can't think of any), I think we should stop doing things that take us further into the territory of having movers that can't satisfy detailed balance, and plan to eventually remove the ones we have. (That's not as urgent as finishing 1.0, but it should be on the list.)

We can still re-use tools from proper path sampling for bootstrapping, but it seems dangerous to create new classes that can't be used in path sampling. I'm picturing a future bootstrapping like this: bootstrapping consists of the same start as single replica (choose based on current ensemble), then within each ensemble there's a sequential of shooting move followed by ensemble hop.

This doesn't set the replica IDs, but I think the trick there is to create a little function called `bootstrapping.get_initial_sampleset(storage)` (or maybe takes a list of `mcsteps`?) which loops over the steps and adds to the replica ID each time a new ensemble is encountered.

To me, this is closer to the spirit of what bootstrapping is doing (the change of replica ID always seemed a necessary evil, and as you well know, it is the source of almost all the frustration in implementing bootstrapping). Moving that to an analysis stage cleans it up a lot.

> Like using plus and times with numbers they are very different but are applied the same way.

I would say the better analogy is using plus/times between lists and number in Python. First off, that `[1, 2, 3]*2 == [1, 2, 3, 1, 2, 3]` is not entirely obvious (why not `[2, 4, 6]`?). Second, `1 + [1, 2, 3]` (very reasonably) doesn't work, even though `[1] + [1, 2, 3]` does. Some things can be combined; some can't. It requires a good knowledge of the tool to understand which combinations are allowed and which aren't. And since path movers and detailed balance are a lot more complicated to the average human than lists/numbers, I think we should err toward something less likely to cause mistakes -- especially since, unlike Python, we don't have an easy mechanism to raise an error if an combination is made which is illegal in some particular context.

> I think you came up with the way to use bootstrapping. So, do I understand that you think this is not a good idea now and we should have a clear distinction between sampling and other trajectory generation?

Repeating for emphasis: ***YES.*** Both that it was me who proposed the direction originally, and that I now think it was a bad idea. At the very least, any moves that can not be valid for sampling should carry a flag so that some user who tries to out-clever the laws of statistics by using them can be warned "it appears your plan does not satisfy detailed balance -- are you sure you want to do this?"

> Especially the Generator Idea is still in there which I need to remove and doubles the number of classes.

I noticed this was part of the problem; I'm glad you agree that this needs to be changed. I was going to suggest not having `Generator` in the name. I think having a class variable (my preference) or having them inherit from some abstract class seems like a better way to make it possible for other parts of code to check this.

The conceptual difference is very important: however, I'm not sure when that needs to be checked computationally.

openpathsampling/pathmover.py
dwhswenson: Revisiting this: I think it may have gotten lost in the shuffle of other topics.

My question here was badly displayed by GitHub. I meant to ask: what is the purpose of the `OneWayExtendMover`? Do we use that? (My intuition says that a 50/50 change of doing this is rarely going to be used. OTOH, each direction, used separately in some particular context, will definitely be used.)

openpathsampling/pathmover.py
jhprinz: Alright, so we agree and it was kind of a misunderstanding. Very good. I actually that the idea wasn't that bad especially at the time. But as you say, we need to be aware of the conceptual difference. I have starting working on the tree. Since this is so closely related to PMC and PM I might make a new PR to explain what I want to do.

jhprinz:
I think I lost track what of this is ready and what not. I think we have advanced beyond this or we are already using it? Was this potentially used in some other PR and should have been closed?

jhprinz:
merged latest PR to cause another test run.

jhprinz:
When it passes please have a look. It will also be part of the later `tree` PR that makes more changes. But this uniqueness of ensembles is important.

dwhswenson:
Will merge as is; making a note of two things we should address in more detail in the future:

1. `ReplicaExchangeMover(ens1, ens2)` should be the same as `ReplicaExchangeMover(ens2, ens1)`. Maybe we can define some sort of canonical sort for ensembles? (Sort based on hash?) This has also been a point of annoyance in a lot of my stuff involving ensemble signatures.

2. We should revisit the issue of movers that can't satisfy detailed balance, but that should be part of a larger discussion on `AcceptorMover`s and automating detailed balance. (I think a chunk of the path mover code will need some cleanup, but we'll need to decide whether that is v1.0 or, likely, later).

PR #281, opened 2015-06-07
Networks cleanup

dwhswenson:
Some cleanup stuff for Networks and Transitions. Should be merged after #268.

- [x] Fully separate generation of ensembles and generation of movers
- [x] Make both `.sampling_transitions` and `.transitions` (analysis transitions) easily available in all networks

This is just some cleanup before doing the `MoveScheme`/`MoveStrategy` stuff: one of the requirements for that will be to clean up the `Network`/`Transition` code (which would probably be a good thing to do no matter what). So that’s what I’m doing here.


dwhswenson:
I'd say the separation is sufficient for now. This is ready for review and merge (as soon as #268 is merged, at least, since it includes #268 in it.)

openpathsampling/analysis/network.py
```diff
@@ -305,15 +312,245 @@ def rate_matrix(self, storage, force=False):
...
+        return ret_dict
+
+    @staticmethod
+    def from_dict(dct):
+        network = MISTISNetwork.__new__(MISTISNetwork)
+        network.movers = dct['movers']
+        network.minus_ensembles = dct['minus_ensembles']
+        network.ms_outers = dct['ms_outers']
+        network.transition_pairs = dct['transition_pairs']
+        network.__init__(dct['transitions'])
```
jhprinz: Quick question here before merging. Is it possible that calling the `__init__` at the end that the restored movers, ensembles, etc get overridden by newly created ones? If so, this should be used:
```
network = MISTISNetwork(dct['transitions'])
network.movers = dct['movers']
network.minus_ensembles = dct['minus_ensembles']
network.ms_outers = dct['ms_outers']
network.transition_pairs = dct['transition_pairs']
```

jhprinz:
If that is fine we can merge. The rest looks very good.

dwhswenson:
Actually, there's a problem related to storage, but not quite what you pointed to.

The overall approach I have here is still the one I want to use (for now): instead of creating a bunch of stuff, and then trying to overwrite it all, we just don't create it. `__init__` checks for the existence of a `_sampling_transitions` attribute, and if it has one, it doesn't create all that stuff.

Problem was, prior to this PR I was checking for something other than `_sampling_transitions`. Also, I renamed `transitions` to `input_transitions`. Between these, the previous version failed to reload the network. I'm working on fixing this now, but since the issues are in storage, I may need some help to get it fixed.

Right now, it doesn't seem to want to *store* the sampling transitions. I can't tell why -- something goes wrong in the `to_dict` function. If I can't figure it out once I've narrrowed the problem down a bit, I'll ask you to take a look.

(There was some reason I didn't use the approach that you suggest; can't remember what it was, and it may no longer be relevant, but I found a situation where I really had to split `__new__` and `__init__` in the rebuilding.)

dwhswenson:
... and fixed, although it'll need your fixes for the new mdtraj before it passes.

Weirdly, the issue was fixed by forcing each of the sampling transitions to have a name. That's not hard to do, but it might indicate some underlying problem in the way the `name` attribute is handled by default. It was crashing on the last line of `saveidx.inner` (in `object_storage.py`):

```
self.storage.variables[self.db + '_name'][idx] = obj._name
```

I'm guessing it might be a problem with the `obj.fix_name()` function for `Transition`s?

Anyway, I left some commented-out stuff to assist with debug logging. The stuff in `todict.py`, in particular, might be a good place to use two different algorithms depending on whether debug-level logging is enabled. The slower approach made debugging a LOT easier.

As to your specific concern, take a look at the last cell of `mistis_analysis.ipynb` -- that answers your question, right?

jhprinz:
I have made a PR for your PR that should fix the naming problem and especially the mdtraj changes. If you think my PR is okay, you can merge this one, too.

jhprinz:
It it passes go ahead and merge.

dwhswenson:
It passes and merging!

PR #285, opened 2015-06-18
MoveScheme/MoveStrategy

### LINKED ISSUE #279: MoveScheme and MoveStrategy
I’ve been thinking about the best way to implement different move schemes and move strategies. At the heart of the idea is this: a network of transitions is really defined by the ensembles involved. But the same ensembles could be studied using different sets of moves: for example, we can study a network using the standard RETIS move strategy, or we can study it using the SRTIS move strategy. That’s where this discussion matters.

#### Definitions of terms

* **move scheme**: for a given simulation, the move scheme is the "move decision tree". Every step of the MC is done by starting with some root move, and tracing a series of decision points to generate (and the accept) a trial.
* **move strategy**: a general approach to building a move scheme (or a subset thereof). SRTIS is a move strategy. Nearest-neighbor replica exchange is a move strategy. All-replica exchange is a move strategy.

So I use "strategy" to talk about the general idea, and "scheme" to talk about a specific implementation of that idea. "Weina worked on the first practical uses of the SRTIS move strategy. In particular, she implemented an SRTIS-based move scheme in her study of villin headpiece unfolding."

#### Implementation thoughts

The approach I’m planning is to create a `MoveScheme` object, which will basically be a container for the current move scheme root, but also has some analysis functions associated with it. (The analysis parts of this object were already created in #266.) Then there will be `MoveStrategy` objects, which will take a `TISNetwork` object in their `__init__` functions, and have a function (maybe called `make_scheme(scheme)`?) which creates a modified version of an old `MoveScheme`. Widely-used strategies will be subclasses of `MoveStrategy`. Most typical strategies (`SRTISStrategy`, `SetAllExchangeStrategy`, `NearestNeighborReplicaExchangeStrategy`) will be idempotent: running `scheme = strategy.make_scheme(scheme)` multiple times in a row should return equivalent (though not necessarily identical) objects. The ones that are effectively orthogonal (switching the allowed exchanges as with `SetAllExchangeStrategy` vs. switching the exchange style with `SRTISStrategy` vs. switching the shooting point selection, for example) should also be commutative. (That may be a bit challenging to implement, though.)

Think of `MoveStrategy` as a factory for `MoveScheme`s.

I thought of several alternatives to aspects of this approach, but I didn’t like any of them as much:

*  *`MoveScheme` could be left as is; part of the `Network`.* I think this is conceptually confusing. This approach clearly distinguishes between the network itself (ensembles) and the way we study it (moves).
* *Each `MoveStrategy` could be a bare function instead of a class.* This is what I originally suggested in #260. But this reduces the flexibility of the factory.
* *There could be one `MoveStrategy` factory class for all strategies.* This means that a user creating new strategies would either have to subclass that overall factory or modify it. I think that subclassing a simpler class is more user-friendly, and allows a user to keep using their strategy as the code updates, even if we don't accept it into master.
* *Instead of having a separate `MoveStrategy` class, different subclasses of `MoveScheme` could apply different strategies.* The problem with this is in chaining several modifications together. Starting from the default, maybe a switch from nearest-neighbor to all-replica exchange, then switch to single-replica sampling. This approach would mean this is `isinstance(scheme, SingleReplicaMoveScheme)` and not `isinstance(scheme, AllExchangeMoveScheme)`, which seems weird to me.

***

That’s my current view on the implementation of this. It's one of the steps toward getting SRTIS working (#260). Some aspects (esp. around commutativity) might not show up in time for 1.0, but there should be a way to easily do the most obvious move scheme modifications.

Any thoughts?

dwhswenson:
Introduces the MoveScheme and MoveStrategy objects, which should clean up the process of making custom move schemes and will simplify implementation and usage of single replica TIS.

It has taken me a while to put up this PR because it has taken me a while to figure out an approach that I think will work. The main ideas are described below. For now, this doesn’t replace the `default_movers` functions in networks, although eventually it will. This PR won’t make it full-featured enough to replace that, because I want to get this merged relatively quickly. This PR will just include basic stuff for shooting, path reversal, and replica exchange.

Much of this will be demonstrated and further described in `move_strategies_and_schemes.ipynb`, a new notebook included as part of this PR.

***

- [x] `OneWayShootingStrategy`
- [x] `NearestNeighborRepExStrategy`
- [x] Applying strategy to scheme works correctly
- [x] `DefaultStrategy`
- [x] `MoveScheme` creates move decision tree correctly
- [x] `AllSetRepExStrategy`
- [x] check that we can switch between nearest-neighbor and all-set strategies
- [x] `PathReversalStrategy`
- [x] `SelectedPairsRepExStrategy`
- [x] `DefaultScheme` creates correct defaults (except stuff involving minus ensemble or MS-outer ensemble)
- [x] ipynb to demonstrate some basic tricks

Resolves #279. For the basic distinction between a `MoveScheme` and `MoveStrategy`, see that issue.

***

The approach I’m taking works like this:

1. Each `MoveScheme` holds a list of `MoveStrategy` instances. Each `MoveStrategy` is essentially a `PathMover` factory with some additional information to provide context to the `MoveScheme`. New strategies can be added to the scheme with the scheme’s `.append` function.
2. Each `MoveStrategy` has associated with it a `level` and a `group`. The `group` is mainly used for organization: the standard `group`s are `'shooting'`, `'pathreversal'`, `'repex'`, `'minus'`, and `None` (for structural aspects), but since these are just strings (used as dictionary keys), the user can easily extend this. The `level` has to do with what exactly is done by the strategy. Certain aspects of the move scheme must be set before other aspects can be built (e.g., you need to have built the shooting movers and replica exchange movers before arranging them in either single-replica or default style). The `level` helps organize this; see below for details.
3. When the `MoveScheme` is asked to give a `move_decision_tree`, it generates the tree by successively applying its `MoveStrategy` objects. There are two factors which determine the order in which the `MoveStrategy` objects are applied: first they are sorted by the `level` associated with each strategy, then within each `level`, they are applied in the order they were appended to the scheme.

***

More details on concepts in this approach:

#### Groups

We often think of our path movers as being sorted into functional groups: e.g., shooting, replica exchange, path reversal, minus, and so forth. In the `MoveStrategy`/`MoveScheme` implementation, I have formalized that a bit. We have several default `group`s, but the user can define others. The `DefaultStrategy` is to randomly pick one type of move (one group), and then to randomly pick one of the moves within that group.

Although, the default behavior is to have each kind of move in its own group, you might have more than one group with the same kinds of moves. For example, if you have two kinds of shooting moves for each ensemble (e.g., 2-way to increase decorrelation, 1-way to get better acceptance) it will probably make sense to think of these as two different groups. At least, you probably care about the acceptance ratio for each subgroup of  shooters more than the acceptance ratio for all of them. The use of dictionary keys for this makes it easy for the user to create such schemes.

#### Ensemble signatures

Every `PathMover` has a list of "input" and "output" ensembles. The pair `(input, output)` is what I call the "ensemble signature" of that mover. This is important when doing replacements: for example, if we replace one-way shooting with two-way shooting, we can use that ensemble signature to make sure we replace each existing one-way shooter with a two-way shooter.

In the implementation, we use a process of build-by-modification. So the list of ensemble signatures does not exist as a separate object. We assume that the we have certain default move types for each functionality (shooting defaults to uniform-selection one-way; replica exchange/ensemble hopping defaults to unbiased replica exchange; minus defaults to `n_l=2` replica-exchange based minus moves). So to create the default signatures, we actually create the default movers that match those signatures.

#### Levels

The easiest way to describe the concept of levels in the MoveStrategy is to consider the process of building the move decision tree. The choice of default levels comes immediately from this.

1. `levels.SIGNATURE` : Define the set of ensemble signatures needed within each `group`. 
2. `levels.MOVER` : Set the specific movers to implement the strategy within each `group`. Note that additions at the `MOVER` level can still affect the signature (by adding specific new movers), but can not remove existing signatures. This can also only replace movers 1-to-1 (so you can replace a one-way shooting with two-way shooting, but you can’t replace a replica exchange with two ensemble hops)
3. `levels.GROUP` : At this level, you can replace all movers within a group in a way that is not 1-to-1. For example, this is where you can switch between replica exchange and ensemble hop.
4. `levels.SUPERGROUP` : At this level, you modify the structure of the `group`s that exist (combining them, or splitting them into two separate groups). 
5. `levels.GLOBAL` : This handles the overall structure of the move decision tree.

Since each strategy has a level associated with it, strategies are sorted into lists for each level as they get appended to the move scheme. When the move scheme generates the path move tree, it does so by first applying all the `SIGNATURE` level strategies (in the order they were appended to the scheme), the applying all the `MOVER` level strategies, etc.

Actually, since each of these levels are just numbers, the user can arbitrarily define other levels to force a certain order to the building of the tree. Internally, the strategies to apply are stored as a dictionary with numeric keys for the levels pointing to a list of strategies. When building the decision tree, we just loop over the sorted keys, meaning the custom key values are allowed. But for the most part, the five default levels should be enough.

***

The current version does not replace the way I’d previously generated move schemes, nor is it used in the test/example notebooks. Before doing that, I need to add support for moves involving minus ensembles and multistate outer ensembles. This also doesn’t do anything for single replica TIS, although this is a pre-requisite for the clean way of implementing that. All of which means that the main code is so far pretty much independent of this PR: integration can hold off until we have full replacement of all existing features.


dwhswenson:
Passes tests, and task list completed. This is now ready for review. 

As mentioned above, this doesn't (yet) replace anything; it just lays the groundwork. And the goal isn't to *remove* the ability to manually edit the move decision tree, just to provide tools to modifications easier for novice users. Other `MoveScheme`/`MoveStrategy`-based PRs will come soon, but I wanted to split them up to make review a little easier.

I would recommend starting with [the new ipynb](https://github.com/dwhswenson/openpathsampling/blob/move_scheme/examples/ipython/move_strategies_and_schemes.ipynb) to get a sense of how this works and what it can do, then moving on to the actual code for review.

dwhswenson:
I'm going to merge #275 before this one: please start to look over these changes, but obviously some of the new tests and new notebooks will be affected by the massive renaming PR. Even if GitHub doesn't mark it as not ready for automatic merge, the tests would fail after the change.

jhprinz:
Still working on going through the logic of this. I remember that wordpress uses these levels to achieve a certain order of execution and I found it very strange to work with. So let me have a more detailed look at the implementation. I think this should work, but it is very different from what I would have done.

Anyway, good job!

dwhswenson:
The reason for this approach is that I can't necessarily enumerate all the "orthogonal directions" in designing a move scheme. I can name a few (setup of the replica exchange network, choice of replica exchange "style" [exchange vs. ensemble hop], type of shooting move, and a few others). But who knows what future developers might invent?

Also, it simplifies changes. If you want to change to an all-in-set replica exchange network, and also change to single replica TIS (meaning ensemble hops), then you have to somehow modify both of those in the right order. This approach uses an automatic ordering that gives the right results regardless of the order of user input. I didn't want to write different functions to handle the different orders (with the permutations of various related effects that might be imagined, that would create an unmanageable code base.) And I wanted to keep the code as close to what we already have as possible, rather than create new levels of abstraction. (That part may change in the future; the goal here is to get publishable quickly.)

Admittedly, the concepts to understand it aren't totally trivial (this is why I wrote the long notebook with lots of examples), but two important points:

1. What comes out is the move decision tree, which is the same thing we had before. Worst case, a custom version of the move decision tree exactly as hard to build as it was before this PR. So this doesn't prevent you from doing it the old way.
2. The details of this implementation are only intended for relatively advanced users -- at least the sort of people who wouldn't find the repex-shoot-repex example to be too terribly intimidating. Simple changes, like switching to all-replica or changing the shooter type, don't require the user to think about the levels -- they're handled automatically.

dwhswenson:
Passes tests again. Fully ready for review/merge!

jhprinz:
This is indeed really cool. The more I think about it. I will merge this, since it provides some really cool features. 

A few ideas I had about the initial concern about using numbers to set the ordering: What I had in mind was something like a graph structure where each change (the strategy) need to know what group it uses and what it will change. Then a graph sorting algorithm will give you an order that will work. 

I am not sure on how to really specify the relation between single strategies which will make this approach not feasible. So I think that the numbering is a good compromise to let the user figure out the order to some extend.

Anyway. I will merge this and we can improve from there (if needed at all).

Great!

PR #287, opened 2015-06-22
MoveScheme support for MultiState and Minus

dwhswenson:
Adds support for moves based on MS outer interfaces and based on minus interfaces to the MoveScheme approach.

- [x] Modify MSTIS network to use `special_ensembles` dictionary
- [x] Modify MISTIS network to use `special_ensembles` dictionary
- [x] MinusStrategy
- [x] Add support for `hidden_ensembles`
- [x] DefaultScheme support for minus
- [x] DefaultScheme support for MS-outer shooting
- [x] DefaultScheme support for MS-outer replica exchange
- [x] Cleanup old stuff from transitions/networks

***

One thing I’m changing here is that I’m creating a more generic way of treating ensembles in the networks. It boils down to this: We have 3 overall "varieties" of ensembles. They are

* ensembles which are sampled and used in the crossing probability calculation (transition ensembles)
* ensembles which are sampled and not used the in crossing probability calculation (special ensembles; e.g., minus ensemble and multistate-outer ensemble)
* ensembles which are not sampled but play a role, perhaps as intermediate ensembles in other moves (hidden ensembles, e.g., the segment ensemble in the minus move)

(By "sampled" I mean they can show up as active samples.)

Previously, networks completely ignored the hidden ensembles (you had to really know the innards of the code to get at them), and treated the two examples of special ensembles as separate things.

The new approach says that these special ensembles are linked in some way with some transition or transitions. So `network.special_ensembles` is a dictionary of `{ ensemble_type_name : dict }`, where the `ensemble_type_name` is a string `’minus’`, `’ms_outer`’, and the dictionary maps the specific ensemble to a list of `Transition`s. You need the context of which kind of special ensemble it is (the name key) to know how to build the mover based on it (MS-outer typically replica swaps with outermost interfaces; minus swaps with innermost).

For example, the dictionary for a simple two-state network would look something like this:

```
{ 'minus' : {minus_ensemble_A : [transitionAB],
             minus_ensemble_B : [transitionBA]},
  'ms_outer' : {ms_outer_ensemble : [transitionAB, transitionBA]}
}
```

Of course, this is all built automatically. This also gives all the networks a more standardized structure. I still include `minus_ensembles` and `ms_outers` as properties, so they’re still easily accessed for these networks.

This approach also leaves room for other kinds of `special_ensembles`; perhaps ensembles that are associated with the other ensembles in a transition, but that aren’t sampled? I can’t think of an example of that, but it is conceptually possible.

***

For now, I’m just having the `DefaultScheme` directly handle the multiple state outer interface. I’m not sure whether that’s the best approach: it might be better to have specialized strategies for that. On the other hand, I feel like it is a bad idea to add a lot of strategies if they’re only barely wrappers for something else — we should be keeping the number of classes small when reasonable.

This includes everything in #285, and so that needs to be merged first.


dwhswenson:
This is ready for review and merge. Includes #285, so it will probably be easier to review that one first, then this.

This should bring the MoveScheme/MoveStrategy to the point where all the features of the existing approach are included (plus, of course, a lot more usability).

jhprinz:
Great. Also merging this one.

PR #289, opened 2015-06-24
MoveStrategy: Balance partners and choice probability

dwhswenson:
- [x] Identify the `balance_partners` for each mover in the group
- [x] Fix to one `balance_partner` (for now): code can find multiple, but will throw warning in fear that other parts of the code can’t handle that
- [x] `DefaultStrategy` sets the `choice_probability` for each mover if not already set
- [x] `DefaultStrategy` uses existing `choice_probability` if possible

This includes #285 and #287, and therefore should be merged after both of those.

This PR also includes some setup toward the `OrganizeByEnsembleStrategy`. These aspects were required first, and were complicated enough that I decided to split them into a separate PR.

***

A few new things in this PR:

* every mover in a scheme is associated with a total choice probability
* every mover in a scheme is associated with its balance partners: balance partners must be in the same group

This will allow us to manage detailed balance correctly without a lot of user effort. See discussion in https://github.com/choderalab/openpathsampling/pull/251#issuecomment-104673472 and following. 

In order to prepare these things correctly, a few rules need to be obeyed:

1. All movers in a given group in the scheme must have their balance partner in the same group.
2. No two movers in a given group can have the same ensemble signature.
3. Any `GLOBAL`-level strategy needs to set the `choice_probability` for all (group-level) movers.
4. Each mover can only be used once in the scheme groups.

What exactly do these mean, and why do they matter? Here’s some more detail.

*All movers in a given group in the scheme must have their balance partner in the same group.* This means that, if you’re doing ensemble hopping, the hop that does A->B needs to be in the same group as the hop that does B->A. This is because I check the signatures within a given group for the balance partners (otherwise, it would be a lot harder to tell that `PathReversalMover(ensA)` is not the balance partner of `OneWayShootingMover(ensA)`). Most movers are their own balance partners; in that case this is trivial.

*No two movers in a given group can have the same ensemble signature.* This is equivalent (when combined with the previous) to the much more convoluted (but more suggestive) statement: "For any mover $M$ within a group, there may be no movers in that group with the signature of the balance partner of $M$ which are not the balance partner of $M$." The point is that we use the ensemble signatures to identify the balance partner, so we need to be able to identify it correctly. In practice, this means that, for example, if you wanted to use both one-way shooting and two-way shooting, you *must* put them in different groups if you want the (not-yet-implemented) automatic identification of balance partners to work. Previously, this was strongly suggested (because it made sense for analysis).

For now we are requiring that balance partners only come in pairs. In principle, I suppose you could have a move which has two different partners, but you would have to account for that in the acceptance probabilities. For example, you could have an A->B hop with one bias, and then a choice of two different biases for the B->A hop. Then you’d have to account for the relative probability of choosing each B->A hop move, as well the relative biases. I’m pretty sure this is mathematically possible, but it is such an unlikely and complicated example that it makes more sense, for now, to restrict balance partners to be 1-to-1. That said, the code does *identify* all balance partners, but throws a warning if the number of partners isn’t 1.

*Any `GLOBAL`-level strategy needs to set the `choice_probability` for all (group-level) movers.* This is so that we can reconstruct the correct `choice_probability` in other strategies. The point is that if you reorganize the move decision tree, it shouldn’t change the probability of doing each move.

*Each mover can only be used once in the scheme groups.* Note that this doesn’t mean it can’t be used as a submove of some other mover. So you can use the same replica exchange in the `'repex'` group and as a submove of a mover in `'my_fancy_sequential_mover'`’s group. However, at the group level, it can only appear once. This is again because each mover must have a unique balance partner.


dwhswenson:
Passes tests. Ready for review and merge.

jhprinz:
merging!

PR #291, opened 2015-06-26
MISTIS Analysis

### LINKED ISSUE #226: Networks
Time to explain a bit of the concept I have for `TransitionNetwork`s. Basically, these are just shortcuts to set up and analyze full network calculations for multiple state and multiple interface set TIS. They are built of `Transition` objects, which can be used later to analyze the rates. For the most part, they should quack a lot of like a `Transition` object, with the exception that, whereas `transition.rate` is a number, `network.rate` is a list (which can be mapped to a matrix).

Usage should look something like this:

```python
mstis = paths.MSTISNetwork(
    [
        (stateA, interfacesA, nameA, default_orderparameterA), 
        (stateB, interfacesB, nameB, default_orderparameterB),
        (stateC, interfacesC, nameC, default_orderparameterC)
    ]
)
# no problem if default order parameter is omitted, though needed for analysis

mstis_sim = paths.PathSampling(
    storage=storage_file,
    engine=my_engine,
    root_mover=mstis.default_movers(),
    globalstate=initial_globalstate
)

mstis_sim.run(5000)

mstis.rate_matrix(storage_file) # 3x3 rate matrix
tcpAB = mstis.transition(stateA, stateB).total_crossing_probability(storage_file)
plt.plot(tcpAB.keys(), np.log(tcpAB.values())) # total crossing probability for A->B transition
```

A few details of how the analysis will look still need to be worked out, partially because I’ll want to reorganize some stuff in the `Transition` objects as well. But the point is to keep the syntax for network analysis as close as possible to that of transition analysis, just that networks return lists. If `Transition.thing` is a number, then `Network.thing` is a list, and there may be `Network.thing_matrix` which is a matrix (e.g., `thing`==`rate`).

For multiple interface set TIS, the setup is:
```python
mistis = paths.MISTISNetwork([retis_transition1, retis_transition2, retis_transition3])
```

where each member of the list is a `Transition` object. The rest would be the same.

Eventually we should add (at least to `MSTISNetwork`) a `.disallow(stateA, stateB)` to remove certain transitions from being allowed. But that will have to come later. 

I plan to split this into several PRs: 
- [x] Building MSTIS Networks #227
- [x] Analysis of MSTIS Networks #266
- [x] Building MISTIS Networks #268
- [x] Analysis of MISTIS Networks #291

dwhswenson:
Adds support for the analysis side of MISTIS.

User still has to explicitly calculate the flux for MISTIS. For now, the notebook is setting the flux to 1.0 so we can get something in the rate matrix. If the flux is not provided, the rest of the calculation is performed, but the rates will report NaN. (The user can still access the total crossing probabilities, though.)

This closes #226.


dwhswenson:
Passes tests. Ready for review and merge!

jhprinz:
looks good, merging...

Issue #288, opened 2015-06-23
treelogic keys should be branches, not trees themselves


dwhswenson:
Why are we giving entire subtrees as keys? I thought the keys were only supposed to be the actual path visited to get to a given mover.

Both approaches give unique keys, but the current version gives unreadable keys. Example: consider the following tree (discussed in a previous conversation):

![image](https://cloud.githubusercontent.com/assets/8178546/8312103/2054fbd6-19db-11e5-8623-1dc75f164ec2.png)

I would expect the blue path to give me

```
[Root, [RandomChoice, [ShootB]]]
```

Instead, for `Root.key(ShootB)`, I'm getting something like

```
[Root, [Sequential, [AxB, ShootA, ShootB, BxA]], [RandomChoice, [ShootA, ShootB]]]
```

This does NOT help me trace the path through the tree to get to that mover!

I mean, here's a real example:

```
>>> print root.key(self.movers['repex'][-1])
[<openpathsampling.pathmover.RandomChoiceMover object at 0x10bbe4490>, 
[<openpathsampling.pathmover.RandomChoiceMover object at 0x10bd8c0d0>,
[<openpathsampling.pathmover.PathReversalMover object at 0x10bbe4c90>], 
[<openpathsampling.pathmover.PathReversalMover object at 0x10bbe4fd0>], 
[<openpathsampling.pathmover.PathReversalMover object at 0x10bbe4d10>],
[<openpathsampling.pathmover.PathReversalMover object at 0x10bbe4d50>], 
[<openpathsampling.pathmover.PathReversalMover object at 0x10bbe4e10>], 
[<openpathsampling.pathmover.PathReversalMover object at 0x10bbe4cd0>], 
[<openpathsampling.pathmover.PathReversalMover object at 0x10bd8c090>]], 
[<openpathsampling.pathmover.RandomChoiceMover object at 0x10bd8c150>, 
[<openpathsampling.pathmover.OneWayShootingMover object at 0x10bbe3f50>, 
[<openpathsampling.pathmover.ForwardShootMover object at 0x10bbe3f90>], 
[<openpathsampling.pathmover.BackwardShootMover object at 0x10bd8c050>]]], 
[<openpathsampling.pathmover.RandomChoiceMover object at 0x10bd8c190>, 
[<openpathsampling.pathmover.OneWayShootingMover object at 0x10bbe4ad0>, 
[<openpathsampling.pathmover.ForwardShootMover object at 0x10bbe4a50>], 
[...]

Issue #284, opened 2015-06-16
Make ReplicaExchange into ConditionalSequential EnsembleHops


dwhswenson:
I think it might be worthwhile to replace the current `ReplicaExchangeMover` with a `ConditionalSequentialMover` consisting of two (unbiased) `EnsembleHopMover`s.

This will give the same results: instead of thinking of replica exchange as a simultaneous swap, we think of it as two independent hops, but the whole move is only accepted if both hops are accepted.

I'd considered this before, but previously didn't see any advantage in it and thought it risked confusion, since that isn't the way people normally think about replica exchange. However, as I'm working on the `MoveStrategy` stuff, I think this might simplify switching between multiple-replica approaches (based on replica exchange) and single-replica (based on ensemble hops). That just becomes a matter of adding/removing the `ConditionalSequentialMover` which makes the replica exchange.

If there's no opposition, I would implement this by making `ReplicaExchangeMover` a subclass of `ConditionalSequentialMover`, so there would still be a separate class for it.

jhprinz:
This is a good idea, but will only work once the "delayed" acceptance works. Once this is done (still working on the PathMove/PMC update) the switch should be very easy and natural. It will also make more complicated replica exchange moves feasible.

Please go ahead.

dwhswenson:
Actually, I don't think it is so necessary for the `MoveStrategy` stuff, now that I'm further into developing that. So I'm back to my original mixed feelings about this, which boil down to:

* **Advantages**: build everything out of the simplest tools (if replica exchange *can* be described as two hops, then why not treat it that way?) It's more structurally and logically satisfying
* **Disadvantages**: no one things of RepEx as a combination of two different moves; everyone thinks of it as one simultaneous move.

Also, you're right that this would require the "delayed acceptance". I hadn't thought of that when I posted this -- definitely a subject we should take up before too long, but since I think I can do the `MoveStrategy` stuff without this, this issue doesn't make the "delayed acceptance" any more urgent than it was otherwise.

jhprinz:
I kind of like the idea to have very simple building blocks and create the rest from it. For everything that is commonly used (like RepEx, Minus, etc...) we will have a convenience mover that does that looks like a single mover, but internally uses the building blocks. This makes it easy to make changes.

At the end, all we want is to build the MoverTree in some way. 

We might even have two views on the tree:
- a simple view that only shows the main blocks
- a complete view that shows all little building blocks

This could actually be connected to the canonical property.

[...]

PR #295, opened 2015-06-30
[WIP] Sets of Trees

jhprinz:
This tries to solve some issues and cleanup PathMovers and PathMoveChanges (once and for all ;)

Addressing issue #284, #288.

Let me give a brief overview about PathMover and PathMoveChanges and their tree structure

#### PathMovers and the PathMoveTree

A PathMover is organized in a tree-like fashion. The idea is that each node is a Mover and gets a ´SampleSet´ (usually called the ´globalstate´) and will output a corresponding ´PathMoveChange´.
A PathMover itself can access the results of underlying PathMovers and do something with them.
The underlying PathMovers are children of a node and here also the order matters, e.g. in a SequentialMover, simply other movers are run in sequence.
This gets complicated when we allow to chose between different possibilities of movers. This means effectively that our PathMoveTree is not a fixed tree, but a 'RandomProcess' that outputs a tree. This gets even more complicated when we allow the decisions of which mover to chose to be based on the actual global state that is passed to the mover.

So, the PathMover itself is providing a set of movers to be run and also the logic on which of these movers to pick.

Now the problem that makes it hard to deal with these trees right now: We have two kinds of leaves / submovers of a PathMover node.

1. The leaves can mean choices of movers to pick, so they are part of the 'tree generating process'
2. The leaves can mean a sequence of movers
3. There is even a mixed one `ConditionalPathMover`

If we clearly separate these two concepts then the problem of keys can be solved and also searching and subtrees make sense.

The example that @dwhswenson gave in #288 hinted me to this. The keys he provided seemed unique and logical, but I could not make out what would happen if we have a sequence of movers. As long as we have choices in a mover like RandomSelectionPathMover there will be only one mover picked from a set of possibilities and hence it is enough in the key to just specify the mover that was picked. For a sequence of movers like 
```
mover = Sequential([shootA, reversal, shootA])
```
we want to be able to say which one we mean, the first or last one. So, in RandomChoices the same mover does not make sense (or could be combined into one), but in a sequence it does. So keys for sequential would look like this:
```
[Sequential, [shootA]]
[Sequential, [shootA], [reversal]]
[Sequential, [shootA], [reversal], [shootA]]
```
or a little simpler
```
[Sequential, [shootA]]
[Sequential, None, [reversal]]
[Sequential, None, None, [shootA]]
```
This key will point exactly to each possible mover and you can also ask
```
# check if a specific mover was called in a particular PathMoverChange
change = st.pathmovechanges[0]
[Sequential, None, None, [shootA]] in change
```
to check if this particular mover was called. So, this idea would solve the problem that we have with too long keys. I have implemented this and it works.

Important is here that these keys make only sense (so far) to be called with `PathMoveChanges` and not with the original PathMover. Simple example why this fails:

```
mover = RandomChoice([
    SequentialA([shootA, reversal, shootA]),
    SequentialB([shootB, reversal, shootB]) # this Sequential is different from the first!
])
# give keys for all movers in mover
print mover.keylist()
>>> [RandomChoice]
>>> [RandomChoice, [SequentialA]]
>>> [RandomChoice, [SequentialA, [shootA]]]
>>> [RandomChoice, [SequentialA, [shootA], [reversal]]]
>>> [RandomChoice, [SequentialA, [shootA], [reversal], [shootA]]]
>>> [RandomChoice, [SequentialB]]
>>> [RandomChoice, [SequentialB, [shootB]]]
>>> [RandomChoice, [SequentialB, [shootB], [reversal]]]
>>> [RandomChoice, [SequentialB, [shootB], [reversal], [shootB]]]
```
This seems reasonable so far. Now check if the last mover is in the mover
```
print [key in mover for key in mover.keylist()]
>>> [True, True, True, True, True, False, False, False, False]
```
The problem is that `in` check for being a subtree starting with the same root and this works for the first few, but when SequentialB is used this is the second node and hence it is not a subtree the way we think about it now. This would be work, but is not, what we want:
```
>>> [RandomChoice, [SequentialA], [SequentialB]]
>>> [RandomChoice, [SequentialA], [SequentialB, [shootB]]]
>>> [RandomChoice, [SequentialA], [SequentialB, [shootB], [reversal]]]
>>> [RandomChoice, [SequentialA], [SequentialB, [shootB], [reversal], [shootB]]]
```
It looks like SequentialA and SequentialB are run in Sequence which is not true. So what is now missing is a subtree check, that nows, that some leaves are sequences and some are choices.
Also this keys will never work in a PathMoveChange!

I have added a hidden function `_leaves` that will give detailed information about the meaning of the submovers. What is missing is to make the `in` function for PathMovers aware of the difference of choice and sequence.

I think it is important that most user do not have to deal with it and it should just work. It is also mostly to have unique keys that work for PathMovers and PathMoveChanges alike. So, ultimately all of this should work:
```
mover = RandomChoice([
    SequentialA([shootA, reversal, shootA]),
    SequentialB([shootB, reversal, shootB]) # this Sequential is different from the first!
])
# change is a concrete run
change = RandomChoiceChange([
    SequentialChange([shootBChange, reversalChange, shootBChange])
]) 
mover_keys = mover.keylist() # see above for list
print mover_keys
>>> [RandomChoice]
>>> [RandomChoice, [SequentialA]]
>>> [RandomChoice, [SequentialA, [shootA]]]
>>> [RandomChoice, [SequentialA, [shootA], [reversal]]]
>>> [RandomChoice, [SequentialA, [shootA], [reversal], [shootA]]]
>>> [RandomChoice, [SequentialB]]
>>> [RandomChoice, [SequentialB, [shootB]]]
>>> [RandomChoice, [SequentialB, [shootB], [reversal]]]
>>> [RandomChoice, [SequentialB, [shootB], [reversal], [shootB]]]

change_keys = change.keylist()
print change_keys
>>> [RandomChoice]
>>> [RandomChoice, [SequentialB]]
>>> [RandomChoice, [SequentialB, [shootB]]]
>>> [RandomChoice, [SequentialB, [shootB], [reversal]]]
>>> [RandomChoice, [SequentialB, [shootB], [reversal], [shootB]]]

# all keys should be valid subtrees of the whole mover
[key in mover for key in mover_keys] 
>>> [True, True, True, True, True, True, True, True, True]

# all keys in the PMC should be in the PMC
[key in change for key in change_keys]
>>> [True, True, True, True, True]

# only the keys of the mover that were used should be true
[key in change for key in mover_keys]
>>> [True, False, False, False, False, True, True, True, True]

# but of course all keys in PMC should be part of the PathMover (since they were generated by it)
[key in change for key in mover_keys]
>>> [True, True, True, True, True]
```

Finally it could make sense to treat `__iter__` as running in order over all 'keys' and not over the nodes. Much like a dictionary runs over all keys and so our special tree will run over all keys and you can use the key to access a specific node.
```
for key in mover:
    node = mover[key]
    assert(key in mover) # will always work
for key in change:
    node = mover[key]
    assert(key in change) # will always work
```

#### Summary
This PR changes the treelogic to be aware of different meaning of child nodes : sequence, choice

It adds
- [x] add `._leaves` to get information about meaning of leaves
- [x] keys are valid subtrees
- [x] keys should work for PMC and PM alike
- [x] change `in` for PathMover to use `_leaves`
- [x] example ipynb
- [ ] add nose tests 

#### Additional stuff
- additional testing


jhprinz:
PR #278 needs to be merge before this.

dwhswenson:
A few comments: Mainly, we need to get our terminology in order.

> This means effectively that our PathMoveTree is not a fixed tree, but a 'RandomProcess' that outputs a tree.

I would say that the move decision tree (please, let's give it a name that doesn't conflict with the red/green/orange trajectory path tree in visualization) is actually a directed acyclic graph. Yes, you can illustrate any path from a DAG as a tree, although it seems more natural to think of it as a sequence.

> subtrees

["You keep using that word. I do not think it means what you think it means."](https://www.youtube.com/watch?v=F8y-qy9N01I)

When talking about the tree data structure, a subtree is a node and *all* its descendants. Nothing in this is a subtree. This is why I have tended to use the word "branch" (I think the technical term is "path", but that's already overloaded enough for us). I intend by that the list of movers actually visited during a move. It looks like this is what you're implementing.

What you used before was some sort of subgraph of the overall DAG, but so is everything else. I think you'd call yours something like the "minimal depth-first subgraph to a given node"?

A word other than "subtree" should be used for whatever we're doing here. Bad enough that we're calling the move decision tree a tree when it isn't quite one. But these "subtrees" have nothing to do with what traditionally bear that name in data structures.

> There is even a mixed one ConditionalPathMover

Added as a hack for bootstrapping. Not used at all. Didn't realize it was even still in there.

Or do you mean ConditionalSequentialMover? Best way to think of that one is as a SequentialMover with an abort switch. That's really what it is -- it stops when there is 0% chance of the sequential move getting accepted. You could go ahead and do the remaining moves, but you'd just waste computational effort. (Very useful for the minus move, where you can abort after a failed cheap move without wasting time doing the very expensive part.)

This trick is only valid if it is true that, when we abort, it is not possible for the whole sequence to be accepted. Another option would be to have some sort of PMC for "didn't actually run" and return that for those moves. Then ConditionalSequential looks just like a normal sequential. Those other PMCs would have to hide in analysis, though.

> ```
> [Sequential, [shootA]]
> [Sequential, None, [reversal]]
> [Sequential, None, None, [shootA]]
> ```

Now we're getting into different definitions of the keys. 

The problem with this is that if I'm looking at the key for the last mover, and I was to ask the question "did I do a `reversal` before `[shootA]`". It's hard to tell that. This is what I was thinking in this context.

> The problem is that in check for being a subtree starting with the same root and this works for the first few, but when SequentialB is used this is the second node and hence it is not a subtree the way we think about it now

I do not understand this sentence or the code that comes after it. What is above it (`[RandomChoice, [SequentialB, [shootB], [reversal], [shootB]]]`) looks perfectly fine to me (personally, I'd group it as `[shootB, reversal, shootB]`, but there might be advantages to this. 

Is this what you are getting? What you want to be getting? Or is there some reason that this doesn't work? I can't tell from what came next. Why does `SequentialA` appear in that?

I don't understand the necessity of `_leaves`. Are you referring to what are normally called "children" of a tree node? Leaves usually refer to the nodes that have no children (these are our generating movers).

Further thought: I think I see what's going on. The problem really is that you're trying to make it a tree when it is a DAG. Look into that: it should make it possible to fix this without adding extra attributes.

As a tree (thinking of it this way will cause problems):
image

As a DAG:
image


More clear why this is a DAG when there's random choice within sequential:
image

Same setup, laid out more clearly as a DAG:
image

dwhswenson:
> Another option would be to have some sort of PMC for "didn't actually run" and return that for those moves.

Actually, this might not be such a bad idea. We didn't have PMCs when we implemented ConditionalSequential, so it didn't occur to us before. But recall that the whole reason these "keys" needed to have structure (instead of just being a list) was that you could have a setup like:

```
Conditional
+- Conditional
| +- SomeMove 
| +- ShootA
+-ShootA
```

Consider the sequence of movers `[Conditional, Conditional, SomeMove, ShootA]`. Is that:
```
[Conditional, [Conditional, [SomeMove], [ShootA]]]
```
or
```
[Conditional, [Conditional, [SomeMove]], [ShootA]]
```

Hard to tell without carefully inspecting the behavior of every move. But if we give these "ghost" moves (marked with an asterisk):

```
[Conditional, [Conditional, [SomeMove], [ShootA]], [ShootA*]]
```
and
```
[Conditional, [Conditional, [SomeMove], [ShootA*]], [ShootA]]
```

are obviously distinguishable. Change the PMC type for the asterisk; keep the movers the same.

I think this saves a lot of hassle with the key definitions.

dwhswenson:
Further thought: I think this can be accomplished to tossing in `None` for any "did not run" move.

I know this only works for PMCs, but as I said in #288, I'm not sure that it makes sense to use the same logic for PMCs and PathMovers. PathMovers form a DAG, PMCs form paths (branches) within that DAG. One is a 2D structure, one is 1D.

jhprinz:
Good points.

Yes, the last pictures illustrate exactly the problem with the tree picture. The problem is that we construct the MoveTree as in Fig.1 but mean something like Fig. 2. One way to remove this is to not use Sequential at all, but just allow a 'next' node. Still, that does not solve the second problem.

So, we need to decide how we want to make this available to the user. Right now, we are writing the structure in code as nested objects which induce a kind of natural tree structure. I think we agree that the for PMD this is much simpler. There is no choice, since these choices have been made and so it is just a chain of events/pathMoves. Here again, we have from the code a tree like structure. 

Question: Do we want to keep the tree structure for PMC? The picture with the red/green/orange shows the actual chain of events and also the tree-like structure into blocks and groups.

For PM this is more complicated. My idea was that if we keep a tree structure for PM, then we can show the actual PMC as a branch(starting at the root). That was my original intend. What I call subtree I think is a actually a branch. Starting at the root means, that the tree and the branch have the same top node.

The reason for me to keep walking for both PM and PMC about a tree is because of the tree structure we were always talking about in our heads and since the code looks a little like it, too. It will also group things together in ways we think about it:
- a chain of move events is what is happening
- a tree like organization of less and less complex moves is what I have in my mind
In Fig. 3 you have natural layers of complexity, in Fig. 4 you don't. If I would give you Fig 4 and ask for a key of the last 'bkwd' mover, what would you say. Fig 3. makes this easier. I could also say the tree adds more information to the structure and tells me which parts I consider not important when running other movers. Everything of same level or higher is part of the key, the rest is not. 

> Now we're getting into different definitions of the keys.

Agreed, I like the one without the None, better and you have the additional benefit to see in the key the steps that lead to this move. Important is that this simplified picture requires the tree-like structure to keep the key small. In Fig. 3 the last backward shooter would have key
```
[SequentialA, [1-Way], [rev], [1-Way, [bw]]]
```

> personally, I'd group it as `[shootB, reversal, shootB]`, but there might be advantages to this.

Me, too. The otherway is simpler to program :) My rule for brackets is if you read from left to right. `[` means one level down, `]` means one level up.

>  ... looks perfectly fine to me...

Yes, that is what I was trying to point out. The only problem is that the above keys will not work in the tree-like structure of a PM yet. This is because of the way I programmed it. In short: Since SequentialMove and RandomMove both have a list of submoves that are treated in the tree the same, there is no way to do `key in mover` for all keys in mover.

> I don't understand the necessity of _leaves. Are you referring to what are normally called "children" of a tree node? Leaves usually refer to the nodes that have no children (these are our generating movers).

My fault. Yes, I mean children. It is my attempt to keep the tree structure but add information about having a sequence of children or just to pick one of a children. Both cases need to be treated differently to create meaningful keys and for `in` / `__contains__` to be implemented correctly

Okay, before I proceed we should decide if we want to keep the tree-like organization (which I think at the time makes sense) because of 3 reasons:
1. from the picture in my mind 
2. the way we write code as a user
3. tree hierarchy provides a way to simplify a key

All I need to add for this is information of eah mover what I consider important for making keys and what not. For sequences (meaning movers of the same level) I use all preceding movers of the same level, while for other movers I do not need information of the same level. The problem remains for DAG. I need something that tells the DAG structure that I split (like in Random) and recombine.

dwhswenson:
> Question: Do we want to keep the tree structure for PMC? 

I think yes, sort of. I would put it this way: 

* The PMC forms a tree, but only as a matter of conceptual convenience. In reality, the only thing that matters is its depth-first traversal order, which gives the branch path taken by the move. The reason to use a tree structure instead of just a list is *only* so that siblings in a Sequential move are easy to identify.
* The key can be a flattened version of that. It does *not* need the tree structure it currently has (which makes it harder to use and understand). The key is just a list of movers visited in the order they were visited. 

> The picture with the red/green/orange shows the actual chain of events and also the tree-like structure into blocks and groups.

I think we're talking about two different pictures here when referring to "red/green/orange". I think you're referring to the visualization of the move decision process (e.g., in `move_strategies_and_schemes.ipynb`). When I said to call it something like the move decision tree, not the "path move tree," I mean that the expression "path move tree" could be confused with this:

image

And you see why the word "tree" is problematic! We've badly overloaded it.

> Important is that this simplified picture requires the tree-like structure to keep the key small.

The problem wasn't that the key was long; it was that the extra length was meaningless. I think the key should list all movers that were actually visited.

For example, it would be nice to be able to take

```
[SequentialA, [1-Way], [rev], [1-Way, [bw]]]
```

and distinguish which of the following it was:

```
[SequentialA, [1-Way, [fw]], [rev], [1-Way, [bw]]]
[SequentialA, [1-Way, [bw]], [rev], [1-Way, [bw]]]
```

This is nice because we're essentially just appending to the key as the move goes on, and I think that it more intuitive than appending and removing things from the key. From the PMC, this should be easy. In fact, as I mentioned above, we can actually use a flat list if we use these "ghost" changes for `ConditionalSequential`.

This is fundamentally different from a key that says:

```
[SequentialA, [1-Way, [fw], [bw]], [rev], [1-Way, [fw], [bw]]]
```
which reads as if *both* forward and backward moves were done in each `1-way`. That was my concern in #288: Not that the keys were long, but that they were misleading.

The difficulty doing it this way is in generating all keys for the PathMover. But I'm not entirely sure that there is a need for that. The question of whether you try every possible branch paths through the graph is unimportant; only whether each move is visited matters. Saving those details in the PMC allows the user to check for any unexpected correlations, but I don't think there's a need to enumerate all the possible branch paths.

That said, plenty of algorithms *do* exist for enumerating the paths in a DAG, so we could find/implement one of those if desired. I think you could even come up with a clever implementation as a generator, which solves some of the problems with the number of them. In addition, for our type of DAG (single source), it is very easy to calculate the number of paths, so a user wanting to iterate over the paths can set a maximum number of paths allowed before an error is thrown instead of running the long iteration.

> It is my attempt to keep the tree structure but add information about having a sequence of children or just to pick one of a children.

When you say "tree structure" do you mean the literal tree data structure or the overall tree-like organization? Again, I think the best way is to really convert the PathMovers to a DAG.

Basically, the choice is between
* hacking a modification onto a tree so that it has special cases that let it act like a DAG
* implementing it as a DAG in the first place

The second sounds a lot cleaner to me. It can be handled by giving every PathMover a `._next` attribute, which returns a list of possible next movers. This also cleans up the fact that some container movers take a single mover (`Filter`, etc) and some take a list (`RandomChoice`, `Sequential`, etc) -- there are a number of checks for whether an a mover has either `.mover` or `.movers` which would just be replaced by `if mover._next is not None`. This default to `None`. In initialization, wrapping movers and random choice movers just put a list of their submovers into `._next`. Sequential movers put the first item in their sequence, and for all movers that aren't the last in the sequence, they trace down to the movers where `._next in None` and modify that mover's `._next` to point to the next mover in the Sequential. (This is probably best implemented with a `set_next(submovers)` function, so that you can get recursion in the process).

If you want to use this as a mix-in, the `._next` property because part of the mix-in superclass, and the `set_next` function gets overridden in specific subclass instances (mainly just variants of `Sequential`).

The advantage of this is that the user sets things up in a more intuitive tree-like approach, but internally we're always working with a proper DAG. This should make it more straightforward to implement things like enumerate all possible branch paths: just look up a good algorithm for doing that on a DAG, and implement it using `PathMover` as vertex and `._next` as a list of directed edges from that vertex. Or better yet, find some library that has that sort of algorithm, and use its data structure.

> Okay, before I proceed we should decide if we want to keep the tree-like organization (which I think at the time makes sense) because of 3 reasons:
> 1. from the picture in my mind 
> 2. the way we write code as a user
> 3. tree hierarchy provides a way to simplify a key

I'm not 100% sure exactly what you mean here. Are you referring to the tree-like treatment of the decision graph, or in the PMCs? As I mentioned above, I don't think these are the same structure: PMCs are 1D, decision graphs are 2D.

For setting up the decision graph, here's how I'd say it: Consider the last two figures in my previous comment. These represent the same system, but are organized differently. The represent the same truth, but with different perspectives.

I think that we want users to think about it like the first version, which is how things are already implemented. I think this is also what you've been saying: basically, it's how things are implemented, and it's also the most intuitive approach.

Internally, when dealing with the fundamental data structure, we need to remember that it is more clearly imagined of in the second version. I think that the keys should reflect this. I think that the choice of what functions are available (iteration, etc) should reflect this.

Technically, you could implement both PMCs and PathMovers as DAGs, (the true tree structure for PMC is also a DAG) but I think that is unecessarily complicated. To me, it makes sense to separate these two data structures.

jhprinz:
I think I see what you are heading for. 

#### how should a key look like and what is its meaning

> I know this only works for PMCs, but as I said in #288, I'm not sure that it makes sense to use the same logic for PMCs and PathMovers. PathMovers form a DAG, PMCs form paths (branches) within that DAG. One is a 2D structure, one is 1D.

That is a very nice analogy with 1D and 2D. Good point: Do we want the same logic for PM and PMC? I though for keys that would be very nice. You get the key for a specific move in your PM and can then check, if that particular mover was called using `key in change`. This feature does not require the whole past of the PMC before each move to work. This would only be useful, if you want to use the key for analytic purposes. And as I understand that is what you are doing.

I would argue against doing that and compare the key to a key in a dictionary. I would not put additional meaning into it than being able to identify the object to want.
The information about what happened before in contained in the PMC and can easily be extracted. Also, If you want to check if two mover coincide in the same PMC just check both
```
if key1 in change and key2 in change:
```

> The problem wasn't that the key was long; it was that the extra length was meaningless. I think the key should list all movers that were actually visited.

That was, what was implemented before. Exactly that, but only for PMCs because for PM there is (as we agree) no unique key, because it depends on what was visited before the mover. The long list of movers in a list you plotted was actually from a PM and this (in the old implementation) does not make sense.

So, I understood you correctly and implemented it that way. I think our confusion was, that you called the key function on a PM and not the PMC.

Just let me say again, that I thought the key should be something to identify a specific point in a PM or PMC.

I think my confusion was a little bit that I thought we want a key that can be derived from the PM and works for every PMC. That was what I was trying in this PR. The only way I see to do this is to not put the whole past of single mover in a PMC into the key, but only so much that it allows to pinpoint a specific mover from the PM and check if any of these ones have been used in a particular PMC.
This optimization toward a short key of course removes information about the past. However, once you have a key, it would be very easy to extract the past of a PMC that lead to that specific mover (needs to be implemented!)

Example:
```
mover = RandomChoice([
    SequentialA([shootA, reversal, shootA]),
    SequentialB([shootB, reversal, shootB]) # this Sequential is different from the first!
])
# change is a concrete run
change = RandomChoiceChange([
    SequentialChange([shootBChange, reversalChange, shootBChange])
])
```
Then getting the key for first shootB in SeqB gives in both cases
```
[RandomChoice, [SequentialB, [shootB]]]
```
but does not contain the past (although in this case it acidentally does). You would call something like
```
print change.past(key)
>>> [RandomChoice, SequentialB, shootB]
```
or whatever long list of things that happened before.

#### Underlying Tree vs DAG

All of the above is independent of the way we implement it. And I have nothing against using DAG under the hood. I still like to think about trees and it would be nice, if could keep the possibility to work with trees the way we do now - so no changes in the way we setup a PM. And I think we agree on that one.
The decision is a little easier for PMC because they are 1D. Maybe it is really a good idea to move to a graph layout and use all the nice librarier available for that. 

#### Movers that do not fulfill detailed balance
What to do with PartialAcceptance, Conditional, RandomAllowed, LastAllowed, etc which do not fulfill detailed balance?

I had the idea to make all the non-balanced movers into strict generators (although we are not using this). It means simply, that you cannot apply the mover to a sampleset, but only to specific samples using the `__call__` function. Balanced movers can be used for sampling and hence be used with sample sets.
This means you can still do an ensemblehop and create a new sample, but you have to do it by hand. Of course also balanced movers can be 'called'. It would also mean that there is no BootstrappingMover and you cannot do Bootstrapping with a PathSimulator. But since each sample knows its origin we will keep track of how the samples were generated.

Put simply, you can use the sample generation tools we have for anything, but for sampling you need balanced movers.

#### Decisions for the refactoring of PMC / PM

These are question that I believe be need to agree upon. I will mark what I think is good with [JH]

- [JH] Make keys smaller and drop "past" and allow only to identify specific node in the DAG/Tree. Information about past would be responsibility of PMC and a PMC would get a function that handles that.
- [JH] Keep code to construct PMs the way it is
- [JH] Remove generator idea not needed
- [JH] Turn non-balanced movers into generators
- [JH] Keep the current tree logic for PMC
- [JH] Keep an extended tree logic for PM. (I fear that constructing a DAG in the init function will mess up storage ?!?, but that is not sure, I would have to think about that)
- [JH] If we keep tree logic have convenience function to return DAG to run with graph libraries
- Change underlying structure of PM to DAG (Although I prefer this from a mathematical point, I think changing the underlying structure will cause lots of problems when recreating the tree information from the DAG, e.g. in loading.)
- [JH] Add more convenience functions to transform the tree structure into e.g. lists, graph, etc...
- [JH] `__contains__` for PM will check, if a given 1D key, or branch or PMC could be generated by the 2D PM as in `Is this 1D PMC element part of the 2D PM structure`

[need a break ;)] but wanted to post for you to read about it.

dwhswenson:
Longer comment coming later, but one quick question for clarification: 

You express concern about using constructing the DAG during init because it reloading. Can you explain that to me? Why isn't it like any other object that has computationally identified parameters? It seems to me like all we have is the `._next` parameter, which does not *create* any new movers, only points to movers. When you restore a SequentialMover, do you not have to restore all its submovers before you finish restoring the Sequential?

jhprinz:
Hmmm, another potential problem I just found with DAG vs tree. Imagine:
```
mover = Sequence(shootA, shootB)
```
Then after initialization shootA needs to know about shootB if we have a `._next` attribute. So the objects cannot be saved or created independently. In theory the storage should be smart enough to load and create objects in the correct order, but I a not sure, if we run into trouble here.

Was it clear, what I mean? In a tree the coupling is in the Sequence Object, while for a DAG a node needs to know its following node.

jhprinz:
> Can you explain that to me?

Well, it was more a hunch that a clear problem. I think we need to make sure that any created or linked properties are saved and restored correctly. This might only be a problem of being thorough.

The bigger problem is what I posted before.

dwhswenson:
> Then after initialization shootA needs to know about shootB if we have a ._next attribute.

First, am I incorrect that storage, by default, completely loads all the objects which are inputs to another object's `__init__` function, then calls `__init__` with those to recreate the object? If that's how it works, then doesn't it do something like the following?

Worst case: assume that shootA and shootB haven't been loaded yet.

1. Load Sequential from storage
2. In the process of loading Sequential, we load all its submovers -- these are required before running `__init__`, since they are input to `__init__`
3. In `SequentialMover.__init__`, we call a function to set `._next` on all its submovers. Details of the implementation of this aren't completely trivial, but should be feasible. It boils down to implementing a function of the same name with slightly different functionality in Sequential and in RandomChoice/Wrapped/others. Call it something like `identify_leaves`, since that's what it does. 

If shootA or shootB have been loaded, it should only be easier -- but it doesn't matter, because by the time you're setting `._next`, haven't the submoves been loaded?

I'm pretty sure this also works with sequential movers that have other sequential movers as submoves: the one lower in the process already has the `._next` for its submoves set before the `__init__` for the one higher in the process is called -- except, of course, the last move in the lower sequence, which is exactly the one we want to find and point to the next move in the higher sequence.

And all of this is just to avoid storing an extra variable that can be computed. If you say it can't be computed, then just store `.next` and there should be no problem.

jhprinz:
I think you are right. Storage should not really be a problem (if I have no bug in storage). It does what you say and will reconstruct objects by calling the init function.

dwhswenson:
#### how should a key look like and what is its meaning

My thinking goes like this: 

1. For PMCs, we need an easy way to get the whole history of movers visited during the move up to and including the given mover. This is what I called the "key". It is both useful as an identifier and it is useful in analysis.
2. For PathMovers, it might be nice to be able to use that same "key" in some way. So the key in the PathMover structure should be the same as the equivalent key for PMCs.

If we can not get this branch-like key, I don't know that if there is actually a use for any other kind of key.

Since since in a DAG there is not necessarily a *unique* path to a node, for PathMovers this either needs to be plural "keys" or not exist at all.

I think it is better to not have "keys" for PathMovers than to have a key which is confusing.

All of this brings me back to the question: what parts of tree mix-in are used by PathMover? Should it even be there?

> This optimization toward a short key of course removes information about the past. However, once you have a key, it would be very easy to extract the past of a PMC that lead to that specific mover (needs to be implemented!)

Yes, but I think it is very confusing for the key given by the PMC to differ from the key used in the PathMover.

Note that there might be some option to have it both ways by using something based on the wildcards (or cats?) you implemented.

#### Underlying Tree vs DAG

It sounds like we're pretty much in agreement that we want the user-facing setup to be as it is now, but internally, we'd like to make that into a DAG. At least, I think we're fully agreed that this is the way we would *like* it to work: are we agreed that it is feasible to make it work that way?

#### Movers that do not fulfill detailed balance

> What to do with PartialAcceptance, Conditional, RandomAllowed, LastAllowed, etc which do not fulfill detailed balance?

If we're not using them, I say remove them before 1.0. Add things as we need them; I think it is very dangerous to ship code with features that we don't have clear ideas on how to use.

We might also group these under some other heading, like "non-balanceable", either with a flag or with an abstract superclass. Then a PathSampling calculation can run through the movers before starting to sample and raise a warning about the use of unbalanced movers (maybe having a special flag that can be set in the calculation to ignore the warning). This would satisfy me when it comes to protecting users from their own stupidity. It has the twin abilities that it stops a stupid user from being stupid, while still allowing a smart user to insist that they want to do something that, on the surface, looks stupid.

Note that also this allows us to check the behavior BEFORE use, instead of only DURING use.

I don't fully understand your "strict generator" approach, but it sounds more restrictive and it also sounds like it is harder to handle reasonable cases. For example, you say:

> This means you can still do an ensemblehop and create a new sample, but you have to do it by hand.

This sounds bad to me. An ensemble hop can have a balance partner. It should work like everything else. The code will complain if it doesn't have a balance partner. It doesn't matter that it is not its own balance partner. It sounds to me like you're proposing something that will make the implementation of various combination movers harder.

There are three types of movers this think of:
* self-balancing : balance partner is self
* balanceable : balance partner is not self, but can exist
* non-balanceable : no balance partner can be made

It sounds to me like you're grouping the second and third together, entirely because of concerns about the third. I think we should treat the third separately from the first two. 

Non-balanceable moves have no place in a path sampling move scheme. They break detailed balance. Balanceable moves can be used in path sampling move schemes, as long as they have balance partners.

We already have the work in place to handle self-balancing and balanceable movers (see #289). So the question here should only pertain to non-balanceable.

#### Decisions for the refactoring of PMC / PM

> [JH] Make keys smaller and drop "past" and allow only to identify specific node in the DAG/Tree. Information about past would be responsibility of PMC and a PMC would get a function that handles that.

As mentioned above, I think there needs to be a clear connection between the keys for the DAG and the PMC, otherwise I don't think the keys for the DAG serve any real purpose.

> [JH] Keep code to construct PMs the way it is

Strongly agree.

> [JH] Remove generator idea not needed

Clarification: you mean here the addition of the word `Generating` to all the movers? I definitely support that, along with anything else that treats these differently. If they're allowed in the move decision tree (DAG), then I think they should be the same general type of node.

> [JH] Turn non-balanced movers into generators

As above, I think there's a better way to do that. Movers that are not (normally) allowed in the move decision process *should* have something to warn of that. But movers that *are* allowed shouldn't be harder to work with.

> [JH] Keep the current tree logic for PMC

Agreed.

> [JH] Keep an extended tree logic for PM. (I fear that constructing a DAG in the init function will mess up storage ?!?, but that is not sure, I would have to think about that)

I disagree -- has the discussion in previous comments addressed your concerns?

> [JH] If we keep tree logic have convenience function to return DAG to run with graph libraries
Change underlying structure of PM to DAG (Although I prefer this from a mathematical point, I think changing the underlying structure will cause lots of problems when recreating the tree information from the DAG, e.g. in loading.)

See previous.

> [JH] Add more convenience functions to transform the tree structure into e.g. lists, graph, etc...

YAGNI. Not opposed, but this isn't worth the time until we find an actual need for it. Besides, we don't know the format we'll want for it until we actually need it.

> [JH] __contains__ for PM will check, if a given 1D key, or branch or PMC could be generated by the 2D PM as in Is this 1D PMC element part of the 2D PM structure

This I like. I think it will be made easier by treating the PMC as a graph and looking for these as subgraphs. There's a little trickiness because you're looking for the same sequence of values in the graph, not the same sequence of nodes, but I'm sure someone has already studied that problem in graph theory.

dwhswenson:
This `._next` attribute I suggested won't work. If the same mover exists at more than one place in the tree, then it doesn't have a unique `._next`. (I literally woke up in the middle of the night with this realization.)

I'll think about better ways to do this. I still think that we really want to use a DAG under the hood, and I think we want to use paths (branches) in the DAG as the keys (making a clear connection between these keys and the chain from the PMC.) The question is how to best implement that.

jhprinz:
Funny, I had a similar thought that the `._next` property will depend on the context and something like
```
mover = Sequential([shootA] * 10)
```
will not work. Right now, I think we want the possibility to analyze the structure as if it were a DAG, but keep internally the extended (2D) tree structure. I will write about my  current idea below.
 
The positive aspect of the tree is that each mover will not depend on its context or where it is embedded. It only depend on its submovers. If we want that then a tree is the natural way to do that and that is probably the reason why we always have this tree picture. it expresses conditional dependencies.
This is somehow lost in the (plain) DAG.

Concerning the keys I still prefer the shortest possible path (branch) that points to a particular mover. This (what I call a key) is only meant to identify a particular point in the DAG/2DTree and 1DPMC uniquely. I would really use here that some movers are (condisered to be) independent. What you call the key I would call a property of the PMC, because it explicitly knows about the complete past inside a PMC that happend before a specific mover (that can be selected using my _key_).

I see, that we potentially will only need my key to create your key and hence it is kind of pointless to have both. But it seems to be the simplest connection between DAG/2DTree and 1DPMC. For example you could ask a PM for all mover keys and then test a PMC which if these were called.

A last thought. Perhaps the question to really pick either DAG or 2DTree is wrong. We might have both. My _key_ is on the Tree side and your _key_ is on the DAG side. Like haveing 2 views on the same thing. DAG is kind of linear and shows really what happens after what. Tree shows the conditional dependencies.

Right now I would just leave the PM structure the way it is or is there a reason to convert it to a DAG to store and load it. As long as we can access it there should not be a problem.

We could have a function `.next` that would work on a PMC like

```
pmc = ...
# this will give the next key in the chain
next_key = pmc.next(key)

# iterate in order
for key, change in pmc.iteritems():
    # key would be my key
    # change would be the 

# alternatively using next
key = pmc.root
while key is not None:
    key = pmc.next(key)
    change = pmc[key]

# or we simplify
key = None
while key = pmc.next(key):
    change = pmc[key]
```

Problem is your _path_ (or what I used to call past) needs a function. Like `.next` we can also have `.prev` and then the past is the list of recursively called `.prev`
```
dag_past = pmc.past(key)
dag_future = pmc.future(key)

future = []
key = None
while key = pmc.next(key):
    future.append(pmc[key])

past = []
key = None
while key = pmc.next(key):
    future.append(pmc[key])

```

I should clarify that the `__getitem__` works with branches and selects the last element, if the branch exists. So, my _keys_ can be interpreted as a small branch, While your keys are larger and contain the _left_ part of the tree (the past). The `__getitem__` is closely related to the `__contains__`. I would say both can also use wildcards for testing specific branches and find the first element that matches a branch-wildcard.

All of this could be implemented pretty fast and allows most of what we want except the underlying DAG structure.

> Generators above...

Yes, I meant remove that. It is legacy and not used anymore.

Any thoughts?

jhprinz:
Hmmm, just deleted a huge post... Because I clicked to early on PRs... I am really annoyed...

Basically,

- yes, I realized the same problem
- I would like to keep the trees we have right now
- keys are short and point to a specific node in the tree
- keys work for PM and PMC
- we add a `.next` and `.prev` function to PMCs like
```
pmc = ...
# get the next key
next_key = pac.next(key)
prev_key = pmc.prev(key)
```
- iterate changes in a PMC exactly like iterating a (ordered) dictionary in python
```
for key, change in pmc.iteritems():
    pass
```
or
```
for key in pmc:
    change = pmc[key]
```
or even
```
key = None
while key = pmc.next(key):
    change = pmc[key]
```
- the `__getitem__(key)` for PM or PMC picks a specific node in the tree indicated by the key
- the key itself is a branch that is unique enough to identify a node
- your _key_ would be accessed using a `.past` function, e.g.
```
past = pmc.past(key)
>>> [shootA, shootB, reversal]
```
- `__contains__` just tests if a branch is in the PM or PMC
- `__contains__` and `__getitem__` can have wildcards in which case the first element found it returned

So, we keep the PMC, PM almost it is now and add DAG capabilities. So you can traverse using next or you can walk the tree (and both is the same!)

Sorry for the brevity, but I did not wanna repeat the long post now.

jhprinz:
Hooray, it got posted this time!

dwhswenson:
Is the long post not the one right above the second post? Seems to say a similar range of things....

dwhswenson:
I don't think leaving it only at the PMC level works. Very simply, PMCs only exist *after* we've done sampling. If we want to check things during the setup of a calculation, PMCs do not exist, but the move scheme does. This is my reason for wanting these things to work based on the move scheme, not the PMC. We need to be able to make a DAG from the move scheme directly, without PMCs.

I think the best alternative might be this: every mover has a function `mover.graph()` (or something like that) which makes a proper subgraph of the DAG starting from that mover. A node in the resulting DAG is implemented in the truly generic sense of having just a `value` attribute (which is the mover) and a list of `edges_to`. Obviously, `root.graph()` is the one we're really interested in.

In this case, note that the same move can be in different places on the DAG. The DAG nodes are a separate class that wrap the mover. I would suggest seeing if there is a good package with a lot of DAG analysis features that already does this.

Again, getting the DAG from the movers, with ***no reference to the PMC***, is extremely important for at least the following reasons (if not more):

1. Check things about the scheme *before* running a calculation
2. PMCs check that you *visited* enough paths: DAG gives you the list of possible paths. You can't know whether a path was missed unless you have the DAG from the move decision tree.

> Concerning the keys I still prefer the shortest possible path (branch) that points to a particular mover. This (what I call a key) is only meant to identify a particular point in the DAG/2DTree and 1DPMC uniquely.

As I said before, I'm bothered by the fact that these keys are not transferable between the DAG picture and the branch/PMC picture. I'll ask again: **is it possible for the DAG/tree picture to use the wildcard approach you implemented to generate a key that will work for either the DAG or the PMC**? This seems to me the best of both worlds: you don't have to enumerate all possible paths in the DAG, but you get something that works with the PMC as well.

> My key is on the Tree side and your key is on the DAG side.

I'm not sure that what I'm suggesting is coming across. To me, a mover in a DAG has a list of keys. No single key is sufficient. Each of the those potential keys represents one of the PMC keys that could reach that mover.

This is why I'd be perfectly happy with wildcards as a shortcut to describe the key.

> Like haveing 2 views on the same thing. DAG is kind of linear and shows really what happens after what. Tree shows the conditional dependencies.

I think this adds confusion. Again, shouldn't using wildcards as part of the key should solve these problems?

***

Also, I don't see anything in your recent notes about my self-balancing, balanceable, and non-balanceable distinction on movers. Does that idea make sense?

jhprinz:
I am back :) So let's see if we can wrap this up now...

> Also, I don't see anything in your recent notes about my self-balancing, balanceable, and non-balanceable distinction on movers. Does that idea make sense?

Yes, it makes total sense. But I would not start a discussion about this in this thread. 

Okay, let me try to clarify a few points

> I don't think leaving it only at the PMC level works

I agree. Did I say so? My view on this is that
- PMC is 1D,
- the full decision tree / PM is a 2D tree
- the full decision tree has an identical DAG representation

So, a mover tree is more than a PathMoveChange tree.

I still mostly think about a Mover as a tree with branches where decisions will be made rather than a DAG. The DAG is correct but looses the tree order we impose on construction, so in my opinion the tree picture has more information than the flattened DAG.

The next problem with a DAG is that during construction you need to have information about moves that are in the tree picture at a higher level. I think we discussed this before and that means that we break the possibility to reuse movers at different points. The reason is simply because the edges leading to the next node are different. The only solution to this would be (as you say) have a DAG and being the movers the nodes. So we would construct a PathMover the usual way (we also agreed that this is what we want) and then there is a `.graph()` or `to_dag()` function that returns a dag with all movers at the nodes. This is totally fine with me. 

> Again, getting the DAG from the movers, with no reference to the PMC, is extremely important for at least the following reasons (if not more):

Also agreed these are "two pairs of shoes" and the PMC must be independent of the PM and thus its DAG representation.

> As I said before, I'm bothered by the fact that these keys are not transferable between the DAG picture and the branch/PMC picture.

I think I do not unterstand that concern or at least I need to ask. What I proposed is a key that works uniquely in the PM-tree view (that has a DAG representation) and in all possible PMCs that can be the result. So you could easily pick a specific mover in your PM tree and check, if this mover was called in your PMC uniquely. There is no need for wildcards. (I still like the cats better :)  The wildcards were only meant to allow to check for more complex move structures. It originated when we could check for arbitrary subtrees. Like check if anywhere a One-Way Shooter was followed by a Backward, etc...
Note that what I proposed is the shortest possible key (although using wildcards might make them even smaller). You also do not have to evaluate all possible paths in the DAG.

If I understand you correctly you want a key that in addition also works in the DAG representation?
And I don't see how the wildcards help. 

Important for my keys are that you can select a specific mover always independent of the others but uniquely. So my naming scheme let's you point to a single mover in the tree. There is no notion of what happened before or after, except the direct path of decisions that lead to this mover. If you want to enumerate all possible paths then what you do is create a set of possible PMCs and so the PMC itself is sort of a key for that particular path (although is has much more information). I don't think that is necessary for a key that selects a mover in a tree.

> You can't know whether a path was missed unless you have the DAG from the move decision tree.

Not totally sure what you mean here by missed. If you want to check, if all possible paths where taken then yes, you need to enumerate all possible paths, but that can be done from the Pathmover or DAG alike. I don't see, why you need a special representation of PMtree or DAG are equal. 

Maybe we are really talking about two thing the whole time. You talk about keys for paths in the PMtree/DAG and I talk about keys to select single nodes. If that is correct you could indeed take your path-key and check if a PMC matches that path. But then wildcards would only be used to remove redundant sequences of movers is that correct. 

> I'm not sure that what I'm suggesting is coming across. To me, a mover in a DAG has a list of keys. No single key is sufficient. Each of the those potential keys represents one of the PMC keys that could reach that mover.

Now, we are getting somewhere. Alright, and I am saying that my key is used to pick a mover in the PM/DAG and it is sufficient. and it is independent of any other decisions. 
You say that each PMC has a unique key representing the path taken in the DAG and then each mover in the DAG can be has several keys given by the keys from the subpaths that lead to that mover.
My idea was to use the structure of the PMtree and cut off all branches that do not contribute to the decisions that lead to a specific mover (an information not present in the DAG) and only use this minimal subtree as a key. This subtree must be also contained in each PMC that uses that mover and so each PMC will use the same key. 

I hope we agree that my solution will allow to pick each mover, but it will not contain information about the full past that lead to a mover. If you want keys that can do that then I understand your concern. 

Would it make sense to keep both concepts? Have a key that can uniquely identify a mover with minimal necessary context in PMC and PM/DAG and a key that can differentiate between the different full pasts that lead to that mover in the PMtree/DAG.

#### Example

A sequence of three onewayshooters

```
fw = ForwardShooter()
bw = BackwardShooter()
rc = RandomChoiceMover([fw, bw])
se = Sequential([rc, rc, rc])
```
There are 10 movers although with repeats only 4 unique ones.

What do you consider (a) good key(s)?
My idea to pick each of the 10 movers would be
```
[se]
[se, [rc]]
[se, [rc, [fw]]
[se, [rc, [bw]]
[se, [rc], [rc]]
[se, [rc], [rc, [fw]]  # example
[se, [rc], [rc, [bw]]
[se, [rc], [rc], [rc]]
[se, [rc], [rc], [rc, [fw]]
[se, [rc], [rc], [rc, [bw]]
```

If we select the second forward shooter then it can have two path-keys that lead to it correct, because the first randomchoice mover can have two outcomes. This means that
```
[se, [rc], [rc, [fw]]
```

can have two pasts and your (path)-keys would look like this:

```
[se, [rc], [rc, [fw]]   #minimal key
[se, [rc, [fw]], [rc, [fw]]  # 1st possible full past
[se, [rc, [bw]], [rc, [fw]] # 2nd possible full past
[se, [rc, [*]], [rc, [fw]] # joining 1st and 2nd using wildcard
```

in my implementation the last wildcard version is identical to my minimal key. So maybe I am already using wildcards?

Note that if you have access to all the subsides you can construct the key directly.

Okay, let's stop here.

dwhswenson:
My thought process runs like this: the DAG is a picture that we have *before* the simulation. It is a representation of possibilities. The PMC only exists *after* the simulation; it is the story of what actually happened.

The problem is that we will have different key to use before and after the simulation. This is weird. If the user wants to either manually generate an expected key, or interpret a computationally generated key, it would be a lot easier if the key looked the same whether you were using it to describe the DAG or the PMC.

So my question becomes the following: when do we use the key from the DAG?

It can be a pain for a DAG to enumerate all the possible paths through a node (all the PMC keys for a PM). However, it should be really easy for a DAG to confirm *whether* a given PMC key is allowed within it (especially if we make the reasonable restriction that the key has to start at the top of the "tree" structure.

The only examples I can think of where we would use a key for the DAG are situations where we want to compare the DAG to the set of PMCs. For example, we might (for a simple DAG, with no long sequential movers) ask the question, "Did our simulation include every PMC that is possible for this DAG?" — this might be a good first test of convergence. It seems to me that the best way to do this is to enumerate the moves in the DAG, and compare them when the PMCs (passed through something that acts like `uniq` to simplify things: probably use the PMC keys as dictionary keys to count the number of times each PMC key occurs). But in that case, we’re not talking about a unique key that identifies a mover: more than one "key" is associated with each mover.

Any other case for a key I can think of would be specifically to identify the various contexts for a given mover within a DAG. In that case, we again want to enumerate the possible PMCs to get you there.

The example keys you give for DAGs are good on their own, but my question is when do we need keys for the DAG that are not compatible with the keys from the PMC. I can’t think of any examples.

***

Addition to link these general thoughts to specific statements above:

> I still mostly think about a Mover as a tree with branches where decisions will be made rather than a DAG. 

The problem is that it *isn't* a tree; at least not on a per-PathMover basis. You only have a tree until you hit the first `SequentialMover`. *You can not analyze these structures as trees, because they are DAGs.*

> I think we discussed this before and that means that we break the possibility to reuse movers at different points.

We will have to encapsulate objects to get the DAG. Wrap movers in some `DAGNode` object. You can probably find a general such object in some other library; there may even be something like that in NetworkX.

> I think I do not unterstand that concern or at least I need to ask. What I proposed is a key that works uniquely in the PM-tree view (that has a DAG representation) and in all possible PMCs that can be the result.

I hope that my explanation above clears this up: the situation I can imagine is that you want to check whether all possible paths in the DAG (perhaps limited to ignoring submovers of `RandomChoiceMover`s within a `SequentialMover`) have been visited in the course of your simulation. This requires some "common language" between the DAG and the PMC.

> If I understand you correctly you want a key that in addition also works in the DAG representation?
And I don't see how the wildcards help.

Yes, I want any keys returned in either case to be "in the same language." If you take the keys that you illustrated above, but took any `[rc]` and replaced it with `[rc, [*]]`, I think that would be what I was talking about. Such a key would work for the PMCs, as you show later, so you should be able to generate all such keys from the DAG and see whether all of them match at least one PMC. This gives an in-between check that doesn't enumerate *all* possible decision paths through the DAG, but checks certain important parts were reached. Ideally, we'd also have the ability to enumerate all paths explicitly, at least for cases where the sequence of random choice movers wasn't too long.

> Maybe we are really talking about two thing the whole time. You talk about keys for paths in the PMtree/DAG and I talk about keys to select single nodes.

Yes, this is what I'm trying to say. I don't understand where we would use keys that select single nodes. Why not just use the nodes themselves? The key only has value because it provides context for the node. If the node exists in more than one context, it *must* have more than one such "key" associated with it. This multiplicity of keys can be represented either by a unique key that uses wildcards, or by listing all such keys.

***

My view is that the DAG (which might be derived from a given root mover using `mover.subgraph()`) needs to have a way to give keys in the same "language" as the PMC keys, maybe as `mover.subgraph().keys(contained_submover)`. The path move "tree" itself has no need for keys at all. The keys that we use should be interchangeable between the two. I think using wildcards gives a compromise that saves us from having to enumerate all possible combinations when you have a sequence of random choice movers.

jhprinz:
Alright, let's go through this and confirm my understanding:

> My thought process runs like this: the DAG is a picture that we have before the simulation. It is a representation of possibilities. The PMC only exists after the simulation; it is the story of what actually happened.

Agreed. Exactly, the DAG (in my 2D Tree representation) is the "factory" and fixed and the PMCs are single "from start til the end" paths in the DAG (or a specific subtree in my 2D tree representation)

> The problem is that we will have different key to use before and after the simulation. This is weird. If the user wants to either manually generate an expected key, or interpret a computationally generated key, it would be a lot easier if the key looked the same whether you were using it to describe the DAG or the PMC.

Yes, that argument is true, but only if we talk about keys for paths and not for nodes what I was aiming at. 

> It can be a pain for a DAG to enumerate all the possible paths through a node (all the PMC keys for a PM). However, it should be really easy for a DAG to confirm whether a given PMC key is allowed within it (especially if we make the reasonable restriction that the key has to start at the top of the "tree" structure.

True, if we have a complicated mover there could be really a lot of possible paths (which can grow easily exponentially). And the reverse is trivial. And this works right now in the current implementation. You can write
```
if my_pmc in global_mover:
    pass
```
it will check, if this particular PMC could have been generated by that PathMover. It does this by checking, if the subtree structure can be placed into the full 2D tree (or the equivalent DAG ) of the PathMover.

> The only examples I can think of where we would use a key for the DAG are situations where we want to compare the DAG to the set of PMCs. For example, we might (for a simple DAG, with no long sequential movers) ask the question, "Did our simulation include every PMC that is possible for this DAG?" — this might be a good first test of convergence. It seems to me that the best way to do this is to enumerate the moves in the DAG, and compare them when the PMCs (passed through something that acts like uniq to simplify things: probably use the PMC keys as dictionary keys to count the number of times each PMC key occurs). But in that case, we’re not talking about a unique key that identifies a mover: more than one "key" is associated with each mover.

I am getting closer to another misunderstanding. I do NOT want to create a key for a complete PathMoverTree. When I talk about a key for a mover then it only means a key pointing to a PathMover instance within the DAG / PMTree (PMTree = 2D PathMover Tree representation). I think I still believe you would like to check if a specific part of your DAG / PMTree was call in a particular PMC. Like you do in the analysis.

```
OneWayShootingMover I'face 0 ran 7.69230769231% of the cycles with acceptance 60/77 (0.779220779221) 
OneWayShootingMover I'face 4 ran 7.39260739261% of the cycles with acceptance 47/74 (0.635135135135) 
...
```

Imagine that you use the same mover several times in the same DAG / PMTree, e.g.  `Sequential([OneWay, RepEx, OneWayShooter])` and you want to compare the acceptance before and after the RepEx. Then how do you do that? My idea was: Use my type of keys and to make this clearer what I mean lets call these locators instead of keys. So each part of the PathMover has a unique locator, the first OneWay has a different locator than the second one. With these locators the above problem could be easily solved like this:
```
pm = ... # this is the PathMover
# these are known of derived from the full PMTree
locator_OWS1 
locator_OWS2

# this would give you back the OneWayShooter
pm[locator_OWS1]
>>> OneWayShootingMover(...)
# which is the same when calling
pm[locator_OWS2]

for pmc in storage.pathmovechanges:
    # this could check, if the first OWShooter was called
    if locator_OW1 in pmc:
        pass
    # this would check, if the first OWShooter was accepted
    if pmc[locator_OW1].accepted:
        pass
```

This is what I always had in mind when I talked about keys. Not to have an identifier that represents the whole PMC. But I understand that this is what you need and I agree that this is useful.

So the testing for a PMC made from a PM works, but that still need the whole PMC and not only its order of PathMover calls. This is what you had in mind and agree. If we have a good key that is small we can use these to store a list of all ran PMC types. Maybe compare how many of all possible ones we have done so far. 

I also agree that in this case the key (and here I mean your PMC key) needs to contain more information than my simple locator. So this idea with wildcards is a very good one. And I think, we can actually have the best of all worlds.

My locators are some kind of partial keys, while the pmc keys represent a full possible result of a PathMover. By using wildcards we can achieve both. The main difference is that the full pmc key need to be complete enough that their order of Pathmovers called is unique. The locator must be complete enough that we can uniquely pick only a single pathmover not all of them.

One question though. When using wildcards it would be enough to only include key points. So for a mover that is just a sequence we could use `['*']`. Is that desirable? It means that since there is only a single possibility for a pmc we do not need a key at all. Problem is that this key works for other PathMovers as well. 

Another example
```
pm = Random([Sequence([Forward] * 5, Forward]) # 1 or 3 steps
```
the locators are clear. What would you chose to be a good pmc key for the two possibilities.
```
# with wildcards
key1 = [Random, [Forward], [*]]
key2 = [Random, [Forward]]
# w/o wildcards
key1 = [Random, [Forward], [Forward], [Forward]]
key2 = [Random, [Forward]]
```

First one I think is nice, but consider the pathmover
```
pm = Random([Sequence([Forward] * 5, Forward]) # 1 or 5 steps
```
we will have the same keys. So wildcards only make it shorter when used with a specific pathmover

For completeness the locators for this case
```
[Random]
[Random, [Forward]]
[Random, [Forward,Forward]]
[Random, [Forward,Forward,Forward]]
```
Interestingly is the locator for the Forward from the first random choice and the second the same. Which makes sense since they are indistinguishable.

> Yes, I want any keys returned in either case to be "in the same language." If you take the keys that you illustrated above, but took any [rc] and replaced it with [rc, [*]], I think that would be what I was talking about.

Okay, then I think I got what you wanted.

> Yes, this is what I'm trying to say. I don't understand where we would use keys that select single nodes. Why not just use the nodes themselves? The key only has value because it provides context for the node. If the node exists in more than one context, it must have more than one such "key" associated with it. This multiplicity of keys can be represented either by a unique key that uses wildcards, or by listing all such keys.

For clarity. I am not interesting in a specific position inside a PMC I only want a locator to be able to talk about a specific location in the PathMoverTree. With this I can later reference for problems I mentioned above. If you want to just check for the occurrance of a specific mover in a PMC, I would just make this a special case and test for it using, e.g.
```
if my_mover in my_pmc:
    pass
```
if you want to generally check for any OneWayShooter you use the class and not the instance
```
if OneWayShootingMover in my_pmc:
    pass
```
These could work in Pathmovertrees and PMC alike. Here I assume that every single PathMover is also a tree (maybe a small one)

For all on this we do not really need a DAG or Tree representation, so I would leave it the way it is constructed right now and add a function to return a DAG representation using a standard graph library so a user can do fancy checks and tests.

Let me try to come up with a proposal for all of this...

jhprinz:
All right, this works and I added a little notebook to show how. It is not super clean, but it combines what we both wanted. A unified way to use keys for single nodes and keys for pathmovechanges. It also allows to generate keys for all possible pathmovechanges in a rather fast way and optimizes for some special duplicates. It also uses trees of tuples instead of lists to make is hashable.

More updates in the next days

jhprinz:
Wrong order of tests. The path mover tutorial relies on mstis.nc...

jhprinz:
Hmm, I need to the visualization changes to this PR and make some little adjustment to the mstis_analysis notebook before this passes...

dwhswenson:
That may be an "insufficient data" error. I've seen it before. Try re-running the test on Travis.

jhprinz:
That and then my visualization did not correctly plot the minus move. If after a minus there is a shooting move, then the plot function of the shooting is looking for the initial shooting point in the last trajectory which is an intermediate sample and hence the plotting fails in this rare case. I have added plotting that (which then resulted in me being confused about the workings of the Minus Move)...

This should work now if there is sufficient data now.

jhprinz:
I disabled something that uses new storage to be faster and I would not like to rewrite it to be overwritten later. So I disabled the code for now.

jhprinz:
This vis error is present in master too so I manually cherry picked the changes and made a new PR that I test right now...

jhprinz:
Ready to be reviewed.

openpathsampling/treelogic.py
```diff
@@ -1,169 +1,247 @@
...
+        return self._deterministic
 
+    @property
+    def unique(self):
         """
-        if self._len is None:
-            self._len = len(list(iter(self)))
+        Return the smallest tree of tuples that uniquely represents this tree
+        """
+        ret = []
```
dwhswenson: Would a better name for this would be something like `minimal_representation`? The name `unique` isn't clear until you read the docstring.

openpathsampling/treelogic.py
```diff
@@ -1,169 +1,247 @@
...
-        yield self
+    @property
+    def is_sequential(self):
+        return len(self._choices) < 2
 
-    def __len__(self):
-        """
-        Returns the total number of Changes mad in a move.
+    @property
+    def deterministic(self):
```
dwhswenson: What is the "deterministic" property?

openpathsampling/pathmover.py
```diff
@@ -320,11 +333,8 @@ def move(self, globalstate):
 
         return paths.EmptyPathMoveChange()  # pragma: no cover
 
-    def __str__(self):
-        if self.name == self.__class__.__name__:
-            return self.__repr__()
-        else:
-            return self.name
+    def __repr__(self):
+        return self.name + '(%s)' % hex(id(self))[2:]
```
dwhswenson: At the very least, at a "0x" before the hex string. That makes it clear that this is a hex representation. Also, I would suggest adding a space before the parentheses.

This whole thing will need to be revisited when we get around to #233.

dwhswenson:
A few minor line notes above, but overall this looks good. I like the idea that you can easily check whether all possible paths were followed -- that was one of the things I thought was important.

For future PRs:

* This needs unit tests. Unit tests check for ***correctness***, not just a lack of exception-raising. Running your tutorial notebook is not unit testing. (I could write `plus` such that `plus(2,2)==5`. The code would run without exceptions, but it would be incorrect.)
* As we discussed on Skype, the number of move paths possible with sequential movers of random choice movers can get out of hand ($m^N$, where $m$ is the number of moves in the random choice, and $N$ is the number of choices in sequence). It might be good to have a way to raise a warning. I think that calculation of the number of paths can be done more cheaply. (It can: [result from a quick Googling](http://cs.stackexchange.com/questions/3078/algorithm-that-finds-the-number-of-simple-paths-from-s-to-t-in-g).) You might do that instead of `len(list(pm.enum))`, and maybe even find a way to run it before really starting `enum` (maybe separate `enum` into an initial part that gets called and checks this, and make a separate recursive function within?) That way, you can also set a `TreeMixin.MAX_ENUM` class variable that `enum` compares against: if the number is too large, `enum` can raise an error instead of actually enumerating. I think something large, but reasonable (good ol' 32767?) is a good default value for that: it prevents the user from enumerating 2 billion paths unless that is *really* intended, in which case the value of `TreeMixin.MAX_ENUM` can be changed.

openpathsampling/pathmover.py
jhprinz: You are right. I think I tried to have something that is short but kind of unique and does not depend on anything that storage uses.

openpathsampling/treelogic.py
jhprinz: Good point. I was looking for a short name since it could be that this property is used more frequently than others. I like to use `minimal` somehow. It is the smallest possible subtree that uniquely represents a tree. I need to think about that. 
The word unique is in relation to the generating (2D) tree. So with respect to all possible outcomes of moves (trees) what is the smallest possible subtree that can only be present in this tree and in no other one. Hence it uniquely identifies this particular PMC.

openpathsampling/treelogic.py
jhprinz: I think I should first add better documentation, but for now deterministic is `True` if the sequence of submoves is fixed in the sense that there is only one way that the movers below will be called. E.g. a `RandomChoiceMover` is not deterministic, but a `SequentialMover` is. This is used to check for the minimal representation since everything that is deterministic does not need to be part of the minimal subtree (besides maybe the first mover to check if the deterministic sequence is called or not).

jhprinz:
I added a function `n_enum` that will only compute the number (which can be big) and `enum` itself return a generator over all of these, but it stops after 1024 using `MAX_ENUM` as a limit.

dwhswenson:
Broad variable name change in here: the singular of "leaves" is "leaf." You regularly use the variable `leave`, which doesn't make sense here. (The code is fine, but this will make it easier to read.)

Of course, don't just replace-all -- you could elsewhere be using "leave" as in "leave behind" or "depart." And if you have "leaves," it should stay as is (I think "leafs" is only used for the Toronto hockey team).

jhprinz:
ups...

dwhswenson:
You also have `leave_n`, `leave_fnc`, etc. ;-) I'd say look for anything with `leave`, and unless it is intended as a verb, you probably want to change it to `leaf`. 

This is assigned to me, but you added a task to the main list, to add nosetests. I'm fine if that's done in a separate PR, but I won't merge if you're planning on adding the tests as part of this PR.

jhprinz:
I removed the wrong `leave`s.

jhprinz:
Is there a reason that everytime I update to the new mstis example the analysis of the rootmover is different? The number of nodes in the PathMover seems to change a lot. Before it was 69 with some movers that I could not explain, like 10 reversals for 9 interfaces, etc. Now this seems to be correct with 63.

dwhswenson:
> Is there a reason that everytime I update to the new mstis example the analysis of the rootmover is different?

I have no idea. If you had an old version where there were movers made by the `MoveScheme` as well as movers made by the `Network`, there might have been weirdness. But that should have been fixed as of #342. That should have been included by the last time you merged master into this branch.

Are you still seeing changes, or is it now steady at 63?

dwhswenson:
Also, which "analysis of the rootmover" are you referring to? Does it show up strangely in the `mstis_analysis.ipynb` notebook, or are there just more movers stored in the netCDF file than you expected?

This is one of the reasons to move everything over to a `MoveScheme` object: you now have access to the `scheme.find_hidden_ensembles`, `scheme.find_unused_ensembles`, and `scheme.find_used_ensembles` to check what ensembles might be defined but not used. Comparing this with the `mover.ensemble_signature` of any mover you're unsure of will tell you whether that mover is sampling an extraneous volume. You can also look at `scheme.choice_probability` or `sum([scheme.movers[group].values for group in scheme.movers], [])` to see which movers are counted as "official" movers and expected to be used.

jhprinz:
Sorry, my question was too unclear. The question was, if you could (easily) confirm that the mover that will show up in the `steps.change` in the mstis example has indeed 63 potential moves. So, if I ask the pathmover to give me the number of possible choices is it 63.

This number was thirty something the first time and I checked by hand that all is working. Then it changed to 69 and now it is 63 which I think is correct, besides that I do not understand why it has 10 reversal movers (is one for one big joined ensemble maybe). 

I made the pathmover tutorial to explain what this PR does and can do and I wrote down the manual way to check for all possible moves.

Main point is that I was thinking to use this as a way to check the Mixin. Otherwise I have to create new objects that allow choices (which is probably what I will do). And second to make sure that there is not something in the constructed tree that should not be there. The second is something that I can not judge easily.

dwhswenson:
Without checking the results, here's what I would expect based on the Default Scheme:

#### "Group-level" (or "official") movers:

* `shooting`: 9 (1 per normal ensemble); total pathmovers = 9 official * 3/official = 27
* `repex`: 9 ((2 within each interface set + 1 ms_outer per interface set) * 3 interface sets)
* `pathreversal`: 10 (1 per normal ensemble, 1 for the ms_outer ensemble)
* `ms_outer_shooting`: 1; total pathmovers = 1 official * 3/official = 3
* `minus`: 3 (1 per interface set); total pathmovers = 3 official * 11/official = 33

total official movers: 9+9+10+1+3 = 32
total movers at the group level: 27+9+10+3+33 = 82

#### Organization-level movers:
* path simulator: 1
* choose move type: 1
* choose within group: 5 (1 per group)

total: 7

Overall total: 82+7 = 89

***

Now, what I get from various things when I reload this scheme:

```
>>> # number of "official" movers
>>> print len(scheme.choice probability) 
32
>>> # all movers except the path simulator mover
>>> print len(scheme.root_mover)
88
>>> # "official" movers per group
>>> for g in scheme.movers:
>>>     print g, len(scheme.movers[g])
pathreversal 10
ms_outer_shooting 1
shooting 9
minus 3
repex 9
>>> # total movers per group
>>> for g in scheme.movers:
>>>     print g, sum([len(m) for m in scheme.movers[g]])
pathreversal 10
ms_outer_shooting 3
shooting 27
minus 33
repex 9
```

All of this matches my expectations, but I'm not clear where you're getting the number 63 from. Is this because there are, e.g., two possible "kinds" of OneWayShooting moves, even though there are 3 movers involved in it? You'd have to also figure out how many "kinds" of minus moves exist, but the same formulae above should otherwise apply.

jhprinz:
Thanks, I will check that with it. And yes, OneWayShooters are counted twice!

jhprinz:
Alright. This should be finished soon.

Since I have more or less finished the update on the visualization there are a few things in here that concern the Pathmover overhaul that @dwhswenson is doing and the implementation that works with 1D/2D (DAG) implementation here.

I fear that some of the stuff will conflict and we should sort this out to avoid messy conflicts. Let me make a little summary below.

jhprinz:
#### DAG and the tree representation

This PR implements a tree-like structure to be used by a `TreeSetMixin`. A tree set is a tree with several choices at certain nodes and hence it represents the same structure as a DAG. The reason why I prefer still a tree picture of the DAG is that we usually picture it or think of it in terms of a tree. The is best reflected in the way that we write nested code.

Since the DAG representation has a starting node and then traverses we need to agree that for the tree we will traverse it in a certain order which I will term `pre`. This is derived from binary trees, where the left child is evaluated first. So we say that we will always run the child nodes in order from _left_ to _right_ and then the head node which uses the results.

In a DAG we have also the possibility to use probabilities to run certain nodes, which we do not have directly in a tree. Or agreement will be that each node will we allowed to chose between an arbitrary subset of children and these will then we run (as before) from left to right. Important is that we do not allow to reverse the order of nodes. The order is pre-determined. Only which movers are run is decided _only_ by the head. This decision can be done before any child is run or after some children have been run. An example for the first would be a random choive move that picks one of many by random. While the seconds would be a Partial Sequential Mover. It would continue and decide after each child what is to do.

The nice property of this idea is. That all possible outcomes of traversal are subtrees of the full tree. This is nice for picturing the process and to visualize it.

#### Enumerating choices

For the generic tree set mixin we need ways to setup a node. The involves specifying (in brackets the example for a pathmover)
1. the node content implemented by the property (the actual PathMover instance `mover`)
2. the node children (the possibly used submovers `submovers`) requires implementation of `._subnodes` 
3. the possible subsets of children (named `.choices`) requires setting `_node_type`

The children are straight forward (and already working as it used to). The choices are done by setting the class variable `_node_type` to one of the following values

```
    NODE_TYPE_NONE = 0
        Means this node has no sub nodes and hence no possible choices for sets of trees
    NODE_TYPE_ALL = 1
        Means that each single child is a single possible choice like in a random choice
        picking among the children
    NODE_TYPE_ONE = 2
        Means that there is only one choice and that is to run all children in order
    NODE_TYPE_ACCUMULATE = 3
        Means there are N choices like in TYPE_ALL but this time the choices are accumulative
        like [child1]  or [child1, child2] or [child1, .., child3], ...
    NODE_TYPE_POWER = 4
        Means all possible combinations of children are possible which relates to the power set
    NODE_TYPE_CUSTOM = 5
        This allows you to make a custom pick and it should return a list of child combinations
        or a callable function that returns a list of child lists
``` 

Note that all possibilities (as mentioned before) are unordered sets of children. So `['child1', 'child2']` and `['child2', 'child1']` are identical. If you need that make two branches and put a randomchoice before.

> QUESTION: Do we need all these cases? Or does getting rid of some moves make some of these redundant (like PartialSequential, etc...)

I would propose that I implement a specific graphic representation for each of these standard cases. This would be used in the `vis` package to show the while pathmove tree and the possible choices at the same time.

#### The difference between `PathMover` (PM) and `PathMoveChange` (PMC)

`PathMover` is the full tree of all possibilities while the `PathMoveChange` is what the `PathMover` generates. So the `PathMoveChange` is a valid subtree of the `PathMover` that generated it and the `PathMoveChange` has no choices!

#### Keys and Locations in the tree

I found it useful to specify specific nodes inside a tree relative to the root node. In a deterministic tree (deterministic means there are no choices of subnodes) specifying a node can be done by omitting all non esential leaves and branches. Since the tree is deterministic we could even just say, give me the n-th node. More informative is it to include the path that leads to a node into that description. 

When using trees with choices this is more complicated and the question to specify a node as before in not clear. There are two ways to answer it:

1. How to find a specific node without past
2. How to find a specific node with past

In a deterministic setting the two answers are the same. In our concrete problem with PMCs and PMs we can ask 

1. Was this mover from our PM called in that PMC?
2. Was this mover from out PM with this specific past called in that PMC?

Note that PMCs are deterministic so we do not need to worry about choice there.

Luckily there is a way out of this problem since PM and their choices (PMCs) are subtrees. 

1. To answer the first question we can omit all non-essential branches as before (these might contain past). This is an acceptable subtree and all we need to check is if this subtree is a valid subtree of the PM or - also possible - if that subtree is also part of the PMC.
2. Do answer the second question we need to extend the subtree so much that all the potential pasts are unqiuely determined.

The tree mixin provides functions to determine subtrees that. I will use the following names for the different types:
1. `key`: the key is past-less and only points to a node in a tree. 
2. `history`: the subtree that specifies a node with past 

Both are of course subtrees without choices. 

To even simplify the description of a key or history we do not use the whole tree but a simplified tree with nodes that only contain a pointer to a mover. This tree class is called `TupleTree`

#### Checking for subtrees in a tree with choices

If you want to check of a tree is a subtree of another this is straight forward if both are deterministic.
If one or both have choices we need to also check all possibilities. This is a little more complicated but still pretty clear.

jhprinz:
My questions would be if you agree with the naming convention (i think that can be improved) and I still need to finish
```
- [ ] unittests
- [ ] implement different functions for .key and .history
- [ ] name for subnodes : subnode or child/children
- [ ] check docstrings
```

and we are ready to go.

dwhswenson:
I will have a longer response -- I really don't have time right now to write anything detailed, and probably won't have time for a while. 

One important point: I think you're still misusing the word subtree.

To anyone who has worked with binary trees, a subtree is the node and ALL its descendent nodes in the tree. I think you're using "subtree" to describe a decision path in a decision tree. "Subtree" is not the right word, and since it has a specific meaning in these contexts, using it leads to confusion.

Unless, of course, you actually are saying that the PMC is the entire subtree, which I would disagree with on the principle that it makes no sense to the user. The PMC gives the history of what movers were visited, and in what order. It is the move decision path (although I like the word "branch" in this case, as "path" is overloaded.)

Again, we very quickly get into the problem that, whether you want to think of it as a tree or not, the mathematical structure is a DAG. Creation can be on "levels" that allow you to think in this tree-like way, but any analysis of the structure in terms of trees will be inaccurate.

It looks to me like what we're doing here is going to be a temporary approach. It will be limited because it isn't designed to work with the correct mathematical structure, and eventually that's going to cause problems. As such, we can merge something along these lines (again, I'll need some time before I can really review this), but we need to keep in mind that this will change at some point.

jhprinz:
Alright. I would go with temporary since for now it provides us with the features we need. If we find something better I am not against it.

In general I think the tree and the DAG are equivalent, it is just a visual rearrangement of nodes and I think I like the treelike one better. So far I assume your "isnt't designed to work with the correct mathematical structure" is wrong in the sense that when both are equal both are right or wrong of course and it is some kind of philosophical decision. I like trees better for now... 
The only caveat is that I require that a tree is traversed and not just ends at leaves. If you want a tree strictly in that sense that then yes, you are right. But then we should stop talking about Trees and also not construct them like one. So then maybe what I want is really a DAG written in the shape of a tree which all the tree features.

About the subtree I just looked through Wikipedia and found that there are 2 definitions: The one you mean from data structures and a second one (that I mean) from graph theory.
>  This is different from the formal definition of subtree used in graph theory, which is a subgraph that forms a tree – it need not include all descendants. For example, the root node by itself is a subtree in the graph theory sense, but not in the data structure sense (unless there are no descendants)

I think of it as the graph theoretical approach obviously, which would then coincide with using a DAG and subgraph that is again a DAG.

Anyway, I would like to wrap this up, but as long as we cannot agree I should either just delete this PR and wait for something better (if we don't actually need it) or we try to do it mostly right so that we can later build upon it.

jhprinz:
Close it for now. Not used and is outdated...

PR #296, opened 2015-07-02
Move scheme cleanup

dwhswenson:
Cleaning up a few things in `MoveStrategy` and `MoveScheme`. First, cleaning up poorly considered redundancies about networks as input. Second, adding a few general functions which might be useful.

- [x] Remove `.network` from `MoveStrategy` (only belongs in `MoveScheme`)
- [x] Add `MoveScheme.sanity_check()`: check sampling ensembles not unused; check choice_probability adds up; check for duplicated movers
- [x] Function to check how many total MC steps expected to get N trials of a given mover

That last one will be particularly useful: we define an MC step as *any* single move. But maybe you'd rather count in terms of shooting moves per ensemble -- this is a better measure of computational cost, because the number of MC moves required for this to be the same is very different in all-replica exchange and in nearest-neighbor exchange. (The best measure is number of MD steps taken, but we can't predict that until we've already sampled the ensembles.)

So if you're setting up your calculation interactively in IPython, using this function will help you figure out how many MC steps to do.

Includes #289, so merge after that.


dwhswenson:
Passed tests; ready for review and merge.

jhprinz:
Great!

PR #298, opened 2015-07-04
LIve visualization

dwhswenson:
Most of the work we do on OPS is designed to make it easier to do real science.

This, however, is just eye candy.

Once it passes tests (which it should), this is ready for review and merge. Try running the `multistate_system_setup.ipynb` notebook to see it in action.


jhprinz:
I think this is pretty cool...

PR #300, opened 2015-08-19
Moving General Storage Logic into separate class

jhprinz:
I thought it would be useful to separate the JSON Pickling and object storage logic from the actual MD/OPS Application. This makes the code more readable and the base file format can be used for other projects. This is what I came up with so far:

#### netCDF Subclass `NetCDFPlus`

I made a wrapper class for the usual netCDF which enhances it with all sorts of nice features, but leaves the basic netCDF API intact. So you can safely switch to nCDF+ without problems.

1. native support for `simtk.unit`
        You can simply create a variable with units and it is automatically stripped off, stored and reattached if loaded. So writing `storage.vars['my_length'][1] = Quantity(1.0, unit.nanometer)` works. The values are converted to the variable units and stored. This way you always know what the numbers in your file mean.
2. Each variable (only numpy and floats) has a unit property
3. Access the enhanced variables using `storage.vars[name]` and the netCDF implementation using `storage.variables[name]`
4. Support for storage of python objects as we do right now. This requires the objects to be subclassed from `OPSObject` (which we should rename). Then you can create a special object variable using
    `storage.add('my_object_var', ObjectStore(options))` and store and access these objects via `storage.my_object_var[index]` or `storage.my_object.save(obj)`. Note that these requires immutable objects to not mess up the complete storage. The ObjectStore can either be specialized to store certain objects effectively or there is a standard version that uses JSON Pickling.
5. More Powerful variable types:
    1. native python types `bool`, `float`, `int`, `str`
        Besides the native implementation of bumpy arrays I fould it useful to move the type casting into the variable itself. So if you create a variable of type `bool` then assigning `storage.vars['my_bool'][10] = True` works and `storage.vars['my_bool'][10]` returns a bool. Similar with other types
    2. There is a variable type that references objects that can be stored. So just create a variable of type `obj.my_object_var` and store or save like any other variable `storage.vars['obj_var'][idx] = obj`. This will  
store the object in the netCDF file using the `ObjectStore` and store in `obj_var` the index to find the object. The actual netcdf type of these indices are `numpy.int32`

All of this is totally generic and independent of our usecase. We just add special Stores for some objects like snapshots, trajectories, etc. and pre-initiate the storage with store for all of our storable objects.

We might at some point move the extended storage into a separate project.


jhprinz:
This also contains the LRUCache already and thus faster loading using the 
`AnalysisStorage(ncfile)`.

All tests passed, which is a good sign :)

dwhswenson:
As JHP and I have discussed before, I like the idea of making parts of OPS more reusable. I've started playing with a side project (a library for certain kinds of semiclassical calculations, inspired by approach we've taken with OPS). So far, it requires OPS, because I reuse our storage (which JHP is separating here) and our engines. But most of OPS is irrelevant to that project.

Of course, all of this is stuff that isn't urgent before 1.0. However, one suggestion: can you generalize the unit type? `simtk.unit` is only used by OpenMM, and isn't (easily) available independent of OpenMM. I seem to remember seeing a conversation where @jchodera looked into splitting the units package off, but ended up using a different units library.

I think creating some simple wrapper that allows various units packages (just as we allow various engines) would be much more valuable than hard-coding to simtk. Would that be hard to do?

jhprinz:
> I think creating some simple wrapper that allows various units packages (just as we allow various engines) would be much more valuable than hard-coding to simtk. Would that be hard to do?

Good point. I was thinking about this issue with the unit support. The approach I have taken is rather simple right now:
Each new variable type has a setter and a getter function to convert it to a netCDF type. E.g. the new python `float` type will do nothing for the setter since netcdf will convert it to numpy.float32 automatically and the getter will just call `.tolist()` which converts it back to python float (or a list of float in some cases).
The unit support is on top of that, but thinking about your concern we could couple this like having types
`float`, `float[simtk.unit.Quantity]` and `numpy.float32`, `numpy.float32[simtk.unit.Quantity]` or some other notation, but then we can explicity specify what we want. Imports can be attached to the getter and setter creation (is only called once) and so we can run it without even simtk.unit installed.

I remember that @bas-rustenburg used a package called `pint` and seemed to like it.

I will add this. The new approach also means that collective variables can have almost arbitrary type now.

bas-rustenburg:
@jhprinz, just to comment on `pint`. If you want to support it, be careful if you're expecting user input in `pint` units. It has [no global units](http://pint.readthedocs.org/en/0.6/tutorial.html#using-pint-in-your-projects).

Perhaps more importantly,  output in pint units is not particularly useful if the user has their own unit registry besides yours.

`Pint` can [parse strings](http://pint.readthedocs.org/en/0.6/tutorial.html#string-parsing), so for some users, storing the unit as a string is more helpful.

jhprinz:
I have made more improvements on a branch named `noidx`. Here is a picture of the memory changes with `lowmemory` setting. I ran 200 x alanine trajectories with 20 frames each. The red line shows the change in disc space after saving the new trajectory (almost constantly decreasing) and the purple line is the memory. A little messy, but overall constantly around zero (I started after all one time objects had been set up). 

mem

This shows that my memory handling seems to work. At least in this first test. This new trick is to explicitly make all snapshots virtual and only have a limited LRU cache. So if you set the cache to be small then OPS itself (by using trajectories) or other internal pointer will not exceed that number at least for snapshots in memory. 
For this to work I had to automatically turn snapshots in trajectories into virtual ones once the trajectory is saved. So as long as the cache is sufficient all snapshots stay in memory. If the cache is full is removes the last used frames. 
Also hard links to other objects are replaced by weak ones once an object is saved to allow the garbage collector to remove not referenced objects from memory. An example is `Sample`. Each `Sample` has a parent the references the originating `Sample`. So if you would load the last sample the would trigger a cascade of loading all parents. This is not effective since you might not need all of them or even none. So what you else do is load a reference to a `Sample` on disc. If you use `my_sample.parent` it will automatically be loaded and returned. If you save an object, a hard link is turned into a weak one to allow the same thing also with newly created objects.
All of these weak references are completely hidden to the user, he will never see these objects. They will always be replaced once accessed. This is the beauty of it. 

Overall the code has gotten cleaner and I am still adding docs and tests, but it finally works and should be ready for review and merge in the next days. Woohoo...

dwhswenson:
Excellent! I've been looking forward to the new caching system.

A few questions about that graph:

* What are the units for the memory? kB? MB? Just want a sense of what is required.
* What procedure did you use to run these? As it, are these actual shooting moves, or just trajectories run sequentially? (Wondering whether there would be artifacts with shooting that don't show up in independent trajectories -- but if I understand the disk plot correctly, this should be shooting, right?)
* Any tests yet (or could you do one) to compare the timing performance (with the toy model) using this? I'm hoping this will improve performance there, too.

But if I understand correctly, the caching is pretty much transparent to the user -- as it was before. It happens, but the user doesn't need to understand how (although perhaps there will be tricks that the user can use to speed things up, like the fast loading we did before for analysis?) To me, that's the ideal: keep memory footprint small; don't make the user think about it. Perfect!

Right now, however, I'm very confused about the many different PRs you have open. I think there's a lot of overlap in purpose, and a lot of mixing between branches. Before we merge anything, can you close the PRs that aren't relevant (for example, it sounds to me like now *this* PR has the LRU stuff in it?) It'll make code review a lot easier.

jhprinz:
Good questions. And some answers first:

- I think this is kB and it is measured using `psutil` and of course includes all sorts of other system related stuff. The storage is almost optimal, it increases about 768kB which is 1600 Atoms * 3 * 20 frames * 4 bytes * 2 (velocity). I just wanted to make sure that the storage removes all internal hardlinks to objects that are stored so that when the user has no reference to an oblect like a snapshot then the garbage collector is free to remove the object from memory. As long as there still exists a user made reference it will also remain in the cache and hence be accessible fast. 
- No shooting moves, just running the engine with a `LengthEnsemble(20)` stopper to have trajectories of the same length and better comparison. Next I will test, if the same works for real shooting moves. I will make a notebook to illustrate and test this and make some better pictures.
- No tests yet, but I will make a whole suit of test for the new netcdfplus.py and the OPS storage specific ones. Not sure though how to test the memory handling. I thought it might be enough to watch the weakref cache which is a good indicator if after a simulator loop, etc the number of weakly linked objects increases with time (which it should not!)

> But if I understand correctly, the caching is pretty much transparent to the user -- as it was before. It happens, but the user doesn't need to understand how (although perhaps there will be tricks that the user can use to speed things up, like the fast loading we did before for analysis?) To me, that's the ideal: keep memory footprint small; don't make the user think about it. Perfect!

Yes, the user does not need to know about caching. Although it helps. There is now a special loading function `AnalysisStorage(filename)` which opens an existing `.nc` file in readmode and automatically does lots of smart caching like loading everything except coordinates and momenta. 

There are still a few tricks that allow to do some stuff faster, but this requires some understanding as before. Like the trick that CVs work very fast with trajectories since we can skip loading the snapshots and only use the snapshot_idx.

Most importantly the user does not need to know about it to make it work. He can make specific things faster by knowing some tricks and of course he can still cause a memory overflow by just storing stuff in local variables

> Right now, however, I'm very confused about the many different PRs you have open. I think there's a lot of overlap in purpose, and a lot of mixing between branches. Before we merge anything, can you close the PRs that aren't relevant (for example, it sounds to me like now this PR has the LRU stuff in it?) It'll make code review a lot easier.

Yes, you are not alone. I am too loosing also my last overview. I will remove unnecessary intermediate steps and make this clearer. Overall I hope that the new code (maybe not the changes) in storage will be much cleaner. Also if someone wants to make changes in the future.

I also started to remove lots of legacy code and functions that have lost their usefulness or are redundant.

Tests already pass and more updates very soon.

dwhswenson:
> No shooting moves, just running the engine with a LengthEnsemble(20) stopper to have trajectories of the same length and better comparison.

Huh... Then I'm not sure what's going on in your disk usage plot above. That's the change in disk space with each run, right? So most of the time, you should be adding some number of frames, leading to a mostly flat line. That part, we see. There might be a few bumps where extra stuff is stored to disk for whatever reason, leading to downward bump. We see a few of those.

But what's happening at the upward bumps? There look to be a few times when nothing is saved to disk, and it isn't followed by a larger than normal save the next time. There's probably a good explanation for it, since it looks like it happens on a pretty regular ~20 trajectory period. Can you figure out exactly what's going on there?

(Now that I think about it, shooting moves also wouldn't really explain that plot.)

> No tests yet, but I will make a whole suit of test for the new netcdfplus.py and the OPS storage specific ones. Not sure though how to test the memory handling.

General suggestion for the memory stress parts of these tests: create a new engine for testing purposes that creates snapshots by assigning random values to all the coordinates/momenta (or even fixed values would work, as long as it is new configuration/momentum objects!) That way you don't have to run the $O(N^2)$ force calculations. It'll make those tests a lot faster on Travis, where we don't have GPUs (and the same for my laptop!)

jhprinz:
The RandomEngine sounds like a good idea.

With the storage bumps the problem is that netcdf does not always write to disc when I expect it to. So the memory bumps are either something else running on my machine or the cache from the netcdf kicks in and store some stuff to disc. This has also to do with the chunk sizes in some way but I do not know what exactly is happening there.

jhprinz:
So, tests pass... Now all I need to finish is testing and documentation.

I wrote a OpenMMRandomEngine that does what you suggest and I tweaked everything that you can observe the storage and the cache and it will keep the caching exactly within its settings. Sofar no side effects although the real cool caching only works with snapshots. These will be removed from memory and so you can run the framework and keep a low memory profile.

I also started removing redundant branches and that is hopefully better now... It will take a few more days to create a least some tests and update the documentation but then we have storage set for 1.0...

jhprinz:
I added a notebook that tests the caching and memory performance of a read PathSimulator (although a simple one) using the RandomEngine for Snapshots. To do that I added secret feature to `paths.base.StorableObject` from which all OPS objects subclass. This will keep a weak reference to all objects that are created if activated so we can really keep track of how many objects are still in memory independent of whether they have been stored or not. Result is that the number of objects of now constant although it depends on the cache settings (of course). We can at some point use this to do more extensive testing of very complex moves if there is an object leak that produces objects that stay in memory and do not get recycled by the garbage collector. Here some figures for the case of a RandomMover of 4 LengthEnsembles (5,10,15,20) using OneWayShooting and an engine that creates RandomEntries, so far only Length Makes sense. We could change this to little changes to allow for other Ensembles to make sense, but the main point is not the ensemble... I use 500 iterations. Objects that do not change over all iterations are hidden. Cache settings use a special caching that keeps only 10 objects per type and weak references.

graph2
 
This first one shows the number that the cache of various types lists it has a reference to. This includes real cached and weak references. The cache has a fixed size and the weakcache does not. 

graph1

The second lists the number of all objects by type that are still in memory. So we have a better idea of the OPS objects what is still in memory and what not.

I hope that this passes now. I have also added a notebook that tests the basic features of the netcdfplus file and need to be tested in strict mode. It saves and loads various types.

dwhswenson:
The constant usage sure looks good, but I'm not sure that I understand exactly what these figures show (partly, colors get re-used, so it's a little hard to follow). Just eyeballing it, it looks like a couple things are about the same in both graphs (total, and maybe PathMoveChange?) Is that expected? If total is the same in both, why isn't everything the same?

Also, have you checked speed performance? I'm hoping that the new structure will also lead to spending less time in the cache lookup during simulation. The ready-to-profile setup from https://gist.github.com/dwhswenson/27de391da5450fc249b1 might be an easy thing to test. (I believe I used gprof2dot to generate the call graph... see also https://github.com/choderalab/openpathsampling/issues/222#issuecomment-97118374 for details on that gist.)

jhprinz:
total is in both plots the same, just to see the overall number of objects. The difference is if there are objects in memory that have never been saved. The will be seen in the second plot but not in the first. 

You are right. The test is not too thorough yet, but it hints that most caching is doing the way it should do right now. It never prevents the user (or us) from keeping references and thus building towards an memory overflow. But now we can watch the objects grow and figure out what might be going wrong. I will run your gist as a test and see what results I can come up with.

jhprinz:
I realized that I need a package called dwhs_profiling to do all tests but overall I realized that it seems a factor of 2 faster than the results in your package. So I assume that me making the code much nicer has not affected performance at least in a negative way. Not sure what the results are. At least the Analysis is much faster now (if you have not used the caching before)

jhprinz:
Let's see, if tests pass now.

jhprinz:
One more nice thing. The ipynb test is now about 1 sec per test faster. Which should be several minutes in total!

dwhswenson:
> I realized that I need a package called dwhs_profiling to do all tests but overall I realized that it seems a factor of 2 faster than the results in your package.

[dwhs_profiling](https://github.com/dwhswenson/dwhs_profiling) -- but the part I was particularly interested in was to see if the call graph generated by gprof2dot (probably on minimal.py from that gist) showed a difference in the relative role of the storage subsystem.

2-fold speedup is good, but that isn't just a comparison to the results in seconds as recorded in the gist, is it? Those timings were either from the mid-2010 iMac at my desk in Amsterdam or from my late-2011 MacBook Pro. Comparing timing in seconds to your newer computers (with SSDs and more RAM) won't mean much.

My main hope is that searching the cache takes a smaller fraction of the total time during production, which would be visible in the [gprof2dot](https://github.com/jrfonseca/gprof2dot) call graph.

> The ipynb test is now about 1 sec per single test line faster.

!!! Is that just due to the little changes in `ipynbtest.py`? That's nice!

dwhswenson:
When merged, this should resolve #258, right?

jhprinz:
Yes, it should make the forget caching obsolete in most cases, but I added a function that each cache can be cleared if so using
```store.cache.clear()```
which will release all cached objects.

**Comment by jhprinz** [comment]
updated the notebooks. Should be ready once we talked it over.

File: openpathsampling/chaindict.py [hunk 1]
```diff
@@ -60,7 +64,7 @@ def __getitem__(self, items):
         return results
 
     def _add_new(self, items, values):
-        self[items] = values
+        pass
```
dwhswenson: ?? purpose of `._add_new` if all it does is `pass`? Is this an abstract that should `raise NotImplementedError`?

openpathsampling/ensemble.py
```diff
@@ -48,12 +47,18 @@ def __init__(self, direction=None):
...
     def bad_direction_error(self):
         raise RuntimeError("EnsembleCache.direction = " +
                            str(self.direction) + " invalid.") #nocover
 
+    def clear(self):
+        self.start_frame = None
+        self.prev_last_frame = None
+        self.last_length = None
+        self.contents = {}
+
```
dwhswenson: I'm not sure if this is ever/should ever be used. Note that the resetting mostly happens in the `check` method, with the option of forcing `reset=True`. I'm concerned that this method would lead someone to use it instead of `check(reset=True)` when the latter is the correct behavior.

openpathsampling/snapshot.py
```diff
@@ -29,8 +30,8 @@ class Configuration(object):
     engine = None
     load_lazy = True
 
-    def __init__(self, coordinates=None, box_vectors=None,
-                 potential_energy=None, topology=None):
+    def __init__(self, coordinates = None, box_vectors = None,
+                 potential_energy = None, topology = None):
```
dwhswenson: Why restyle this incorrectly? PEP8 has no spaces for default arguments.

dwhswenson:
I literally *can't* review this. The net diff for this PR is so large that GitHub won't display it, probably because it [exceeds 1MB in diff data](https://help.github.com/articles/what-are-the-limits-for-viewing-content-and-diffs-in-my-repository/). (In the future, let's try to keep PRs smaller!!!)

Two things you could do to make more lines visible:

1. Empty the attic. Your "rename-all" stuff changes things in the attic. Also, you can just remove files instead of storing then in the attic. Attic directories are a vestige of circa-1999 CVS, and were made obsolete which everyone switched to svn, before everyone switched to git.

2. Clear output on the ipynbs. Normally we like to keep that output, but for this huge PR, clearing it might let me actually see the changes in another file or two. (A few single-cell changes show up a multiple differences).

(If those don't do it, I'll have to pull your branch down locally, view diffs here, and I don't think I'll be able to attach comments to specific lines on GitHub. But at least I'll be able to see the changes in the last dozen files.)


The parts GitHub *does* show me look mostly okay. You did some weird reformatting of `snapshot.py` that breaks PEP8, so please undo that, and I don't think the `EnsembleCache.clear` is needed (and if it isn't needed, it isn't a good idea; it can only add confusion). 

Could you write up a small summary from a user perspective of how this works now? I see `AnalysisStorage`, but there's no discussion of it. I assume that obsoletes the old tricks to speed up loading?

openpathsampling/snapshot.py
jhprinz: I think I changed the default settings in my IDE at some point but changed it back to PEP and forgot to check everywhere :)

jhprinz:
You are right. Trying to reduce the unnecessary overhead. Maybe we should agree to remove all output in the notebooks to keep comparison simple. I am always annoyed by trying to fix merge problems due to the changed output. 

I found some approaches to strip away the output when pushing/committing on git, but for testing we need the output (at least for my netcdfplus test notebook). Also for producing the docs.

If we agree that notebook should only be executable we could do that and force sphinx to rerun the notebook. Problem is that everytime the notebooks will change...

One person in stackoverflow mentioned the possibility to split the outputs to a second file, so that it would be easier to compare these or even treat them as binary. So we could keep the content but have no problems...

Any ideas? For this PR I will remove all output in most notebooks and we have to redo them in the next PR.

jhprinz:
Hmmm, this is really annoying I agree. Deleting files makes things much worse, since all the deleted files are listed with all deleted lines and there is no way to just look at not deleted files.

Cherry-picking seemed to be too complicated so I made two new branches on my fork storagemain and storagenb which are the storage main code changes and the other only the notebooks (which are basically not changed besides reruns and deleted outputs -- this also makes things worse...)

Anyway, you can now see the whole diff in storagemain.

jhprinz:
Do I need to add another PR for you to see the diffs? I will try to keep storagemain uptodate with number types to reflect changes I did.

jhprinz:
fixed lots of PEP8 problems almost exclusively in storage and affected files.

jhprinz:
Still need to write the little summary from the user perspective.

dwhswenson:
> Do I need to add another PR for you to see the diffs?

Through GitHub; yes I think so. 

It looks like it might be easiest for me to just check out this branch and check the diffs locally. Have you otherwise frozen the code? If you change a file after I review it, I won't know.

This review will take me a few days, and probably won't start until this weekend; a bunch of other stuff going on right now.

jhprinz:
I just fixed the little comments you had about the PEP8 and cleaned up more code that way. Hope it is better now. Just updated also storage main

jhprinz:
I am sorry that this is such a hassle. Without the ipynb and the deleted files it is readable...

dwhswenson:
I assume changes in most files are probably just renames of the base class (which, of course, affects every file), so it might not be too painful. But I have a lot of job application deadlines tomorrow, so I won't start on this until Friday.

Can you plan to freeze any changes here before Friday? That way, I can start the big manual review then, and I can see any remaining fixes by looking at commit-by-commit diffs.

jhprinz:
Yes, I will do that. GOOD LUCK! Don't worry about this. Applications first!!!!

jhprinz:
my notebook tests all passed in less than 8 minutes... Yeah

jhprinz:
marked this as ref docs becuase it potentially contains stuff that goes into docs for storage.

Here is a small summary as how a user would use the new storage:

### Production

For production you would create a storage in the usual way using
```
from openpathsampling.storage import Storage
storage = Storage('filename.nc', mode='w')
```
by default we have a "should work for normal sized systems" setting that also allows some analysis speedup using caching. If you really want to run production and keep memory usage small you could change the caching behavior using:
```
storage.set_caching_mode('production')
```
and it will keep caching to a minimum which also means that analysis will be VERY slow.

### Treatment of collective variables

Collective variables can be associated with a cache on the storage. Since we do not want to write every single value separately you need to call a sync function. Note that you can only store cvs of snapshots 
that are actually stored to disk. There is an internal list of values that can be stored to keep this fast.
```
cv = CollectiveVariable(...)
storage.save(cv)
cv.sync() # sync mem_cache with storage_cache
```
mem_cache is the cache for all snapshots (independent of stored or not) and storage_cache is a cache that allows faster access to disk.

The saving of the cv will create a variable in the storage for the values automatically if the parameter `store_cache=True` is part of the CV creation. This is the default and for CV you do not want to store you can set this to false and values are cached in memory but never stored to disk. Useful for very complex functions that have output you cannot save.

The normal way to create a CV is by using a name and a template (snapshot)
```
cv = CV_Function('cv_name', my_func, template)
```
the cv then runs 
```
test_single = my_func(template)
test_list = my_func([template])
test_multi = my_func([template, template])
```
and tries to figure out 
1. the dimension (vectors, scalar, etc...), 
2. the potential simtk units, 
3. the number type (int, float, bumpy, etc...)
4. and if it is a storable type, dimension combination
5. if the cv accepts lists of snapshots (like mdtraj) requires

you always override these values but it should not be required. This means that now very complex functions are supported. Since the check is done only once in the beginning it is still fast.

Last thing that is still complicated are functions that are user-defined since we cannot access variables from the global context an example
```
center = 1
def dist(snapshot):
    return (snapshot.coordinates - center)

cv = CV_Function('distance', dist, template)
storage.save(cv) # does not work because we do not know center
```
[to be continued...]

### Sync NetCDF

At the end of writing you should make sure that all data has been synced with the netCDF file (the internal C code has also some caching that depends on chunk sizes.) and if you try to read there might
so values missing otherwise.

### Computing Analysis

When doing Analysis it makes sense to use caching optimized for analysis. This means no snapshots (if possible), but everything else. Do do that and fill caches automatically just open a storage using
```
from openpathsampling.storage import AnalysisStorage
storage = AnalysisStorage('filename.nc')
```
and storage will be opened read-only and most stuff you might need is cached as much as possible.
Thus opening a file is slower but analysis is afterwards orders of magnitude faster.

### Caching

If you are still in your notebook of the production and want to switch to analysis you can do so by opening a new analysis storage but it will keep the old stuff around and potentially the memory full. The second possibility is to change caching behavior
```
storage.set_caching_mode('analysis')
```
and then cache what you need. `AnalysisStorage` does the following:
```
self.samples.cache_all()
self.samplesets.cache_all()
self.cvs.cache_all()
self.volumes.cache_all()
self.ensembles.cache_all()
self.pathmovers.cache_all()
self.pathmovechanges.cache_all()
self.steps.cache_all()
```
which can be executed using
```
AnalysisStorage.cache_for_analysis(storage)
```
and you are set for analysis. It will automatically shrink caches appropriately and keep the last entries.

jhprinz:
I changed the handling of snapshots from trajectories in the last commit. So now, if you get a snapshot from a trajectory you will get a real Snapshot object and not a lazy one. This way the user will never see lazy objects at all!

dwhswenson:
Is this PR locked enough for review now? Sent off 4 applications yesterday, and have a little time with relative quiet, so I can review if it is ready.

dwhswenson:
Also, the now-even-faster tests: is this because the production mode is so much faster as well?

jhprinz:
Hmmpf, I was about to write that it works and you can review it, but when I removed a little bottleneck that should speed it up I realized that there are some problems with the internal CV handling down to problems that the `.index` function used in EnsembleCaching is not compatible with my caching and use of lazy snapshots. Turns out it is not a big deal but I need another day or two.

Sorry, but this seemed serious for a little time and took me quite while to understand what actually is going wrong...

jhprinz:
Good news is, that it is actually faster than I thought :) once I did it right ...

dwhswenson:
If you end up modifying anything in the `EnsembleCaching`, please double check that it hasn't affected the speed scalings with more ensembles in sequence or with longer trajectories. The tests for that are in  [my gist on sequential ensemble timings](https://gist.github.com/dwhswenson/27de391da5450fc249b1), although you'll have to update some of the source code line numbers for some of the profiling checks. The [`dwhs_profiling` package is on GitHub](https://github.com/dwhswenson/dwhs_profiling).

jhprinz:
The problem is easily explained and I have a workaround that does not require changing your code, which I think is best. This ensemble caching was quite complex and fiddling around with it is not a safe option...

When a trajectory is stored then all snapshots in it are saved and then inside the trajectory the snapshots are replaced by references to the caching. This way I free the memory and only caching takes care which  snapshots are loaded fast. This works very nicely and assures that during production memory does not overflow. Problem is that EnsembleCaching keeps a reference to previous frames and if they are replace only inside the trajectory, then the test `trajectory.index(cached_snapshot)` fails since the cached_snapshot is only virtually in the trajectory. This happened never before since I did not use these proxy objects and it can only happen, if you test an unsaved trajectory for an ensemble, then save the trajectory and then test the same ensemble again, before the ensemblecache is reset.

I fixed it this way this way: Add a weak link to the real snapshot in the proxy and override the `__eq__` function. This will also speeds up getting objects from the cache.

jhprinz:
Consequence of all of this was, that I simplified the code some more and suddenly the tests ran about 10% faster... I also could remove the `__init__` in the old OPSObject so that we do not have to call an additional init every time an object is created.

jhprinz:
Seemed to be a necessary detour. Let's see if I fixed everything. I also updated the visualization to include Extend, Subtraj, RepEx, etc... So you can really see how your trajectories evolve.

There was one little thing that I realized in the current MinusMove Implementation:
The Minus (without the RepEx) does an Extend and then pick a SubTraj which will create decorrelated samples in the first interface. Now we pick randomly to extend forward or backward and then pick randomly first or final subtraj. Thus it can happen that we run an expensive extend and then throw this away and keep the old initial trajectory with obiviously no decorrelation. And this has nothing to do with the RepEx.
Is this intended or just an issue of me maybe misimplementing it. It does observe detailed balance still, but does not seem optimal.

This should part of the evolution of a trajectory in Interface0 of stateA. In the middle you see that it works nicely: extends and picks at the other side. On the bottem same thing but it extends backwards and picks on the "wrong" side and thus we loose the decorrelation...

dwhswenson:
That is a ***severe*** problem with the minus move. When was this changed? Is that part of this PR? It should not be merged in, if it hasn't already (my local version looks like it *should* be doing a conditional sequential as described below).

The minus mover works like this. Please think of it in the context of MISTIS:

1. Randomly select either the first or last "excursion" in the minus trajectory.
2. Randomly select an innermost interface from one of the interface sets associated with this minus.
3. Attempt a replica exchange of the two trajectories above. If that exchange is rejected, the total attempt ends and the move is rejected; otherwise continue.
4. Extend the trajectory that has been moved into the minus ensemble so that it satisfies the minus ensemble. This should succeed with probability $1.0 - P_{transition}$, where $P_{transition}$ is the probably of a transition spontaneously occurring (basically 0 if the event is truly rare).

If you have changed it so that it does not do exactly that, then I'm pretty sure you're either wasting computation (which, in MISTIS, can be a huge portion of your total time) or you're not satisfying detailed balance (which is a little tricker in MISTIS). My implementation was carefully done to include both normal minus and multiple interface set minus. I explicitly had notes in the code pointing to the reference for this (my 2014 JCP with Peter).

Implementing this in OPS required auxiliary ensembles: IIRC, a helper ensemble for the minus move that is *almost* the union of the innermost interfaces, but only allows A-to-A paths. We use that to select part of the path, and this is what the replica exchange is with. I'm pretty sure this has to be this way. When I make something complicated, it's because I couldn't find a simpler solution. Be careful around the minus mover, because if you're not thinking about the multiple interface set minus move, things will break.

dwhswenson:
Context on the severity of that problem: a correctly implemented minus move was about 25% of computation time in my previous code. MISTIS expects a 1/N acceptance on the minus move, where you have N interface sets for a state. In the correct implementation, only the accepted moves have any contribution to the total cost. So let's say your calculation took 100 hours. 25 hours were spent doing the minus move *correctly*. If you have 3 interface sets, then you would do 3 times as many extensions, for a total of 75 hours instead of 25.

In other words, in a realistic situation, this would increase computation time from 100 hours to 150 hours. That is not acceptable.

jhprinz:
I thought that it was a problem, but not that severe in terms of waste of ressources. I am not sure where this happened, but my assumption is that when streamlining the minus code some time ago that happened.

To make it clear. The ONLY difference to what you want is, that the final choice of sub trajectory after the extension is done a the random end and not the one that was extended to. The fix is very easy and I will add it to this PR. My question was mostly to know that this is an error and not be not getting reasons for detailed balance. Thanks fot the clarification!

jhprinz:
Hmmm, sorry for causing all the trouble. Everything was correct, that is why all the test still passed... I just misinterpreted the above picture. Let me explain: If you look closely at the picture then you see that the order of Subtrajectory and Extend is reversed. That confused me at firstm but it just means. That our Subtraj in the Minus move need a sample from the minus ensemble, which in the case of the picture had been generated by an extension move before (this is why the order is reversed!) And in that that case _sadly_ we picked that end where we originally started the extension in the previous minus. I think this cannot be circumvented because of detailed balance.

jhprinz:
Let me pick up on the replicaID question. Beacuse that it important here, too. Do we want to have replicaIDs that are also non integers? If so that is not a problem at all, but I need to change the variable type for sample.replica to str or so. Otherwise saving a sample with a str id might so something weird.

dwhswenson:
> That our Subtraj in the Minus move need a sample from the minus ensemble, which in the case of the picture had been generated by an extension move before (this is why the order is reversed!) And in that that case sadly we picked that end where we originally started the extension in the previous minus.

Sorry, I'm not following what's going on here. It sounds like you're saying that the minus mover is still behaving as it should, but I'm not clear on what the confusion was.

The minus ensemble has some number of excursions from the state (which I've also occasionally called "loops"). In our implementation there need to be at least 2: the first and the last. (This is not technically required for the original minus move, which was just an extension, but I believe it *is* required for detailed balance in the multiple interface set minus move.) It shouldn't matter which excursion you pick, but from your words above, it sounds like it does.

If you're running MSTIS, you should still be seeing a total acceptance of the minus move that is something like $1.0-P_{transition}$ (in an equilibrated ensemble). So if we're seeing minus moves get rejected, we should look into why.

> Let me pick up on the replicaID question.

Let's keep the discussion in #309 for clarity (but this reference here will help us track the conversation in the future!)

dwhswenson:
chaindict.py:

* docstring: why only `Configuration` indices?
* `_add_new` and `_set`: why are these empty functions? And `_get` is `return None`?
* overall, *really* needs more comments/documentation. I can't really follow how this is working; I trust that it does, but I think it'll be hard to maintain without more descriptive stuff

dynamics_engine.py

* `n_frames_max` is a required option. You can't know a good choice without considering the system. I think setting 1000 is a bad choice. Setting 0 means it won't run at all if you don't give an option. The other choice is to give some absurdly large number. But 1000 means you can run paths that typically range around 700 frames and you might not notice that you're cutting off the tail of your distribution.

jhprinz:
#### chaindict.py
- Completely redid the docstrings with example and little cleanup
- _get and _set and "empty" because that is the default behaviour of a chaindict. Return `None` for a key when `my_chaindict[key]` is called and not do anything when `my_chaindict[key] = value` is called

jhprinz:
#### dynamics_engine.py

Some thoughts:
- We could not have this option and let the user set this number before he can initialize and use the engine. 
- We could also have a time constraint rather than a steps constr
- We could have the default be zero which means unlimited.
- I would like to have this set on creation so that when we know that a Sample is created by this engine we know that this sample cannot be longer than `n_frames_max` because of that particular engine.

dwhswenson:
#### dynamics_engine.py

(Specifically, the default value of `n_frames_max` in the `options` dictionary)

If we simply remove the default value, it looks like the current code already gives an infinite length trajectory as the option. If you think that's the desired behavior, then just make the default dictionary empty.

> We could not have this option and let the user set this number before he can initialize and use the engine.

Do you mean to have it set separately from instantiation? I think we both dislike that idea.

> We could also have a time constraint rather than a steps constr

I don't want to add dependence on additional packages (even standard ones such as `time`) if not needed; especially if we're only requiring them for one special case.

> We could have the default be zero which means unlimited.

I don't like "numbers are that number except for zero which is infinity." Not having the attribute already creates an unlimited result.

> I would like to have this set on creation so that when we know that a Sample is created by this engine we know that this sample cannot be longer than n_frames_max because of that particular engine.

I agree. This is in the `options` dictionary given to the engine when it is created. The only question is what to do if you *don't* explicitly include `n_frames_max` in that dictionary. I think the reasonable choices are:

* treat that as infinite
* don't run at all

I don't think setting it to some arbitrary number is a good choice. Within "don't run at all," you can either error or fail silently. Within "infinite", I think the trick we use now (the "stopping" ensemble is `FullEnsemble` instead of a `LengthEnsemble`) is probably best.

#### chaindict.py

Thanks. Those docstrings make it a lot easier to follow what is going on.

dwhswenson:
#### ensemble.py:

* What’s with `trajectory.lazy()` in `Ensemble.split`? Add a comment in the code explaining why there’s all this extra code for what seems like it should be easier. (I think I get it from the context of this PR, but if I just looked at that function 2 years from now, I would have no idea why it was implemented like that.)

#### base.py

* Comments/docstrings to explain the lazy wrapper (also, rename the decorator to something like `lazy_loading`: that’s a lot less ambiguous than just `lazy`)
* Rather than requiring lazy-loading classes to manually add the `_lazy` dictionary, [take a look at this trick](http://stackoverflow.com/questions/681953/python-class-decorator/682242#682242): Do you think you might be able to use that as part of the decorator to set an empty `_lazy` dict? If I understand it correctly, that trick lets you prepend some commands to the existing `__init__` function. In this case, it could make it so that all the implementation of lazy-loading is in the decorator. Our case is even easier than that example, because we don't need extra arguments to `__init__`.

#### pathmovechange.py:

* Why are you fixing `repr` to fixed width? Use that with `str` for pretty-printing; `repr` for unambiguous description

dwhswenson:
Comment for tracking purposes: since we can't use line notes for this PR, I'll keep editing this comment to contain a checklist of the files that warranted discussion, and I'll check them off once we've addressed everything in each file. (Most of the 100+ files that were changed aren't going to show up here, because there's no discussion needed.)

- [x] chain_dict.py
- [ ] dynamics_engine.py
- [ ] ensemble.py
- [ ] base.py
- [ ] pathmovechange.py

jhprinz:
#### ensemble.py
I added a lengthy comment at that point to explain it. To put the reason short is that. The split function does not care, if I pass a real Trajectory object or a list of snapshots. What lazy does is return the trajectory internal list of snapshot proxies (if the trajectory is already saved). So in that case and if your CVs are also cached. This little trick avoids loading again snapshots from the storage to run split. It will of course load a snapshot if a CV needs attributes of it (which then will be cached and then it will be fast again). You can always use this trick if you might only need the snapshot ID and not the snapshot itself.
Only difference between these proxy objects is operator overloading and a different `type()` result (`__class__` works though!)

jhprinz:
#### base.py
> Comments/docstrings to explain the lazy wrapper (also, rename the decorator to something like lazy_loading: that’s a lot less ambiguous than just lazy)

Good point. Renaming will do and add docs...

> Rather than requiring lazy-loading classes to manually add the _lazy dictionary, take a look at this trick: Do you think you might be able to use that as part of the decorator to set an empty _lazy dict? If I understand it correctly, that trick lets you prepend some commands to the existing __init__ function. In this case, it could make it so that all the implementation of lazy-loading is in the decorator. Our case is even easier than that example, because we don't need extra arguments to __init__.

Yeah. I thought about that. But this will make the `__init__` slower. I was not sure what the best way is and then decided to at least remove the `._lazy` dict from the main `StorableObject` class to keep this simple. You are right we could do that that is actually what I used to set of loading and saving in stores.
If you think that is not a problem of speed then I can easily add this. It will prevent potential users from wondering why it does not work (when they forgot to add the dict...) If you have no objections then I will add this automatically.

jhprinz:
#### dynamics_engine
Agreed. Not putting `n_frames_max` in the dictionary will look first in the class definition for a default_options attribute. Since storing is not an issue we can actually have
```
default_options : {
    'n_frames_max' : 'inf' # int('int') is not supported
}
```
alternatively we can use `None` to indicate that there is _none_ maximum frame number. Think both is fine and the unbound ensemble should be the default.

jhprinz:
I just realized that I have extensive type testing which is based on the default_options. This means that you have to put all possible options into the `default_options` dictionary otherwise they will be ignored.
Then also the type of the variable has to match. So when we put in a string then you can only put strings (right now). If you put `None` then everything is allowed. 

Not sure anymore if this is flexible enough for us. For now I will change the default to None and treat None as meaning no max frame number.

A last idea that just came to me is that we could parse the docstrings of an engine for the possible options. So you have to only update the docstring and then the options match. (Although better generate it the other way round :)

jhprinz:
Last thought I had was to replace the max frame by a simple
```
if len(trajectory) >= n_frames_max:
    break
```
instead of creating a complicated stopping Ensemble that is then used with `can_append`.

More ideas:
- Replace `running = ` with `ensemble = ` in `engine.generate` and have two functions `generate_forward` and `generate_backward` that will call `can_append` or `can_prepend` resp. on the ensemble.
- in the above case I would have the general `running =` renamed into `generate_until` or so.

jhprinz:
I made a little PR #340 to add these changes, if we want to. This might make the PathMover overhaul easier, too.

dwhswenson:
#### base.py

(On letter the decorator add the `_lazy` dict)

> If you think that is not a problem of speed then I can easily add this. It will prevent potential users from wondering why it does not work (when they forgot to add the dict...) If you have no objections then I will add this automatically.

How bad is the speed hit, in practice? If this significantly slows down `Snapshot`, it might be a problem. (Check speed on analysis as well as production -- the slowdown will be less noticeable when you're alloc'ing memory for configs and momenta during production.)

dwhswenson:
#### dynamics_engine.py

This discussion has moved to #340, right? All the changes discussed above are managed in that PR, which is independent of this one?

dwhswenson:
#### trajectory.py

* Again, with `lazy` : `trajectory.lazy()` needs a better name. Lazy what? `Trajectory.as_proxies` and `Trajectory.iter_proxies` would be clearer.

#### todict.py

* Rename file to `object_json.py` ?

#### testorderparameter.py

* in `setUp`: tests/test_data already contains an alanine trajectory as PDB. Load that instead of generating a new one. This will speed up tests (Travis doesn’t have GPUs!!)

#### teststorage.py

* move `assert_close_unit` to test_helpers.py so it can be used by other tests
* again, why are the tests running real trajectories? If possible, load the PDB trajectory and then use it.

dwhswenson:
#### ensemble.py

> I added a lengthy comment at that point to explain it.

Haha -- "lengthy" is right! Someday we'll want to make it a little more condensed. But that's enough for me to call `ensemble.py` done.

dwhswenson:
#### collectivevariable.py

I strongly disagree with this interface for `CVFunction`. This is a user-facing function; indeed, one of the most important user-facing functions. I feel that there are two major problems with it:

* Bad defaults
* No “canonical” initialization 

Also, the docstrings will need a major overhaul. I recommend doing that after we figure out what to do about the other problems.

***

##### Bad defaults

Defaults are shown in Python’s `help` and in documentation. When there are reasonable defaults, they should be explicit. If the default `var_type` is `'float'`, write `var_type='float'`, not `var_type=None` and change it later. I’d also rather see default `dimensions=1`: when does it start being that instead of `None`? 

***

##### No “canonical” initialization

I really dislike the idea of having many (possibly conflicting) options in `__init__` for functions that are user-facing. You set it up so that the template overrides, but is that necessarily obvious to a user? 

I think there’s a better way to do this.

1. Change `_init_from_template` to a  function `options_from_template` that returns the dictionary of changes instead of setting them with `self`.
2. Add a class method `from_template`, which *only* allows the `template`; none of the `dimensions`, `var_type`, etc. This just calls `options_from_template` and then `__init__`s with the resulting dictionary.
3. Remove `template` from the arguments to init (in all functions)

When all is said and done, this should work just fine (for appropriate input functions)

```python
my_normal_cv = CVFunction("normal", myfunc)
my_cv_from_template = CVFunction.from_template("template_version", myfunc, template_snapshot)

my_weird_cv = CVFunction("weird", other_func, dimensions=2, var_type="string", fnc_uses_lists=True)
my_weird_cv_from_template = CVFunction.from_template("weird_template", other_func, template_snapshot)
```

No ambiguity, no question of “which initialization approach should I use.” Any reason this wouldn’t work?

dwhswenson:
I've now reviewed the whole of the PR to date. This is an epic effort, with a LOT of really useful cleanup that I thought we wouldn’t get around to until just prior to release!

The remaining stuff to be dealt with is in the checklist in https://github.com/choderalab/openpathsampling/pull/300#issuecomment-150684165. I also submitted a couple minor changes in jhprinz#7.

jhprinz:
#### base.py
- speed problems I mentioned. The speed problem is negligible since it is comparable of subclassing. It will just add one line of code to inits that use lazy. I first thought this fix is per construction but it is only done once per class creation. So that should be alright if we agree that subclassing is okay from a speed perspective. 

#### dynamics_engine.py
Yes, changes will be in #340 . 

#### todict.py
Also agreed, this is a better name, although it is used to mainly create dictionaries from objects and this dicts are used to turn them to json. Since we are not using the dict itself, but ony for json I would rename it.

#### trajectory.py
Yep. Done

#### testorderparameter.py
Great, I though this was just a single frame pdb... Fixed that.

#### teststorage.py
Jup... Working on it

dwhswenson:
#### base.py

> I first thought this fix is per construction but it is only done once per class creation.

Wow... Python decorators are awesome. 

I also thought it was per construction. Subclassing is small enough overhead that we can definitely pay that cost -- I assume it would be hard to measure that cost compared to fluctuations in the cost of dynamics!

If the cost is minimal, let's definitely use that! And this has been implemented now, right? I think so -- marking base.py as done in the checklist.

#### dynamics_engine.py

Done with that in here. (Close to done in the other PR)

#### todict.py, teststorage.py, testorderparameter.py, trajectory.py

Looks like the last few commits have made the changes for all of these!

***

This leaves us with `pathmovechange` (I'd suggest switching from `repr` to `str` or something else entirely) and `collectivevariable` (which is the hard one).

jhprinz:
> I first thought this fix is per construction but it is only done once per class creation.

Just to be precise. What it does once it create the replaced init function. In each construction it will then just call the new init which is basically in complexity equivalent to calling the super method of a superclass.

#### pathmovechange.py

This is easy to fix and will do soon

#### collectivevariable.py

I really like the factory idea. That makes the trouble of deciding what parameters to use easier. The change is actually not that difficult. My main question would have been if the automatic testing of the function is useful or not. I like using a template to setup the parameters. And load and save uses all parameters anyway so all of that is already working. I think I can finish this tomorrow...

dwhswenson:
#### collectivevariable.py

> My main question would have been if the automatic testing of the function is useful or not.

Do you mean whether the template-based initialization is useful? I think it is great, but as a secondary creation approach. I suspect that 98% of the time, I won't use it. But for that 2% of the time when I create a CV that maps not to a single float but a vector or a non-float, I would *much* rather use the template setup than try to remember which of the options I need to set.

Also, please double-check various docs. I've always described a CV as mapping a snapshot to a single float. You made that much more flexible by making it possible to map a snapshot to pretty much any type (at least, any of a significant list of types). The docs (in-code and out) should reflect this improvement!

(Aside: While you're working on reshaping `Snapshot`, a similar factory-based approach might be a good idea there. Most of the time we just initialize a `Snapshot` with its features [positions, velocities] directly, but the various other approaches might be convenient as factory options.)

jhprinz:
#### cv.py

- `dimensions=1`: This is tricky. dimensions is now a tuple like it is used for numpy or better netcdf. For a scalar the dimension is the empty tuple `(, )`. I added `None` as an alternative. I think None for dimensions if there are none is not a bad choice. Better would be to put is as `tuple()`. Your proposed `dimensions=1` would correspond to an array of one dimension of length 1, which is not a scalar.

- There is a common issue in using mutable classes (list, dict, etc) as defaults in python parameters which comes from python working different than expected. Imagine you have the following method in a class
```
class MyDict(...):
    def set_dict(self, dct = {}):
        self.dct = dct
```
than I would have assumed that calling `my_instance.set_dict()` would set `self.dct` to `{}` the empty dictionary. Yes, that is what happens, but consider this case
```
a = MyDict()
b = MyDict()
a.set_dict()
b.set_dict()
a.dct['key'] = 'something'
```
so setting up two instances. Use set_dict on both and then add a key to the `dct` attribute of one instance. Now the problem
```
print b.dct['key']
>>> 'something'
```
Now `a` and `b` use the same dictionary. This is because the default is evaluated at class creation. So only once and hence the `={}` will once create a dict and use it for all instances the same. Recommended is the usual `def func(..., dct = None):` and then, if the parameter is `None` replace it with a real new dictionary.

Luckily tuples are immutable and so we can use it.

Bottom line: I would use `dimensions=`
1. `None` or `tuple()` for a scalar
2. `int` or `tuple(int)` for a one dimensional array
3. `tuple of ints` for a multi-dimensional array

jhprinz:
Lots of changes in cv.py...

This is now much better and has more or less correct docstrings. I decided to do one thing though:
Right now we just pass ass function parameters which are not used yet as kwargs to the function.
Like this
```
cv = CV_Function('name', fnc, dimensions=2, var_type='float', param1=value1, param2=value2)
```
Since we have so many other parameters and to keep the docstring clean I now use
```
cv = CV_Function('name', inc, f_kwargs = {'param1' : value1, 'param2': value2}, dimensions=2, var_type='float')
```

jhprinz:
If we keep it the way it is, I need to kind of always take care of all possible parameters that superclasses use to make sure that the user does not pick e.g. the variable `dimensions` for his function and gets an error without reason.

jhprinz:
#### pathmvechange.py

I reverted the `__repr__` to the original one. We will have another PR that will redo the `__str__` and `__repr__` of all objects.

jhprinz:
Second question is that I woule like to keep the following parameters for all classes the same
```
CV_Callable(name, c, c_kwargs, ...)
CV_Function(name, c, c_kwargs, ...)
CV_Class(name, c, c_kwargs, ...)
```
which looks okay for `CV_Callable` and `CV_Class` but for function better would be
```
CV_Callable(name, c, c_kwargs, ...)
CV_Function(name, f, f_kwargs, ...)
CV_Class(name, c, c_kwargs, ...)
```

the above will keep the code much cleaner. The second has better names but clutters the resuablilty for storage and requires stupid redundant code.

jhprinz:
I fixed it now that we have actually `f_` for the function, but it does not look too nice...

dwhswenson:
> `dimensions=1`: This is tricky.

And this is ***exactly*** why we need explicit default parameters. I would have expected `dimensions=2` to work; if the default said `dimensions=tuple(1,)`, I would have realized that I needed a tuple.

The after the function name, function signature is the first thing people see. It should be as clear as possible.

> There is a common issue in using mutable classes (list, dict, etc) as defaults in python parameters 

Okay, and good point, but what was the context for this?

> If we keep it the way it is, I need to kind of always take care of all possible parameters that superclasses use to make sure that the user does not pick e.g. the variable dimensions for his function and gets an error without reason. This is why I prefer it using dicts.

I still don't like this change. In its simplest form, this will be used by novice users every day. Adding the dictionary syntax is a layer of complication that I'd like to avoid if possible. The better approach, in my opinion, is to give names to our arguments that are unlikely to be used by a function -- preface them with `cv_`: `cv_dimensions`, `cv_var_type`, etc. Documentation can say that these words are disallowed from subfunctions, and I'm not terribly concerned that they would be used.

> Second question is that I woule like to keep the following parameters for all classes the same

If I understand correctly, the issue is that you want to change the arguments of the `__init__`?

First, keep in mind that function will be the most commonly used case. The arguments for it should make sense. Second, this is a required argument, so users are less likely to use its name. I think `callable` is an accurate general term for this.

***

The overall disagreement here is that you want to have all these abilities, and I want something simple enough for the weakest student to use without needed it explained. Why don't we separate these into two classes? `CV_Function` that behaves as it used to, and `CV_GeneralizedFunction` that has all the bells and whistles? I think that gives us the best of both worlds.

dwhswenson:
One other point: does netCDF call that tuple of dimensionality `dimensions`? numpy calls it `shape`; if I saw `shape` instead of `dimensions`, I would be more likely to expect the tuple. If netCDF also calls it `shape`, then we should call it `shape`.

jhprinz:
> dimensions=1: This is tricky. And this is exactly why we need explicit default parameters. I would have expected dimensions=2 to work; if the default said dimensions=tuple(1,), I would have realized that I needed a tuple.

What do you want to say with that? What is the context. Do you mean that bad docstrings are bad? :) I completely agree. My question was, if you agree that there is a difference between `dimensions=1` and `dimension=None` or `dimensions=tuple()` and that this makes sense. (assuming that the docstrings are correct, or course) I think it does make sense to have `None` as a default meaning a scalar.

> The after the function name, function signature is the first thing people see. It should be as clear as possible.

I agree. Hence updated docstrings.

> > There is a common issue in using mutable classes (list, dict, etc) as defaults in python parameters
> Okay, and good point, but what was the context for this?

I thought this might be relevant when using tuples as defaults (this is the context) and I saw you doing that a few times in code and though it would be a good time to just mention that.

> I still don't like this change. In its simplest form, this will be used by novice users every day. Adding the dictionary syntax is a layer of complication that I'd like to avoid if possible. The better approach, in my opinion, is to give names to our arguments that are unlikely to be used by a function -- preface them with cv_: cv_dimensions, cv_var_type, etc. Documentation can say that these words are disallowed from subfunctions, and I'm not terribly concerned that they would be used.

Hmm, the `cv_...` would be an option but still makes the function calls potentially very long, a dict would be easier to define before calling the CV_Function. I remember you saying that you don't like lots of parameters in functions, which I tried to avoid and keep clearer.

The `cv_` prefix would be okay and I think I go with that. I agree that there are _two_ most common cases. Simple functions for toy models and use of md_traj or other featurizers for MD which will have oftern more than one value and be not scalar. We should have simple functions for this. I would start with the general case though and get more specific. So my proposal would be

1. Keep it almost the way it is and prefix the (mostly storage related) arguments with `cv_` to make clear that these are not for functions. 
2. All others (or better ones that do not start with `cv_` will be excluded and considered for the called function kwargs
3. `CV_Function` will be the general one with defaulting to scalar and float
4. (optional) `CV_ScalarFunction` will be a single float that is the super easy to use function for beginners. This is not necessary since in the normal case it is identical to `CV_Function`
5. There will be specialized versions for `CV_MDTraj_Function`, `CV_MSMB_Featurizer` and `CV_PyEMMA_Featurizer`

jhprinz:
> One other point: does netCDF call that tuple of dimensionality dimensions? numpy calls it shape; if I saw shape instead of dimensions, I would be more likely to expect the tuple. If netCDF also calls it shape, then we should call it shape.

That is a good point. Yes, netcdf calls it dimensions that is why I kept it that way. So that you can use the same in creating a netcdf variable. It might be clever to use this, since the user will not like have to deal with that. 

To be precise in netcdf it is called dimensions and these are normally names of named dimensions so that you know that some variables share a dimension, so you tuple would normally look like `tuple('snapshots', 'atoms', 'spatial')` And somewhere is defined that snapshots is growing, atoms is some fixed number and spatial is 3. What I did is extent this for cvs and saying tuple(2, 5) creates dimensions for 2 and 5 and adds `snapshots` as first entry. Then it creates the variable.

So shape is the current actual size, while dimensions is kind of the concept. 

After writing that I agree that shapt would be a better name, since this is not the dimensions. I think I just change it. netcdf does actually have a shape property (just checked) although it is not used during creation. 

Agreed...

dwhswenson:
> Do you mean that bad docstrings are bad? 

It's not the docstring I'm worried about, though -- in my development setup, my autocomplete will show my the `def` line from a function when I autocomplete. That means I also get what the *actual* default values are. So I want those default values to be as clear as possible. I'm fine with `dimensions` being a `tuple` (even moreso if it is now `shape`, or `cv_shape`): my point is that, in that case, the default should be `tuple(1,)` so that it is clear from the code definition, before even getting to the docstring! In this example, it would have immediately prevented my from using a integer.

And if that's the case, I don't think you need to support `dimension=1` or `dimension=2`; a user should use `cv_shape=tuple(1,)` or `cv_shape=tuple(2,)`.

Although, as I think about this more, a single value doesn't have a shape, since that implies the wrapper around it. I think, with the name `shape`, it makes as much sense to use `None`, although the empty `tuple()` might be even better, to emphasize what the type should be.

>  I agree that there are two most common cases. Simple functions for toy models and use of md_traj or other featurizers for MD which will have oftern more than one value and be not scalar.

I think you overestimate how often the user's desired CV for real MD has already been implemented in MDTraj, etc. (Example: custom order parameters for DNA, based on a non-standard definition of the base opening angle -- something I've used before.) Using functions of your own design is not just for toy models.

But yes, both of these approaches will be used. I guess my assumptions were (1) some of these properties (e.g., `cv_uses_lists`) would always be the same for any CV based on that package; (2) the user would either easily know parameters describing the return type, or they would set it up with a template snapshot.

> So my proposal would be

The proposal sounds good to me. If `CV_Function` with no optional arguments behaves as `CV_ScalarFunction` would, then I don't see a need for the latter.

Also, with the `cv_` prefix, maybe `cv_return_type` and `cv_return_shape`? That makes it really clear *exactly* what these are.

jhprinz:
> > Do you mean that bad docstrings are bad?

Sorry for the kind of not constructive question. I understood what you were referring to, but the old defaults were part of the case where we could not distinguish between scalar and array of length 1. My main concern was, that we use None all the time as defaults and then the docstring is important and in the case of dimension/shape a single type was also not helpful. I agree that a default like `tuple(1, )` would have been better than `1` but I wanted to use either `tuple()` or `None` instead. `tuple(1, )` was out of the question and using arrays of size one rather than scalar did not seem to make sense to me.

Anyway. I implemented the cv_ prefix and agree that CV_Function is the most common case. In the current implementation I realized that the JSON pickling for classes and other code is practically identical so I reused that, hence the strange CV_Callable class, which I would keep abstract and not add the `__init__.py` 

> Also, with the cv_ prefix, maybe cv_return_type and cv_return_shape? That makes it really clear exactly what these are.

Agreed, that looks even better.

dwhswenson:
This is very close; I think I only have a few more (very simple) suggestions:

* I noticed a function `_interprete_num_type`: I think that has an extra 'e' in "interpret". (Although it is not better `_identify_num_type` or even `_identify_type`?)

* Order of arguments as listed in `CV_*`: let's put everything that deals with output together, and everything that deals with input together. (This basically means moving `cv_simtk_unit` before `cv_requires_lists`.)

* Is the `single_as_scalar` only relevant for MDTraj? Sounds like it might be more broadly useful.

jhprinz:
Good points...

> I noticed a function _interprete_num_type: I think that has an extra 'e' in "interpret". (Although it is not better _identify_num_type or even _identify_type?)

Well, my mistake, I will change that to `_identify_var_type
`
> Order of arguments as listed in CV_*: let's put everything that deals with output together, and everything that deals with input together. (This basically means moving cv_simtk_unit before cv_requires_lists.)

Will do.

> Is the single_as_scalar only relevant for MDTraj? Sounds like it might be more broadly useful.

Well, kind of. The point is that mdtraj and msmbuilder often give you a array of dim 1 instead of a scalar, beacuse they are intended to be used for more than one output variable. This option is for using these methods that return `[1.0]` instead of `1.0` and you do not want to wrap that in a function to fix. This will just reformat the output quickly by reshaping. If you write your own function I expect that the user produces output the he wants it too. And if he returns `[1.0]` then it will be saved as such.

jhprinz:
I also changed the name of `cv_simtk_unit` to `cv_return_simtk_unit`

So, while looking through cvs I fixed another bottleneck that should make storing cvs faster and less painful. This includes an option to make a CV reversible using `cv_reversible=True` which tells the CV that the reversible copies that are always created will have the same value and hence do not need to be computed again. This should increase CV speed by a factor around 2. It is especially nice if you reverse a trajectory then you immediately have all cvs. 

It will also take care of the case now that you computed some cvs, then store the snapshots (which means now you can also save the values and not before) and then do a sync. It will just grab values for all newly stored snapshots from the main cache and store these, too.

Last question about the name of the option. Right now I have
`cv_reversible` but it could be `cv_symmetric` or so. Then I will make sure that tests pass and we have a go.

dwhswenson:
> Last question about the name of the option.

`cv_time_symmetric` : the others alone don't mean anything to me without reading the docstring.

Note:

* This should default to `False`. It may be `True` most of the time, but if it defaults `True` then a user can create an incorrect CV by omitting this. The user chooses `True` to get the speedup.
* You can't autodetect this from a template. If the template snapshot has all velocities 0 (which I think will be a common template), then any CV is time-symmetric. So this needs to be an option (default `False`) in the build-by-template approach.

jhprinz:
I changed the defaults and added a simple test to the `from_template` that will check, if
```
cv(snapshot) == cv(snapshot.reversed)
```
holds. If so we assume by default that it is time_reversible. You can of course override this. It is also this simple equals, so if you have a numpy array then this will fail and fail means not time_reversible. So (right now) for numpy array you need to set it yourself.

Finally I added more tests to the `test_cv` notebook.

dwhswenson:
> if `cv(snapshot) == cv(snapshot.reversed)` holds. If so we assume by default that it is time_reversible.

A user is likely load the template snapshot a PDB, with velocities initialized zero. This test will say, incorrectly, that the CV is time symmetric.

I think this is a serious bug that we should avoid.

jhprinz:
Sorry, I read your previous post wrong. I understood velocities zero in the template should mean is time-reversible, since the choice of zero is not really a template. I would not call this a serious bug, if the docs are correct. It is maybe a convention that could be dangerous, but initializing a function that depends on velocities using a snapshot that has only zero velocities is maybe _not smart_. Doing that for coordinates it would be more obvious that chosing only zeros is not in the spirit of having a template.

But alright. I will remove this. It causes some troubles anyway that make the code more clumsy. 
Thanks for the remark!

dwhswenson:
I think it's better to have a false negative and run slower than a false positive and run incorrectly.

I tried to think of another approach that would still auto-identify this property, but there are other problems once you add other `SnapshotFeatures` than just velocity and position. I considered the possibility that you could use the test you suggested, but also test for all-zero velocities. If template velocities are all zero, you're stuck saying you can't predict, but otherwise maybe you could. However, this is still a problem once we include other `SnapshotFeatures`: you might have a snapshot which doesn't depend on velocity, but does depend on some other feature that is affected by the direction time's arrow. (There are examples of this in semiclassical dynamics -- at least, in certain implementations that I find potentially useful.)

One possibility that I think ***would*** work is to check whether the CV is only dependent on positions. You could do this by making a copy of the template snapshot, setting all snapshot features except positions to `None`, and trying the CV. If it doesn't raise errors (probably `AttributeError`s), then the CV only depends on positions and is therefore symmetric with respect to the arrow of time.

This auto-checking would miss some velocity-dependent CVs that happen to be independent of time's arrow (e.g., kinetic energy) but it would catch most of the CVs you care about, and wouldn't give any false positives.

jhprinz:
I was thinking... The return_type, shape and unit only matter for saving. So we could say that this is part of storage and remove the choice of the user from it. 

Main point is that the user can only do things wrong. Example: The user has a function that returns a numpy scalar and uses the default. This is float and since these are compatible you will not see an error due to the power of duck typing. When you load it will return a python float and if you then expect something that only a numpy variable can do this will give an error. If it would be handled automatically it would return the type you created.

For the `time_reversible` I would be on the safe side and set it to false and allow the user to switch it to true if he explicitely says so. Everything else could go wrong.

We could even remove the necessity from templates, since if you want to store you already have a template in the storage you could use. Only caveat is that the testing if the function uses lists or single values needs a template. If we would say that the user needs to know about this if he writes a function that supports lists, we can remove `from_template` completely or there is no need for it. Default for requires_lists is false.

I think that it the best solution. All these parameters are somehow confusing.

dwhswenson:
That all sounds very good to me.

If I understand correctly, this proposal is:

* All the output-related parameters get figured out in `storage`. We need to know these things to store them correctly, but the first time this CV tries to store a result, we'll necessarily have a "template" snapshot and you can use the tricks in `from_template` to figure these things out.
* `time_reversible` and `requires_lists` remain as optional parameters in the CV setup, defaulting `False`. If `time_reversible` should be `True` but isn't, then the calculation runs correctly, but slower. If `requires_lists` should be `True` but isn't, then using the CV would raise an error (it would be nice if we could catch this and append to the error message to say "Maybe you want your CV to be `requires_lists=True`".)
* the `from_template` method would be removed, because all its work would be shifted to storage

I do think you might still consider using the depends-only-on-position test I suggested above as part of the first check of storage stuff -- that way a user doesn't *have* to specify `time_reversible=True` to get that behavior, in the most common case. The only downside of doing that would be that less experienced users might use a kinetic-energy-like CV and it would run slower because they don't know about the `time_reversible` flag; the trade-off is that beginning users don't have to worry about it in order to get the speed-up.

> I think that it the best solution. All these parameters are somehow confusing.

Agreed on both points.

jhprinz:
> All the output-related parameters get figured out in storage. We need to know these things to store them correctly, but the first time this CV tries to store a result, we'll necessarily have a "template" snapshot and you can use the tricks in from_template to figure these things out.
Yes.
> time_reversible and requires_lists remain as optional parameters in the CV setup, defaulting False. If time_reversible should be True but isn't, then the calculation runs correctly, but slower. If requires_lists should be True but isn't, then using the CV would raise an error (it would be nice if we could catch this and append to the error message to say "Maybe you want your CV to be requires_lists=True".)
Yes, but I added two more option that are present now and alter the output of a CV and are generally useful.

1. `wrap_numpy_array`, default: False: if true this will always wrap the output in a numpy.array so that an output of a list of numpy.array will be a single numpy.array. 
2. `use_numpy_scalar`, default;False: if true this will check, if the last dimension of the output is of length one and will remove this dimension. Meaning that an array `[[1], [2], [3]]` will be turned into `[1, 2, 3]` which is in most cases what you expect.

Both options will only affect the outcome and add some convenience to work with the result. They are mostly used woth MDTraj outputs, etc...

> the from_template method would be removed, because all its work would be shifted to storage
Yes.

> One possibility that I think would work is to check whether the CV is only dependent on positions. You could do this by making a copy of the template snapshot, setting all snapshot features except positions to None, and trying the CV. If it doesn't raise errors (probably AttributeErrors), then the CV only depends on positions and is therefore symmetric with respect to the arrow of time.

That is the best idea yet, although there might be cases where a `None` might pass (which in general it should not). We could turn this idea around and say that everything that explicitly depends on `velocities` defaults time-reversible to false. So set only momenta to None and test again. Problem is that a snapshot in the future might not have velocities. In this case your idea works better.

Another approach could be to scan the code, if it uses a get attribute of .velocities in some way. But I guess that is even more unsafe.

For now I would leave this out and change this, once we have support for `SnapshotFeatures`. As you say in the worst case it will slow down access to cvs of reversed snapshots a little, but that might be okay .

dwhswenson:
> `wrap_numpy_array`, default: `False`: if true this will always wrap the output in a numpy.array so that an output of a list of numpy.array will be a single numpy.array.

Just to be sure this is clear: this means that if a CV returns a 5-item `list` where each item is of the list is an `np.array` with shape `(2,)`, the result is an `np.array` of shape `(5,2)` (or maybe `(2,5)`)? As opposed to flattening it to `(10,)`. 

> `use_numpy_scalar`

I understand what this does, but I don't think the name is clear. Maybe `scalarize_numpy_singletons`?

> there might be cases where a `None` might pass (which in general it should not).

Do you have an example? I tried to think of one when I proposed it, and the only things I could imagine were cases where the user did something kind of nonsensical (like writing the CV such that if it lacks `velocities`, it pretends the velocities were all zero).

(My idea was to also compare the result from the `velocities=None` calculation, if it passes, with that from the template; so this would only be a problem is the template snapshot has the same velocities as assumed by `velocities=None`.)

> For now I would leave this out and change this, once we have support for `SnapshotFeatures`. As you say in the worst case it will slow down access to cvs of reversed snapshots a little, but that might be okay.

Agreed. This PR needs to get merged ASAP.

***

Something else to keep in mind, related to a lot of this. The double-storage of snapshots as snapshot and its reversed version is entirely because of the existence of (and relatively high frequency of) path reversal moves. It bothers me that the entire structure of storage is so effected by one aspect that is particular to this kind of calculation.

Not for this PR, but:

* Is it possible to have a distinction between storage that automatically stores the reversed snapshot and storage that doesn't? My guess is that this might best be two different classes (rather than an attribute, which, if a user changed it after starting to write the file, would probably cause all sorts of problems).
* Is there another way to efficiently manage reversed snapshot pairs, without actually creating them in pairs? Maybe using an extra dictionary that maps each snapshot to its reversed partner?
* Do we actually save the reversed momenta, or is it just saved as a pointer with a flag to reverse them? (This would make a significant difference in storage space. I assume it is just the pointer and flag; otherwise, I think we should make it so as part of the `VelocitySnapshotFeature`)

jhprinz:
> wrap_numpy_array, default: False: if true this will always wrap the output in a numpy.array so that an output of a list of numpy.array will be a single numpy.array. Just to be sure this is clear: this means that if a CV returns a 5-item list where each item is of the list is an np.array with shape (2,), the result is an np.array of shape (5,2) (or maybe (2,5))? As opposed to flattening it to (10,).

Yes, instead of a `shape=(5,2,1) is will return `shape=(5,2)`. It just removes the kind of unnecessary third dimension.

> use_numpy_scalar: I understand what this does, but I don't think the name is clear. Maybe scalarize_numpy_singletons?

Sure, I wasn't happy with the name myself. 

> Do you have an example? 

Well, just one where the function treats None as zero velocities (which is dumb, I agree)

> Something else to keep in mind, related to a lot of this. The double-storage of snapshots as snapshot and its reversed version is entirely because of the existence of (and relatively high frequency of) path reversal moves. It bothers me that the entire structure of storage is so effected by one aspect that is particular to this kind of calculation.

Well, it seemed the simplest solution to the problem that you want
```
snapshot.reversed.reversed is snapshot == True
```

This is why I create them always in pairs. Storage is really minimal, just a few bytes more. And if you do it correctly there is no real speed issue, too. It basically stores snapshots in pairs where one is reversed the other isn't. All of this is only relevant for analysis to make sure that you do not overcount or have to check if two snapshots are the same. In theory this would be easier but I wanted it to be completely naturally work with CVs. That means that I want to store the CV in a table with the same dimension as the snapshots. The advantage is that you can use the same index for all snapshot data and all cvs alike. I found that to be very useful if someone wants to read the netcdf file and analyse the data themselves.

> Is it possible to have a distinction between storage that automatically stores the reversed snapshot and storage that doesn't? My guess is that this might best be two different classes (rather than an attribute, which, if a user changed it after starting to write the file, would probably cause all sorts of problems).

You would have to replace the snapshot_store with another that does not do auto-create the reversed ones. You also have to add a cache for remembering which object was the reversed of which one and you have to store and load that efficiently, too.

> Is there another way to efficiently manage reversed snapshot pairs, without actually creating them in pairs? Maybe using an extra dictionary that maps each snapshot to its reversed partner?

I think not really. You can do what you propose but it appears (so far) more difficult and is probably slower.

> Do we actually save the reversed momenta, or is it just saved as a pointer with a flag to reverse them? (This would make a significant difference in storage space. I assume it is just the pointer and flag; otherwise, I think we should make it so as part of the VelocitySnapshotFeature)

No, you do not save reversed momenta and if you load a snapshot also the reversed one is not created, there is only a stub. If you access the reversed it uses potentially cached reversed ones to make construction faster. 

All of that is kind of complex under the hood, but in storage and to the user quite fast and well hidden.

jhprinz:
I have just discovered a few little things that slowed the code down: 

I was running the notebooks again to make sure they work now when I was wondering why some cells take quite long to evaluate. Turned out that the use of the proxies is quite fast, but I wrote the `__getitem__` of Trajectory in a way that will automatically load the real snapshot so that a user will never see the proxy objects (unless he uses `.as_proxies()`). That means that when you write
```
t = storage.trajectories[0] # loads list of proxies and is very fast
t2 = Trajectory(t) # iterates over all REAL snapshots and this triggers loading
```
you will have gained nothing the first time since all snapshots are loaded. So I fixed parts of ensemble and trajectory so that Proxies will stay proxies if possible. This is very nice and gave a big speed boost -- mostly when you use cached cvs and since volume and ensemble depend all on that stuff, they all got faster. 

#### Example

```
trajs = {}
for ens in [t.ensembles[-1] for t in mstis.sampling_transitions]:
    trajs[ens] = next(s.trajectory for s in old_sampleset if ens(s.trajectory))
    
assert(len(trajs)==3) # otherwise, we have a problem
```
took 1.4s without using the proxy change and 0.14s with the changes. I checked that now all of the function will not trigger loading. Even for already loaded snapshots this is faster since it will not test for proxies etc. it will just do list.append, extend, add, etc...

jhprinz:
Look how fast the `mstis.ipynb` will run until you hit the actual production. Nice...

jhprinz:
This is really the last change...

dwhswenson:
Whoa ... can you double check the correctness of the IPython notebook runs? Could you just do runs of 5000 or so steps in both the MSTIS and MISTIS examples? Run the analysis, and check that it gives answers (5000 steps should be sufficient to have answers for everything, although those answers won't be converged).

In the [most recent test](https://travis-ci.org/choderalab/openpathsampling/builds/88683488), `mistis_setup` took 9 seconds. That doesn't seem credible to me. (On the other hand, `mstis` took a little under 3 minutes, which is a lot faster than it used to be, but within a believable order of magnitude). Since I know that ipynbtest has been missing some errors lately, I'm concerned there might be a problem that we're not seeing.

However, I suppose that MSTIS has live visualization, while MISTIS doesn't. That *will* add a decent bit of overhead, multiplied by 300 steps.... But I would be a bit surprised if we can attribute more than a minute of the difference to the live vis.

jhprinz:
You are right. I had not tested mistis only mstis, which actually runs faster now. There was a missing renaming in the notebook. I will run a long test tomorrow and check that it passes.

jhprinz:
I also need to figure out why the ipynbtest is not working correctly. Guess this has todo with all the ipython jupyter updates.

jhprinz:
ups, I left the 5000 steps in the notebook and it did not pass...

jhprinz:
much better just figured that the tests ran faster than usual...
```
Duration: 17 min 22 sec
```

And all seems fine. This speed up is mainly due to avoid unnecessary loading of objects (I think). So this is a gain that happens once per notebook but is maybe not so present, when we run longer simulations.

jhprinz:
Ran the MSTIS example with 5000 steps and realized that the cache for cvs was too small. With too small cache it took 7 1/2 minutes to compute the rate_matrix. Increasing the cache to hold all cv values let the computation of the rate matrix down to 24seconds. I think that is reasonable. Even for bigger systems that should give times that we can deal with.
This used that all cvs are precomputed and stored!

jhprinz:
Now running MISTIS with 5000 steps. This is actually running much faster. Don't really know why.

jhprinz:
Second test complete. MISTIS works (as far as I can understand the result)
458902 snapshots. 6352 samples.
Loaded file in about 20seconds. Rate computation with force=true took first time about 1 min 30s where I believe some stuff needed to be loaded or some CVs were missing.
Second try with force took 27seconds.

jhprinz:
Alright, ready for review.

dwhswenson:
> Rate computation with force=true took first time about 1 min 30s where I believe some stuff needed to be loaded or some CVs were missing.

Rate computation requires loading all trajectories in the minus ensemble, and determining which volume each snapshot is in (for the flux). So it will always be a little slow. It probably isn't worth calculating that for *all* snapshots during production, though. 

> Now running MISTIS with 5000 steps. This is actually running much faster. Don't really know why.

Much faster than before, or much faster than MSTIS? Much faster than before is a good sign that this has sped things up significantly. Much faster than MSTIS would be a bit strange, since they take about the same amount of time on Travis.

dwhswenson:
I'm concerned about the error in the [most recently committed version of `mstis_analysis.ipynb`](https://github.com/jhprinz/openpathsampling/blob/d1897c195a129b2b7e806a6475b9a3d8b4d31d61/examples/ipython/mstis_analysis.ipynb). It seems to find 0 minus movers for some minus ensemble. That would indicate something weird going on. Can you re-run this to check whether that was just some weird fluke? I assume that you would have seen it in your longer runs if it was a real problem, but please double-check anyway.

jhprinz:
Hmm, this could come from copying the notebook from another branch. In the 5000 steps examples everything was fine. No errors. I will rerun the notebook tests and see, if that happens again. If I recall it did not happen in the 5000 step test, but it happens once in a while for the 300 steps.

jhprinz:
Just reran my tests and did not get the error. I think it was one of these rare cases where you do not have all necessary minus. Could have already happened in the mstis.ipynb

jhprinz:
This is my last notebook

https://github.com/jhprinz/openpathsampling/blob/numbertypes/examples/ipython/mstis_analysis.ipynb

jhprinz:
Had an error in the last test_cv notebook which runs strict. needed to update this again after I mistakenly checked out an older version and committed it.

dwhswenson:
It took 2 re-runs on Travis to get this to actually pass the tests.

That said, I'm just going to merge it, and then I'll look at the notebooks locally to try to figure out what's going wrong.

* 204 commits
* 118 messages in the thread (once I post this one)
* 122 files changed (how many files do we even *have* in the repo?)

Are you sure it is long enough? A little longer and it would be Tolstoy's War and Peace. ;-) This is one epic of a PR.

jhprinz:
Yes... I imagine us in 50 years sitting in a cottage by the fire and remembering the legendary PRs from the past while having some good whisky....

Issue #299, opened 2015-07-11
Disallow calling abstract methods


dwhswenson:
We have a few classes (`PathMover`, `Ensemble`, `VolumeEnsemble`, and I'm sure others) which, if we were implementing this in C++, would be abstract classes. Core Python doesn't have the concept of abstract classes; any class can be instantiated.

Currently, we do some things that aren't very correct; namely, returning reasonable values for abstract methods that need to be overridden by subclasses. If a user were to instantiate an `Ensemble` (instead of a subclass), they would get something which might be called the "infinite-length trajectory ensemble": `__call__` always returns false, while `can_append` and `can_prepend` always return true.

I suspect that if a user instantiates `Ensemble`, it was probably an accident. I think it is extremely unlikely that the user wants this infinite-length trajectory ensemble. Since this is almost certainly user error, I think it makes sense to raise an error. In addition, raising errors will make it less likely for developers to forget to implement some function (e.g., imagine what would happen if a dev implemented an ensemble, but only overrode `call` and `can_append`, forgetting `can_prepend`: code would run without error, and without careful testing, the naive developer might put bad code into production!)

We should do one of the following:

* abstract methods return `NotImplementedError`
* use the `abc` (abstract base class) module to forbid instantiation of abstract classes.

I think either option is reasonable. Using `NotImplementedError` would be easier to do (using `abc` will definitely require some modifications of the tests, and I'm not sure what effect it might have on storage). On the other hand, using `abc` makes easier for developers to catch unimplemented functions. Note that even if we use `abc`, some aspects should probably still raise `NotImplementedError` (e.g., `Ensemble.oom_matrix`, which isn't *required* in order to do path sampling, but may be required for certain analyses). 

Thoughts?

jhprinz:
In general a must to avoid stupid mistakes. I am not sure what I prefer, although I assume that in the derived classes we use some functionality of the abstract base class. Just adding `NotImplementedError`  will break some of the functionality and I assume that `abc` can cope with this more elegantly. In the worst case we have to reimplement a canonical behavior in all derived classes. 

If I read the documentatin from `abc` correctly, you cannot call super to run initialization code from abstract classes. So, we need to make sure that there is no problem.

Concerning the OPSObject this might be a problem since we now require all base classes to be subclasses of OPSObject and all of these need to call super initialization. This would mean that the 
`NotImplemented` approach is easier to deal with given the current storage. 

[...]

PR #302, opened 2015-08-24
[WIP] Add abstract base classes

jhprinz:
This implements the discussion about abstract base classes. See Issue #299.


jhprinz:
Ready for review and merge. Will be just some little tweaks.

dwhswenson:
1. I'm not sure about the usage here. It seems weird to `raise` a function. I mean, I know the function gets evaluated and returns the exception which is then raised. But wouldn't it be cleaner to either (a) use a function `self.raise_abstract_class_error()` or (b) subclass `NotImplementedError` -- you could even subclass it so that it took `self` as argument and generated the error message you have from it, i.e., calling as `raise OPSAbstractClassError(self)`.

2. Would it be possible to change the `AbstractBaseClassError` so that the suggested classes exclude any abstract subclasses? For example, you could have:

```python
class AbstractClass(OPSNamed):
    def foo(self):
        raise self.AbstractBaseClassError()

class AbstractSubclass(AbstractClass):
    pass

class ConcreteSubClass(AbstractClass):
   def foo(self):
       print "ConcreteSubClass!"

class ConcreteSubSubClass(AbstractSubClass):
    def foo(self):
        print "ConcreteSubSubClass!"
```

Then, I think the current implementation gives something like:

```
>>> bar = AbstractBaseClass()
>>> bar.foo()
Abstract base class AbstractBaseClass cannot be called, please use derived class like [AbstractSubClass, ConcreteSubclass]
```

If we're going to suggest classes, it would be nice if we didn't suggest other abstract subclasses. (But if this would be difficult to implement, we could either remove the suggestion or it could even be merged as is.)

jhprinz:
1. I like the idea of b) since then it is really an Exception class and it looks like it always looks in Python. Only thing is that one needs to pass the class to get the information about subclasses.

2. You are right and I think we have some cases where this happens. We need some way to mark classes to make this work. We can either use a decorator for abstract classes and/or add a wrapper for methods
```
@abstractclass
class MyAbstractClass(OPSObject):
    @abstractmethod
    def my function(self, arg1, arg2, ...):
        """
        There is a docstring here that matters!
        """
```

the abstractclass marks a class as abstract (somehow) so that the abstractmethod can detect which subclasses are not abstract. This is probably the nicest way to do it.
Last question would be how to mark a class as abstract. All class member variable pass on so it will be difficult to detect these, but we could add a "mixin" class and this is easy to detect...

dwhswenson:
1. Good point... but no reason not to do `raise OPSAbstractClassError(object=self)`, and extract the information from there, right?

2. If you're going to mark classes as abstract, use Python's `abc` module instead of reinventing it. You can still throw a custom exception to identify the concrete subclasses.

jhprinz:
2. I just re-read the abc implementation and it seems it does what we want -- I though it would not allow code in abstract classes to be used, but it does.

jhprinz:
Almost done. Turns out this is good for testing if we correctly subclass.

I was thinking about changing the standard error when you try to create an abstract instance, which looks like
```
TypeError: Can't instantiate abstract class xyz with abstract methods abstract_method1, ...
```
This is not really informative, but it turnes out this is not really changeable since it uses some internal cython abstract stuff which depends on a hidden variable that lists the abstract methods that need to be implemented. If this list is not empty you cannot create the class. 

This mechanism is hardcoded into python and all `abc` does is setting up the list at class creation. There is nowhere the text `Can't instantiate abstract class...` be found and hence I assume that besides recompiling your own python version no way of changing it.

Good news is, it works as expected. I had to change some tests that depended on abstract classes though.

jhprinz:
Will be replaced by PR #375.

PR #304, opened 2015-09-03
Support for PyEmma Featurizer

jhprinz:
This is add support for usage of the PyEmma featurizer. This complements support for mdtraj and msmbuilder featurizer. All of these can be stored (using a little trick) since storing code is general a little tricky.

This PR was build on some of the other PRs before because it needed the better storage and CV support that is coming.


jhprinz:
This has to be merged after #300 .

jhprinz:
This is also ready to review, but it still requires pyemma to be installed. Do we wanna do that or should I rewrite some code to make it work without testing pyemma?

jhprinz:
I remove the import of pyemma from the top and put it into the PyEmma_CV function. As long as we are not testing this should be mergeable...

dwhswenson:
So, the following is how I would think we want to make PyEmma "optional." I assume it is close to what you've implemented.

* If the user doesn't have PyEmma installed, OPS works fine (as long as you don't try to use PyEmma CVs, of course!)
* If the user does have PyEmma installed, the PyEmma CVs work without any effort beyond installing PyEmma.
* The default conda install puts PyEmma in the requirements. (I think this is worth doing.) This also means that the Travis environment has PyEmma, and includes tests for the PyEmma CVs.

Is that what you're thinking, too?

jhprinz:
> If the user doesn't have PyEmma installed, OPS works fine (as long as you don't try to use PyEmma CVs, of course!)
Yes.

> If the user does have PyEmma installed, the PyEmma CVs work without any effort beyond installing PyEmma.
Yes.

> The default conda install puts PyEmma in the requirements. (I think this is worth doing.) This also means that the Travis environment has PyEmma, and includes tests for the PyEmma CVs.
No sure, but we can. This would also mean using `conda install openpathsampling` will install PyEmma, too. If we want that (and same should be true for msmbuilder) we should include it and make a test. Otherwise, just provide the support, but no installation.

dwhswenson:
> The default conda install puts PyEmma in the requirements.

I keep going back and forth on this one. I definitely think it needs to be included in the Travis tests, but that can be done by installing it separately. That's probably the best approach for now: change the `.travis.yml` to include installation of PyEmma (and other things that are not required, but are supported). Since it doesn't matter whether the user installs PyEmma before of after OPS (the linking happens at runtime), it might be best to make the user install it separately.

 (Can't tell what you're doing now because this PR also shows all the changes from the big storage overhaul, so I won't be able to really review this until that is merged.)

jhprinz:
I am confused why the tree branch ended up in here, too... I cannot recall merging this one in.

jhprinz:
So we wait until #162 is resolved and then I will fix this. With the new storage this is only a few more lines of code.

jhprinz:
I think there is an issue installing PyEmma on travis. Not sure what is happening.

dwhswenson:
How is the `import` statement giving you different output (`diff` return), instead of an error, if there is a problem? I would expect a problem with installation would raise an exception and cause an error.

Maybe run that notebook verbose?

franknoe:
Please check with @marscher

Am 19/11/15 um 19:38 schrieb Jan-Hendrik Prinz:
>
> I think there is an issue installing PyEmma on travis. Not sure what 
> is happening.
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/choderalab/openpathsampling/pull/304#issuecomment-158150263>.
>


-- 

----------------------------------------------
Prof. Dr. Frank Noe
Head of Computational Molecular Biology group
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de
Mail:  Arnimallee 6, 14195 Berlin, Germany
----------------------------------------------

marscher:
Massive PR :+1: 

Verbose testing output on Travis would be a nice standard behaviour. So it seems that pyemma is installed properly by conda-build. 

You can just add a dummy import statement in travis, right before invoking the notebook tests to see what is going on (I'd guess you clean this branch up anyway before merging).

``` 
python -c "import pyemma"
 ```

marscher:
emma imports succeeded on the copy of this branch: https://travis-ci.org/marscher/openpathsampling/builds/92180454#L1134 (coordinates package is being imported implicitly)

jhprinz:
Let's see if this works now...

jhprinz:
Alright now the pyemma test passed although I have no idea why the `import pyemma.coordinates` produces a diff. I assume that the call send some internal (not displayed) logging messages which are interpreted as more messages in the generated notebook compared to the stored one. This would result in a diff due to different number of messages returned.

The last cell did not pass was my mistake. The computed CVs are different everytime and need to be treated as such. 

I reverted the order of notebooks to be tested back and once this passes tests it should be ready.

jhprinz:
The tests keep failing due to statistics I presume...

jhprinz:
ready for review and merge...

It fixes a few glitches for advanced CV storing and adds pyemma support with a integration test notebook.

jhprinz:
PyEmma has some nice features to easily get a bunch of useful CVs from a snapshot in a very compact form.

dwhswenson:
PyEmma was branched off of some early stuff from the tree branch. I'm holding off on this until we finish #295.

jhprinz:
I removed that (I hope). So this false merge should not be present anymore. Besides notebook updates this has only code changes in `chaindict.py` and `collectivevariable.py`

dwhswenson:
Part of my concern is that it isn't clear to me that the undoing of those commits here won't also undo things in the "tree" PR, messing everything up. Let's finish that one first, then merge it into here, to make sure that everything still works.

Git sees that you made commit A as part of this PR. You also made commit B, which undid A. Another branch contains C, which builds off of (and requires) A. When you merge it all together, what do you get? A, B (undoing A), then C (which needs A)? I think you can fix this, but it requires rebasing (and honestly, rebasing gives me a headache).

The net changes you see here aren't what is important in this case. The whole story includes "undo what I did in A", which means that when you merge the other branch, it will try to undo it what was done in the common commits. Let's play it safe by finishing #295, then rebasing if need be.

jhprinz:
You are right, this is why in the other cases I made a new 2nd PR that redid the changes.

jhprinz:
ready for review and merge once it passes. 
It fixes a few glitches in storing MSMBuilder Featurizers and adds support for pyemma as a featurizer.

dwhswenson:
Following on Skype discussion the other day, I'll merge this and we'll hope it doesn't cause headaches because of the include/undo parts of the "tree" PR.

PR #309, opened 2015-09-10
Add iterables support for samplesets

jhprinz:
This makes it possible to write the following in `SampleSet`s.

```
sset = st.samplesset[2] # load a sampleset for demonstration
sset[0:3] # return a list of all samples with 0 <= rep_id < 3
sset[[1,3,4]] # return samples with rep_ids 1, 3 and 4
sset[[ens1, ens2]] # return (first) samples in ensemble1, 2 
```

in iterables all mixtures are allowed
```
sset[[1,3, ens1, ens2]]
```
Iterables or Slice will return a generator expression to keep memory consumption small.


jhprinz:
this can be reviewed or better commented. I though this is a useful feature, but that depends.

dwhswenson:
I'm not so sure on this one.

* Replica IDs are arbitrary, aren't they? I didn't think they had to be restricted to integers, but maybe you've hard-coded that in storage. In general, they're just some way of identifying the replica so we can follow it when analyzing sampling properties. Note that the docstring ***explicitly*** says "any type is allowed as a replica ID except Sample or SampleSet". If this isn't true, update the docstring.
* I don't understand why you would need this behavior. Saying "I want this one replica" makes sense, but I can't think of any context in which you would want to think about replicas in terms of ranges or slices.
* This will slow down retrieval. I'm not sure by how much (because I don't know whether this is the bottleneck), but you should test that.

If it has no effect, then basically it doesn't hurt anything, so we can merge it (assuming we've decided that we're throwing out the idea of using non-integer replica IDs; I don't much care about that, although it might be nice to mark a replica with the name of the file the initial path came from). But I don't think the benefits of this feature (I can't imagine a use case where this helps) will outweigh any real cost to speed.

I think that last part, which is my biggest concern, can be dealt with by re-ordering your statements. If-else blocks should be in order of the likelihood they will be used. When getting a sample from a sample set, I suspect the order of likelihood would be:

1. `sset[ensemble]`
2. `sset[replica]`
3. anything else.

You have "anything else" as the most important case.

jhprinz:
I thought Replica IDs would be an int and that is how it is hard-coded in the storage. It can be changed of course, but int are very fast to compare. What else would you use as a variable type? Would a string be sufficient? We should check that.

Range or slices just came for free. It seemed nice to say: "Give me the sample with repID 1 and 5" instead of writing give me repID 1 and give me repID 5 one by one. Mostly to make code more readable.
```
[sampleset[replica] for replica in rep_list]
```
is now
```
sampleset[rep_list]
```
since rep_list can be anything that is iterable also slices and ranges are allowed. Where this made sense was for comparing two sample sets in tests like 
```
sset1[range(0:10)] == sset2[range(0:10)]
```
since iterating over replica ids (assuming they are ints) might not give them in the same order every time since dicts have no inherent order.

I think I had this problem once in a test case and found it useful to have.

Timing problems are minimal and I prefer this convenience over some performance and you are right, we should move the iterables part to the bottom. Good point.

jhprinz:
I was wrong about the slice. This did not come for free, but considering again ints as replica ids it might be useful in some cases like initializing or setting up first samplesets.

jhprinz:
By changing the order there is no performance change compared to the old code. Using ensembles or replica_ids (int) there is no change. I agree that before rep_id could be anything and if that is the case we will need to check for iterables and this causes a slowdown for rep_ids compared to before.

dwhswenson:
Are we dropping the ability to use non-integer replica IDs? I'm fine with that, but if so, please change the docstrings to correctly reflect the code.

jhprinz:
I would say yes, since we did so _silently_ so far. At least I set storage to use ints and we never realized.
Negative ints are allowed though, which was nice for the minus interface.

dwhswenson:
I'm fine with that. I've been trying to think of situations where a replica ID has semantic value, and there really aren't many. I suppose it could be the file name of the initial trajectory; but I think the user can just as easily keep a dictionary of filename-to-integer if they really need to track that.

Please correct the docstrings to reflect what the correct state of the code is, and this will be ready to merge.

jhprinz:
If this passes it is ready to be merged.

dwhswenson:
The docstring change I'm asking for is to remove the sentence: `Any type is allowed as a replica ID except Sample or Ensemble.` (line 22-23) That sentence is not true.

dwhswenson:
(Returning assignment to @jhprinz : see previous comment for desired change)

jhprinz:
Done.

dwhswenson:
Only took about 5 re-runs, but it finally passed tests. (I think we need to make the MISTIS run a little longer.) Merging!

PR #317, opened 2015-10-01
Clearer error message for sanity check

### LINKED ISSUE #315: Add better error message for `SampleSet.sanity_check`
If the assertion (that each trajectory is in the ensemble it is associated with) in `SampleSet.sanity_check` fails, the error message should state which ensemble failed. Currently, you can find this information in the log if you have info-level logging enabled, but if the exception is raised, it should give more info.

This just involves wrapping the current `assert` in a `try: except:` where the `except` clause rewrites the error message and reraises. I'll do this soon in a small PR.

(This came about because Abel, @bolhuis's undergrad who is playing with OPS, came across this error and didn't know how what to do about it. When I get an email from him, I take it as an opportunity to improve our user experience!)

dwhswenson:
Resolves #315. Ready for review and merge as soon as it passes tests.

### DIFF
--- a/openpathsampling/sample.py
+++ b/openpathsampling/sample.py
@@ -192,7 +192,17 @@ class SampleSet(OPSNamed):
             #assert(sample.valid)
             logger.info("Checking sanity of "+repr(sample.ensemble)+
                         " with "+str(sample.trajectory))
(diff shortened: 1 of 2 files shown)


jhprinz:
Very good.

PR #318, opened 2015-10-05
Jupyter support

jhprinz:
Add support for Jupyter / IPython 4.0.0 (ala "the big split") and add a few little improvements to the ipynbtester...


jhprinz:
Since it passes tests this should be ready to review and merge.

dwhswenson:
I like the fact that this keeps support for both versions (even if only the newer version got properly tested...) Merging!

Issue #294, opened 2015-06-30
`PathMoveAcceptor`s and detailed balance


dwhswenson:
This is a continuation of discussion from #251. Where that left off, I think we'd agreed that we should separate acceptance from generation of trials in our Monte Carlo. Here we should discuss details on the implementation of that.

One thing we'll have to keep coming back to is the question of detailed balance. Maintaining detailed balance MUST by the job of the `PathMoveAcceptor`. New tools implemented as part of `MoveScheme` should make that easier to accomplish.

### Detailed balance

Fundamentally, we need to know the probability of generating the trial state from the original state and vice versa. Call these `P_o_n` and `P_n_o`. At each decision point in the move decision tree, we modify these probabilities. Note that we can mark some level of the move tree as the "official mover", and these probabilities split into a probability of choosing a given "official mover" and the probability of making the change within that "official mover." I will call the probability of choosing a specific "official mover" the "choice probability". I will call the probability of generating a given trial from a given initial condition using that official mover the "trial generation probability". The product of these two gives us `P_o_n` or `P_n_o`.

Since we're usually interested in the ratio of `P_o_n` and `P_n_o`, we can often obtain shortcuts by directly calculating that ratio. However, we should leave other possibilities open as well.

Each path through the move decision tree has a "balance partner". These partners are pairs that can satisfy microscopic reversibility. If a given mover is used to calculate `P_o_n`, then you must use its balance partner to calculate `P_n_o`. Note that in many cases (shooting, replica exchange) a mover is its own balance partner: basically, any time the input ensembles equal the output ensembles. However, in cases like ensemble hopping, a different mover is the balance partner.

In principle, a mover can have more than one balance partner. For now, we restrict it to only one.

### Proposal: where to put the "official movers" split

In `MoveScheme`, I've defined groups of movers based on what we'd normally consider the move type, such as "shooting", "replica exchange", "path reversal", "minus move", etc. I propose that the movers included in these lists be the "official movers" in the sense discussed above.

#### For the root mover

Using the "official mover" split I describe above, the `MoveScheme` automatically determines the choice probability and the balance partners for the root mover.

#### For bundled moves

If creating what I have previously termed a "bundled" move, we again need to define some "official submove" point on the subsequent subtree. My suggestion is that by default each (topmost) submove of the bundled move is to be considered the "official submove". In that case, each submove must be its own balance partner, and we only need to calculate the trial generation probability.

[...]

PR #320, opened 2015-10-06
Organize by ensemble strategy

dwhswenson:
*What this PR does*: Makes it possible to rearrange the “global” structure of the move decision process. Standard behavior is to select a move type, then a specific move within that. Now we can select an ensemble first, then a move within that. We can also switch between these.

*Why we should have this*: Ensemble-first organization is a step toward SRTIS. The ability to conserve move probabilities when switching selection order makes comparison between SRTIS and normal RETIS more straightforward, and makes it easy for users to switch from one approach to the other.

*What is hard about this*: My intuition expects different behaviors between ensemble-first and move group-first organization. Details below, but figuring our that there was a difference, and then figuring our exactly what the difference was took some effort (and a major rewrite after my first draft of this).

Tasks:
- [x] EnsembleDictionaryMover
- [x] OrganizeByEnsembleStrategy
- [x] OrganizeByMoveGroupStrategy (rename of Default)
- [x] Check that we get the correct weights regardless of top-level organization strategy
- [x] Update docstrings

***

### Background 

Before getting into the details of what this PR contains, let me refresh a few essential background points. A few of these came from #289. The big thing is that movers can be uniquely defined by `(groupname, ensemble_signature)`, and that the main thing that comes out of the highest-level organization strategy is the `choice_probabilities`, which is a dictionary of `{PathMover : probability}` . 

### Goals

* code should allow different intuitive behavior for move-type organization and for ensemble organization
* `scheme.choice_probabilities` should be preserved (unless explicitly overridden)
* general schemes should have a reasonable default weights

#### Intuitive behavior differs by global organization type

This is one of the weirder parts of this PR, but it basically boils down to this: when I set the move-type weights in a move-type-first organization scheme, I think of a dictionary like this:

```python
group_weights = {
    ‘shooting’ : 1.0,
    ‘repex’ : 0.5,
    ‘pathreversal’ : 0.5,
    ‘minus’ : 0.2
}
```

The tricky bit is that there may be 5 shooting movers for every minus mover. If the weights represented “probability of choosing this move type”, then we’d do 5 shooting moves for every minus move. But since there are 5 different shooting movers, that would mean each individual shooting mover was used as frequently as the minus mover.

That’s not what I want. I want the minus move to be done less frequently than shooting: about 1/5 as often. So rather than make the user do the math to get that to work (as my old code did), I make it so that the code figures it out automatically. This means that (assuming all ensembles have the same weight) *each* shooting move will be 5 times as likely to occur as a minus move. That’s what we want.

On the other hand, ensemble-first organization does just what you’d expect: first you pick an ensemble with the random probability given by `ensemble_weights`, then you pick a move which has that ensemble as one of its initial ensembles.

#### Ensemble-first organization does not have a unique solution

The quantity that is conserved is the total choice probability for each ensemble. However, some ensembles (e.g., replica exchange) have more than one input ensemble. This means that they show up in more than one place in the ensemble-first decision process. There is not a unique way to split these weights, so the approach I’ve used is to say that each ensemble contributes the same to the *total* probability of choosing that mover. Details on this problem, and the approach I used to solve it, are in [this gist](https://gist.github.com/dwhswenson/5d5b18ba8e811cbe21da).

#### Reasonable defaults

On one hand, it would make sense to normalize most of these results: so the probabilities should be normalized probabilities. On the other hand, that’s harder as a way to think about it -- and obviously doesn’t work with the `group_weights` described above.

So the approach I’ve taken is the following:

1. `choice_probability` really is a normalized probability
2. `group_weights` is rescaled such that `’shooting’` is 1.0 (we usually think in terms of shooting movers); if there is no group called `’shooting’`, then we rescale so that the most common value in the `group_weights` dictionary is 1.0.
3. `mover_weights` are rescaled so that the most common number within each second-level decision is 1.0.

These are so that, if you say “I want this twice as much shooting in `ensembleA` as `ensembleB`,” you can just change the default 1.0 to 2.0.

### Caveats when overriding

There are two levels of weights to override -- the weight that is sorted by (either move group or input ensemble) and the weight within that sorting group.

For the sorting groups, overriding the weights are easy: the input dictionaries take the group name or the ensemble and map it to a weight (`group_weight` or `ensemble_weight`).

For the second level, this is a little more complicated. First, note that these weights should only be used by pretty advanced users -- typical needs are met without them. These weights are referred to as `mover_weights`, and the shape of the dictionary keys is different between the two types of organization strategy. For `OrganizeByMoveGroupStrategy`, the keys are `(groupname, ensemble_signature)`; for `OrganizeByEnsembleStrategy`, they are `(groupname, ensemble_signature, ensemble)`.

In both cases, the `mover_weights` default to 1.0, as does `ensemble_weights`. However, `group_weights` defaults to the version shown above. This means that to get the standard default behavior with ensemble-first organization, your scheme needs an `OrganizeByMoveGroupStrategy`, followed by an `OrganizeByEnsembleStrategy`. (This will eventually be hidden to SRTIS users, who will just ask for an SRTISScheme that handles all of that automatically.)


dwhswenson:
Finally! I'm calling this one ready for review and merge. I've only been working on this since, oh, June.

openpathsampling/pathmover.py
```diff
@@ -384,6 +384,8 @@ def metropolis(cls, trials):
         accepted = True
         probability = 1.0
 
+        # TODO: This isn't right. `bias` should be associated with the 
+        # change; not with each individual sample. ~~~DWHS
```
jhprinz: What would you call this then? The underlying idea was that since we accept samples we check, if all samples can be accepted and otherwise we multiply the `bias` which I think is something like the forward proposal probability.

openpathsampling/analysis/move_strategy.py
```diff
@@ -361,182 +398,235 @@ class DefaultStrategy(MoveStrategy):
...
-                movers = scheme.movers[group]
-                group_weights[group] = sum([scheme.choice_probability[m] 
-                                            for m in movers])
-            try:
-                normalizer = group_weights['shooting']
-            except KeyError:
-                normalizer = sum(group_weights.values())
-            for group in group_weights:
-                group_weights[group] /= normalizer
+                    sortkey_w[skey] = 1.0
```
jhprinz: I found a nice trick to write this in one line.
```
sortkey_w[skey] = self.default_goup_weights.get(key, 1.0)
```
use the `.get(key, default)` method

jhprinz:
Hmm, my post disappeared... Anyway just a quick question to confirm.
Choosing a different organization will completely rearrange the move tree and have first a random choice mover that selects the ensemble or the move group and then have a next set of random choice below that. So the move tree will look very different from what it used to if you choose ensemble first, right?

Anyway, I think this is ready to be merged, yes?

openpathsampling/pathmover.py
dwhswenson: This should not be a method. That's my point. It breaks the ability to maintain detailed balance for arbitrary moves. The overall acceptance is not of a sample. It is of an entire active sample set. In most moves, most of the sample set doesn't change, so this doesn't matter. However, the current structure of path movers will not work in the general case. See #251, #294.

openpathsampling/analysis/move_strategy.py
dwhswenson: Wish I'd known about that sooner! I use that idiom all the time. Added a commit here to switch this one; over time we can replace the many other times I've used that.

dwhswenson:
> Choosing a different organization will completely rearrange the move tree and have first a random choice mover that selects the ensemble or the move group and then have a next set of random choice below that. So the move tree will look very different from what it used to if you choose ensemble first, right?

Precisely. Single replica TIS needs to be based on a "choose-the-ensemble-first" strategy (because the ensemble is given, so there's no "choice" to be had). This lets us switch between the two approaches.

The important thing is that we maintain the same probabilities for each mover. The original approach was "choose a move type, then choose a mover." From this we can get an absolute probability of choosing each individual mover. The ensemble-first strategy is "choose an ensemble, then choose a mover." Again, we can get the absolute probability of each individual mover. The smart part of this PR is that when we change from one organization strategy to the other, those absolute probabilities remain the same.

> Anyway, I think this is ready to be merged, yes?

As soon as the newest commit passes tests, yes, it should be.

jhprinz:
Great, this is really cool. Looking forward to seeing this in the move tree ...

Merging...

PR #321, opened 2015-10-07
Fixed a couple docstrings

dwhswenson:
The first docstring I looked at was for OneWayShootingMover -- looks like that was actually the only one that @jhprinz hadn't corrected in #278. Anyway, this should be pass tests and be ready for merge.


jhprinz:
merging.

PR #322, opened 2015-10-07
Clean up ipynb tests

dwhswenson:
Lately, tests have been failing due to Travis’s 50-minute time constraint. Some of this might be fixed by changing the caching behavior as done in #300, but even so, our integration tests include redundancies which should be removed.

Most obvious redundancy: we call `sliced_sequential` twice. That’s just silly. In fact, I don’t think we need to run that one at all. If we need a check on that, it should really be in the nosetests: that notebook was intended for documentation, not testing.

The same is true of the `langevin_integrator` notebook. That was me (1) using visual inspection of distributions to check that the integrator was working correctly; and (2) learning how to use IPython notebooks and matplotlib. That test doesn’t actually test anything extra unless you visually inspect the results. I’ll remove it from the tests.

Finally, we have the `toy_bootstrapping`, `toy_tis`, and `toy_analysis` notebooks. Functionally, everything that is done in there can be in the MSTIS and MISTIS examples. We have to run both the MSTIS and MISTIS data-generating runs to test those approaches, but then don’t have to also run the stuff in `toy_tis`. Part of this PR will be ensuring that everything from the toy examples (especially analysis) is also done in the MSTIS example.

- [x] Remove `sliced_sequential` and `langevin_integrator_check`
- [x] Split `multistate_system_setup.ipynb` into `mstis_bootstrap.ipynb` and `mstis.ipynb`
- [x] Reloading trajectories from mstis_bootstrap in mstis; test `map_trajectory_to_ensembles`, and `translate_ensembles`
- [x] Create `PathSampling.run_until(stepnum)` which runs until the given number of steps are stored
- [x] Use `PathSampling.run_until` for test scripts
- [x] Add move scheme checks to `mstis_analysis`
- [x] Add `PathTree` visualization to `mstis_analysis`
- [x] Add visualization of individual trajectories
- [x] Add path length plot to `mstis_analysis`
- [x] Add crossing probability plots to `mstis_analysis`
- [x] Add direct plotting of max lambda to `mstis_analysis`
- [x] Add TCP result plots to `mstis_analysis`
- [x] Add lists with the fluxes for each state in `mstis_analysis`; CTP for each transition
- [x] Fix `repex_networks.ipynb` to work with new input data
- [x] Remove testing calls of `toy_bootstrapping.ipynb`, `toy_tis.ipynb`, `toy_analysis.ipynb`
- [x] Add text to `mstis_bootstrapping` and `mstis` to explain what is going on

***

Highlighting a few specific points:
* `PathSampling.run_until(stepnum)`: We need ~200-300 MC steps for analysis to work. Sometimes it takes more than 10 minutes to run 300 steps. So we have 250, and restart. Then the next run, we get another 250 (for 500 total), and restart -- we could have stopped at 300, but instead we try to get 300 within 10 minutes each time. This new function figures out how much is already in storage, then calls `run` for `stepnum - n_steps_in_storage` steps.
* `mstis_analysis` will be the main analysis showcase, showing all the tricks that were previously in the `toy_analysis` scripts. The MSTIS runs will also show how to do bootstrapping in a separate file from the main run, to illustrate how to load samples from a file. The combination of these facts allow us to remove the toy_bootstrapping, toy_tis, and toy_analysis scripts.

Also, I re-ordered the tests a bit so that the things which I think are currently most likely to fail come earlier (mainly, that means I moved `alanine.ipynb` much later in the process, because I think the stuff in there is a lot more stable now than in the mstis/mistis.) If the test is going to fail, we might as well make it fail quickly. The rest of the test notebooks still run, but we as developers can see a problem sooner (if we’re watching Travis).


jhprinz:
Thanks for taking care of this. It is really a waste of time to run all of these the current way. Let me know if I can help with e.g. parts of the visualization.

dwhswenson:
Oh hey, [look at that](https://travis-ci.org/choderalab/openpathsampling/builds/85191117)! Travis tests ran in less than 24 minutes. I haven't seen ***that*** in a while. :-)

Over the next day I'll add some documentation to the MSTIS test scripts, then this will be ready to merge.

jhprinz:
Yay, great...

I was wondering if we could merge this before storage stuff in #300 , since it makes testing so much faster!

dwhswenson:
This is ready for review and merge.

In the future, I'll add more to the descriptions of what is going on in the tests (which will also be examples in documentation), but at this point those descriptions have parity with the old `toy_*` examples. 

I'm particularly pleased with the updates to the [`mstis_analysis.ipynb` notebook](https://github.com/dwhswenson/openpathsampling/blob/ipynb_tests/examples/ipython/mstis_analysis.ipynb). It needs more explanation, but it really shows a lot of what can be done in terms of analysis.

jhprinz:
Lgtm, will merge. A few comments to keep in mind.

You made some additions to the live_vis and for a talk I had also some local modifications that could be interesting. Remind me that I will post an issue about that.
I like the `run_until` idea, but I am not sure if it should stay with the engine, since it is related to the storage and so it would be clearer if it is a function of the number of steps and storage. We can change that later. For now that works fine.

dwhswenson:
> You made some additions to the live_vis and for a talk I had also some local modifications that could be interesting. Remind me that I will post an issue about that.

I was thinking that some styling options would be nice -- I didn't take the time to add that in this PR, but it could be useful. Some way to set with width of accepted/rejected trajectories, or maybe an option to do the color scheme based on something other than the ensemble. (The details of the color scheme can already by changed.

The modifications in here were just to make it so that I could plot an arbitrary list of samples, instead of requiring that it plot a `step`. Previously, the `ToyPlot` object was used to plot trajectories from storage. Using live_vis makes it reusable on systems other than the 2D toy models.

> I like the `run_until` idea, but I am not sure if it should stay with the engine, since it is related to the storage and so it would be clearer if it is a function of the number of steps and storage.

I don't understand this argument. First, `run_until` is not with the engine, but with the `PathSampling` object. `PathSampling.run_until` is a replacement for `PathSampling.run`, in cases that you've already run some number of steps. Why would it be `Storage.run_until`? If there is no `pathsampling.storage`, or if no steps are stored, then `run_until` should behave exactly as `PathSampling.run`.

jhprinz:
Comment on my `run_until` comment: It is a function of a PathSampling and thus forget the complain about engine and storage. But would it be useful to have two numbers like run 100 steps now until a maximum + stored ones is like 300 steps? It seems that now it runs 300 and next time it does run none.
Is that correct?

jhprinz:
Sorry, realized that before you posted...

dwhswenson:
Ha -- good timing. Just to be clear:

Let's say `calc = PathSampling(...)`. In that case `calc.run(300)` will run 300 steps, no matter how many are in storage. On the other hand `calc.run_until(300)` will run until the step count reaches 300 (which typically means there are a total of 300 steps in storage). If the step count is already 300, then it will not run any steps.

So:

```python
calc1 = paths.PathSampling(storage1,...)
calc1.run(50) # 50 steps in storage1
calc1.run(50) # 50+50 = 100 steps in storage1


calc2 = paths.PathSampling(storage2,...)
calc2.run_until(50) # 50 steps in storage2
calc2.run_until(50) # no new steps, still 50 steps in storage2
calc2.run_until(100) # 100 steps in storage2
```

Instead of having one function with two inputs, we have two functions. This is more straightforward, because otherwise the parameters are described as "number of steps, unless you reach the maximum number of steps in storage, which overrides this one" and "maximum number of steps in storage." In this case, in one function the input parameter is just "number of steps" and the other function the input parameter is "maximum number of steps in storage."

jhprinz:
Thanks, that explained it. I did not think of using several commands of `run_until` which does what I want.

dwhswenson:
Ooh, I just realized (and maybe this is what you're doing) but this could be a really good way to make the notebook useful for both testing and preparing real analysis-ready results. I can't remember exactly what your notation was to skip cells in `ipynbtest.py`, but we can do something like:

```python
mstis.run_until(300) # for testing
#### cell break
#!skip
mstis.run_until(5000) # for interactive users
```

jhprinz:
Yes, that should work. `#! skip` (whitespaces do not matter) will not execute the cell at all.

Good point. Maybe we should move the ipynbtest into a separate project and include it so we can better make changes? We also thought about adding the possibility to set global variables from the command line and maybe have a package that allows to add more output manually like
```
import ipynbtest as ipyt

if version < 1.0:
    ipyt.write('Version too small, will skip this test')
    ipyt.fail('Wrong version')
```

Don't know, there might lot's of useful or not useful stuff...

PR #324, opened 2015-10-09
Travis debugging

### LINKED ISSUE #323: Travis problems with ipynbs
Travis has just started acting weird with ipynbs. I'm having this problem in #322; see for example:

* https://travis-ci.org/choderalab/openpathsampling/builds/84549142

(Also confirmed that I get the same problem in other PRs, too)

I've written a very simple test in that PR, [`stupid_test.ipynb`](https://github.com/dwhswenson/openpathsampling/blob/ipynb_tests/examples/ipython/stupid_test.ipynb), the entire code of which is:

```python
%matplotlib inline

cv_x = paths.CV_Function("cv_x", lambda x : x.xyz[0][0])
volume = paths.CVRangeVolume(cv_x, 0.0, 1.0)

print volume.name

from toy_plot_helpers import ToyPlot
plot = ToyPlot()

plot.plot()
```

This runs with no problem on my machine, either in a notebook or using `ipynbtest.py`. However, Travis fails `print volume.name` with `name 'volume' is not defined` and fails `plot.plot()` with `name 'plot' is not defined`. Looks to me like the installation isn't working correctly. I don't know what changed to break that.

Nosetests all run fine. It's only in the ipynbtest stuff that we have problems. Is it possible that there's a scope issue in ipynbtest, or that it is somehow using the wrong python? This seems to have just cropped up today, so it's probably related to a change in Travis or one of our requisites. Any ideas?

dwhswenson:
To see if I can figure out what's going on with Travis. Resolves #323.


dwhswenson:
OK... almost there. The problem (or at least, ***a*** problem) is that Conda has upgraded to numpy 1.10, while Travis still has numpy 1.9. For some reason, within the ipython notebook tests, Travis's `site-packages` directory comes before the conda directories. This only happens when running inside the IPython kernel; if we just call normal Python, this isn't a problem -- in fact, the Travis directories aren't even in `sys.path`. So something is screwing up `sys.path` when running in the IPython kernel.

dwhswenson:
The problem seems to be a weird conflict between Travis's virtualenv, installating conda within that, and some `sys.path` rewriting that IPython does when it finds itself in a virtualenv. The workaround is to `deactivate` the virtualenv immediately on starting the installation.

I put more details in https://github.com/ipython/ipython/issues/8898.

Once again, we're passing Travis, so this is ready to merge!

dwhswenson:
Huh... Something seems to have changed in conda; now it is giving us numpy 1.9 (over the weekend, it was giving 1.10).

I still think we should merge this. Otherwise we're stuck keeping the same version of numpy as Travis does, no matter what. This PR fixes that. Besides, I think we should try to support the most recent version of numpy, which this should allow.

jhprinz:
Agreed. Will merge and we go from there...

PR #325, opened 2015-10-12
allow external defined classes to be stored

jhprinz:
This adds a quick fix to update the internal list of storable objects. This will be the same in the new storage.

You can store locally created classes as well provided they are subclassed from `StorableObject` which is the case for all built-in storable OPS objects.

The list is created when a storage file is opened or created. If you define your class after opening a storage file you can call
```
storage.updated_storable_classes()
```
to update the list manually.

Note that classes are stored by class name and so no class names can be the same. Not sure what trouble this will cause. We sould add some testing if a user by accident creates classes that already exist.

### DIFF
--- a/openpathsampling/storage/object_storage.py
+++ b/openpathsampling/storage/object_storage.py
@@ -123,7 +123,6 @@ class ObjectStore(object):
         self.has_uid = has_uid
         self.has_name = has_name
         self.json = json
-        self.simplifier = paths.storage.StorableObjectJSON(storage)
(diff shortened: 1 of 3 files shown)


jhprinz:
This is a fix to one of the problems that Abel had.

dwhswenson:
This passes tests (took a couple re-runs to make it happen). Is it ready to merge?

jhprinz:
Yes, should be ready to merge. It is a quick fix, but should do what we want. #300 will make similar changes.

PR #329, opened 2015-10-16
EnsembleHopStrategy and ReplicaExchangeStrategy

dwhswenson:
Adds the ability to switch between EnsembleHop and ReplicaExchange implementations of ensemble changing. This is a big step toward Single Replica TIS.

- [x] EnsembleHopStrategy
- [x] ReplicaExchangeStrategy
- [x] Check that we can switch back and forth

***

The approach here is to use the same "group" (repex) for both ensemble hop and replica exchange (although this can be customized in the strategy if the user really wants to). This is where some of the multiple-level stuff about move strategies (see #285). For example, a user can define a certain specific replica exchanges as being important using a `SelectedPairsRepExStrategy`, and then convert those to `EnsembleHop`s using the `EnsembleHopStrategy` -- and, of course, they can be switched back again by the `ReplicaExchangeStrategy`. It doesn't matter the order in which the user adds these strategies to the scheme doesn't matter; the `strategy.level` idea organizes them correctly.

### DIFF
--- a/openpathsampling/analysis/move_scheme.py
+++ b/openpathsampling/analysis/move_scheme.py
@@ -149,6 +149,10 @@ class MoveScheme(OPSNamed):
                             self.movers[group][idx] = mover
                     else:
                         self.movers[group].append(mover)
+        elif strategy.replace_group:
+            if strategy.from_group is not None:
+                self.movers.pop(strategy.from_group)
+            self.movers[group] = movers
         else:
             try:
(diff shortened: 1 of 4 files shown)


dwhswenson:
This should be ready for review and merge.

jhprinz:
As far as I can see through the complexity of the move_scheme approach this looks good.

jhprinz:
merging.

PR #334, opened 2015-10-22
Fix Visualization

jhprinz:
This should fix problem with failures when using PathTree visualizations


dwhswenson:
Re-ran Travis on the above; still fails in the same way. But if this is just cherry-picks from #295, I think that one is close enough to ready to merge that it is easier to fix the small things over there and just merge it. (If so, we can close this PR.)

jhprinz:
Seems to pass now (in my fork it did). Needed to rename binstar to anaconda in the install script. For some reason it just stopped working.

jhprinz:
All this PR adds is to update the PathTreeVisualization so that it will not fail anymore.

dwhswenson:
Well, now that it passes I'll merge this one!

PR #340, opened 2015-10-26
Engine improvement

jhprinz:
Little updates to the engine...

It adds two function that run the engine foward or backward in a given ensemble.

Since this is what we probably only (99%)  use an engine for it might be nice not always have to write
'.can_append'

It also adds `.prepend` for Trajectory objects to be consistand with 
```
can_append -> trajectory.append(...)
can_prepend -> trajectory.prepend(...)
```


dwhswenson:
Funny... I was cleaning up old branches yesterday, and came across a branch where I had started to implement something along these lines. Decided not to make a PR of it, because I couldn't remember how complete it was (since it consisted of about 2 commits from April).

Is this ready for review?

jhprinz:
Ready for review...

jhprinz:
Even better. Have a look and let me know what you think. Nothing big. Just some convenience.

dwhswenson:
Context comments and overall stuff.

This is based on the stuff discussed in a few comments, starting with https://github.com/choderalab/openpathsampling/pull/300#issuecomment-151092995. There are some differences between what was said in those comments and what is implemented here. I like what is implemented here more.

Specifically:

> More ideas:
> 
> * Replace running = with ensemble = in engine.generate and have two functions generate_forward and generate_backward that will call can_append or can_prepend resp. on the ensemble.
> * in the above case I would have the general running = renamed into generate_until or so.

I don't like that idea. We've discussed separability and reusability of some parts of the code. I think the `dynamics_engine` subpackage could be really nice; it would be good to keep that code independent of the ensembles. The current version just takes a list of functions that take the current trajectory and must all return `True` to keep running. That is very reusable.

(On the other hand, the `generate_forward` and `generate_backward` parts are reasonable.)

> A last idea that just came to me is that we could parse the docstrings of an engine for the possible options. So you have to only update the docstring and then the options match. (Although better generate it the other way round :)

Mixed feelings, but overall it strikes me as "too clever for our own good." I like the idea of the simplicity, and I like the fact that it enforces docs representing code. But adding new engines is something where I hope we'll get a lot of contributions from people who otherwise don't contribute a lot to the code, so I want it to feel like the kind of programming they're already comfortable with.

openpathsampling/dynamics_engine.py
```diff
@@ -296,6 +294,9 @@ def generate(self, snapshot, running=None):
             logger.info("Starting trajectory")
             log_freq = 10 # TODO: set this from a singleton class 
             while stop == False:
+                if self.options.get('n_frames_max', None) is not None :
```
dwhswenson: Do we need to use `get` here? It seems like either we need to include `n_frames_max` as in `default_options`, or we need to have a back-up in case the key isn't in the dictionary. But we don't need both, do we?

dwhswenson:
One possible change: It has always bothered me to use `forward==False` as an equivalent to `backward==True`. I usually prefer to use an integer `direction`: `direction > 0` is forward; `direction < 0` is backward, and `direction == 0` indicates some error.

Up to you whether you want to change that.

Other than that, this looks good. One line note asking for clarification (the code should work, but if the redundant check is necessary, that might indicate something more serious has gone wrong and so maybe it should error instead.) Otherwise, this is just waiting on a decision about https://github.com/jhprinz/msm-tis/pull/6 ;-)

openpathsampling/dynamics_engine.py
jhprinz: Yes, this is a fail safe. Right now all dynamic engines must contain this option and if we keep it that way we can remove since in the constructor it will be checked.

jhprinz:
> I don't like that idea. We've discussed separability and reusability of some parts of the code. I think the dynamics_engine subpackage could be really nice; it would be good to keep that code independent of the ensembles. The current version just takes a list of functions that take the current trajectory and must all return True to keep running. That is very reusable.

Right, I kept the general `.generate` function the way it is and just made two convenience function that at the same time will take care of the of the correct reversal. Is the implemented way actually correct?
I mean the backward will start with the last sample and then simulate forward with reversed momenta. So the resulting trajectory will have the initial snapshot as last frame. And all momenta are reversed. I think we would need to reverse all frames at the and. Because we assume that time runs from first to last frame and in that case we have momenta in the wrong way. Let me check that.

> One possible change: It has always bothered me to use forward==False as an equivalent to backward==True. I usually prefer to use an integer direction: direction > 0 is forward; direction < 0 is backward, and direction == 0 indicates some error.

Interesting. I was wondering what would be a good choice for that and had `direction = 'forward'` first. And realized that flipping is difficult. Best would be a binary class like True/False. I like the +1 / -1 since you can write `backward = -forward`. I think I will pick that.
Can we add somehow
```
    FORWARD = 1
    BACKWARD = -1
```
maybe as some immutable class attribute?

dwhswenson:
> I like the +1 / -1 since you can write backward = -forward. I think I will pick that.

I used to use a signed `int` as the number of steps between saved frames. So I would tell my integrator `run_steps(direction*n_steps)`. I always liked that bit of code.

> I mean the backward will start with the last sample and then simulate forward with reversed momenta. So the resulting trajectory will have the initial snapshot as last frame. And all momenta are reversed. I think we would need to reverse all frames at the and. Because we assume that time runs from first to last frame and in that case we have momenta in the wrong way.

I think what you have written here is correct (except I think your fingers typed "last sample" when your brain said something like "first frame" or "initial snapshot"). If the implementation is wrong, shouldn't that show up in the tests? If not, there should be tests to check for it.

> Can we add somehow `FORWARD = 1; BACKWARD = -1` maybe as some immutable class attribute?

I can't count how many times I've wished that Python had something akin to C's `#define`. As far as I know, there's no proper equivalent. But `if direction == DynamicsEngine.FORWARD` reads well to me (or `self.FORWARD` in the engine, or `engine.FORWARD` if an engine is in scope -- all sound good). So I'd be in support of that.

dwhswenson:
Merging this (I assume it is ready -- it looks like you implemented everything else that was discussed in there!)

PR #335, opened 2015-10-22
Use MoveScheme in the example notebooks

dwhswenson:
So far, `MoveScheme` and `MoveStrategy` have been developed without being used in the main example notebooks. They have solid test coverage, so integrating them should be relatively easy. This PR is pretty small, and just covers the process of switching over to using `DefaultScheme`s. It will be followed by a separate PR to clean up the redundant stuff in `Network`s/`Transition`s.

- [x] Switch MSTIS to DefaultScheme
- [x] Switch MISTIS to DefaultScheme
- [x] Switch MSTIS equilibration to scheme with only shooting
- [x] Check for expected choice probabilities in analysis
- [x] Clear notebook output for easier merging

PR #342, opened 2015-10-26
[WIP] Cleanup: Networks/Transitions without movers

dwhswenson:
Cleanup that was enabled by #335. We no longer need movers or move schemes to be made by networks/transitions. This also means there’s no longer a need for a distinction between `TISTransition`s and `RETISTransition`s, since the only difference was in the move scheme.

- [x] Remove move building from Networks/Transitions
- [x] Join TISTransition and RETISTransition into one object

Later:
* Make separate `transition.py` to build the transition objects separate from other analysis stuff


dwhswenson:
I'll save the file reorg for a later date, since I think that'll cause annoying conflicts with the storage PR.

dwhswenson:
Ready for review and merge. This is mainly some cleanup of the stuff that's no longer needed, since the `MoveScheme` gives better convenience functions for building movers than the idea of having the networks build the movers.

Part of the point was that the only real difference between TIS and RETIS is the choice of move scheme. Therefore these have been made into one TISTransition object.

jhprinz:
lgtm... merging...

Issue #292, opened 2015-06-27
Counting MD steps run in each PMC


dwhswenson:
It would be really nice if each path mover (or at least each path mover that uses an engine) could report the number of MD integration steps it ran. If comparing the efficiency of two methods, I think the best approach is to use the number of MD steps as a measure of the total effort (it assumes overhead from everything else is can be made negligible, but is also independent of the computing platform).

From a user-facing standpoint, this might be best as a property of a PMC. I see it like this:

```python
>>> pmc.n_timesteps_used
{ engine1 : 125 }
```

I suggest a dictionary return value in case multiple engines are used.

Internally, that would probably be something like:

```python
# in PathMoveChange:
@property
def n_timesteps_used(self):
    self.mover.n_timesteps_used(self) # maybe takes self.details?
```

where the implementer of the path mover has implemented the `pathmover.n_timesteps_used(pmc)` to get the number of MD steps taken during a given pmc. 

For PMCs with sequential subchanges, the correct answer is of course to add to any key that already exists:

```python
@property
def n_timesteps_used(self):
    sub_counts = [sub.n_timesteps_used for sub in self.subchanges]
    # this implementation avoids branching
    engines = set(sum(sub_counts.keys(), []))
    step_count = {e : 0 for e in engines} 
    for sub_count in sub_counts:
        for engine in sub_count:
            step_count[engine] += sub_count[engine]
    return step_count
```

(or at least something like that should work)

By making all of this a property, we require no additional storage, and implementers of path movers just need to add one function to get the correct value. It also has the nice side effect that, if a user implements a path mover without implementing this function, and only discovers the mistake *after* running a huge calculation, then (assuming everything needed was stored in the `pmc.details`) the user can add this to the code after the fact and use the existing computational results.

dwhswenson:
This should be part of the information returned during an `EngineMover`'s move. Should be handled in pathmover overhaul.

jhprinz:
This should be very easy now using the `snapshot.engine` property. Do you want to fix this or should I do it. It should be very quick...

dwhswenson:
Why is this easier with `snapshot.engine`? The question here is how many new steps the `PathMover` created. The `PathMover` has to be changed to make this possible, and it already knows its engine.

Already in progress in #394.

jhprinz:
True, this is more complicated...

PR #343, opened 2015-10-26
visualize_frequency and timing hack

### LINKED ISSUE #339: Frequency for Livevisualisation
Hi, I don't know whether this should go here. The new live visualisation function in the mstis notebook looks great. However, there is not much changing at each time step while running. This is probably due to the many replia exchanges and the the rejections. I suggest plotting only with a certain frequency. is it possible to add a frequency variable? e.g something like mstis_calc.live_visualization.background = 100?  This will show only a new trajectory set every 100 moves. What do you think?

dwhswenson:
Resolves #339 

Now you can set the visualization frequency using, for example:

```python
mstis_calc.visualize_frequency = 50 # redraw every 50 steps
```

This also contain a little hack related to an email @bolhuis sent me earlier, asking about computational cost for different kinds of moves. It turns out, it is easier to get the actual time than the number of MD steps used (still desired; see #292). However, this temporary hack lets you get timings from `step.change.details.timing` (for any step except the first, which doesn't have a mover associated with it anyway).

So,

```python
import numpy as np
repex_timings = np.array(
    [s.change.details.timing 
     for s in storage.steps 
     if s.change.canonical.mover in scheme.movers['repex']]
)
print repex_timings.mean(), repex_timings.std()
```

***Aside***: when I tried this out, I noticed that I had a few *very slow* repex moves (half a second?) even though most were less than 0.1 seconds. Might be worth looking into what's going on there. Could just be long trajectories, but I don't know. (RepEx in OPS has to loop all frames in each trajectory to check acceptance, because we don't take short-cuts based on assuming our ensembles are TIS ensembles).

Same works for other cases where mover group definition is the same as canonical mover definition (minus, path reversal). However, for some reason `OneWayShooting` isn't canonical (I think we decided this, but I'd still suggest changing it). You can still extract those timings using something like:

```python
shooting_timings = []
for step in storage.steps:
    for shooter in scheme.movers['shooting']:
        if shooter in step.change:
            shooting_timings.append(step.change.details.timing)
shooting_timings = np.array(shooting_timings)
print shooting_timings.mean(), shooting_timings.std()
```

Note that in the toy models, nearly the entire cost is in the overhead. So you'll find that shooting moves cost the same as replica exchange, because both loop over each frame of the trajectory once. Saving metadata about the trajectory and the ensembles could speed this up in the specific case of TIS, but so far we've focused on keeping everything as generic as possible. The overhead should be roughly constant with system size, so even a half-second per replica exchange isn't much when running dynamics in large systems.

### DIFF
--- a/examples/ipython/mstis.ipynb
+++ b/examples/ipython/mstis.ipynb
@@ -649,7 +649,8 @@
     "background = ToyPlot()\n",
     "background.contour_range = np.arange(-1.5, 1.0, 0.1)\n",
     "background.add_pes(engine.pes)\n",
-    "mstis_calc.live_visualization.background = background.plot()"
+    "mstis_calc.live_visualization.background = background.plot()\n",
+    "#mstis_calc.visualize_frequency = 25 # speeds things up, but isn't as pretty"
    ]
   },
   {
@@ -687,7 +688,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython2",
-   "version": "2.7.9"
+   "version": "2.7.10"
   }
  },
  "nbformat": 4,
--- a/openpathsampling/pathsimulator.py
+++ b/openpathsampling/pathsimulator.py
@@ -351,7 +351,7 @@ class PathSampling(PathSimulator):
         initialization_logging(init_log, self, 
                                ['move_scheme', 'globalstate'])
         self.live_visualization = None
-        self.visualization_frequency = 1
+        self.visualize_frequency = 1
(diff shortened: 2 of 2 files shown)


dwhswenson:
Oh, and this also changes the `run` function so that, if you're using the live visualization, the last frame stays visible. I hadn't done this when I first implemented it because I wanted it to be a surprise when @jhprinz actually ran it for the first time -- no need to keep it secret now!

This is ready for review and merge. As mentioned, I'd like a better place to stick the timings (possibly making them an optional feature?) But it works for now.

jhprinz:
Looks good. We should clean up the timing later, which is a very useful feature...

PR #347, opened 2015-10-27
FullBootstrap

### LINKED ISSUE #346: Move `full_bootstrapping` code into main package
This is the function that takes a snapshot, and runs until the innermost interface is satisfied, and then does the bootstrapping until the given interfaces have trajectories. So it takes a snapshot as input, and gives you a sample set as output.

I think I'll create a `FullBootstrapping` path simulator to do this. At some point, we might combine the `FullBootstrapping` with the `Bootstrapping` (which doesn't include the process of getting the first path from the snapshot). However, for now the `FullBootstrapping` will actually create its own `Bootstrapping` subobject (since that's what the function does.)

It won't be the prettiest, but at least the ugliness will be hidden from the casual user, instead of having it in a notebook.

dwhswenson:
Resolves #346 

- [x] Move `FullBootstrapping` into a class
- [x] Use new class in mstis_bootstrap
- [x] Use new class in mistis_setup
- [x] Add max_ensemble_tries


dwhswenson:
A few more details. The approach we'd be using in tests was to run the bootstrapping for a relatively small number of steps at a time (up to 20), and putting that in a `while` loop until all ensembles were filled. That overrides the ability of the bootstrapping to ever stop before all interfaces are filled.

This returns the ability to stop, but optionally. First, it lets us set how many runs per "round" (default 20). Then it lets us set the maximum number of rounds allowed without progressing to the next ensemble (default `None` means no limit).

There's also an instance variable `fullbootstrapping.error_max_rounds`, which defaults to `True`, meaning that if we hit `max_ensemble_rounds`, we raise a `RuntimeError`. If `False`, it only logs a warning and does a `break` (then returning the not-fully-filled `globalstate` and allowing the script to continue).

Again, these two bootstrapping classes should eventually be combined into one class with the features of both, but this is enough to get it working now.

Context for the included features:

@sonyahanson and I have been talking about using this bootstrapping to test possible reaction coordinates for her systems. One of the concerns is that we can end up in a dead-end; in that case, we want this to be able to abort. We might also want it to abort with only the warning, instead of a full-out error. And we might want our script to continue its analysis even if the attempt to generate a crossing was aborted (in fact, we'd probably write scripts under the assumption that most attempts *would* be aborted).

In the tests, on the other hand, we want it to (1) not have a limit, so it never stops, and (2) if we did set it up so stopping was possible, we would want to raise an error at that point instead of continuing with bad initial data.

This implementation should make all those options possible.

dwhswenson:
Passes tests. Ready for review and merge.

jhprinz:
Great. That looks much better in the notebooks!

PR #348, opened 2015-10-28
update README.md

jhprinz:
First attempt to update the README.md.

### DIFF
--- a/examples/ipython/README.md
+++ b/examples/ipython/README.md
@@ -4,10 +4,93 @@ This folder contains some example python notebooks to learn and understand the w
 For this to work you need to have `ipython` and `ipython notebooks` installed. For help consult the 
 [Installation Guide](http://ipython.org/install.html).
 
+> Note there has been a major change in ipython notebooks which now run using jupyter server. 
+
 Then open a terminal, walk to this folder (in your local copy) and run 
 ```
 ipython notebook
 ```
+or since version 4.0 you need to run a jupyter server using
+```
+jupyter notebook
+```
 
 ## Examples
-- Visualization [Notebook](http://nbviewer.ipython.org/github/jhprinz/msm-tis/blob/visualization/examples/ipython/Visualization%20Example.ipynb)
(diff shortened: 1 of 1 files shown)


jchodera:
Can you use single quotes ``` to protect filenames so they appear like `filename.ipynb`?

You can also hotlink them like 
```
[`filename.ipynb`](http://link/to/notebook)
```

dwhswenson:
I think you should have a "First steps" section which points to the 3 `mstis` notebooks, followed by `alanine`. In my opinion, this is the way to learn this: you start with a system where the dynamics are trivial, so you can focus instead on what it means to do path sampling. That's what the 2D systems are for. Then you graduate to the level where you want to use a real system: that's what the alanine example is for. (This is also why I don't see a point to using the `openmmtools` alanine -- the point of the toy notebooks are to do path sampling without worrying about what the system really is. The point of the alanine example is to teach you how to set up the kind of system you'd actually care about.)

This "first steps" could be as easy as:

***

## Getting started with OpenPathSampling

The first notebooks to start with are the MSTIS examples, `mstis_bootstrap.ipynb`, `mstis.ipynb`, and `mstis_analysis.ipynb`. These will show you how to use OPS on a system so simple that you can use it interactively. Then look at `alanine.ipynb` to learn how to use OpenMM within OPS to study a more realistic system.

Other features of OPS are shown in the other notebooks in this directory. Here is a list of the notebooks and their purposes:

***

(followed by the list you made)

The way it is set up now, it looks like the first thing someone should look at is either `alanine`, which I don't think is the place to start, or in the tutorials with `fast_sample_loading` which ***definitely*** isn't the first thing to look at.

jhprinz:
Thanks for the comments. I just had to wait for a test to finish and started writing a little about every notebook as a start. I think I cannot really finish this until we have the list. I made some changes and we should merge this soon so that we start improving on this.

Let we know if you have more suggestions.

jchodera:
>  I made some changes and we should merge this soon so that we start improving on this.

+1 for merging now and then improving later.

dwhswenson:
> +1 for merging now and then improving later.

I agree. I hadn't merged it yet because I figured @jhprinz would assign it to me when he was done making changes for now.

jhprinz:
please merge at will!

Issue #331, opened 2015-10-20
Single replica multiple interface set minus move


dwhswenson:
The more I think about it, the more it seems that combining the multiple interface set minus move and single replica TIS is not so trivial as I’d hoped.

I think I have a couple possible solutions, so here’s the long description.

Keep in mind the things a good minus move should do:

* satisfy detailed balance
* allow trajectories to switch between all interface sets (easy if there's only 1)
* (almost) never do an extension that isn’t accepted

### Two ideas that fail

When there’s only one interface set, the single replica minus move just amounts to extending until we have a new interface crossing, and then only keeping the new excursion as the trajectory in the innermost interface. Very simple, satisfies detailed balance, and near 100% acceptance.

How might we adapt this for multiple interface sets (multiple innermost interfaces)? I see two obvious possibilities, but neither works:

* *Select which interface set after the extension*: This will not give us 100% acceptance. The excursion may satisfy more than one interface set. So even if we only choose among interface sets that are satisfied by this, we need to account for the ratio of ensembles satisfied by the old excursion and the new one. We could find ourselves rejecting a lot of extensions. (Formally, this is how I will implement the `SingleReplicaMinusMover`: in the event of only 1 innermost interface, you *do* get 100% acceptance, and it satisfies detailed balance. It just isn't optimal for MISTIS setups.)
* *Select which interface set before the extension*: This does not satisfy detailed balance. If I’m coming from an excursion satisfying interface set A and I stop at the first excursion satisfying interface set B, what if the trajectory made an excursion across A in between? We can’t get the original trajectory back.

#### Keeping the minus interface as an extra replica (as with a replica reservoir)

This requires initially creating many replicas in the minus ensemble, selecting one at random, and doing a normal multiple interface set minus move with it. Of course, this gets the 100% acceptance rate of the MISTIS move, and because it maintains the same number of trajectories in the reservoir, you have detailed balance.

The only downside is that you have to create the initial reservoir. However, you typically need to run MD in order to define your states (and even if we define states adaptively with MSMs, the initial trajectories will include several examples of minus trajectories). The number of trajectories in the reservoir is a parameter to experiment with: we could use as few as 1, although I think the decorrelation would be faster with more.

(And yes, we could use this replica reservoir idea in the regular TIS too… probably wouldn’t even be that hard to add in OPS.)

#### Splitting the minus move and really using the minus segment ensemble

[...]

PR #350, opened 2015-10-29
Single Replica Minus Mover

dwhswenson:
The SRMinusMover is a little different from the standard minus mover: instead of keeping an active minus ensemble, the minus ensemble is only used as an intermediate. 

As before, we define the innermost ensemble, the minus ensemble, and the "segment" ensemble, which is like the innermost except it only allows A->A paths.

The basic process for the single replica minus is like this:

1. Hop innermost->segment
2. Extend segment->minus
3. Select subtrajectory in segment
4. Hop segment->innermost

In practice, this is split into a forward minus and a backward minus: forward minus extends forward and selects final subtraj; backward minus extends backward and selects first subtraj. The total `SingleReplicaMinusMover` has a 50/50 chance of selecting its forward or its backward submover.

So the normal `MinusMover` is a replica exchange between the `minus_ensemble` and one of the `innermost_ensemble`s. On the other hand,the single replica minus mover is a move from `innermost_ensemble` to (hypothetically another) `innermost_ensemble`.

As mentioned in #331, this approach is really only useful for single innermost interfaces. In fact, the code as implemented will only be correct in that case (and raises a warning if you aren't).

- [x] SingleReplicaMinusMover
- [x] Tests for SRMinusMover
- [x] SingleReplicaMinusMoveStrategy
- [x] Tests for SingleReplicaMinusMoverStrategy

### DIFF
--- a/openpathsampling/__init__.py
+++ b/openpathsampling/__init__.py
@@ -24,18 +24,18 @@ from collectivevariable import CV_Function, CV_MD_Function, CV_Featurizer, \
     CV_Volume, CollectiveVariable
 
 from pathmover import (
-    BackwardShootMover, MinusMover, RandomChoiceMover, ForwardShootMover,
-    PathMover, PathMoverFactory, PathReversalMover,
-    ReplicaExchangeMover, ConditionalSequentialMover, EnsembleHopMover,
-    PartialAcceptanceSequentialMover, ReplicaIDChangeMover, SequentialMover,
-    ConditionalMover, RestrictToLastSampleMover, PathSimulatorMover,
-    PathReversalSet, NeighborEnsembleReplicaExchange, SampleMover,
-    StateSwapMover, FinalSubtrajectorySelectMover, EngineMover,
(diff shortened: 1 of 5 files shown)


jhprinz:
Is this ready to be merged?

dwhswenson:
No ... I'm going to add some more tests, and I think I'll go ahead and add the associated `MoveStrategy` and its tests as part of this PR.

dwhswenson:
Now this is ready for review and merge.

Note: the SRMinusMover is useful beyond just single replica TIS. It is essentially the original implementation of the minus move. Instead of keeping a separate minus ensemble in the active sample set, just uses it as an intermediate ensemble. We might eventually switch to using this as the minus mover for MSTIS -- it should work there, and removes the need to manually populate the minus ensemble.

(The minus ensemble has to be manually populated for MISTIS, however, because we need to keep the minus trajectory in the active samples to get interface switching.)

jhprinz:
Very nice. I always habe to smile a little when I realize how clear you can write down a new mover in ops...

Will merge this.

PR #351, opened 2015-11-02
Correct readme links

dwhswenson:
The links in examples/ipython/README.md were pretty bad. They weren't working, and even if they had been working, the would have pointed to a branch of JHP's fork instead of choderalab/master.

Since GitHub can show notebooks natively, there's no reason to use nbviewer.ipython.org. It also didn't really make sense to use nbviewer for the plain python file.

Quick fix; this should be ready to merge immediately.


jhprinz:
lgtm...

PR #359, opened 2015-11-06
Making Shoot/Extend movers conveniences for one EngineMover

dwhswenson:
This handles an issue discussed as part of #251, and previously partially implemented in the aborted #273. I’m implementing this as part of the path mover overhaul (reducing need for future code by consolidating code here).

Also, the previous version of shooting had a significant bug. It used the ratio of the sum of the bias weights as the metropolis acceptance probability. It should be the ratio of the shooting point selection probabilities. If the case of either (a) uniform selection probability, or (b) only one valid shooting point, this doesn’t make a difference (that would be all the cases that we have used so far). I will correct it here.

- [x] Make both of Shooting and Extension use the same underlying `_run` function
- [x] Clean up newly simplified code to minimized extraneous lines of code
- [x] Shooting point selectors should have function for probability of choosing a given snapshot: `selector.probability(snapshot, trajectory)`
- [x] Shooting bias should be the ratio of the probability of selecting the points from the two trajectories


jhprinz:
This appears ready ?!

dwhswenson:
Almost -- I'm going to try 1 or 2 other things, and there's a little cleanup of the code to be done. Should be done today, though.

dwhswenson:
Decided against trying anything fancier. I thought that I might replace some of the if/else stuff by having a class variable called `make_trajectory`, which would be set in each forward/backward mover to `_make_forward_trajectory` or `_make_backward_trajectory` (using the way Python allows you to assign functions to variables). It was a cute way to simplify some code, but my concern is that I don't know how well various documentation tools can handle that. I might try it out in the future and see if I can keep the code readable with it.

Anyway, this is ready for review. The net is over 100 lines in code reduction, all in and around the shooting/extending movers.

openpathsampling/pathmover.py
```diff
@@ -546,225 +538,105 @@ def __call__(self, trial):
...
-            shooting_point.trajectory[0:shooting_point.index] + \
-            partial_trajectory
-
-        trial_point = paths.ShootingPoint(
-            shooting_point.selector,
-            trial_trajectory,
-            shooting_point.index
-        )
+    def _make_forward_trajectory(self, trajectory, shooting_index):
+        initial_snapshot = trajectory[shooting_index].copy()
```
jhprinz: Question: Why do we make a copy of the snapshot? It does not change the configuration, velocities or reversed state, right? Then it is virtually the same as the old one and we don't need a copy. I am not sure, but I think that the engine cannot alter the given snapshot so it will create a redundant snapshot everytime. Has this something todo with potential two-way shooting?

openpathsampling/pathmover.py
```diff
@@ -546,225 +538,105 @@ def __call__(self, trial):
...
+                                               trajectory[0:shooting_index]
+                                              ).can_append
+        partial_trajectory = self.engine.generate(initial_snapshot, 
+                                                  running=[run_f])
+        trial_trajectory = trajectory[0:shooting_index] + partial_trajectory
+        return trial_trajectory
 
-        return trial_point
+    def _make_backward_trajectory(self, trajectory, shooting_index):
+        initial_snapshot = trajectory[shooting_index].reversed_copy()
```
jhprinz: Same thing as before. And if we can use a shallow copy, then the `.reversed` property is the best way to get the reversed snapshot and not create a redundant copy.

jhprinz:
Looks good. Besides the question about the snapshot copies.

dwhswenson:
Handling all the snapshot-copy questions in the main thread instead of line notes:

> Question: Why do we make a copy of the snapshot? It does not change the configuration, velocities or reversed state, right?

Main reason? Because this is how the code did shooting before. I wondered a bit of the same, but I kept it as close to the way it was as I could. I just reorganized to reduce code.

> Has this something todo with potential two-way shooting?

This might make some implementations of 2-way shooting possible, but I think 2-way shooting can still be implemented without the copies. (The idea would be that the step that modifies the shooting point needs to create a new snapshot to use as the point you shoot from -- given that this step, by definition, creates a new snapshot, that seems logical.)

Note that this might also solve the problems I mention in https://github.com/choderalab/openpathsampling/pull/363#issuecomment-154542176. I'll try a version without the `copy` statements. It passed nosetests locally; pushing now to test ipynbtests.

dwhswenson:
Looks like the `copy` stuff wasn't needed (at least, after removing it, all tests pass). I still have it there in comment form in case we need to go back (funnily, all I have to was to comment out the end of the lines with `.copy()` and `_copy()`).

I think that means this is ready for merge!

jhprinz:
Yes. Very good!

PR #363, opened 2015-11-06
Shooting Point clean-up

dwhswenson:
In conversation about the `ShootingPoint` container objects, @jhprinz and I decided that the need for those had been overridden by `MoveDetails` object as a general move information container. (When `ShootingPoint` was created, the idea was for each move to have its own class of details; but JHP’s storage has, since even its first incarnations, made it so we don’t need that extra complexity.)

Anyway, it’s time to remove them, since they generally make other things harder. This just involves restructuring a bit of the way the shooting point selectors work. Now `pick` will return the index to shoot from; everything else can be determined inside the shooting mover.

- [x] Modify EngineMover to work based on only looking at the index from shooting point
- [x] Add relevant information from shooting point directly to details
- [x] Modify ShootingPointSelectors to that `pick` return the index of the snapshot
- [x] Remove ShootingPoint object

This builds off of #359, so that one needs to be merged first.


dwhswenson:
@jhprinz, I’ll need your help to fix something in the move history tree visualization. Quick explanation of what I think broke it:

Shooting movers consist of joining two partial trajectories together: one old, one new. There are two snapshots at the shooting point, one in each trajectory. Previously, we always kept the version of the snapshot from the new trajectory segment. I switched that to keep the old one, so that the input and output trajectories of shooting would have the shooting point in common.

I think this should be pretty easy to fix, but I just wanted to make sure it was fixed correctly (I’m not sure, for example, whether the shooting point is supposed to be a part of the segment drawn for the new trajectory or not.)

Note that this change might also be a bugfix: what would have happened in the following example

original `[A1, A2, X, A3, A4]`
fwd from X  `[A1, A2, X, B3, B4]`
bkwd from X `[B1, B2, X, B3, B4]`

where `X` is (coincidentally) the shooting point in each move? Because we previously didn’t see this as the same snapshot (at least, when I did `assert(shooting_snapshot in new_trajectory)`, it failed), I think this case might have been a false positive when we test whether the trajectories are decorrelated: “original” and “bkwd” above still share one frame, but it seems the `X` snapshot in each trajectory wouldn't have given `X_old is X_new` (or even `X_old == X_new`). It would have looked different to the code, even though it was the same.

dwhswenson:
Code passes, but the problem I described above about the move history tree visualization remains. I even switched back to the old combining rule; no change. I think removal of the copy had the same effect (now the shooting point is the same snapshot, not a new one), so merging the last updates from #359 didn't make any difference.

@jhprinz : assigning this to you for a fix. Up to you whether you'd like to fix and issue the PR to `dwhswenson/clean_shooting_point` or if you want to review this, merge, and issue the fix in a later PR.

jhprinz:
> assigning this to you for a fix. Up to you whether you'd like to fix and issue the PR to dwhswenson/clean_shooting_point or if you want to review this, merge, and issue the fix in a later PR.

alreadry working on a fix, but this will take probably until tomorrow since I want to switch to using CSS instead of all those parameters. Also the way I implemented scaling, etc is very clumsy and unnecessarily complicated.

dwhswenson:
This can is ready for review and merge now. It had been waiting on the visualization bugfix that came in #379.

PR #366, opened 2015-11-07
Fix for state names in 2-state MSTISNetwork

### LINKED ISSUE #361: Find issue in MSTIS notebook
I am trying to get to grips with the mstis notebook, and prepare for tps. I do this by changing the 3 well  potential to a 2 well potential. This reduces the number of states to A and B.
Bootstrapping works. However I get the following problems:

stateA = old_store.volumes.find('A')[0]
stateB = old_store.volumes.find('B')[0]
KeyError                                  Traceback (most recent call last)
<ipython-input-21-5ffa48458735> in <module>()
----> 1 stateA = old_store.volumes.find('A')[0]
      2 stateB = old_store.volumes.find('B')[0]

/Users/bolhuis/git/openpathsampling/openpathsampling/storage/object_storage.pyc in find(self, name)
    221                 self.update_name_cache()
    222 
--> 223             return self[self.name_idx[name]]
    224 
    225         return []

KeyError: 'A'

It seems as if the state cannot be found in the old volume.
This is very strange, as the 3 well systems works fine.
I circumvented this problem by setting explicitly:

stateA = paths.CVRangeVolume(opA, 0.0, 0.2**2)
stateB = paths.CVRangeVolume(opB, 0.0, 0.2**2)

which is a bit dumb of course, but works.

However, the first approach should also work. Can you look in to this?

dwhswenson:
Resolves #361

This was easier than I thought. I just needed to check whether a custom name had been set before renaming stuff to "all states except". 

Ready for review and merge as soon as it passes tests.

### DIFF
--- a/openpathsampling/analysis/network.py
+++ b/openpathsampling/analysis/network.py
@@ -187,7 +187,9 @@ class MSTISNetwork(TISNetwork):
                 state.name = name
             other_states = states[:state_index]+states[state_index+1:]
             union_others = paths.volume.join_volumes(other_states)
-            union_others.name = "all states except " + str(name)
+            # don't rename if custom name is set (probably 2-state)
+            if union_others._name == "":
+                union_others.name = "all states except " + str(name)
 
             this_trans = paths.TISTransition(
                 stateA=state, 


dwhswenson:
Self-merging, because (1) @bolhuis needed this, and (2) it's only one line of code and an indent. I have an idea for a way to approach this problem more generally (and more cleanly), but that will come in a later PR.

jhprinz:
Sorry, yes good work. I was not available on the week-end.

Issue #362, opened 2015-11-06
ipynbtest fixes?


dwhswenson:
Posting this mostly to prod @jhprinz, since I know he's already aware of the issue:

The `ipynbtest.py` script isn't catching all thrown errors. We think this has something to do with the ipython/jupyter "[Big Split™](http://blog.jupyter.org/2015/04/15/the-big-split/)" at IPython version 4.0; at least, that's when we started noticing the problem.

Some errors can propagate for a long time before `ipynbtest` registers them: I've seen errors in bootstrapping that only showed in the replica exchange networks -- after production runs and analysis! As a result, the errors that are raised aren't very informative, since they don't represent the actual point of failure. This makes debugging annoying.

This should get fixed ASAP. Also, I would personally like to see `ipynbtest` in its own repo. I plan to re-use it for other projects, and there's nothing specific to OPS in it. It will continue to grow and evolve, so giving it its own issue tracker is probably worth doing.

jhprinz:
this will be fixed by #367 or better the updated ipynbtest. Turns was a stupid bug that did not test cells that had no output.

dwhswenson:
Solved by #367 (or, at least, by updates in the new repo introduced by #367)

PR #367, opened 2015-11-09
Move ipynbtest to external package

jhprinz:
This moves the ipynbtest to a separate project / package and installs it using conda.


jhprinz:
I moved the testing script to [ipynb-test](https://github.com/jhprinz/ipynb-test) and made it conda installable. There is still be problem, that I need a new working Anaconda TOKEN to make this automatically. So far I uploaded a recent version directly to anaconda.

You can now simple use
```
conda install ipynbtest
```
when omnia is in your package channels. After that just run
```
ipynbtest.py -h
```
to see the help. I hope that this way we can keep updating this much more clean.

jhprinz:
This will also resolve issue #362. Problem was that a cell with no output stored in the notebook would not be tested. This is corrected now. The updated ipynbtest has a few additional features.
1. `--abort-if-fail` will stop the test after the first cell fails. This will improve speed when checking if all tests pass without knowing exactly which ones.
2. `--eval [CODE]` will run the CODE before the first cell is run. This way you can set some variables for specific cases, number of samples, etc...

dwhswenson:
Great! Let me know when these changes are ready to merge. I have every expectation/intention of using this in many non-OPS projects. I've already starred, watched, and forked the repo. :-)

jhprinz:
Yeah, my first star... :)

jhprinz:
I think this is ready for review and merge. We can make updates in the external repo and then I can upload the new version to anaconda.

All this does is install the ipynbtest package and remove unnecessary files.

PR #368, opened 2015-11-10
TPSNetwork / TPSTransition

### LINKED ISSUE #360: TPS Network
To do TPS with a similar workflow to the way we've set up TIS, we should have a TPS network. I'm inclined to have a the TPS network be a general multi-state approach, since the multi-state TPS reduces to 2-state TPS pretty easily.

Question, particularly for @bolhuis: which of the following do you think is the most useful way of setting it up?

Version 1: Useful if you're likely to only define a few specific pairs of states as transitions:
```python
tps = paths.TPSNetwork(transition_pairs=[(stateA, stateB), (stateB, stateA), (stateA, stateC), (stateC, stateA)])
```

Version 2: All states are allowed as both first and last state:
```python
tps = paths.TPSNetwork(states=[stateA, stateB, stateC[)
```

Version 3: Define the initial states, define the final states; all (non-self) transitions are allowed:
```python
tps = paths.TPSNetwork(initial_states=[stateA, stateB, stateC], final_states=[stateA, stateB, stateC])
```

I think my personal inclination is to make version 3 the standard way of setting it up (allowing, for example, sampling of all transitions out of stateA without including transitions out of B or C), with an extra method to allow version 1 (which allows you to select exactly which transitions are allowed). Version 2 is the simplest in the all-to-all case, but has no flexibility. It also doesn't make it possible to study only A->B (without also doing B->A).

Any strong preferences on this?

dwhswenson:
Resolves #360

Current implementation is as version 3 in #360, with classmethods for version 1 (`from_state_pairs`) and version 2 (`from_states_all_to_all`).

Note that, currently, TIS uses the network and transition objects to hold analysis functions as well. None of those are yet implemented for TPSNetworks and Transitions.

- [x] TPSTransition
- [x] TPSNetwork, default build of ensembles
- [x] `TPSNetwork.from_state_pairs`
- [x] `TPSNetwork.from_states_all_to_all`
- [x] Tests for the construction of `TPSNetwork` using each method
- [x] Store and reload of `TPSNetwork`, `TPSTransition`

Also, this will only have a little bit of testing: I only checked that the desired number of transitions are created. It sounded from email like @bolhuis wanted to use something like this soon, so I wanted to toss something together quickly.

For the most simple TPS network, going from `stateA` to `stateB` (without the reverse process), you can do any of:

```python
versionA = TPSNetwork(stateA, stateB)
versionB = TPSNetwork(initial_states=[stateA], final_states=[stateB])
versionC = TPSNetwork.from_state_pairs(state_pairs=[(stateA, stateB)])
```

All of those will give you the same result.

### DIFF
--- a/openpathsampling/__init__.py
+++ b/openpathsampling/__init__.py
@@ -73,7 +73,7 @@ from analysis.tis_analysis import (
 from analysis.move_scheme import MoveScheme, DefaultScheme
 
 from analysis.network import (
-    MSTISNetwork, TransitionNetwork, MISTISNetwork
+    MSTISNetwork, TransitionNetwork, MISTISNetwork, TPSNetwork
 )
 
 from analysis.replica_network import (
--- a/openpathsampling/analysis/network.py
+++ b/openpathsampling/analysis/network.py
@@ -1,6 +1,7 @@
 import logging
 
 import pandas as pd
+import itertools
 
 import openpathsampling as paths
 from openpathsampling.base import StorableNamedObject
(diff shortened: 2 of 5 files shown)


dwhswenson:
Ready for review and merge.

jhprinz:
Looks good. Only thing we (I) need to make better is the handling of named objects...

dwhswenson:
> Only thing we (I) need to make better is the handling of named objects...

I'll have a PR coming later today with a (small) improvement along these lines. It isn't a perfect fix, because it requires good behavior from code contributors, but it should at least go a way toward avoiding object-renaming errors like the one that Peter was getting this weekend.

PR #369, opened 2015-11-11
Safe .named

dwhswenson:
This is inspired by the problem @bolhuis had in #361. That problem occurred because my network objects were trying to be too helpful. Under the assumption that the user wants `print` to give results that are easy to read, the code was automatically assigning names to objects. However, sometimes the object that the code was trying to name already had a *more* helpful name.

Here’s the basic conflict: 
* We don’t want names to be forcibly set names on initialization, both because that’s weird for the user, and because that makes object initialization messier.
* Foolish renaming (as I did before @bolhuis caught that problem) can lead to confusion (as @bolhuis experienced)

The approach I’m taking here involves a slight modification to the `.named()` method. In cases where `.named()` was already used (chained to an initialization) it makes no difference. However, if cases where the user had already defined a name, now `.named()` will not override the existing name.

The point is, if your factory class is naming objects as a convenience to the user, you shouldn’t override any names the user has already given. So any algorithmic name assigning should use the `object.named("name")` function instead of `object.name = "name"`. Users can still use the latter, and that will *force* whatever renaming they want.


dwhswenson:
Ready for review and merge, although it kept getting hit with that move history tree problem in the alanine notebook.

jhprinz:
I saw the alanine problem and it will be fixed by the visualization update.

jhprinz:
I think that is a quick fix for now, in the long term we should do something about naming (still trying to get my head around it). Maybe we need to split the user naming and the automatic string description.

PR #370, opened 2015-11-11
LockedMoveScheme

dwhswenson:
This is a way to create a `MoveScheme` object that is incompatible with `MoveStrategy`s. This might be generally useful for user-defined schemes -- in general, you just need to give this a root mover (you can also give it a network). We can use this to wrap some single-move “pseudo-simulations.” For example, if we find that our initial trajectory for the minus ensemble is A-to-B, we do shooting until it is A-to-A: this will let us wrap that in a MoveScheme.

If we can't find a better way to do SRTIS, this can also be the basis of SRTIS.

- [x] LockedMoveScheme forbids various strategy-related functions
- [x] Tests for LockedMoveScheme
- [x] Use LockedMoveScheme for one-off schemes in notebooks


dwhswenson:
This should be ready for review and merge. (Although it took several re-runs of tests to work....)

jhprinz:
Lgtm. The only thing we should make sure is that the use of schemes stays optional and is not necessary to do a simulation. Especially in simple cases this could add an unnecessary layer of complexity. So I understand it that if you have a setup that uses schemes you can now use it also with movers that are not compatible with strategies.

PR #371, opened 2015-11-11
Better input behavior for MSTISNetwork, MISTISNetwork

dwhswenson:
The input format for both MSTIS and MISTIS networks should be a little cleaner. I’m moving both to a list of tuples (maybe in the future this could optionally be named tuples?). I’m removing the state naming from the MSTIS tuple; the user can either name a state on their own, or the states are automatically named “A”, “B”, “C”, etc., in the order they are found.

Now it will look like this:

```python
paths.MSTISNetwork([
     (state_1, interfaces_1, orderparameter_1),
     (state_2, interfaces_1, orderparameter_2)
])

paths.MISTISNetwork([
    (state_1, interfaces_1, orderparameter_1, state_2),
    (state_2, interfaces_2, orderparameter_2, state_1)
])
```

Two major changes:
* state names are not part of the arguments
* MISTISNetwork initialization looks a lot more like MSTISNetwork initialization

If names are not already set, they will be given automatically. If `state_1` was not already named, it would be named “A”. Similarly, `state_2` would be named “B”. However, if they already have names, the names would not be changed.

In MSTIS, the order in which state get their named is obvious: the same as the appearance in the list. For MISTIS, it goes through all the starting states first, then through all the ending states.

If there are more than 26 unnamed states, then: (1) The names continue as “_A”, “_B”, “_C”, etc., then “__A”, “__B”, …; (2) The grad student doing this should seriously consider quitting before the project drives him or her insane.

This requires #369, and should be merged after that one.

- [x] Switch MSTIS to not use names as tuple field
- [x] Add auto-naming for MSTIS
- [x] Add tests, including cases where default names are used by user
- [x] Switch MISTIS to use tuples; auto-naming


dwhswenson:
This can be reviewed, but be aware that it involves a major change in the way networks (both MSTIS and MISTIS) are created, and so will affect anyone else's code once it is merged (@asagodi, @bolhuis)

jhprinz:
I would merge this. If there are no objects?!? But as David said most notebooks need to be changed afterwards.

jhprinz:
Or I say this can be merged at will. So David please go ahead. No objections!

dwhswenson:
Sent an email to Abel and Peter to warn that this could break other stuff they have. They're warned -- merge at will!

dwhswenson:
Back to mergeable status... @jhprinz, you want to do the honors on this one?

jhprinz:
Sure...

PR #374, opened 2015-11-12
Use MoveScheme as input to PathSimulator

dwhswenson:
Switching so that, instead of using the `root_mover` as the input to a `PathSimulator` calculation, we take a `MoveScheme`. After the `LockedMoveScheme` is included, that gives a quick way to create a `MoveScheme` out of any mover. 

Using the scheme itself means that more complicated schemes, and their associated networks, will be automatically saved. Right now the user has to save that separately, so this should simplify the typical user experience, and all the older advanced methods can be managed by adding one wrapping to the old approach.


dwhswenson:
Ready for review and merge.

jhprinz:
merging!

Issue #349, opened 2015-10-29
Detailed balance, SRTIS, and separating trial generation from acceptance


dwhswenson:
In our Skype call the other day, I mentioned that there’s a serious need to overhaul the path movers if we want them to support SRTIS (and for general purposes anyway). At the heart of that process is that we need to separate MC trial generation from trial acceptance.

The purpose of this issue is to explain that problem, and identify whether there is a simpler approach worth doing to hack together a working SRTIS. The hacks I've come up with so far break compatibility with the MoveScheme/MoveStrategy approaches.

***

Here’s a clearer explanation of the problem. Let’s start with a simple single replica scheme with multiple state TIS. Imagine SRTIS on a 3 state system, with states `A`, `B`, and `C`, and each state has two interfaces, e.g. `A0`, `A1`, plus there’s a multiple state interface `m`.

### Why there’s a problem

First let’s just look at the a subset of this: all the interfaces that allow states beginning in `A`. I’ll draw the move decision as tree I’ve done before. I’ve labeled the lines with letters (`a`, `b`, `c`) to mark the probability of each choice; usually within each interface, we have the same probability of choosing any available replica exchange.

![image](https://cloud.githubusercontent.com/assets/8178546/10825384/2b8aa482-7e3b-11e5-9ef8-af5ea27f671e.png)

Detailed balance requires calculating the ratio of the probability of choosing the mover and its detailed balance partner. In the picture above, this means that we need, for example, the ratio of the probability of choosing the A1->m move, given that we’re in `A1`, and the probability of choosing m->A1, given that we’re in `m`. If both values of `a` were the same, this ratio would be 1. However, it is obvious that they can’t be the same (`3a+b=1=2a+b` implies `a=0`).

We need a way to account for that ratio in our acceptance. The approach that best represents the mathematics of Monte Carlo is to separate the acceptance from the trial generation, which is the essence of the path mover overhaul I propose. There may be hacks to solve this in specific cases, but they won’t work in general.

### Bad hacks as possible solutions

Note that in all of the hacks listed below, we lose all the nice features of MoveSchemes and MoveStrategies. You can no longer customize the relative probabilities of various moves. You can’t go back and forth between single replica and regular TIS. I’m not sure whether you’d be able to customize the replica exchange network, and I think it might even become difficult to customize the shooting movers.

[...]

PR #375, opened 2015-11-13
Abstract Base Class

jhprinz:
This PR should replace PR #302. It uses `abc` from python.

This is ready to review and merge.


openpathsampling/pathmover.py
```diff
@@ -559,25 +549,32 @@ def _make_backward_trajectory(self, trajectory, shooting_index):
                             trajectory[shooting_index + 1:])
         return trial_trajectory
 
+    # direction is an abstract property to disallow instantiation of the EngineMover unless we use
+    # a concrete subclass that sets this. This is not super elegant but is the way to do it with
+    # abstract classes
```
dwhswenson: Actually, I think that *is* a pretty elegant way of doing it.

openpathsampling/pathmover.py
```diff
@@ -802,10 +812,9 @@ def __call__(self, sample1, sample2):
 # SUBTRAJECTORY GENERATORS
 ###############################################################################
 
-
-class RandomSubtrajectorySelectMover(SampleMover):
+class SubtrajectorySelectMover(SampleMover):
```
dwhswenson: I also like the separation of the abstract `SubtrajectorySelect` from `RandomSubtrajSelect`

openpathsampling/tests/test_helpers.py
```diff
@@ -232,4 +239,41 @@ def current_snapshot(self, snapshot):
 
     def generate_next_frame(self):
         self._current_snapshot = None
-        return self.current_snapshot
\ No newline at end of file
+        return self.current_snapshot
+
+
+class IdentityPathMover(paths.PathMover):
```
dwhswenson: `IdentityPathMover` might move to the main `pathmover.py`. This is the no-op move I've talked about in #349, so we might use it there.

dwhswenson:
Still waiting for Travis to actually pass it, but this looks pretty good. Two tiny things:

* Do you want to move `IdentityPathMover` into `pathmover.py` as part of this PR? If you don't do it here, I'll probably have to do it later.
* Could you rename `test_pathsimulator.py` to `testpathsimulator.py`? For the most part, our test files don't have the underscore (I think we should also rename `test_toy_dynamics` to `testtoydynamics` to keep it consistent).

openpathsampling/tests/test_helpers.py
jhprinz: Yes, I had it first in PathMover and then I moved it since it was only part of the test...

jhprinz:
> Do you want to move IdentityPathMover into pathmover.py as part of this PR? If you don't do it here, I'll probably have to do it later.

Sure, no problem. Is `IdentityPathMover` a good name? It was my first idea and seemed to work. It could also be `EmptyPathMover` or `NoPathMover`

> Renaming

Good point. When I named the file I was actually wondering whether to name it with or without the underscore. I would not like to rename all the other files, since it needs to be merged into other branches. So we should do the other renamings later.

jhprinz:
Done...

dwhswenson:
> Is IdentityPathMover a good name?

I think it works well enough. The only thing that is weird about it is that it seems to me like the name suggests an *accepted* no-change. Whether "no change" is accepted or rejected is just semantics, although I'd rather keep it rejected.

I'll merge once tests pass.

dwhswenson:
Genuine test failure: I think you need to remove `paths.` from `IdentityPathMover`'s inheritance.

dwhswenson:
tests pass now! merging.

PR #379, opened 2015-11-14
quick fix for visualization bug

jhprinz:
This should fix the problems with PathTreeVisualization for now.

It fixes a problem in SamplePathMover that will set the correct index for the trial shooting point.


jhprinz:
Ready to review.

PR #380, opened 2015-11-14
Add feature support

jhprinz:
This will allows us to use different types of snapshots. Right now there is
`ToySnapshot` and the ordinary `Snapshot` which we might rename to MDSnapshot.

The feature store is also working and can represent these two types already.


jhprinz:
this was a stupid bug: Turns out if you create an abstract class then the type of the class is not type, but abc.ABCMeta because it is created using a metaclass other than type. This made the JSON to_dict fail... 
Finally got it to work again...

jhprinz:
This PR will add the following features to the storage. All of these is necessary to implement support for different Snapshot classes. The idea is that the `Store` that handles loading and saving needs to be chosen or be adaptive to the used `Snapshot` class. This provides maximal flexibility and speed. 

Right now we support three types.

1. abstract `AbstractSnapshot`: This is the abstract and most simple Snapshot we can think of. It only implements the handling of the reversed snapshot and it references a topology (this might be replaced by a reference to an engine).

2. regular `Snapshot` class (might be renamed to MDSnapshot) : These use two additional objects Configuration and Momentum that store coordinates, box_vectors and potential_energy and velocities with kinetic_energy. It uses simtk units to be compatible with openmm

3. the `ToySnapshot` class: for one particle in N dimensions. Stored is only position and velocity, no units.

#### Changes

- the topology is now always property of the snapshot and not (as before) a property
- all stores (that allow storing of python objects) are now stored as well. This is similar to creating variable types like compound variables or variable length variables. It also means that if you create a store you can (almost always) load it, since it contains the complete structure. 
- all files related to storage have been split into parts necessary for netCDF and the OPS specific parts.There is now a new folder named `netcdfplus`. This way all the stuff that is not really OPS is out of the way.

#### Snapshot feature support

This provides the possibility to use different kinds of snapshots. Currently it provides two ways to do that.

1. There are specific stores for each type and when you create a storage the appropriate store is chosen and stored in the storage.
2. There is also a feature based implementation that works using `Features`. A feature contains code to initialize storage necessary for the feature content and a list of class variables that are to be stored. An example is:
    ```
    class Coordinates(Feature):
        _variables = ['coordinates']

        @staticmethod
        def _init(self):
            self.init_variable('coordinates', 'numpy.float32',
                           dimensions=('atom', 'spatial'),
                           description="coordinate of atom '{ix[1]}' in dimension " +
                                       "'{ix[2]}' of configuration '{ix[0]}'.",
                           chunksizes=(1, 'atom', 'spatial')
                           )
    ```

    to add feature support in a `Snapshot` class just add
 
    ```
    __features__ = ['Velocities', 'Coordinates']
    ```

    this way the Snapshot class does not need to import any storage related modules but lists only the features by name, while the storage knows how to interprete specific features.

jhprinz:
I have combined these PR with the saving of stores since these two belong together and the saving is only used for feature support. Feature support can be disabled. This is handled by setting
```
Storage.USE_FEATURE_SNAPSHOTS = True or False
```

The PR affects to many files due to the renaming of some files.

dwhswenson:
> I have combined these PR with the saving of stores since these two belong together and the saving is only used for feature support.

Please try to break this into small, step-by-step PRs instead of one giant one. This looks like it risks becoming another merging nightmare like #300.

jhprinz:
The actual changes are not so big. 

Problem is the amount of work to make it work in little steps since implementing one thing requires fixing other things only to make it pass. Then remove these changes and go from there, add the next thing. Fix everything, remove and so on. I could try to restructure everything to break it down, but that will have to wait since I have to figure out in which way I can make this work so that is does not into a break it down into too many merging nightmares...

I will leave this the way it is for now. Hopefully all the little PRs in the meantime will not make this obsolete... Each PR require fixing merges because of the renamings... Last time the renaming was a constant fixing merge errors...

Anyway. This was a quick 1 day attempt to implement features, and it is not 1.0...

dwhswenson:
> The actual changes are not so big.

Okay... just, if you see a way to split the functionality, please try to. It makes merging easier for me.

A couple discussion points while it is still work-in-progress:

#### Naming

Could the `feature.py` and `Feature` class be explicitly `SnapshotFeature`? Even within the `paths` namespace, I feel like "feature" is a too-generic term for it to be obvious to a new contributor.

#### Topologies

So there's a problem I had while while working on [octomore](https://github.com/choderalab/octomore). This may already get solved as a part of this PR (do I understand correctly that `snapshot.topology` is now settable?) But I want to make sure that this use case remains in your mind, since I think it will be a common one.

I was trying to take a trajectory loaded through MDTraj, and use it to get initial conditions to run with an OpenMM engine. But because MDTraj and OpenMM use different topology classes, it took me a few tries to figure out how to simply and efficiently assign the topology to the right place. (You can see my solution as part of https://github.com/choderalab/octomore/pull/10.)

I think this will be a common practice: assigning either a first snapshot for a new engine from a snapshot with a different class of topology, or even taking a whole trajectory and assigning it as an initial path for the new engine. In particular, I suspect we'll do this frequently after loading via MDTraj. If `snapshot.topology` is settable, that ought to fix the problem. You might add conveniences in the `Trajectory.from_mdtraj` to handle this as well. (Although I noticed you've currently deleted `Trajectory.from_mdtraj` ... so please explain I how would accomplish the above with this.)

Anyway, this PR seems like an appropriate place to make this process easier.

jhprinz:
> Okay... just, if you see a way to split the functionality, please try to. It makes merging easier for me.

Yes, sorry. Had a rough weekend :) and tried to patch this quickly together and I did not realize that it looked so bad. The restructuring of the files (which looks pretty nice now) cause almost all files to show up as changed although it is mostly a rename in the import statements, nothing else.

I have also moved some classes to other files which made this even worse. The actual new code is not so much and pretty straight forward.

I think I can move the refactoring into a separate PR and this should make it much better.

jhprinz:
> Could the feature.py and Feature class be explicitly SnapshotFeature? Even within the paths namespace, I feel like "feature" is a too-generic term for it to be obvious to a new contributor.

Sure. Any suggestions like this are helpful.

### Topologies

Yes, I realized a few problems that are related to this. Let me try to explain the way this works in this PR and what our possibilities are to make this work better. This implementation is more a fix to make topologies work the same way across all snapshots classes.

I think a snapshot needs to know a topology. This is not required to make OPS work, but it seems natural that coordinates are associated with a topology that we should keep these bundled. Before the ToySnapshots had a topology attribute while the normal snapshots used a configuration object that was referenced by the topology.

The topology was never stored with a snapshot, but only per storage file once and all snapshots or configurations you load get an attribute `topology` that points to the topology in the storage, which itself points to the template in the storage which itself points to a topology object... So complicated.

For newly generated snapshots from the engine the engine also has a template and thus a topology. In a weird way if you would use another topology for the engine than for the storage there would be no way to see that problem. Say engine uses topology A and storage B. Then the generated snapshots from the engine, stored and loaded again have not topology B and not A.

#### Possible solutions

1. **Store a topology with every snapshot**: This is the simplest solution. Each snapshot class has a topology object and this will be stored (by reference).
    _Pro_: Easy and clear
    _Contra_: Additional 4 bytes per snapshot and additional time to create objects (which will not affect lazy objects!)

2. **Store an engine reference with every snapshot**: We want to have this anyway (we agreed upon that) and since an engine always needs a topology (at least an openmm engine uses a topology to be created) we could take the route 
    ```topology = snapshot.engine.topology```
    _Pro_ : Less additional storage (only for the engine)
    _Contra_: If you do not create a snapshot using an engine it does not have a topology. Hence loading from pdb only works if you turn from_pdb into an engine that you need to store. It also forces storage of the engine and not only the topology. It would be a nice way to bundle all creation of snapshots. It could be a little tricky to make sure that not every load of a trajectory from mdtraj will result in a new engine.

3. **Initialize a global topology for all**: Setting something like 
    ```openpathsampling.topology = ...```

    and this is used by all classes, but I really don't like this idea. There is too much that can go wrong
    _Pro_: Easy
    _Contra_: Dangerous and ugly style

4. **No topology for snapshots**: No topology for snapshots or trajectories. A function that requires it (like an MDtraj CV) needs to be passed a topology explicitly.
    _Pro_: Easy and saves storage
    _Contra_: More difficult to use certain (useful) features.

I consider only the first 2 reasonable. 1 is okay. if you think the engine idea is better I think we need to have these _SnapshotFactory_ engines as well, which I think is nice. These could also contain the from_mdtraj which is effectively a engine that generates a trajectory from a file.

#### `from_mdtraj`, where did it go...

I wanted to remove everything that is storage related from Trajectory. from_mdtraj is a utility that is useful only for specific types of simulations and so I moved it (actually it was already upto one function call) to tools. These will be split into parts per engine.

dwhswenson:
> Yes, sorry. Had a rough weekend :) and tried to patch this quickly together and I did not realize that it looked so bad.... I think I can move the refactoring into a separate PR and this should make it much better.

No worries. And yeah, I think that making the refactoring into a separate PR will help.

### Topologies

Of your suggestions, I also prefer the first or second. The trade-off I see between them is:

* **every snapshot has a topology**: Now there are two topologies, one at `snapshot.topology` and one at `snapshot.engine.topology`. This is a chance for serious confusion. Suggestion: during the setter for `current_snapshot`, maybe we include a test that `snapshot.topology is engine.topology`? (Or at least `==`?) When generated, there should be no possibility of confusion.
* **only use engine topology**: while this fixes the previous problem, it does mean that we would need to treat MDTraj as an engine, which seems weird to me.

So maybe the first, but with that little safety check when assigning a snapshot to the engine. And it would be nice to have convenience functions to convert a whole trajectory from, e.g., being based on MDTraj topologies to being based on OpenMM topologies. `Trajectory.retopologize(self, new_topology)`? Up to the user to confirm that `new_topology` is sufficiently equivalent to the old one, because I think there's only so much we can do to ensure that.

### `from_mdtraj`

> I wanted to remove everything that is storage related from Trajectory.

Well, I wouldn't call `from_mdtraj` "storage-related," but I agree with the overall point of making each "section" of OPS as independent from other parts as possible, to enable later reuse. It sounds like that's the real goal here?

jhprinz:
This PR will be replaced by #383.

jhprinz:
I am confused. Was the merged because #382 was merged? So effectively it was not merged, yes?
If so, that it is correct.

dwhswenson:
I think so. At least, I only recall merging one thing today!

If all the commits included in #380 were also in #382, GitHub thinks that merging #382 is equivalent to also merging #380.

PR #383, opened 2015-11-17
[WIP] Add `Snapshot` features 

jhprinz:
This is a copy of PR #380 but now based of or PR #382. Let's hope that I did not forget any files :)

PR #384, opened 2015-11-18
Use Snapshots with features

jhprinz:
This implements an idea that @dwhswenson and I had:
All simulations are in principle independent of the type of the snapshot used. For toy dynamics this will be very simple snapshots with only coordinates and velocities (I excluded the potential), for MD these might include units, Configuration / Momentum objects with box_vectors etc...

Features are a way to construct certain classes of Snapshots out of building blocks which are called features. Each feature must be orthogonal to other features used, but not all features must be orthogonal. You could have something that only stores coordinates and a snapshot that stores coordinates and velocities. So the code for coordinates is reused. 

[more to come...]


jhprinz:
this will replace PR #383.

jhprinz:
This will introduce some changes to the way Snapshots are used.

- There are different classes of type `Snapshot` which all derive from `AbstractSnapshot`. 
- The old `Snapshot` is now the snapshot meant for MD, while toy systems should use `ToySnapshot`, which can be stored much faster.
- I removed support for `unit-system`, which was too much. Now each variable can have a `simtk_unit` and on creation the units for snapshots are chosen from the template snapshot. Thus, if you create a snapshot without units it will assume that your coordinates do not have units! 
- ToySnapshots do not have box_vectors or energies, only coordinates or velocities of `N` dimensions. And they are not stored using `Configuration` or `Momentum` which makes it much faster to store.
- As a fall back there are still explicit implementations of the snapshot stores for all types so that features are not necessary.

jhprinz:
This is ready for comments and (merge). I assume we will discuss a few things before merging. I am not 100% sure if the use of classes for features is best.

dwhswenson:
This branch is dependent on another branch called `addfeaturesagain`: is that branch to have a PR, or should it be deleted?

jhprinz:
correct, that was the original branch. And then I removed lot's of features and reapplied these. Same thing as we had before. It is merely remains of my clumsy attempt to split the PR.

openpathsampling/storage/features.py
```diff
@@ -0,0 +1,240 @@
...
+import simtk.unit as u
+import numpy as np
+
+from openpathsampling import Configuration, Momentum
+from openpathsampling.netcdfplus import ObjectStore
+
+from openpathsampling.tools import units_from_snapshot
+
+
+class Feature(object):
```
dwhswenson: Rename to SnapshotFeature for clarity.

openpathsampling/storage/snapshot_store.py
```diff
@@ -28,35 +58,14 @@ def _load(self, idx):
...
-            configuration=configuration,
-            momentum=momentum,
-            is_reversed=momentum_reversed,
-            reversed_copy=LoaderProxy(self, reversed_idx)
-        )
-
-        return snapshot
+    def _put(self, idx, snapshot):
+        self.vars['momentum_reversed'][idx] = snapshot.is_reversed
+        self.vars['momentum_reversed'][idx ^ 1] = not snapshot.is_reversed
```
dwhswenson: why is this called `_put`?

openpathsampling/storage/storage.py
```diff
@@ -25,6 +25,8 @@ class Storage(NetCDFPlus):
     simulation. This allows effective storage of shooting trajectories
     """
 
+    USE_FEATURE_SNAPSHOTS = False
```
dwhswenson: Does this mean that none of this is being tested yet (at least, for storage)?

openpathsampling/tests/test_toy_dynamics.py
```diff
@@ -177,57 +175,19 @@ def test_sanity(self):
...
-        snap = Snapshot(coordinates=np.array([[1,2,3]]), 
+        snap = ToySnapshot(coordinates=np.array([[1,2,3]]),
                         velocities=np.array([[4,5,6]]))
-        # note that we truncate the z-direction, regardless of what it is.
-        # This is for a 2D model!
         self.sim.current_snapshot = snap
-        assert_items_equal(self.sim.positions, [1,2])
-        assert_items_equal(self.sim.velocities, [4,5])
+        assert_items_equal(self.sim.positions, [1,2,3])
+        assert_items_equal(self.sim.velocities, [4,5,6])
```
dwhswenson: Truncating to 2D was the original intended behavior, and the reason for this test. Is the intent to change that behavior?

dwhswenson:
* `features.py` : rename to `snapshot_features`
* There's a lot of `idx ^ 1` : I don't know what that is supposed to do. Seems a bit like obfuscated code -- is there a more readable way to do that? Even if you just define a function called `effective_index(idx)` that return `idx ^ 1`, that would probably make it clearer (and you could use that function's docstring to explain why this is there and what it is doing).

openpathsampling/snapshot.py
```diff
@@ -313,22 +258,137 @@ def inner(self, *args, **kwargs):
 
     return _has
 
+@lazy_loading_attributes('_reversed')
+class AbstractSnapshot(StorableObject):
+    """
+    Simulation snapshot. Contains references to a configuration and momentum
+    """
+
+    def __init__(self, is_reversed=False, reversed_copy=None, topology=None):
```
dwhswenson: If all subclasses of `AbstractSnapshot` require an `engine` parameter in their `__init__` functions, shouldn't `AbstractSnapshot` require it too?

openpathsampling/snapshot.py
```diff
@@ -313,22 +258,137 @@ def inner(self, *args, **kwargs):
...
+        """
+        Returns a shallow copy of the instance itself. The contained
+        configuration and momenta are not copied.
+
+        This will also lead to a new reversed copy when using reversed!
+
+        Returns
+        -------
+        Snapshot()
+            the deep copy
```
dwhswenson: Correct the docstring: this is a shallow copy, right?

openpathsampling/snapshot.py
```diff
@@ -313,22 +258,137 @@ def inner(self, *args, **kwargs):
...
+    # ==========================================================================
+    # Utility functions
+    # ==========================================================================
+
+    def copy(self):
+        """
+        Returns a shallow copy of the instance itself. The contained
+        configuration and momenta are not copied.
+
+        This will also lead to a new reversed copy when using reversed!
```
dwhswenson: Please clarify this statement in the docstring. What is meant by "when using reversed"?

openpathsampling/snapshot.py
```diff
@@ -313,22 +258,137 @@ def inner(self, *args, **kwargs):
...
+        contained configuration and momenta are not copied and the momenta
+        are marked reversed.
+
+        This will also lead to a new (non-)reversed copy!
+
+        Returns
+        -------
+        Snapshot()
+            the deep copy
+        """
```
dwhswenson: Same as with regular `.copy()`: docstring says deep copy, but should be shallow, right?

openpathsampling/snapshot.py
```diff
@@ -313,22 +258,137 @@ def inner(self, *args, **kwargs):
...
+        configuration and momenta are not copied.
+
+        This will also lead to a new reversed copy when using reversed!
+
+        Returns
+        -------
+        Snapshot()
+            the deep copy
+        """
+        this = AbstractSnapshot(is_reversed=self.is_reversed, topology=self.topology)
```
dwhswenson: Why does `AbstractSnapshot` implement this? Seems kind of weird, since this snapshot shouldn't be possible to instantiate, if it is actually abstract.

openpathsampling/snapshot.py
```diff
@@ -575,7 +586,103 @@ def subset(self, subset):
...
+        coordinates : simtk.unit.Quantity wrapping Nx3 np array of dimension length
+            atomic coordinates
+        velocities : simtk.unit.Quantity wrapping Nx3 np array of dimension length
+            atomic velocities
+        box_vectors : periodic box vectors
+            the periodic box vectors
+        potential_energy : simtk.unit.Quantity of units energy/mole
+            potential energy
+        kinetic_energy : simtk.unit.Quantity of units energy/mole
+            kinetic energy
```
dwhswenson: Update docstring to represent the changes you made. This is completely false for `ToySnapshot`.

openpathsampling/snapshot.py
```diff
@@ -575,7 +586,103 @@ def subset(self, subset):
...
+
+
+    # TODO: Can be removed. We don't need this anymore. Still here for legacy reasons
+    @property
+    def xyz(self):
+        """
+        Coordinates without dimensions.
+
+        """
+        return self.coordinates
```
dwhswenson: The toy engine exists as a simple model that should be analogous to MD. Therefore, the `ToySnapshot` should implement as much of the same API as `Snapshot` as possible. Keep .xyz around (although it can be a one-line alias `xyz = coordinates` now).

dwhswenson:
Overall, I think the structure looks reasonable. What's the case against having `SnapshotFeature`s as classes? When it comes to extensibility, I think that's the way to go.

One question: Why make `__features__`  strings? Why not just classes? That would seem more straightforward, and cleans up some of the `import features as ft` and `getattr(ft, feature)` stuff. Again, easier for extensibility (a new `SnapshotFeature` shouldn't need to be part of the core of OPS in order for a user to use it).

openpathsampling/snapshot.py
jhprinz: Alright. Then I just leave it the way it is.

openpathsampling/snapshot.py
jhprinz: Yes, well, I implemented the AbstractSnapshot so that it will also work. So you can have even a simulation without any data... This is why I implemented all the functions. It was more a test to see if it would work consistently (which it should). I agree that it is probably never used and maybe also should not be used (since it does not make sense).

openpathsampling/snapshot.py
jhprinz: Ups, I think that is legacy from trying to remove the engine from the PR.

openpathsampling/tests/test_toy_dynamics.py
jhprinz: Well, the new ToySnapshot supports, whatever dimensions you put into the template. So if you use 2D then 2D will be stored and using 3D coordinates will fail. I am not sure though if that is what you would like to have to be consistent with using 3D spatial coordinates somehow.

openpathsampling/storage/storage.py
jhprinz: I tested both cases but I have not yet implemented a test that compares both cases. Will do that once we agree that the overall structure is reasonable. I can also add a speed comparison now and see if the feature implementation is slower.

I realized though that the new ToySnapshot seems to be faster, but have not tested it. In my feeling the MSTIS examples are running faster now. This is because all the energy handling and creating of Configuration objects is not necessary anymore.

openpathsampling/storage/snapshot_store.py
jhprinz: Hmmm, get and put. Load and save was taken. get and put is what is used in databases I think. Do you have a better name? This is the function to get or put the data from the storage.

openpathsampling/storage/features.py
jhprinz: Did that already.

openpathsampling/tests/test_toy_dynamics.py
dwhswenson: The idea from the original is that file formats are often explicitly 3D (needed for visualization in VMD, for example). So if I wanted to take a toy system that actually has 2 spatial dimensions and get data in/out from, e.g., XYZ format, I need to remove the z-coordinate at some point. That was what this did.

I frequently did this sort of thing with my 2D "Geisslerene" model, so that's the motivation. Maybe there's a better way to do this, but that was the whole purpose of that specific test.

openpathsampling/storage/storage.py
dwhswenson: > I realized though that the new ToySnapshot seems to be faster, but have not tested it. In my feeling the MSTIS examples are running faster now. This is because all the energy handling and creating of Configuration objects is not necessary anymore.

Primarily energy calculation. Keep in mind that, with only 2 degrees of freedom, energy calculation is about half the cost of force calculation (and is not calculated as part of the force calculation). So you probably see ~66% time cost. This is why we don't usually do that with simple models; I added it because it was a part of the original snapshot.

openpathsampling/storage/snapshot_store.py
dwhswenson: Is there a good docstring in the superclass? I kind of get why you have these names, but I don't think the roles are obvious from a code readability standpoint. OTOH, I'm just reading the changes on GitHub, as opposed to using Python's `help` and other code browsing tools, so good docstrings elsewhere would clear this up.

jhprinz:
Thanks for the comments. I did not really update all the docstrings to make sure I am on the right track. I will now iron the docs.

> Overall, I think the structure looks reasonable. What's the case against having SnapshotFeatures as classes? When it comes to extensibility, I think that's the way to go.

Hmm, yes I don't know. All the classes are basically a list of create_variable and create store (which could be stored differently) and a list of variable names. A class seemed appropriate although you will never create an instance of it. This is what I found weird somehow. 

One question: Why make __features__ strings? Why not just classes? That would seem more straightforward, and cleans up some of the import features as ft and getattr(ft, feature) stuff. Again, easier for extensibility (a new SnapshotFeature shouldn't need to be part of the core of OPS in order for a user to use it).

Yes, another thing that I could not easily decide. I thought that having a simple API that will not need to import stuff from storage would be nice. The feature that you import contains specific code to store in a netcdf file and my understanding of a feature was just saying `This snapshots uses coordinates` without specifying what coordinates are. It might be better to just make a list of variables that should be stored. That is kind of how it works internally. Although a feature is not a variable. 

> features.py : rename to snapshot_features

Yes, that would be okay.

> There's a lot of idx ^ 1 : I don't know what that is supposed to do. Seems a bit like obfuscated code -- is there a more readable way to do that? Even if you just define a function called effective_index(idx) that return idx ^ 1, that would probably make it clearer (and you could use that function's docstring to explain why this is there and what it is doing).

This function picks the paired index so if you have 1 it gives 0, 4 gives 5 since the pairs are (2n, 2n + 1). The old implementation is
```
paired_idx = 4 * (idx / 2) - idx
```
but the XOR is much simpler and faster. Since it effectively need the last bit to be flipped. Hence `idx xor 1` or in python `idx ^ 1`.

#### Simplification idea

I was thinking to unify the feature and variable storage idea. Which is nice and makes it really easy for the user, but it might introduce performance decreases. I might make an issue about this idea at some point. General idea is that using features and in PR #398 a simpler way to store an object the unifyinf idea is to say "These are the class attributes I wish to store and these attributes have a certain variable type and dimension and unit (and maybe a few hints for efficient storage like chunk sizes)" Then inprinciple storing this object can be completely automated. You would only say these objects should be stored using JSON and these objects (like snapshots) should be stored using attribute to netcdf variable approach. An the user does not have to build a new storage at all (unless he does weird stuff like handling reversed snapshots, etc...). An example
```
class Snapshot(StorableObject):
    __storables__ = {
        'configuration' : Variable(type='obj.configurations'),
        'momentum' : Variable(type='obj.momenta'),
        'reversed' : Variable(type='bool')
    }
```
to create a store to save these just do
```
storage.register('snapshots', VariableStore(Snapshot))
```
and all the rest is done automatically. Even the `_init` function. To make this even more cool we could use a metaclass instead of storable object, but that is not necessary in this case.

The only part missing is that this assumes that a store for Configurations is already present. And in the features we have right now this is nicely bundled in the `_init` function.

For now I would leave it as it is and maybe for storage 2.0 we make it simpler?

dwhswenson:
#### `Feature`s as classes

> A class seemed appropriate although you will never create an instance of it. This is what I found weird somehow.

Hmm... good point. It looks to me like there's no need for inheritance here. [So the more pythonic approach is probably to put each feature in a separate module, making `._init` a free function.](http://stackoverflow.com/a/12735474)

Also, if there's no instance, the use of `self` as a variable name is really confusing (not to mention that I should have caught that it was in a `staticmethod`) -- that should be some external `storage`, right?

#### `__features__` attribute

> Yes, another thing that I could not easily decide. I thought that having a simple API that will not need to import stuff from storage would be nice. The feature that you import contains specific code to store in a netcdf file and my understanding of a feature was just saying This snapshots uses coordinates without specifying what coordinates are. It might be better to just make a list of variables that should be stored. That is kind of how it works internally. Although a feature is not a variable.

I don't think I explained my point. I'm thinking about how people can extend OPS. Right now, new features need to be implemented in the `openpathsampling.features` namespace. That means that if someone wants to distribute an OPS engine with a new feature, they need to modify that namespace. My opinion is that we should restrict the `openpathsampling.*` namespace to things we intend to support. Contributors should be able to add new things that work with OPS, but exist outside of our namespace (and it isn't our job to support that).

Example: I want to make an engine to do semiclassical dynamics. This requires an extra feature (the monodromy matrix). There's no reason to add that to OpenPathSampling for everyone else, because that won't be useful to everyone else, and it will just clutter the OPS code. So I should be able to register a feature `semiclassical_engine.MonodromyMatrix` without touching any of the official OPS code.

The current approach fixes the namespace. My suggestion is to change it to

```python
__features__ = [paths.features.Configurations, paths.features.Momenta]
```

instead of strings that, as far as I can tell, are only used to unwrap into that namespace. Then I could have:

```python
class SemiclassicalSnapshot(AbstractSnapshot):
    __features__ = [paths.features.Coordinates, 
                    paths.features.Velocities, 
                    semiclassical_engine.MonodromyMatrix]
```

If someone does this, it is their job, as the developer of `semiclassical_engine`, to maintain and support `MonodromyMatrix`, not our job, as the developers of OpenPathSampling.

#### `idx ^ 1`

My problem here is still with the readability. A user gets an error, and the stack trace passes through code that looks like this. This code looks pretty cryptic, and the user has trouble figuring out whether this might be related to the bug.

It seems like the options are:

* comment every line explaining why this is
* replace with an extra line giving `reversed_version = idx ^ 1`
* replace with a function `def reversed_idx(idx): return idx ^ 1`

Any of these are possible, but this should be more readable. At the very least, put an explanation of it in the class docstring.

#### Simplification

> For now I would leave it as it is and maybe for storage 2.0 we make it simpler?

Yes. This sounds like a potentially good idea, and might solve some of the above, but we shouldn't do it now.

jhprinz:
> `Features` as classes

Very good. I did not see this possibility, but that is exactly what it is. A module. Should be do that?

> `idx ^ 1`

I replaced it by a function `paired_idx` and explained it in the docstrings.

> Features

Very good point. That is really useful and looks good if you read the features. And it should be easily extendable. Will see how it turns out

jhprinz:
Okay, I made modules. It now looks like you proposed.
```python
import features

class ToySnapshot(AbstractSnapshot):
    """
    Toy snapshot. Contains references to a coordinates and velocities only
    """

    __features__ = [
        features.Coordinates,
        features.Velocities
    ]
```
features is now a _package_ at main level. It seemed not to belong into `storage`.

jhprinz:
For the same reason I moved ConfigurationStore and MomentumStore to the features package since this is the only place these are used.

jhprinz:
I placed Configuration and Momentum into a file named `snapshot_content` which is okay but not great. Suggestions?

jhprinz:
Just tried to also move Configuration and Momentum to the corresponding files in features which failed due to import cycles (which I could not resolve)

jhprinz:
Just making sure that we agree on making feature into a sub-package. It looks nice and works fine.

dwhswenson:
> Just making sure that we agree on making feature into a sub-package. It looks nice and works fine.

Definitely. Sorry, I thought my support for that was clear from before. So we'll have, for example:

```python
paths.features.Coordinates
```

Two things to clarify (and to serve as documentation):

1. Should it be `Coordinates` or `coordinates`? Since it is technically a module, the standard would be lower case. That also informs people that it can't be instantiated. But it also still feels a lot like a class -- specifically, like `Coordinates` is the endpoint of the namespace that the user cares about. Lower case hints to me that I will want to look inside that namespace to access the thing I need, which isn't true here. Anyway, I can be persuaded either way on this, we just need to make sure we're consistent.

2. Just checking that I understand things correctly: In the above, `features` is a subpackage and `Coordinates` is a module. Any feature-module must contain a module-level variable called `_variables`, which is a list of strings (is this always a length-1 list?) and a function `_init(store)` which adds a store for this variable. If the feature being stored is a complex object (i.e., not just a simple predefined type like an `np.array`), then the module should also include a class that creates a netcdf+ `ObjectStore` for it (which will be used in the module's `_init` function). Is that description correct?

jhprinz:
> Just checking that I understand things correctly: In the above, features is a subpackage and Coordinates is a module. Any feature-module must contain a module-level variable called _variables, which is a list of strings (is this always a length-1 list?) and a function _init(store) which adds a store for this variable. If the feature being stored is a complex object (i.e., not just a simple predefined type like an np.array), then the module should also include a class that creates a netcdf+ ObjectStore for it (which will be used in the module's _init function). Is that description correct?

Yes, in detail
1. features is a subpackage (`openpathsampling.features`)
2. Coordinates is a module, like the other features. So one file named `features/Coordinates.py`
3. Each feature module has
    i. A list of one or more variable names, e.g.
        ```
        ['coordinates', 'velocities']
        ```
    ii. An `_init` function that knows how to create the variables (of same name) and potentially needed stores that can hold new variables types (like complex objects)

4. You can reuse a store but you cannot redefine it in two features. Example: Imagine you have two feature that both store `Configuration` objects. Then (so far) only one can create the `ConfigurationStore` to hold the objects. One way to make this possible is to check for existing stores with the same parameters. So far this case does not happen since all features are orthogonal.

jhprinz:
About the upper/lower-case problem. I asked myself that question. I think I prefer using lowercase. Not sure if I can give a good reason for it.

dwhswenson:
> About the upper/lower-case problem. I asked myself that question. I think I prefer using lowercase. Not sure if I can give a good reason for it.

Do you want to switch it to lowercase before I merge then? Files ares currently features/Coordinates.py, etc.

Other than that, this is ready to merge, right?

jhprinz:
Ready to merge when it passes...

jhprinz:
??? all tests pass on my machine

dwhswenson:
You need to use git to rename the files. It didn't register a local change in capitalization in the file names.

jhprinz:
Me neither but my IDE usually does that and my git status says all files are checked in.

dwhswenson:
Files listed here include capitals. Maybe this will help http://stackoverflow.com/questions/10523849/changing-capitalization-of-filenames-in-git

jhprinz:
no way, how stupid is that...

jhprinz:
crossing fingers

jhprinz:
Yes, this worked and is using featured snapshots now.

PR #385, opened 2015-11-18
Add possibility to save stores itself...

jhprinz:
This was part of the original PR #380 and is only the part that affects netcdfplus by adding the possibility to save the list of used stores within the storage itself. This provides better backward compatibility and will allow to always open a store as a file. 

This is in total a much better way to define a file format.

Of course the used stores require implementations which need to be available and these code is not stored!


jhprinz:
Ready for review

dwhswenson:
Also depends on `addfeatureagain`; same question as in #384.

openpathsampling/netcdfplus/dictify.py
```diff
@@ -205,7 +211,7 @@ def to_json(self, obj, base_type=''):
         return json.dumps(simplified)
 
     def to_json_object(self, obj):
-        if hasattr(obj, 'base_cls'):
+        if hasattr(obj, 'base_cls') and type(obj) is not type and type(obj) is not abc.ABCMeta:
```
dwhswenson: When do you have `type(obj) is abc.ABCMeta` while trying to store the object? My intuition would have been that it was impossible for that to occur, since `abc.ABCMeta` shouldn't get instantiated.

openpathsampling/tools.py
```diff
@@ -205,24 +205,20 @@ def units_from_snapshot(snapshot):
         the simtk.unit.Unit to be used
     """
 
-    units = {}
+    units = {'velocity': None, 'energy': None, 'length': None}
```
dwhswenson: This assumes (as it did before) that your snapshots have a concept of velocity, energy, and length. Is there a way to avoid that assumption? Or do we just make that assumption, and then not use those units if they don't describe snapshot features?

dwhswenson:
Would it be reasonable to add `from storage.storage import Storage` to the OPS `__init__.py`? I find the extra lines of `import openpathsampling.storage as st` to be a little annoying, since the user will only ever directly use one class from that subpackage.

dwhswenson:
Aside from comments above, this all looks reasonable. I'll plan to merge this one first, since it has a number of code overlaps with #384; I'll review that after this is merged. Let's finish the discussion on this one as soon as possible.

openpathsampling/netcdfplus/dictify.py
jhprinz: The is something that is cause by the use of a metaclass to implements abc. In the classes that use abstract I placed a `__metaclass__ = abc.ABCMeta` which replaces the function `type` by the function `abc.ABCMeta` as the class constructor (not to be confused with the class instance constructor which is the `__init__` method). When building any class that is a potential subclass of this abstract base class the metaclass is also inherited. That means that all subclasses lead to `type(class) == abc.ABCMeta` and not `type(class) == type`. It took me some time to figure that out. 

This part was not needed before since usually we do not store a class, but rather instances and their attributes. The reference to a class was only necessary to provide a store with the possibility to store for which class this store is meant. This is necessary for all object_stores. 

If you have a better idea to check if an obj is a class or an abstract class that would be great. The more pythonic way is probably to use `inspect.isclass()` but that does more or less the same thing and check for more cases and is potentially slower. But I can change that to remore the necessity to import `abc`.

openpathsampling/tools.py
jhprinz: This is still in there because I split the PR. In the next steps this should be removed and only be used for MDSnapshots. 

I still need this in here because the current implementation uses this to determine that toy snapshots do not have a unit an MD snapshots have. 

This is one of the reasons why the PR became so big. I realized to get rid of this little unit thing I need to implement support for different Snapshot Classes and hence the possibility to store stores(that know which Snapshot they store), etc... one lead to the other.

jhprinz:
Yes, this is independent of all other PRs and should be merged first. The `from storage.storage import Storage` is good as long as it does not create import cycles (which I will check). I agree that this is annoying.

jhprinz:
Works. You can now use
```
paths.Storage()
paths.AnalysisStorage()
```

jhprinz:
This passed tests so should be ready to merge.

openpathsampling/netcdfplus/dictify.py
dwhswenson: I understood what the code was doing, but I still don't quite understand why. (And I think that, for what it is doing, that's a fine way of doing it). But I don't get what you're saying here:

> This part was not needed before since usually we do not store a class, but rather instances and their attributes. The reference to a class was only necessary to provide a store with the possibility to store for which class this store is meant. This is necessary for all object_stores.

Can you explain what is going on that we are storing classes, as opposed to the attributes of an instance of a given class?

openpathsampling/tools.py
dwhswenson: So these units should be removed in #384?

dwhswenson:
> Works. You can now use
> ```
> paths.Storage()
> paths.AnalysisStorage()
> ```

Excellent! That definitely feels more natural.

Merging this, but please respond to the line notes above. They're mainly me asking for clarification; the code looks fine and runs fine.

openpathsampling/tools.py
dwhswenson: Whoops, just saw the note in #384 that this is coming in a later PR (I assume that's #396?). Nevermind.

openpathsampling/netcdfplus/dictify.py
jhprinz: I see. The underlying idea was that a store (the class instances that stores objects) should be storable objects themselves. In doing so the stored file contains most of the structure in it, not only the data. I think that was clear. Now to save a store I needed to add a new variable type for the to_json functions. The thing that I named simplifier. It takes a dict (presumable the one from `to_dict`) and converts all objects that cannot directly turned into json into a little dict that will be reconstructed when loading. So 
```
object >(to_dict())> dict >(simplifier))> simple_dict >(json.decode())> string
```
The simplifier never needed to encode a class itself. And my implementation checked for `type(obj) == type` and if the class was subclassed of storableobjects (we do not allow storing unknown classes). In that case we take the name of the class and store it with a `_class` key. When reconstructing it realizes the `_class` key and picks the appropriate class from the list of registered storable objects. 

Now to your question. The point is that before it just never occurred that an instance had an attribute that pointed to a class. Now for storing stores themselves the store has an attribute that points to a class. Namely the base class to be stored, which I call content_class.

Hence the need to update the simplifier to handle now abstract classes as well. This would have failed before too, but since we never stored abstract classes the error never showed up.

dwhswenson: That explains it. Thanks for the clarification!

Issue #260, opened 2015-05-15
Single Replica TIS


dwhswenson:
The basic approach I'm planning to take here is to use an unbound function `single_replica_scheme(network)` which can take a network (MSTIS or MISTIS) and returns a move scheme based on the single replica TIS approach. It will also replace the `movers` object in the `network` with an appropriate dictionary for the SR move scheme.

This is on hold until completion of #142.

PR #386, opened 2015-11-18
Poor SRTIS

dwhswenson:
It isn’t the pretty version that I’d like to have, but I think it will work. Approach described in https://github.com/choderalab/openpathsampling/issues/349#issuecomment-156745744 (and other comments, esp from @bolhuis, in that thread). My opinion on this approach, as I expressed it previously: "It definitely fell out of the ugly tree, but at least it didn’t hit every branch on the way down."

- [x] PoorSingleReplicaStrategy
- [x] Tests for PoorSingleReplicaStrategy
- [x] Add the ability to use the `real_choice_probability` for estimating number of moves
- [x] SRTISMoveScheme (for now, using PoorSRStrat)
- [x] Test notebook for SRTISMoveScheme
- [x] Add support for bias to `EnsembleHop`
- [x] Test that biased ensemble hopping works
- [ ] Analysis for SRTIS scheme

Future versions of the SRTISMoveScheme will switch to the better approach, once we have it. But I think this will work in the short term.

`choice_probability` is still preserved, but is no longer the correct way to estimate the number of total steps needed to get a certain number of trials in a certain mover -- now that will require another thing called `real_choice_probability` which gives the result (assuming equal sampling of each ensemble).

I'm still not sure that everything in this is safe for all possible move strategies, but it should give us something that is temporarily workable.


jhprinz:
+1 for the name of the PR

dwhswenson:
Quick update on this: so.... I completely forgot that I still need to add support for biases in the ensemble hop. Oops! That will come next.

However, other than that, I seem to have something that at least *says* it is doing SRTIS. I haven't checked the results at all, but at least the code runs without error.

Also, adding the SRTIS notebook to the testing suite.

dwhswenson:
Important stuff on the implementation of bias functions:

##### Bias functions take a (provisional) change and a (previous) sample set to determine the bias. 

This is very important for three reasons: (1) It allows us to use many different kinds of bias functions;  (2) It allows us to (eventually) wrap multi-sample changes in a bias function; and (3) It is the most generic approach, allowing the most flexibility later.

Right now, the focus is on a limited type of bias: we have a single sample from an ensemble hop, and the bias value is determined by the new and old ensembles.

However, it should be possible to have a bias on *any* kind of move, and we should be able to extract that bias from information from the sample other than the ensemble. For example, my proposed CITIS method puts a bias on the acceptance of the shooting move, based on the maximum value of the order parameter for the trial trajectory.

Initial implementation only allows 1 sample at per change; fixing up that part depends on how things look in the pathmovechange, and @jhprinz and I will have to chat to figure out the best way to handle that.

dwhswenson:
@jhprinz : This leads to some file format changes (in particular, `MoveScheme` has new attributes). In order to run longer calculations (which, hopefully, we won't have to repeat), it might be useful to merge this while it is only partially completed.

Probably the best order is (1) I finish reviewing & merge your big storage update, (2) I fix any conflicts with this, and then (3) you review/merge this while it is partially complete. I think I'm close to finishing everything in this, but I don't want waiting for this change on file format to be a bottleneck for running full examples for the paper.

I removed the auto-close of #260 from the top post; that will officially close in a follow-up PR after this is merged.

dwhswenson:
This is ready for review as a partially-complete PR. It affects some file-related stuff (initialization/storage of `MoveScheme`), so better to merge it sooner than later.

After this is merged, I'll open a new PR to complete the goals of this one (see #260). It is pretty close.

jhprinz:
I think this looks good although I do not completely understand the reason for all the `NullMoves`. Is it to be consistent with having reversible pairs and mimicking acceptance smaller than 100%. 

If you agree I will merge this and we can continue.

dwhswenson:
Removed [WIP] tag (the goal is in-progress, but I guess the incarnation in this PR isn't). Merge at will!

> although I do not completely understand the reason for all the NullMoves

Short answer is "detailed balance." Long answer is in #349.

jhprinz:
Allright. Will merge...

PR #387, opened 2015-11-19
[WIP] Add support for variables with arbitrary objects

jhprinz:
This is an implementation that add another var_type `obj` to the `create_variable` function.

While creating a variable `obj.[store]` will create a variable that can store objects that are stored in the given store the type `obj` takes any object and stores a reference to the object and an index.

A similar possibility is to store a JSON string that references the class and the index. This is the fast way and storage cheap way. No JSON pickling involved.

This provides one way to store arbitrary trajectories. Imagine you have two types of snapshots one in the file and one external. You could have two stores one that stores snapshots in your file, while the other only stores references to external frames in other files. 

So if properly stored loading a trajectory will give you a list of snapshots loaded from different places. Nothing changes, the store will just realize that some snapshots use this store while others use the external one.

Next, I will add an example on how to use this.


jhprinz:
This was already implemented in PR #409.

PR #391, opened 2015-11-23
[WIP] Add support for external stored content

jhprinz:
This shows how to allow to store `Configuration` objects outside of the main store file.
It extends the storage with a list of external files that is used to store coordinates of frames.
The current examples implementation can only use Configurations and not momenta. This is due to the fact that mdtraj (which I use to implement reading of external MD trajectories does not support velocities for reading, it can be used for writing I think). I presume to be compatible across several formats.

I am still looking for a way to read velocities for say `.trr` files. This problem is independent of the current implementation and will only make it accessible for the use with Gromacs, etc...

#### Implementation 
The external way works the following.
1. Keep the internal way to store configurations and rename it to `storage.intconf`.
2. Create a new store `storeage.extconf` that only referenes an ExternalFile
3. Create a new store `storage.mdfiles` that keeps track of external files so that we do not have to store the file everytime we reuse the same file.
4. The `storage.configuration` will work exactly as before, if you do save normally. Hence it is completely backwards compatible and there is not problem with using cvs.
5. There is a new `storage.configuration` that replaces the old one and stores a reference to `intconf` and `extconf`. This allows to store a configuration internally or externally or both!
6. The are a few additional functions 
    i. `storage.configuraion.save_as_external(mdfile, frame)` which stores a configuration which is already saved as a frame in an mdtraj readable file. This takes only the file and a frame.
    ii. `storage.configuration.persist(idx)` will copy (if necessary) the externally stored configuration to the internal store and references the internal for faster access. This means you can store everything in the main file if you wanted to. 

This PR is based on PR #385 .


jhprinz:
This will never be merged and is only open to remind us of the feature. With the changes in storage I have to reimplement it, which is fine and not that hard. So I will close this PR and copy the content to an issue. That will reduce the open PR to what is really necessary.

PR #393, opened 2015-11-23
[WIP] Store Engine with PathMover

jhprinz:
This will store the engine explicitly with a PathMover. It will not make any direct changes. Only that to set the default behaviour you need to set
```
paths.EngineMover.default_engine = ...
```
this used to be `.engine`, but was only used in `alanine.ipynb` anyway.
When you create an engine it will be set as default already and hence all PathMover created after that will use that engine. You can also specify your own engine as an optinal parameter for all PathMovers that use an engine (shooters, extenders and minus)


jhprinz:
I removed the test in minusmover since the way we create movers this cannot happen. It uses either the default or the explicitely set engine for both shooters.

jhprinz:
This is ready for review and merge, when it passes.

dwhswenson:
in meeting. Have different approach. Hold onthis.

jhprinz:
Sure, was a quick attempt. I can only work on this tomorrow... No time today...

jhprinz:
I think this was adressed by #465 ? Can we close this now?

dwhswenson:
Haha -- funny enough, I think I even changed my original `EngineMover.engine` to `EngineMover.default_engine` because of some potential confusion I found.

Main difference is that I don't automatically set default: only if you use certain simulations. `PathSampling`, in particular, has no default and requires the engine to be set on any engines that use it.

(I think I'd said at one point in a Skype call "nevermind on the hold: it's close enough to what I was thinking" -- and obviously, we ended up in the same place.)

Yep, #465 definitely duplicated this (and extended it a bit). Closing.

PR #394, opened 2015-11-23
[WIP] Count MD steps run in EngineMover

### LINKED ISSUE #292: Counting MD steps run in each PMC

dwhswenson:
Resolves #292. 

In principle, I think the 2-line fix in here does the job. However, I'll add some testing for it later.


dwhswenson:
Correction ... current stuff does a limited implementation of the goal. The idea proposed in #292 is better than what I've implemented so far here. But this is a necessary start.

jhprinz:
Can we move this to an issue. Only if we abandon this PR anyway and will create a new one later.

dwhswenson:
> Can we move this to an issue. Only if we abandon this PR anyway and will create a new one later.

The PR is still pretty close to up-to-date. No reason to abandon it.

(Already mentioned this in person, but adding for the record)

dwhswenson:
Closing this as outdated and unlikely to be cleaned up for merge (more likely to be completely rewritten). The linked issue is suggests a better implementation than what was started here.

Issue #330, opened 2015-10-17
Substorage: Different "views" on storage


dwhswenson:
[This may already be a part of storage (or the new storage); if so, we can close. If it isn't already implemented, let's make a new PR of it (don't add any more features to #300).]

Pretty frequently, we think about just using a subset of the total steps from a calculation. For example, if I'm using block averaging to get estimates for the errors, I may want to split the steps into 10-20 subgroups, and do the analysis on each of those. Or I might want to compare a long (10+k steps) run of a MSTIS run with a shorter SRTIS which gets its fixed bias from the first 1k steps of the MSTIS.

It would be nice to have a simple notation for this. At the very least, I think we could do something like:

```python
substore = storage.substorage(step_range=slice(0,2000,2))
```

and it might even be possible to do

```python
substore = storage[slice(0,2000,2)]
```

Importantly, such a substorage should be **read-only**. Otherwise there's ambiguity about what should be saved. It should also still have access to all the ensembles, etc. in the storage. In my mind, the only thing that changes is how the iterator over steps works. In the examples above,  I would like for:

```python
for step in substore.steps:
    analyze(step)
```

to be the same as

```python
for step in storage.steps[slice(0,2000,2)]:
    analyze(step)
```

Another option might be to change the iterator so that `for step in storage` is the same as `for step in storage.steps`; I think that's going to be the most common usage, but I don't know if that's necessarily safe to do.

It might also be worth thinking about the possibility of allowing ensembles/lists of ensembles or replica IDs/lists of replica IDs to define substorages as well. Again, this would be a convenience for analysis, which often wants to track one replica ID (for replica exchange) or one ensemble (for ensemble property analysis). We have ways to do that already, but this might be cleaner and more user-friendly. This might be implemented by having:

```python
# in Storage
def substorage(self, step_range=None, ensembles=None, replicas=None):
    """If more than one argument is defined, the combination is treated as an 'and'."""
```

jhprinz:
Alright. This is very neat. And the first where only `mcsteps` is affected is almost trivial. I can make a kind of proxy that gives you all stores and override the `mcsteps` in the proxy. As long as no other stored object will reference any mcsteps this is not a problem. Otherwise a stored object will call the altered store to get the objects and use wrong indices.

So, if you do that then effectively `st.steps[10]` will call `st.steps[20]` 

If you do the same for ensembles this will be a problem. 

[...]

PR #395, opened 2015-11-24
Allow a view of the steps property of storages

jhprinz:
You can now get a lazy copy of a storage with altered iteration behavior for the `.steps`.
This makes it possible to run all the analysis on only a subset of steps.

All this does is really change the `__iter__` behaviour of a store. So do
```
view = StorageView(full_storage, range(0,100))

for step in view.steps:
    # this runs over steps 0 to 99
```
you can also pass this to analysis functions like
```
mstis.rate_matrix(view)
```

Nothing is copied or kept more than once in memory!


jhprinz:
This adresses Issue #330 .

jhprinz:
Ready for review and merge once it passes.

All it does is adding a special `StorageView` objects that mimics an existing open store, but if
you iterate over `.steps` it will only pick the indices you set it to.

dwhswenson:
Finally catching up on old stuff... This looks good, although I think the tests should be part of nosetests, not the ipynb tests (this looks more like unit testing than integration testing to me).

Will merge as is; test location can be discussed later.

PR #396, opened 2015-11-24
Cleanup of unitsystem use

jhprinz:
The removes using arbitrary unit and unitsystems from mdtraj based snapshots.
It mostly affects tools.py


jhprinz:
To be merged after PR #384.

jhprinz:
This can be reviewed and merged once it passes tests.

It will remove the use of unit systems completely. Mostly fixing the snapshot from pdb function to rely on mdtraj using nanometers, etc. as units and create appropriate snapshots.

jhprinz:
Ready to be merged. It will only remove some legacy stuff that would handle different unit systems. This is not used anymore since we have multiple Snapshot types.

PR #397, opened 2015-11-24
Add Version to netcdfplus and storage and setup.py cleanup

jhprinz:
This just adds a version number to all files and logs if you are loading an old version or loading a file from a newer version.

There is a version from netcdfplus and a version from ops_storage.


jhprinz:
To be merged after PR #385.

jhprinz:
add a `__version__` to the netdfplus sub package and use the ops.__version__ for the ops part

jhprinz:
There is a problem to get the version from `openpathsampling.version.short_version`.
Locally it works, but on travis this fails. No idea, what really gets installed. For now I will put in the version number manually to get it to work.

dwhswenson:
Try adding a couple extra lines to `devtools/ci/git_hash.sh` so that it also prints the short version. That will tell you what Travis thinks it has. (I added full version output in Travis a while ago as a check on things.)

I think something might be wrong in a setup.py. I see things in output claiming v0.4.0 when students here install it. We should be 0.0.0 everywhere.

jhprinz:
Let's see. These little problems can be really annoying...

dwhswenson:
Removing `alanine.ipynb` is going to cause problems for `test_pyemma.ipynb` and `storage_tutorial.ipynb`.

jhprinz:
True...

dwhswenson:
Also: 
```
Wigner:~/ops dwhs$ grep -n "0.4" setup.py
98:    setupKeywords["version"]           = "0.0.4-alpha"
```

So setup.py line 98 has the wrong version listed.

dwhswenson:
(I'm working on a PR to remove alanine, and keep the functionality in the other two notebooks)

dwhswenson:
Conda package version listed as 0.0.3 in `devtools/conda-recipe/meta.yaml`

jhprinz:
Restarting a second time...

jhprinz:
Finally it passes...

jhprinz:
So far I put in the version manually. So it could pass, or we leave all of the PRs open and wait until we figure out what is wrong.

dwhswenson:
Did you check if the problem is that the version it is seeing isn't the *installed* version of OPS? `openpathsampling.version` only exists after you run setup.py. It is possible that the installed version of OPS is not the one that `nosetests` sees, since `nosetests` is run within the git directory. I think we install with `python setup.py install` instead of `python setup.py develop`. That might enough to fix it.

dwhswenson:
Also, please make versions consistent. See above about conda version and setup.py version.

dwhswenson:
>  I think we install with python setup.py install instead of python setup.py develop. That might enough to fix it.

Other option: in `__init__.py` we have a `try: except:` around importing `version`. You could just set the necessary version in the `except` part of that, and see if that fixes the problem.

jhprinz:
I cleaned up the current setup.py and made it use a setup.yaml which seems to be much easier to maintain.
This also allows us to run a `devtools/ci/create_conda_meta.py` script that will create the `meta.yaml` for the conda recipe from the setup.yaml file. This contains all the information necessary for setup.py and looks like:

```yaml
name: openpathsampling
version: 0.1.0

authors:
  - David W.H. Swenson
  - Jan-Hendrik Prinz
  - John D. Chodera
  - Peter Bolhuis

emails:
  - dwhs@hyperblazer.net
  - jan.prinz@choderalab.org,
  - choderaj@mskcc.org
  - p.g.bolhuis@uva.nl

description: 'OpenPathSampling: A python package to do path sampling simulations'

long_description: |
  OpenPathSampling (http://github.com/choderalab/openpathsampling) is a
  python library to do transition interface sampling.

classifiers: |
  Development Status :: 3 - Alpha
  Intended Audience :: Science/Research
  Intended Audience :: Developers
  License :: OSI Approved :: GNU Lesser General
  Public License v2.1 or later (LGPLv2.1+)
  Programming Language :: C
  Programming Language :: Python
  Programming Language :: Python :: 2
  Topic :: Scientific/Engineering :: Bio-Informatics
  Topic :: Scientific/Engineering :: Chemistry
  Operating System :: Microsoft :: Windows
  Operating System :: POSIX
  Operating System :: Unix
  Operating System :: MacOS

download_url: http://github.com/openpathsampling/openpathsampling
url: http://www.openpathsampling.org

package_data: []

packages:
  - openpathsampling
  - openpathsampling.storage
  - openpathsampling.tests
  - openpathsampling.analysis
  - openpathsampling.netcdfplus
  - openpathsampling.engines
  - openpathsampling.engines.features
  - openpathsampling.engines.openmm
  - openpathsampling.engines.openmm.features
  - openpathsampling.engines.toy
  - openpathsampling.engines.toy.features

platforms:
  - Linux
  - Mac OS X
  - Windows

released: false

requires:
  - python
  - numpy
  - scipy
  - pandas
  - netcdf4
  - openmm
  - openmmtools
  - pymbar
  - docopt
  - pyyaml
  - mdtraj
  - svgwrite
  - networkx
  - matplotlib

license: LGPL 2.1 or later
license_file: LICENSE
```

We could add running the update script before building, it only maes sure that the version number of the conda package is the same as given in setup.py and written in version.py

jhprinz:
Versions are also taken from the version.py files now

dwhswenson:
> I cleaned up the current setup.py and made it use a setup.yaml which seems to be much easier to maintain.

Is this based on any standard approach to handling setup files in Python repositories? I think it is very dangerous to get too far from what other developers will be familiar with. Are you sure it will work with other install approaches that users might use?

If this doesn't install with all of the following then it needs to be fixed:

* `python setup.py install`
* `python setup.py develop`
* `pip install -e .`
* `easy_install .`
* `conda build devtools/conda-recipe`

These are standard one-line installs that, if we have everything set up correctly, should work. If they don't work, then we lose a group of users who expect installs to work that way. I personally use the first 3 (depending on the machine and circumstances).

You now require a non-standard package just to ***run*** `setup.py`. This seems like it would potentially scare off users, and like it makes adding other distribution mechanisms more difficult. I'm afraid this might make it literally impossible to make OPS `pip` installable: As I understand it, `pip` obtains the requirements from [the `setup.py` file's `install_requires` keyword](http://python-packaging.readthedocs.io/en/latest/dependencies.html). If you can't run `setup.py`, you can't even get that keyword.

I get that you're trying to put all the package metainformation in one place, but I think we should trade our inconvenience for the convenience of others coming to the project. Or at least find something based on a more standardized approach: rolling your is probably reinventing the wheel, and will only make it harder for others in the future.

jchodera:
> I think it is very dangerous to get too far from what other developers will be familiar with. Are you sure it will work with other install approaches that users might use?

I agree with @dwhswenson on this. It sounds awesome, but there are so many ways that people install things that it would have to work with all of these methods correctly to solve more problems than it causes.

It does look like a very nice solution to a really common and painful problem, though!

jhprinz:
So, if I understand correctly now we suddenly want OPS to be pip installable, too. Yes? And where is the problem to have steup.py to require `pyyaml`? OPS needs pyyaml, too?

> Are you sure it will work with other install approaches that users might use?

No, why do you think it won't? Did anyone of you ever try to install OPS using 

`pip install -e .`
`easy_install .`

The others obviously work since travis passes and it works locally with running setup.py.

Apparently there is no time to waste on this solution and I have to time to check if that will work in these cases. If you want it to work with easy_install, I can remove the last commits and keep it confusing and leave the consistency of verions between setup.py, version.py and the conda build to another PR.

dwhswenson:
TL;DR: It's not that I dislike this. It is a good solution to the obnoxious problem of multiple different distribution methods, and I find the yaml to be very readable. My concern is that it steps far from the standard behaviors, and for things that have standards (even multiple and stupid standards) we should stay close to the standards.

Basically, I wish that Python had taken an approach like this in the first place. But they didn't, and I think we should stick with standard behavior for things like installation procedures. It isn't pretty to have the metadata in multiple places, but for now, I think it is safer.

> So, if I understand correctly now we suddenly want OPS to be pip installable, too. Yes?

Eventually pip intallable: yes. Reason: this is how I keep getting funded to work on OPS.

Explanation: Right now, we have a bunch of requirements that are actually optional to have the core of OPS functionality (e.g., OpenMM, pyEmma, maybe even MDTraj). Example: you can run the toy models without any of these installed.

My thought: the omnia distribution of OPS should include all of those extras. After all, you *can* run OPS using only the toy engine, but you can't do any interesting science with that. The omnia distribution is the shortcut to doing actual science with OPS, and is highly recommended.

But omnia shouldn't be the only distribution mechanism. In particular, I'd like to be able to distribute it through the project that is currently funding me. But they might not want to use `conda`-based distribution. So it is best for me to be able to put a version of OPS on to pypi, and distribute through `pip`. (If we can distribute through pip, we can distribute through anything.)

We should be able to do that by making a few things that are currently requirements into recommendations (in particular, OpenMM, which I don't think is pip-installable). Then most other requirements could be correctly handled by `pip`.

If we set it up correctly, `pip install openpathsampling` should install OPS regardless of existing requirements. Occasionally, there are things that need to be hacked around (numpy and scipy are common culprits), but those tools are working on improving their install procedures. `pip` will find requirements in `install_requirements`. However, it needs to be able to read the `setup.py` for that. If `setup.py` requires `yaml`, I'm not sure that it can be read.

Anyway, maybe you can keep this aside for the future? I think it is a nice solution, but right now I think that it has the potential to confuse users by more than it helps us. On the other hand, it seems like there's a need (more general than OPS) for a good generalization to handle multiple install methods, so maybe the idea here could be reused in a broader context?

jhprinz:
replaced the yaml with a python dict inside the setup.py. That removes dependency from pyyaml.

Ready for review

dwhswenson:
Could you remove cell 17 from `storage_mem_test.ipynb`? There's no reason to add 113,000 lines there.

I think there's another approach that works better than setting the OPS version in a `try: except:` in `storage.py`. The current `try: except:` is kind of pointless, since there's no reason to get it algorithmically if you still have to update the file with every version. But it will be easier for me to just merge this and write up my proposal as a separate PR.

jhprinz:
Okay, this needed the mem_test_fix... So we probably do not need that...

Let's fix the other idea later.

jhprinz:
This shoule work now.

dwhswenson:
Looks reasonable -- merging!

PR #398, opened 2015-11-25
Use a generic object store for simple (enough) objects

jhprinz:
This is merely reusing redundant code making it easier for the user to write own stores.

It simplifies the `SampleStore`, `SampleSetStore` and `MCStepStore`. The new way to define the store looks like this

```
class MCStepStore(VariableStore):
    def __init__(self):
        super(MCStepStore, self).__init__(
            MCStep,
            ['change', 'active', 'previous', 'simulation', 'mccycle']
        )

    def _init(self, units=None):
        super(MCStepStore, self)._init()

        # New short-hand definition
        self.init_variable('change', 'obj.pathmovechanges')
        self.init_variable('active', 'obj.samplesets')
        self.init_variable('previous', 'obj.samplesets')
        self.init_variable('simulation', 'obj.pathsimulators')
        self.init_variable('mccycle', 'int')
```

and I think this is very readable. All you have to do is 
1. to call the super init with the class to be stored and a list of attributes of the class that are used. 
2. write a list of type definitions on how to store each variable.


jhprinz:
This needs PR #384.

dwhswenson:
This one looks fine to me. Merging.

Issue #337, opened 2015-10-24
Visualization questions and cleanup suggestions


dwhswenson:
Looking more closely at these new visualizations, and there are a few issues I'd like to see resolved.
1. This should be one line per **_canonical**_ move. Separating the minus move into multiple steps is just confusing. At the very least, there needs to be an option for this.
2. How do I turn off showing replica exchange? Or turn off showing minus moves? This is what we'd usually want to look at. And if, after a minus move, the next path might be completely decorrelated: that's fine. It should be like starting a new tree with the decorrelated segment of the minus trajectory.
3. What are the numbers to the left? Can I turn them off? The representation of the move should be labelled with the step number, and possibly a short string to identify move type (I mostly agree with your selections: "f", "b", "r", "x", and I'd use "-" for the minus move; possibly "f-"/"b-" to differentiate direction.) Other numbers are just confusing. I think the number you have next to the path is actually _not_ the step number -- step number is at the far left now, right? (All these differences from the behavior of old scripts will only confuse users -- let's keep things as similar as possible.)
4. How do I get this to show rejected moves? That was one of the most useful new aspects of this visualization in a previous version. Now I don't see any difference when I set `tree.rejected=True`.

I think you might be losing sight of the way we actually use these "path tree" visualizations. The purpose is _not_ to be a log of every substep of the move, tracking provenance. Rather, it gives important information on how paths are decorrelating. The current version of these includes a bunch of visual "extras" that distract from the point.

Examples of the kinds of information that we get from these pictures:
- _Does the path length oscillate between different characteristic lengths?_ This could indicate switching between two mechanisms.
- _Do we see blocks where one direction of shooting is consistently rejected?_ This tells us something about early-vs-late barriers in the system.
- _Do rejected shooting moves from a particular region have much longer trajectories that still end in the states?_ Again, this suggests multiple channels, and may indicate that we're completely missing one channel.

These are just a few sorts of things I've thought of in past while looking at these plots. That doesn't mean the list is exclusive. The point is that asymmetries in the tree can inspire us to ask questions about why. Those are easier to catch if the visualization is less information-dense.
---

As an aside, #335 includes a subclass for the move tree visualization in order to simplify for the common case. It makes the user's tasks (which are what the notebooks are to illustrate) much easier.

jhprinz:
Will address this in #400 .

jhprinz:
Can this be closed? I am not sure if I cover exactly everything in here in the Vis update PR.

[...]

PR #400, opened 2015-11-27
Implement better SVG creation in PathTreeBuilder

jhprinz:
This is the second round of the VIS update. I made it even interactive now. But I still have to fix the MoveTree part. 

Basically you can now fully configure the view to your liking. I even coupled it interactively with the MD viewer from mdtraj and then the configuration changes when you touch/hover over a certain snapshot. But that is still a bad hack. Nevertheless it works!

So there is still the update of MovePathTree missing and then this is ready. The code has simplified a lot (I think).


jhprinz:
One question for the PathTree and reversal moves. How do you display these. Which is better
```
    1 2 3 4 5 6 7
fw        +-8 9    
rv  9 8 4 3 2 1
fw        +-A B C
```
this would keep fw in one direction in the plot although effectively after a reversal move fw means backwards (in a time sense), so we could do this
```
      1 2 3 4 5 6 7
fw          +-8 9    
rv    1 2 3 4 8 9
fw  C B A-+
```
which reflects better what we actually do, but might be confusing with fw and bw appearing on both sides alike!

Ideas? Especially @bolhuis might help here.

dwhswenson:
Personally, I prefer to keep the direction of time always the same. Remember that this visualization comes from TPS, where you would always have the initial state on the left, and the final state on the right.

Under what circumstances would the second version help you understand what's going on in your simulation? The first version seems much clearer to read, in my opinion. Example: If I see that shooting points are only accepted near the early frames of a trajectory (near the initial state), that might be meaningful. In the second version, that information would be obscured.

jhprinz:
Your answer was kind of the explanation already.

Because in TPS you would not reverse since then you go from A > B paths to B > A paths. Is that correct? 

The first picture will do exactly that so for A>B paths: It reverses the direction of the order parameter and I thought that would be nice to see left is the initial state and right is the end state. If you reverse like in picture 1, you loose that. This is why I prefer 2. Because it monitors progress and crossing the barrier.

For TIS when most paths go A>A, this is not so important. But still seems to make sense. If I interpret this consequence of CV direction correctly then a reversal move would also reverse the direction of the CV? Is that true? This is why I pointed the second fw shoot in the opposite direction after the reversal. If I shoot forward in time in a reversed trajectory it looks like a backward shoot in CV space.

This is why I would say, to this make sense for TPS we need to use the second plot. If you strictly want time to go from left to right then the first. Hence my question what is preferred.

This question originated in a problem I had when I tried to figure the overlap between to trajectories after a reversal move. They have the same length so it makes sense to place them aligned under each other. Problem is that the idea of taking shared frames and place these under each other does not work
```
1 2 3 4 5 6 7 8
          |
      8 7 6 5 4 3 2 1
```
If you would reverse timedirection in the plot after a reversal move this problem disappears. My initial idea was to place blocks whenever a frame is new in the sense that the trajectory above has the same frame at the same x position like we do in fw and bw shoots. This works for all moves like extension, subtraj, shoot, etc, but not for reversal
```
1 2 3 4 5 6 7 8
          |
          + 9 A B
```
and so I have to repeat all blocks so that subsequent shooting moves can build on that. Third point was the decorrelation between trajectories, which is also not well defined for reversal moves. When looking for overlaps in trajectories two reversed trajectories have an overlap of n times 1 frame or looking from transitions they do not share any transition and are hence completely decoupled. Still that share all
frames when we ignore the direction. And we assume they are strongly correlated.
So using plot 2 we can keep looking for overlaps to check for decorrelation like we do for all other moves.

#### To summarize the Pros/Features
1st
- consistent time direction fw in time is left to right

2nd
- consistent CV direction
- less block (less cluttered reversal moves)
- visual correlation for reversal moves

I would go with the second since we allow for reversals anyway (meaning we do not care so much about time direction :)

Hope that explained the initial question.

Of course it is trivial to implement both with an option to choose which you want. So this question simplifies to what default we want.

dwhswenson:
> If I interpret this consequence of CV direction correctly then a reversal move would also reverse the direction of the CV? Is that true? 

I don't understand this question. What is the "direction of the CV"? The CV is just a function that maps a snapshot to a number.

A path reversal move takes the trajectory, and replaces it with a trajectory with the frames in the reverse order and with the velocities reversed. I think that's what the visualization should represent. 

> This question originated in a problem I had when I tried to figure the overlap between to trajectories after a reversal move. They have the same length so it makes sense to place them aligned under each other. Problem is that the idea of taking shared frames and place these under each other does not work

It should look like this:

```
1 2 3 4 5 6 7 8
8 7 6 5 4 3 2 1
```

If you want to draw some lines, make a big X from 8 to 8 and from 1 to 1. Not sure how that will look for long trajectories, but that's what you're doing here.

You ***have*** to treat different move types differently. This is why these visualizations were previously only done for TPS, where you have one ensemble and you're just tracking the shooting move.

This is how it looked before. The way it looked before was good.

jhprinz:
> I don't understand this question. What is the "direction of the CV"? The CV is just a function that maps a snapshot to a number.

Come on. It's not that hard. The CV (or order parameter) in TPS is in the best case 0 for state A and 1 for state B and interpolates in between. Thus, if I have a trajectory that goes from A to B the CV increases (not monotone, but overall). That is the direction of the CV: From A (left) to B (right). Keeping the time direction in the reversed trajectory it will go from B (left) to A (right)  of course. Or is this wrong? Hence obviously the same frames will point in the reversed direction (as you write yourself) and their CVs will go from right to left. Which is in TPS NOT what you are interested in. Correct?

> You have to treat different move types differently. This is why these visualizations were previously only done for TPS, where you have one ensemble and you're just tracking the shooting move.

Of course. But to quote you: I think a user could be confused if every move is visualized differently (of course he could also be confused by reversing the time). But I do not see why the why you prefer is better and 

> This is how it looked before. The way it looked before was good.

Is no argument. This is just lazy. But if we all agree and this is stupid what I proposed I will not waste time implementing it of course. I just did not find any good argument to do it this way.

jhprinz:
Almost ready. Needs some more docs, but should otherwise ready to merge.

jhprinz:
Phew. Tests pass...

dwhswenson:
> Phew. Tests pass...

Does this mean I should be reviewing it now, or just that it's been a nightmare to get tests passing? (I've been having a ridiculous number of problems with `alanine.ipynb` timing out lately...)

jhprinz:
Well, the latter. I changed the Alanine Test. Apparently for some reason it had a `n_steps: 50` in it. Which results in long runs. It should be almost done. It needs more documentation though.

jhprinz:
More changes. I need to restructure a lot to make this work properly with rejected samples.

dwhswenson:
What's the status on this? I'm working on some things that might require changes to the move tree visualization, so I'm waiting to see what this provides before starting that.

BTW, if you're still working on this, please make it so that it doesn't depend on `type(mover)`. One of my upcoming projects (I want to mock in a shooting mover) shows exactly the why that's typically discouraged. Instead, it makes more sense for a mover to have to implement a `_visualization_type` class variable. That allows duck-typing the correct behavior, or inheritance of the correct behavior from a superclass.

openpathsampling/visualize.py
```diff
@@ -1,502 +1,394 @@
...
         with open('tree_xxx.html', 'w') as f:
             f.write(self.to_html())
 
-        bash_command = "wkhtmltopdf -l --page-width " + page_width + \
-                      " --page-height " + page_height + \
-                      " --disable-smart-shrinking " + \
-                      "-B 1mm -L 1mm -R 1mm -T 1mm tree_xxx.html " + file_name
+        bash_command = "open tree_xxx.html " + file_name
 
         os.system(bash_command)
```
dwhswenson: Can we find a way to do this that doesn't rely on a bash command? I believe that `open` is Mac-specific, and the previous `wkhtmltopdf` doesn't exist on my system. In general, I'd suggest avoiding the approach of writing a temporary html file (which you don't remove?!) and using an external process to convert from that. It's hard to make it platform-independent.

If we can't write to PDF directly, then we just give the options of writing to SVG or HTML. If the user wants to convert that to PDF, they can google "convert HTML to PDF" and find a solution for their system.

jhprinz:
almost ready... added a tutorial and documentation.

jhprinz:
Let's see...

I made some additional changes and cleanup to make the code more reusable. This means I added in visualization two helper classes
`SampleList`: an ordered dict that wraps a list of Samples and can compute lots of relations between them including the overlap and shift. This works for all sorts of moves and uses the parent property to find relations between the samples. `SampleList` has two basic settings

1. `time_symmetric`, if true will assume that we treat a snapshot and its reversed alike
2. `flip_time_direction`, something I find useful, but you might not. So default is false and it will try to flip trajectories, if that means to get an overlap. That is useful for moves that reverse the overlapping part like path reversal.

This object will also create a `SnapshotMatrix` which is really just to cleanup the code. It is a sparse matric of snapshots and you can check for relations, like father, root, etc...

The move type will affect only visualization. The SampleList analysis runs completely without needing the move type. It will only look at the trajectory. So the PathTree knows about move types because these will be visualized differently. The type is determined either from the class name and can be overridden by a `._visualization_class` class member string as you requested.

The different visualization works like this:

Each trajectory is split into parts forward, middle and backward. Where the middle part can mean overlap with the previous/parent trajectory, or also being new. So fw and bw only makes sense if there is an overlap somewhere. Overlap can also be reversed. That means there are 4 types:

1. `new`: These are the new parts, fw or bw or everything if it is the full trajectory.
2. `overlap`: the part in common with the previous
2. `full`: The part in common, if it is the full trajectory (useful, if you want to hide overlap but still see something if it is the full trajectory
4. `reversed`: The overlap part, if it is reversed.

All the parts are mutual exclusive and depending on the move only some of these will be present. For each you can chose one of 3 visualization types. There is a default, but you can override this per mover using the 
```
PathTree.options.movers['moverClassName][param] = value
```

There are 3 possibilities
1. `line` will make a thin line with the name of the mover in the middle
2. `block` will make one big block, so single snapshots can be seen. This is fast.
3. `single` will make little blocks per snapshot. Each little block, can be given a function for a text and color. This will result in large SVGs but with lots of nice effects.
4. `none` will do nothing

Default is, that new parts will be blocks and repeated parts are either ignored or plotted as a line.

As an example: The shooting move will make new parts as blocks, and hide common parts. That really cleans up the picture. The extend move will show new parts as blocks and the common parts as a line. This visualizes better the extension. ...

I will add the notebook to the tests, but am not sure if that makes sense.

You requested 

1. not using external programs for pdf generation. Done.
2. coloring of snapshots. Done. There is a `.coloring` attribute where you can set a coloring function

Last comments:

I completely split the visualization and the generation of SampleLists. That seemed cleaner. So you can use 
```
samples = SampleList.from_steps(steps, replica, accepted)
samples = SampleList.from_ancestors(sample)
```
to cover the most common cases. Note that the sample list does not contain the steps used.

##### If there are no steps, how to check for accepted?

This is up to the user to generate a list of samples what he wants. For now this is replaced by looking at the parent relations. What does that mean? The last sample in the list will determine the "goal". And all parents to this goal will be in the direct line of "evolution" of this sample and hence all of these are accepted. Other samples in the list must then be "rejected" because they are not important for the goal. Which does not mean that these samples could not have been derived in an accepted move. Especially, if we will have lots of little moves that will comprise a single move. 

This idea of "necessary for the goal" is easier to understand than coloring after accepted/rejected. If you are interested which parts of a move were accepted this is another question and we should address this in another visualization.

Note that this concept results in the same accepted / rejected coloring what we had before. But it allows for more complex cases.

#### PathMover vis

This works almost as before. Just using CSS now.

#### missing / future

A visualization for an actual step / move. I think that could fill in the gap we might have in the PathTree right now

jhprinz:
Should be ready for review when it passes tests.

dwhswenson:
A few questions for clarification before I look through the code:

> You requested
>
> not using external programs for pdf generation. Done.

Did you find a way to do it internally, or do we just say "we export SVG; the user can figure out how to make that a PDF"? I'm fine with either approach, just want to know what to look for.

> coloring of snapshots. Done. There is a .coloring attribute where you can set a coloring function

Does the coloring function have access to the "parts" like `new`, `overlap`, etc? And the question of `fw` or `bw`? In other words, can I write a coloring function that says "use `colorA` if we're backward and in `volume`, `colorB` if we're backward and not in `volume`, `colorC` if we're forward and in `volume`, and `colorD` if we're forward and not in `volume`, and show nothing unless we're `new`. (That's exactly what I'd like to do with these for the fixed-length TPS).

> I completely split the visualization and the generation of SampleLists. That seemed cleaner.

I was thinking about making some sort of trick for selecting certain samples for analysis, based on various criteria. Essentially, something that creates a generator that chooses things based on properties like "accepted", "canonical_mover", "replica ID", etc.

In the long run, it might make more sense to have that be the input to `SampleList`, instead of having `SampleList` do its own filtering by replica. (For now we leave as-is, but if we have a general way to filter steps down to relevant samples later, we should keep this in mind.)

> This is up to the user to generate a list of samples what he wants. For now this is replaced by looking at the parent relations. What does that mean? The last sample in the list will determine the "goal". And all parents to this goal will be in the direct line of "evolution" of this sample and hence all of these are accepted. Other samples in the list must then be "rejected" because they are not important for the goal. Which does not mean that these samples could not have been derived in an accepted move. Especially, if we will have lots of little moves that will comprise a single move.

What happens if you want to visualize only shooting moves, with no minus moves? My feeling is that, if the shooting is from the decorrelated part of the minus move, you should basically start a new tree separated from the first. I'm unsure how this "goal" based approach would work in that case -- especially if you chose the `SampleList` to only include shooting moves, which I think is a relatively likely thing for a user to try.

> A visualization for an actual step / move. I think that could fill in the gap we might have in the PathTree right now

Could you expand on this? I don't know what you mean here.

jhprinz:
Closes #337

jhprinz:
Added a `from_steps` to PathMoveTree. Now it is really a one liner. I also hid ReplicaExchangeMoves by default. Updating the last notebooks now.

jhprinz:
We really need to get rid of `alanine.ipynb` I probably restarted about 10 times in the last few hours. This costs so much time...

dwhswenson:
> We really need to get rid of `alanine.ipynb`

Agreed. For now, do you want to disable it and the tests that depend on it? I've started on a branch to remove it; I'll put that as a high priority (although much of today is/has been meetings on miscellaneous projects.)

`storage_tutorial` and `test_pyemma` depend on it; I don't think either (certainly not pyemma) are in danger of being broken by anything we're working on.

jhprinz:
This is the 5th time I restart... Man...

jhprinz:
if it passes should be ready to be reviewed.

I added a `tutorial_pathtree_visualization.ipynb` that shows some of the new tricks. And also the other notebooks should have been updated, if not merging screwed that up :)

jhprinz:
Last thing. I would like to place the CSS in a separate file which would then end up in `openpathsampling/visualization`. I will do this later when we finally fix the directory structure.

jhprinz:
Passing...

openpathsampling/engines/trajectory.py
```diff
@@ -290,6 +290,49 @@ def next(self):
 
         return ObjectIterator()
 
+    def index_symmetric(self, value):
```
dwhswenson: Could you add a brief comment to the docstring as to why one would want this? I'm not sure what this would be needed for.

dwhswenson:
Overall, looks good, and like there's a lot of customizability. We can finally make path trees that make sense to people with red-green color blindness! And individuals have a lot of design choices, letting them put their stamp on the output. I like that.

A few things to do/question of whether it is possible to do better:

1. Note that this has conflicts to resolve before merging.

2. Tutorial notebook: Why are the sample labels in cell 17 hex? The label I would expect there is a step number.

3. Remove the `vis_attic`. There's no reason to keep "attic" code around; it can always be recovered by git.

4. Can you clean up `tutorial_visualization.ipynb`? Most of it seems to have been superceded by `tutorial_pathtree_visualization.ipynb`, and there's a lot of uncommented code in there that just looks like development tests.

5. A couple minor things about the linker between trajectories (the vertical lines): Is it possible to change the colors of those? Could they be set to be in the background instead of the foreground? When they're relatively large, e.g. cell 19 in `tutorial_pathtree_visualization`, I think they'd look better in the background (or maybe just setting `join_blocks=False` would help).

6. Maybe switch the decorrelation marker to an I-beam (go through the vertical line to the other side, instead of just going in one direction)? Again, when the lines are large (cell 19), the very top marker could be smoothed to look better.

Specific notes on `visualization.py` (which was too large of a diff to do line notes):

* usage of `.c` as a method: it is unclear what this is supposed to do
* usage of `cls` as a variable: Note that `cls` is often used in place of `self` for class methods. In the same way that it is legal, but really confusing, to use `self` for something other than the standard “this instance” pointer, it is very confusing to read code that uses `cls` as something other than a class type. See [PEP8](https://www.python.org/dev/peps/pep-0008/#function-and-method-arguments). Could you change this variable name to something more descriptive?
* in general, more docstrings and comments would help. I only skimmed this, and it is hard to get a real sense of what is going on in the 2100 lines of code, especially with a lot of very abbreviated variable/method names that don't have docstrings.

jhprinz:
I remembered that you really didn't like the implementation not using steps so I changed that, too. 

```py
pt = PathTree(
    steps,
    ReplicaEvolution(
        replica,
        accepted
    )
)
```

Instead of `ReplicaEvolution` there is also `EnsembleEvolution(ensemble)` and `SampleAncestors(sample)`.

All the rest is hidden besides all the options. I added some more examples in the tutorial and ` tutorial_visualization.ipynb` is to be removed.

To fix the foreground/background issue I completely turned all subparts of the plot

1. shooting hooks
2. trajectory labels
3. snapshots / blocks
4. legend

into sub functions and all of these can be changed separately and the order in which they are plotted. I think it is a little slower now but all the analysis is only computed if necessary.

> which was too large of a diff to do line notes

Yes, a diff would not make sense when I rewrite the whole file completely.

> Note that cls is often used in place of self for class methods. In the same way that it is legal, but really confusing, to use self for something other than the standard “this instance” pointer, it is very confusing to read code that uses cls as something other than a class type. See PEP8.

No comment...

jhprinz:
Ready for another round to review...

jhprinz:
Just as a comment. I think the way with steps it works now is really nice and easy to use. It also produces quite nice pictures with lots of flexibility if you need it.

I am quite happy with it now. Have fun using it.

dwhswenson:
Ah! `css_class`! Seriously, that little change made the code SO much more readable. I had been utterly confused before when I kept thinking `cls` had something to do with Python classes, but then the code made no sense. :-)

> Instead of ReplicaEvolution there is also EnsembleEvolution(ensemble) and SampleAncestors(sample).

This may have overlap with stuff in `analysis/replica_network.py`. We should be careful we're not doubling work here, and should probably combine some of those functions. (Not needed for this PR, but keep in mind that there are already a lot of tools for analysis.)

> To fix the foreground/background issue I completely turned all subparts of the plot

Nice -- I think the new version looks even better.

One thing I really like about this is that there's a nice balance of "the OPS signature look" with "the user can customize." I mean, we'd know from looking that a user was probably using OPS to generate these, based on the kinds of things that can be visualized. But with the flexibility in colors and presentation, users can also put their own stamp on it.

Issue #70, opened 2014-11-13
compatible to Python 3


jhprinz:
Coule we change the code so that it runs also under Python 3?
I had the idea to add a Siegtank converter which would allow to import a complete Siegetank project into our netCDF Storage and continuously sync it, as it progresses.
The idea was to just get the actual status of a siegetank project. Get all the trajectories one by one, strip water and store the essential part (I assume that I don't need water and or momenta) in our storage.
Unfortunately Siegetank requires Python 3 and to avoid having to seperate scripts it would be great to keep out code also runnable (if not optimal) under Python 3. 

This seems to involve mainly
- change `print` to `print()`
- take care of unicode strings
- rewrite some iterables (e.g. iterates() does not work in Python 3)

which is not too much work. Most of it can be done automatically.

Would that be a good goal? I mean importing from Siegetank is a nice gimmick but potentially not relevent unless we want to use Siegetank at some point to run our trajectories (which would be compatible withe the next simulation generator scheme)

jhprinz:
I just saw that my idea has a compatibility check mode and it will highlight all parts that are not 2.7 and 3.5. I will have a look at how complicated that is...

dwhswenson:
This has been superseded by #714. Closing with the intent of keeping further developments there.

PR #404, opened 2015-12-13
Updates to the install process

dwhswenson:
Just adding a little more to the docs for installing. I've been telling people that OPS is ready for them to look at; it would probably help if the docs gave accurate instructions on how to install.

Ready to merge. Only changes are in docs.

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -24,7 +24,7 @@ omnia channel. ::
 To install the most recent release of OpenPathSampling with conda, use the
 following commands ::
 
-  $ conda config --add channels http://conda.binstar.org/omnia
+  $ conda config --add channels omnia
   $ conda install openpathsampling
 
 If you want the cutting edge of what OpenPathSampling can do, you can
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -14,12 +14,22 @@ Install with Conda
 ------------------
 .. _install-with-conda:
 
-`conda <http://www.continuum.io/blog/conda>`_ is a python package manager built for scientific python. Unlike ``easy_install`` or ``pip``, it handles binaries and binary dependencies, which are critical for most scientific workflows. If you're a ``conda`` user, you can install OpenPathSampling by adding the omnia channel. If you're not a conda user, you should look into it. ::
+We recommend using `conda` to install OpenPathSampling.  `conda
+<http://www.continuum.io/blog/conda>`_ is a python package manager built for
+scientific python. Unlike ``easy_install`` or ``pip``, it handles binaries
+and binary dependencies, which are critical for most scientific workflows.
+If you're a ``conda`` user, you can install OpenPathSampling by adding the
+omnia channel. ::
 
-To install OpenPathSampling with conda, use the following commands ::
+To install the most recent release of OpenPathSampling with conda, use the
+following commands ::
 
   $ conda config --add channels http://conda.binstar.org/omnia
   $ conda install openpathsampling
+
+If you want the cutting edge of what OpenPathSampling can do, you can
+install the development version using `conda install openpathsampling-dev`.
+(Also, if you're reading this prior to the first official release.)
 
 .. note:: ``conda`` will automatically install all of the tricky dependencies
     from binary packages automatically! This includes pytables / numpy / scipy!
@@ -37,11 +47,13 @@ Then, in the directory containing the source code, you can install it with ::
 
   $ python setup.py install
 
+You will have to manually install the requirements and dependencies.
+
 Dependencies
 ============
 
 To use OpenPathSampling, the following libraries and software will need to
-be installed.
+be installed. (TODO: this list is partial)
 
     Linux, Mac OS X or Windows operating system
         We develop mainly on 64-bit Linux and Mac machines. Windows is not
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -23,6 +23,7 @@ requirements:
     - mdtraj
     - svgwrite
     - networkx
+    - matplotlib
 
   run:
     - python
@@ -39,6 +40,7 @@ requirements:
     - mdtraj
     - svgwrite
     - networkx
+    - matplotlib
 
 about:
   home: https://github.com/choderalab/openpathsampling


docs/getting_started.rst
```diff
@@ -14,13 +14,23 @@ Install with Conda
...
+scientific python. Unlike ``easy_install`` or ``pip``, it handles binaries
+and binary dependencies, which are critical for most scientific workflows.
+If you're a ``conda`` user, you can install OpenPathSampling by adding the
+omnia channel. ::
 
-To install OpenPathSampling with conda, use the following commands ::
+To install the most recent release of OpenPathSampling with conda, use the
+following commands ::
 
   $ conda config --add channels http://conda.binstar.org/omnia
```
jchodera: You just want to use `omnia` instead of `http://conda.binstar.org/omnia` now.

In fact, you may just tell them that they want to do `conda install -c omnia openpathsampling`

dwhswenson: Is that `conda install -c omnia openpathsampling` as a replacement for both the commands we had previously? I'll update the docs to use that, but want to make sure they're correct (I don't know since I don't use conda, myself).

Mainly, I wanted this to have a mention of the `-dev` package, since that's the only thing that is actually currently available.

jchodera: > Is that conda install -c omnia openpathsampling as a replacement for both the commands we had previously?

Yup!

>  I'll update the docs to use that, but want to make sure they're correct (I don't know since I don't use conda, myself).

It's worth testing, and won't impact your existing installation:
```bash
wget --quiet https://repo.continuum.io/miniconda/Miniconda-latest-MacOSX-x86_64.sh
bash Miniconda-latest-MacOSX-x86_64.sh -b -p miniconda
export PATH="$HOME/miniconda/bin:$PATH"
conda install -c omnia openpathsampling-dev --yes
```
Are you sure you don't want to create a non-dev version, even if it's just a preview release?

Also, what's up with no py3.x support?

franknoe: Isn't the difference that in ```conda install -c omnia ...``` the 
channel is just used in this install command, rather than being added to 
your conda config? In other words, if you install by this one-liner and 
want to update subsequently, can you just do ```conda install 
openpathsampling``` or do you have to add ```-c omnia``` again?

Am 13/12/15 um 17:36 schrieb John Chodera:
>
> In docs/getting_started.rst 
> <https://github.com/choderalab/openpathsampling/pull/404#discussion_r47446925>:
>
> >
> > -To install OpenPathSampling with conda, use the following commands ::
> > +To install the most recent release of OpenPathSampling with conda, use the
> > +following commands ::
> >
> >    $ conda config --add channels http://conda.binstar.org/omnia
>
>     Is that conda install -c omnia openpathsampling as a replacement
>     for both the commands we had previously?
>
> Yup!
>
>     I'll update the docs to use that, but want to make sure they're
>     correct (I don't know since I don't use conda, myself).
>
> It's worth testing, and won't impact your existing installation:
>
> wget --quiet https://repo.continuum.io/miniconda/Miniconda-latest-MacOSX-x86_64.sh
> bash Miniconda-latest-MacOSX-x86_64.sh -b -p miniconda
> export  PATH="$HOME/miniconda/bin:$PATH"
> conda install -c omnia openpathsampling-dev --yes
>
> Are you sure you don't want to create a non-dev version, even if it's 
> just a preview release?
>
> Also, what's up with no py3.x support?
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/choderalab/openpathsampling/pull/404/files#r47446925>.
>


-- 

----------------------------------------------
Prof. Dr. Frank Noe
Head of Computational Molecular Biology group
Freie Universitaet Berlin

Phone: (+49) (0)30 838 75354
Web:   research.franknoe.de
Mail:  Arnimallee 6, 14195 Berlin, Germany
----------------------------------------------

jchodera: Also, you're missing some dependencies in your conda `meta.yaml`.  After installation on `osx`:
```
>>> import openpathsampling
  File "/Users/choderaj/miniconda/lib/python2.7/site-packages/openpathsampling-0.0.1a0-py2.7.egg/openpathsampling/__init__.py", line 69, in <module>
  File "/Users/choderaj/miniconda/lib/python2.7/site-packages/openpathsampling-0.0.1a0-py2.7.egg/openpathsampling/analysis/__init__.py", line 1, in <module>
  File "/Users/choderaj/miniconda/lib/python2.7/site-packages/openpathsampling-0.0.1a0-py2.7.egg/openpathsampling/analysis/histogram.py", line 3, in <module>
ImportError: No module named matplotlib.pyplot
```

jchodera: To update, I think you'd have to do `conda update -c omnia openpathsampling`.

It's better in general to add the channel (`conda config --add channels omnia`), but that's a bit more advanced than the one-liner needed to install.

dwhswenson: > Are you sure you don't want to create a non-dev version, even if it's just a preview release?

I have no opposition to this, but we hadn't done it. I'm just trying to make the docs reflect reality.

> Also, what's up with no py3.x support?

See #70. We decided this wasn't a priority for 1.0, but will be done soon after.

> It's better in general to add the channel (`conda config --add channels omnia`), but that's a bit more advanced than the one-liner needed to install.

I don't think a 2-line install is ***that*** much worse than a 1-line install, especially if it makes updating easier for the complete novice (which is pretty important for the pre-release versions). I'll switch it to just `--add channels omnia`.

> Also, you're missing some dependencies in your conda `meta.yaml`.

Added `matplotlib`. @jhprinz : Can we *use* the conda install for OPS on Travis instead of this separate `conda-requirements` file? That would warn us of problems like that. We would also need to find a way to distinguish between required and recommended package -- something we need to do in any case, to update the non-conda install instructions.

jhprinz: The travis part is 2-fold: 
- A conda package is build using the `conda-requirements` file. 
- Then a new miniconda environment is created and the previous installed package in installed from _memory_. 

This second installation is used for testing.

Once all tests pass and only PRs on the master branch the docs are built, uploaded to amazon and the new version is uploaded to anaconda.

So we need the requirements to build the package but not afterwards. If this is still the case this should be changed.

Not sure how to have recommended packages. I presume that conda does not support this natively. We could add this to setup.py somehow?

There is also still the question if we want to add the example notebooks to the package or only put them online.

dwhswenson: My point is that our Travis tests don't check whether you can actually `conda install` OPS, because they don't use the `meta.yaml` that conda uses when it installs stuff. At least, that's how I understand the fact that @jchodera got a "missing matplotlib" complaint.

The problem is that `conda-requirements` is not tied to `conda/meta.yaml`. Is there a way to fix that? If I'm understanding correctly, `conda-requirements` are not actually requirements, and not actually used by conda directly. They're "things we install using conda on Travis before we run OPS." Does conda itself ever make use of the `conda-requirements` file, or only the `meta.yaml`? Don't we want to actually install using the stuff in `meta.yaml` and make sure that we can run based on that?

My thought with "recommended packages" is that these would be other packages that aren't part of the conda `meta.yaml` but are part of the test suite. We might want to separate the testing into stages (or even separate tests) to make sure that all works correctly.

Short version:
* Currently, our tests don't check whether the conda installation will actually work. From what John had, it appears that if a user just tried to install OPS, with nothing else, the result **would not work**. We need to fix that. What's the best way to fix it?
* We will have some packages that are truly *required* for OPS to run, and some others that we want to include in testing to check that other features work. How do we manage that?

jhprinz: Sorry, then I got your comment wrong. If I read the documentation of `conda build` correctly then we do not need the `conda-requirements` at all. The build will download all the files itself. Unfortunately I cannot test this on my machine becuase the require `conda-build` can only be installed in the root environment...

I will make a PR and see if that is the case. We could then simplify the procedure and remove conda-requirements

PR #405, opened 2015-12-14
remove conda requirements

jhprinz:
testing


dwhswenson:
You might merge #404 first: I added stuff to conda/meta.yaml there

jhprinz:
Merged #404 into this and did the following changes:

1. Remove `conda-requirements`. This was apparently used to install conda packages because the `conda install path_to_local_package/.../openpathsampling-*` did not install the locally build package with dependencies but only the packae itself.

2. Changed the conda install to 

    ```shell
    conda install --yes --use-local --update-deps openpathsampling-dev
    ```
    This should install openpathsampling as any other package with dependencies (the update part should not be needed and is just to be sure) but from the current local build environment path.

3. Some cleanup of code in `.travis.yml` that is not needed anymore. 

4. The tests require also `msmbuilder` and `pyemma` to run so I added these to the _run_ requirements in `meta.yml`


It will now really install in a fresh conda environment and hence check if all dependencies are present.

jchodera:
You might check out some of our more recently updated .travis.yml files for
projects like torsionfit. These build the conda package and then only
import it in the test: section of the meta.yaml, but then activate the
conda _test environment to do the actual testing in .travis.yml in an
environment that matches what would be installed on a user machine.

dwhswenson:
> The tests require also msmbuilder and pyemma to run so I added these to the run requirements in meta.yml

This is where I think we need to figure out how to handle recommended packages. A user doesn't need to have msmbuilder or pyemma installed to use OpenPathSampling, and I thought the plan was that these would not be conda requirements either. However, they need to be used for the complete tests. Same with all the jupyter/ipython stuff -- not required to use OPS, but highly recommended, and absolutely required for the test suite.

I think we should test two cases: one where we have a minimum setup, and one where we have all the recommended packages. We could do this by putting the extra tests (which make use of recommended packages) in a separate directory, and doing the following sequence:

1. Minimum install: OPS as a conda user would see it, with no extra packages
2. Nosetests for minimum install
3. Install recommended packages in conda environment
4. Nosetests for extra things that are now available
5. Integration tests (only with everything installed)

Technically, we could use a build matrix in Travis to do this, but I think that is overkill -- we just want to make sure that a first-time conda user can install, `import openpathsampling`, and do some simple stuff without errors. We don't need to do a full test suite (including relevant integration tests) on the minimum install -- that will get harder to manage.

That stuff in step 3 might be just installing using a text file list of recommended packages like the old `conda-requirements` file -- my main concern is that we need to make sure that, in step 2, a first-time user doesn't get an error when they `import openpathsampling`.

jhprinz:
John had the solution for this problem at least most of it:

Use the test section of conda build process. Conda build will create two additional environments always `_build` and `_test` and you can set which packages to install in which. So I moved all essentials to `run` and `build` which should be the same besides `build` needs setuptools and moved `msmbuilder`, `pyemma`, `jupiter_client` (still needed for ipynbtest), `nose`, etc... to being installed in the `_test` env only. We only need to make sure that you can build the package correctly.

Only missing point is that if a user installs ops this way. it will work, but features from pyemma etc will not be available. So some output or info in the conda package would be useful.

jchodera:
While I understand the desire to have a "minimal" package that has only what you need for core functionality, encouraging people to install this package might leave them disappointed or frustrated when the tutorials don't work.  

What are the real pressures for having a "minimal" package only?  Is it worth the user frustration this might cause?

An alternative might be to encourage people to install a [conda metapackage](http://conda.pydata.org/docs/commands/build/conda-metapackage.html) you put in https://github.com/omnia-md/conda-recipes that pulls in the minimal package plus all the bells and whistles that make for a nice user experience.

dwhswenson:
There are two questions here, and although they're related, we should keep the distinctions in mind:

* What are the base packages that come with the "default" OpenPathSampling install?
* How do we manage tests for packages that are outside the "default" install?

Metapackages might be a partial solution, although I think we still need to think about how to test code with dependencies that aren't conda-installable.

Context for my comments: I'm thinking about the 3-5 year plan for OPS, at which point I hope we'll support many engines, including OpenMM, Gromacs, NAMD, LAMMPS, QChem, CP2K, and probably others. There's no reason to require that a user install *all* of those just to run the basic tests. At some point, requiring too much install space will turn users off (and some engines are commercial software, meaning installation *must* be the user's responsibility).

Similarly, we'll have wrappers for collective variables from many sources, including MDTraj, PyEmma, MSMBuilder, MDAnalysis, Curves+ (via pyna), and possibly many more. Again, no reason to require a user to install all of those packages just in order to run the basic tests.

Metapackages at the level of "minimal", "default", and "full" might be a good start, but keep in mind that some software we should be able to integrate with (e.g. QChem) will *never* be conda-installable. We still need a way to run tests that include supported, but not required, dependencies.

jchodera:
I'm just encouraging you to think about the user experience, and how you will instruct the user to install what is needed to do what they want.  Maybe it's sufficient to print an error message that gives instructions for how to install the missing bits if they run into some component not being installed?  Or maybe you can offer to shell out to `conda` and install it for them if it can be easily done from inside the Python code (assuming that they are using things interactively, of course)?  Or maybe there could be an `openpathsampling.automatically_install_dependencies = True` global option that will try to do this automagically if it can?

jchodera:
Just noticed this, BTW:
* https://anaconda.org/rmcgibbo/qchem-utils
* https://github.com/rmcgibbo/qchem-utils

dwhswenson:
> Maybe it's sufficient to print an error message that gives instructions for how to install the missing bits if they run into some component not being installed?

This was my thought. There needs to be a check for failure if something isn't installed. If you try to use `QChemEngine` and there's no `qchem` on your system, you should get a clear error. The only question is whether this should happen at instantiation of the engine or on first use. We can point them to QChem's website if they want to use that, or to the Gromacs website if they want Gromacs and it isn't installed. (It does need to be a *very* clear error message, saying "I looked `here` for `this thing` and couldn't find it -- that can help catch cases where the install is in an unexpected location).

And specifically for user-friendliness, I want such features (if we officially support the engine) to be installed with the default package. We could always put every engine into a separate install process, but that's not user-friendly (`conda install openpathsampling openpathsampling-qchem openpathsampling-gromacs` is ugly). The problem is that those features don't work unless something else is installed.

> Or maybe you can offer to shell out to conda and install it for them if it can be easily done from inside the Python code (assuming that they are using things interactively, of course)? Or maybe there could be an openpathsampling.automatically_install_dependencies = True global option that will try to do this automagically if it can?

These are cool ideas, but limited in practice. It could work for things that are freely downloadable (Gromacs, Curves+) but not for things that aren't (QChem). There's also a need to keep the automagic install up-to-date with filename updates/directory restructures on anything where we pull the code from elsewhere. Finally, this can fail due to things outside of our control: e.g., the Curves+ website was down for a few months this summer (and the dev didn't even know until, about a month in, I emailed him about it).

> Just noticed this, BTW:

Interesting. I'll have to look it over to see how well it supports MD within QChem -- or might contribute that feature.

jhprinz:
> Or maybe you can offer to shell out to conda and install it for them if it can be easily done from inside the Python code (assuming that they are using things interactively, of course)? Or maybe there could be an openpathsampling.automatically_install_dependencies = True global option that will try to do this automagically if it can?

I found a potentially nice way to handle this. For once `conda` is also a python package which you can run from within python and check installed packages install new one, etc... This might make it easy to add this to `setup.py` if we wanted to.

Another quite elegant way is to use `ipython notebook extensions`. These are little hack of Ipython that add features like search/replace within notebooks, add a TOC to the notebook for quick access of parts, etc... Very nice is that it is conda installable and it ships with a special extension that add a config panel to Jupiter where you can select which extensions you want active and you can set parameters. You access this by opening `localhost/nbextensions` in your browser

https://github.com/ipython-contrib/IPython-notebook-extensions

Yesterday I tried a little to tweak this and it can be used to show all available conda packages and show which ones you have installed. Far from being perfect but we could adapt this to verz convenientlz show possible extentions to ops and let the user install these with a click. This is an example

conda

I would think that in general this would be a nice addon for conda users...

jhprinz:
To close this for now. I think the new rearrangement of packages etc should fix the initial issue. I have also updates ipynbtest to require jupiter for testing and travis seems to pass the installation process.

One last nice thing I learned about `meta.yml` is that when you want to build the package locally and you want to mimic checking out the current branch and build it there is a trick to do so. The advantage is that it will not copy you full source folder to a build environment and do the building there, but it will only copy files add in the git repository. Meaning it will exclude all .gitignore files conveniently. To do so just change in `meta.yml`

```
source:
#  path: ../../
   git_url: ../../.git
```

and you can run `conda build` without problems.

jhprinz:
Ready to for review and merge. Note that this includes changes from #404 and shoudl close this one too!

jhprinz:
One it passes of course!

dwhswenson:
> Another quite elegant way is to use ipython notebook extensions.

This is pretty cool, although I don't think we would want to require users to use that (just as we don't *require* them to use IPython for anything). But that seems like a really convenient trick for practical use.

As for the changes in here: they looks good; just keep re-running tests until they pass (any idea what has made tests less likely to pass lately? Stochastic failures seem worse than they used to be).

dwhswenson:
Finally tests passed! (I have no idea how many trials that took -- we need to look into what has changed there.)

Merging!

Issue #372, opened 2015-11-11
Improved handling of named objects


jhprinz:
This is one of the two major points I think needs improvement in the storage (the other is about making stores storable -- different story): Better handling of named objects.

So far this works that we explicitely specify which objects (e.g. PathMovers, Engines, etc...) can have names and which ones don't (like Snapshots, Configurations, etc...). In general objects that contain running data like Samples, Snapshots, Trajectories do not have names while objects that create or operate on data (like engines, cvs, Movers, volumes, etc) do have names. For most object types only a small subset of objects will actually have a name while the others do not.

The setting of one object the allow for names will create an additional `[object]_name` variable in the netcdf file which will store names with the objects and most of the names will be empty or not used. It also requires handling and loading of all the names. Caching, etc...

I was wondering if it might be better to just add a special store to the storage which is called `.named`. It acts like a dictionary and references objects of _any_ type. While usual stores are more like a list `int -> object` this would be `str -> obj`. I would propose that names have to be exclusive, so one name can only point to one object or anything that can be pickled while the same object can have different names! Exactly like in a dict. 

When a file is opened the name dictionary will be cached so that access is fast.

#### Loading and saving of named objects

Usual load access would work like a dictionary
```
stateA = storage.named['stateA']
```
or using attributes
```
stateA = storage.named.stateA
```
and saving should work the same
```
storage.named['tis_AB'] = my_tis_ensemble
```
or we use a save function (different possibilities)
```
storage.named.save_as('tis_AB', my_tis_ensemble)
storage.save_as('tis_AB', my_tis_ensemble)
my_tis_ensembles.save_as(storage, 'tis_AB')
```

After saving there will be an entry in the store using two dictionaries
```
name_key : str
    contains the key of named object idx
name_json : str
    contains the json description of names object idx
```

#### Advantages

1 it's fast
2. requires minimal storage
3. You can store any object, like specific snapshots or even python dictionaries if you want, like meta information about a simulation. Anything that the ObjectJSON.simplifier can turn into JSON is acceptable.

#### Possible problems

The only problem I so far see with this approach is the detachment of name and object. If you load an object from a store and it has name then you do not know about it, unless you check the name cache
```
stateA = storage.volumes[0]
stateA.name  # ??? This will not work easily
```

We can of course check naming when an object is loaded, but that would trigger a search in the cache for every load operation which is not from cache. I think that is not optimal and should be avoided.
To deal with this we could

[...]

PR #406, opened 2015-12-18
A dictionary like store

jhprinz:
This solves issue #372 and allows to add a dict like store where you can do this:
```
storage.my_dict_store['xyz'] = [any storable object]

#examples would be
storage.tags['allstates'] = [stA, stB, stC]

#loading
stA, stB, stC = storage.tags['allstates']
```

it also contains some improvement and cache fixes to work with loading and saving named objects in general. Now you can also do loading by name. It will naturally pick the last object with that name. This way you can even overwrite definitions if you want to.

```
stA = storage.volume['A']
```


dwhswenson:
> `stA = storage.volume['A']`

`volume` or `volumes`? (Assuming this might just be a typo -- if you're just adding a new access method to the existing store, it is already called `volumes`, right?)

jhprinz:
Typo, yes sorry... Just what you would expect to work... So these three are equivalent now
```
stA = storage.volumes['A']
stA = storage.volumes.load('A')
stA = storage.volumes.find('A')
```

dwhswenson:
That sounds perfect. Does this then require that any (user-defined) named object be unique within that store (I think we'd discussed this and decided that was the way to go). By that, I mean that you can only have one volume named "A", but you could also name an ensemble "A" if you wanted to.

jhprinz:
So you can now add a dictionary store to you storage file using
```
storage.create_store('tag', DictStore())
```
after that you have a dictionary name `.tag` and you can set
```
storage.tag['somekey'] = someobject
myobject = storage.tag['mykey']
```
I thought of having two of these, one for meta data and one for the user to name global stuff.

The dictionary can be unique but does not have to be. If you use
```
storage.create_store('tag', DictStore(immutable=True))
```
You can only have one object with a unique name
```
storage.create_store('tag', DictStore(immutable=False))
```
allows you to have more. Then you can _overwrite_ using the same name. You will always get the last stored back unless you use
```
storage.tag.find_all('name...') 
```
and you get a ordered list of all objects. Oldest objects first.

i want to add the same functionality to nameable stores. Basically it is a little test that when you use an existing name it does not store the object and throws an error. 

My question would be if we want really an error or just a warning and store anyway without a name?

Last added feature is that you can also do now
```
storage.save(volumeA, 'stateA')
#or
storage.volumes['stateA'] = volumeA
```

so saving with optional string index will try to name the object first and then save it as usual
```
storage.store[key] = obj
# will effectively do
obj.name = key
storage.save(obj)
```

I also improved logging.

dwhswenson:
If I understand this correctly, you're saying that if you try to give the same name to two objects, you append instead of overwriting (or throwing error). I think that will be confusing. If I write:

```python
storage.some_dict['name'] = my_obj
```

then I expect that, no matter what else happens, I'll get

```python
>>> storage.some_dict['name']
my_obj
```

If somewhere else in my code, I did `storage.some_dict['name'] = other_obj`, then my intent is ambiguous -- maybe I want to replace that name (which does weird things if we've already stored) or maybe I don't know that name is already in use. If user intent is ambiguous, I think it is better to not guess what the user intends, and to throw an error.

Imagine you and I were working on setting up a simulation by passing the `.nc` file back and forth (a really cool possibility with the storage tricks you've developed!) If you gave one object a name, and I gave a different object the same name, then there's going to be all sorts of confusion as to which name is being used. Good reason to throw an error.

jhprinz:
I agree, making the naming unique is better. I just kept it to be backwards compatible. I was also afraid that in some cases automatic things could try to name things again and then the code would break. 

So far I believe this has not happened so I can just set the default to unique no matter what. For special cases of storage a user that really know the details could turn of uniqueness. 

For example overwriting the last access or opentime (although there are better solutions) would be something that makes sense like
```
storage.tag['lastaccess'] = dateoftoday
```
This would work out of box with a mutable dict. And you keep the list of opened dates also...

Anyway. We should make sure that it defaults to a case where no problems can occur.

jhprinz:
Last question: For a DictStore it seems to make sense to implement `__iter__`, `.keys()` and so on. Also `__iter__` would then loop over all names/keys and not over the objects, since it makes no sense to loop over unrelated objects without the name.

dwhswenson:
> Last question: For a DictStore it seems to make sense to implement `__iter__`, `.keys()` and so on.

Definitely.

> Also `__iter__` would then loop over all names/keys and not over the objects, since it makes no sense to loop over unrelated objects without the name.

It should match the standard behavior of the Python dictionary (which I had to test to remind myself what it is: it loops over keys). But I think the main motivating factor here is to act like a `dict`.

jhprinz:
Correct. So it acts like a dict now and I added convenience `to_dict()` method to get the complete dictionary (for whatever reason).

jhprinz:
So, close to last updates. I added the possibility to force a store to have unique names. All works automatically and even in nested saving this is taken care of. So we have to decide which stores should have unique names. I would like to set this for all named stores as default.

This has one problem though. When I set all stores to use unique names then suddenly tests fail since we have not taken too much care of unique names. At least in some cases. E.g. The `TPSNetwork` breaks when initial and final states are the same. In construct a joined initial and final states and uses the same naming convention. For now I added a check for equality and then use the same joined volume instead of two distinct ones. Another way would be to add `final` and `initial` to the names itself.

Thoughts?

jhprinz:
I decided to set `unique_name` to default to False to make this pass. Especially in Volumes and Ensembles it seems to happen quite often that we generate objects with duplicate names. I guess we have to think about this.

jhprinz:
This is part of the unified PR #409. And since it passes tests I call this done. I will be changed later anyway.

dwhswenson:
> I decided to set unique_name to default to False to make this pass. Especially in Volumes and Ensembles it seems to happen quite often that we generate objects with duplicate names. I guess we have to think about this.

This is something I've been meaning to change -- the problem is that a lot of the ensemble names stay the same regardless of the transition they're associated with. The name should be clearer (look through some of the pandas output in the example notebooks and you'll see cases where this is unclear.) It's on my to-do list, just not as high on it as other things.

Will merge this as is. Any other discussion will be in #409.

PR #412, opened 2016-01-08
Doc updates

jhprinz:
I started correcting the docstrings and fixing all the errors in the sphinx build to the current version. 

To be merged after PR #409 since this PR had alrady much changes of docstrings in it. Once that is done this PR should be easy to merge and will not affect any extisting code only docs.

- fix sphinx extension to include ipython nodebooks and markdown files


dwhswenson:
Looks ready to merge. A few questions/notes:

* Question: Why do only some docstrings get marked as raw with `r"""string"""` -- should we be doing all docstrings this way? (I'd rather not, if not necessary)
* Question: How does some of the extra markup in docstrings (e.g., `:class:` markers) affect in-Python documentation reading (from IPython's `?`, for example)? I would hope that doesn't make things harder to follow; if so, we might look for another way around it. I expect I'll mainly use the in-Python stuff.
* Several of the docs that I've started and that you include within the web-accessible docs (mostly under `guides`) are very incomplete, which is why I hadn't included them yet. We might remove them from easy web access until they're a bit more complete.

Still, I'll merge this as is so we can move forward.

PR #415, opened 2016-01-12
[WIP] Features to CVs (or engine native CVs)

jhprinz:
### Summary

We want to treat some Snapshot features as CVs, like energies. This is already implemented in this PR. But...

We could remove the distinction between Snapshot features and Snapshot CVs and treat them the always same. The would remove the need for different Snapshot Classes and we do not have to define features. The old features are treated as CVs and CVs know how to store themselves. The user API to interact with snapshots, configurations, cvs, etc does not change.

Do we want that? 

#### Pros
1. It is more elegant and easier for the user since he does not have to define features or even worry about them. The engine gives a base Snapshot type with all features it requires (or supports) and the user can add to that. 
2. We could keep the current API the same. No obvious changes 
3. Some cool caching possibilities, like we do for snapshots so far
4. WE could keep the different Snapshot classes as predefined Templates of `Snapshot()` + CVs
5. The engine takes responsibility to define the base Snapshot, but in a very transparent and extensible way.

#### Cons
1. It could have some performance overhead, we would have to test that.
2. The engine takes responsibility to define a base Snapshot in a very extensible way. More difficult to check if the given Snapshot is of the correct type

### So far

kinetic and potential energy are now also CV and handled as such. The base class CollectiveVariable now implements a user settable CV that can 

```
my_snapshot.my_cv = value
```

be accessed using the normal setter method of snapshots.

### Motivation and Discussion of realization

For reasons of 
- generality of CVs
- compatibility with more MD engines
- generality

we should allow an engine to return also CVs with their returned frames. The simplest example for this are energies or even box_vectors.

These could be optional and fall into the category of a CV. In the most geenral sense this is also tro for the coordinates and velocities, too. Like OpenMM supports reporters for selected computed quantities, we would say, that there is the potential abstract snapshot plus CVs that depend on this snapshot.
Now I think we should keep the snapshot as the basic object that has engine dependent basic necessary properties, all optional properties are CVs. Still, we could think about making _all_ properties into CVs and the snapshot is empty. 

This would not change the way the user interacts, but change the internal responsibility for saving information about snapshots. It would also replace the necessity of using features. Since in essence a feature is a CV. 

One consequence is that that we will not have different Snapshots anymore and hence the creation of Snapshots by the user looks complicated. Basically it would then look like this:

```python
# These would by default be provided or set by the engine!
Snapshot.add_cv(
  CollectiveVariable('configuration')
)
Snapshot.add_cv(
  CollectiveVariable('momentum')
)

my_snapshot = Snapshot(
    coordinates = my_coordinates,
    box_vector = my_box_vector
)

storage = paths.Storage('name.nc', template=my_snapshot)
```

So, the `CollectiveVariable` command replaces setting the feature specific part for each Snapshot class. Since CollectiveVariable can (in most cases) figure out how to effectively store themselves you do not need to set anything. For the user this is completely transparent and simple as before. 

What is missing to make this work is that the following lines in the feature have no counterpart yet:

```python
store.storage.create_store('configurations', ConfigurationStore())

store.create_variable(
    'configuration',
    'lazyobj.configurations',
    chunksizes=(1,)
)
```

The second part could be easily (except the `lazy` flag) inferred from the object itself. The first cannot - we have to define which ObjectStore (here `ConfigurationStore()`) is supposed to handle `Configuration` objects. This is where we (or a potential power user) has to provide a store that is added automatically when a CV referring a certain type of objects is stored.

```python

# get the engine specific Snapshot type
ActualSnapshot = engine.snapshot_class

# will call internally
ActualSnapshot.add_cv(
  CollectiveVariable('configuration'),
)

# as usual
my_snapshot = ActualSnapshot(
	configuration = Configuration(...)
)

storage = Storage('name.nc', template=my_snapshot)

# will call internally
all_cvs = my_snapshot.cvs
for cv in cvs:
	# this will store all not yet stored cvs
	storage.cvs.save(cv)

	# the following code is part of cvs.save
	
	var_type = estimate

	# this will figure out for the one cv we have
	# that is will store a `Configuration` object
	obj_class_to_store = type(cv(my_snapshot))
	
	# figure out which store is needed for this
	store = storage.find_store(obj_class_to_store)
	
	if store_to_use is None:
		# no store exists, look in list of available stores!
		# and create it		
		store = storage.create_default_store(obj_class_to_store)

	# this will create a variable to store the data of the object type
	storage.cvs.create_variable(
	    cv.name,
	    cv.cv_loads_lazy ? 'lazy' : '' + obj.' + store.name,
	    chunksizes=(1,)
	)
```

So the still missing part for a power user would be to define which store to be used for a specific object type like so:

```python
default_store_factory = {
	# this line the user needs to add, e.g.
    Configuration : lambda : ConfigurationStore().named('configurations'),
    Topology : lambda : NamedObjectStore(paths.Topology).named('topologies')
}

# this will do the creation
class Storage(...):
	def create_default_store(self, obj_cls):
		store = self.default_store_factory[obj]()
		self.create_store(store.name, store)
```

We still support different Snapshot class, but they will not need other properties, since the only thing that differentiates Snapshot classes are attached CVs. In the most general case you could have several types that share any combination of CVs. To make your own class on the fly use

```python
# Make the ToySnapshot
class ToySnapshot(Snapshot):
	pass
	
# These would by default be provided or set by the engine!
ToySnapshot.add_cv(CollectiveVariable('coordinates'))
ToySnapshot.add_cv(CollectiveVariable('box_vector'))

my_snapshot = ToySnapshot(
    coordinates = my_coordinates,
    box_vector = my_box_vector
)

storage = paths.Storage('name.nc', template=my_snapshot)
```

One point I did not mention is that there will be a second basic type of CVs. There are now 

1. CVs that are user/engine set. There is not way to recompute these values (examples are coordinates, velocities, ...)
2. CVs that a a function of the first type. Like (energies, angles, distances, ...)

For the first type you can use `__setitem__` to set your non-function cv
```
my_snapshot.cv = value
```

To make it easier for the user we register (almost all) cvs as an attribute of a snapshot. So also for the second type you can do

```python
psi_cv = CV_Function('psi', ...)
print my_snapshot.psi
```


jhprinz:
rejected by myself. We decided to attach Snapshots to engines and this makes this obsolete. And in the current implementation you get into trouble when using more than one file because there is no way for the descriptors to identify which CV (from which file) is the right one.

PR #420, opened 2016-01-16
Getting the actual git hash for editable installs

dwhswenson:
When working with alpha testers (i.e., Abel) it can be very helpful to check that the installed version of OPS is actually on the git commit that the user thinks it is. `version.py` is only set when the user runs `setup.py`, so we don't want to require that every time. I added a little function to `__init__.py` that gives the git commit hash if the file is in a git repo (it will raise an error otherwise, I think). This is useful for debugging when we can't tell exactly which version of the code the user is using. We might move it to another module, but I wasn't sure which module was appropriate (or if we need a new one).

Anyway, this should be ready to review at any time.

### DIFF
--- a/openpathsampling/__init__.py
+++ b/openpathsampling/__init__.py
@@ -98,3 +98,10 @@ from pathmovechange import (
 )
 
 from storage.storage import Storage, AnalysisStorage
+
+def git_HEAD():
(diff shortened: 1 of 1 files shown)


jhprinz:
I like the conditional import. Merging.

PR #422, opened 2016-01-19
Coveralls

dwhswenson:
[Coveralls.io](http://coveralls.io) gives a GitHub-integrated web overview of test coverage. This makes it automated and easy to identify what parts of the code are not yet covered by unit (nose) tests.

The PR gives us coveralls support, and slightly cleans the nosetests. After it is merged, we should also have a working badge on the project README for test coverage.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -2,7 +2,7 @@
 omit = 
     */python?.?/*
     */site-packages/nose/*
-    */openpathsampling/tests/
+    */openpathsampling/tests/*
 exclude_lines = 
     pragma: no cover
     def __repr__
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,9 @@
+[report]
+omit = 
+    */python?.?/*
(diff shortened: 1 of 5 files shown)


dwhswenson:
Ready to merge. Example of the coveralls output [here](https://coveralls.io/builds/4773665).

jhprinz:
Great. Will merge...

dwhswenson:
I changed the parameters on coveralls.io so that coverage only gives a fail (to mark the PR with a red X) if we either lose 10% of existing coverage or drop down to 50% total coverage. Either would indicate a pretty serious change compared to our current status, so I think it is safe to call that an error. Before, it was flagging any loss of coverage as a fail (so if you added one line of uncovered code...) -- that seemed a little extreme. Later, we might make those values a bit more strict.

See examples of previously failing, now passing, in #406 (which lost 0.2% coverage -- really, not a difference worth calling "failing".)

Issue #128, opened 2015-01-28
License?


jchodera:
What license do you guys want to use?

How about the same LGPL that OpenMM is released under?  That's reasonably permissive and consistent with other projects in the lab.

dwhswenson:
I'm fine with LGPL. Can we release under "LGPL version 2.1 or later"? This maintains compatibility with GPLv2 projects as well as GPLv3. (LGPLv3 cannot be used in a GPLv2 project, if I recall correctly.)

dwhswenson:
Actually, now that I think a bit more, there's a very thin separation here between "application" and "library." That makes it a bit hard to understand how LGPL applies.

Example: say I'm a user who adds a new path mover. I could either implement this as a separate package (`my_package.MyNewMover`) or I could implement it so it is directly part of OpenPathSampling (`openpathsampling.MyNewMover`). Both are reasonable approaches. 

But I'm not sure when copyleft is invoked. I'm pretty sure the separate package is treated as application-layer, and thus could use OPS in a proprietary application without requiring that the code to `MyNewMover` be released. But if it is implemented inside OPS, is it library-layer, and therefore subject to copyleft? Or does that only become true when the user modifies `__init__.py`, or some other file in our directories? Is there a difference in legal responsibilities depending on whether the user implements tests in `testpathmover.py` vs. in a separate test script?

Somewhere along that spectrum, copyleft applies. But I don't know where that line is drawn, which makes the LGPL confusing. In principle, I like LGPL both for what it allows and for the protections it provides, but I find it confusing that the same code, implemented in two different files (or possibly saved in different directories!) falls under different license provisions.

(I doubt any of us plan to be litigious, so it probably doesn't really matter anyway. And I don't have a license I like better than LGPL; just pointing out that this is weird.)

dwhswenson:
@jhprinz and I are agreed on "LGPL 2.1 or later"; we'll close this, assuming no disagreement from anyone else.

jchodera:
LGPL 2.1 or later sounds great!

PR #423, opened 2016-01-19
License

dwhswenson:
I noticed (in a Travis log) that `setup.py` claimed we were under GPL 2.0. Per #128, we should be under LGPL 2.1 or later. I also added the LGPL 2.1 `LICENSE` file to the main directory.

Merge as soon as it passes tests; this changes no real code.


jhprinz:
Finally passed. Merging!

PR #426, opened 2016-01-21
Include `version` in `__init__.py`

### LINKED ISSUE #417: Add .version to __init__.py
This would allow us to write
```
print paths.version.full_version
```

dwhswenson:
Resolves #417. If the installation includes `version.py`, this imports it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,9 +19,7 @@ script:
 
   # Run tests in _test environment
   - source activate _test
-  - python -c "import openpathsampling as paths"
-  - python -c "import openpathsampling.version"
-  - python -c "import openpathsampling as paths; print paths.version.full_version"
+  - python -c "import os; os.chdir(os.environ['HOME']); import openpathsampling as paths; print paths.version.full_version"
   - source devtools/ci/nosetests.sh
   - source devtools/ci/ipythontests.sh
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,8 @@ script:
 
   # Run tests in _test environment
   - source activate _test
+  - python -c "import openpathsampling as paths"
+  - python -c "import openpathsampling.version"
   - python -c "import openpathsampling as paths; print paths.version.full_version"
   - source devtools/ci/nosetests.sh
   - source devtools/ci/ipythontests.sh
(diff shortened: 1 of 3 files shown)


jhprinz:
there is something weird going on. travis shows this error message
```
/home/travis/miniconda2/envs/_test/lib/python2.7/site-packages/mdtraj/geometry/order.py:25: UserWarning: Module openpathsampling was already imported from openpathsampling/__init__.py, but /home/travis/miniconda2/envs/_test/lib/python2.7/site-packages/openpathsampling-0.0.1a0-py2.7.egg is being added to sys.path
```

Apparently this means that there are two openpathsampling installed. Is it possible that it does not use the built one but rather the old one from anaconda?

dwhswenson:
The problem was this:
```
| $HOME
| - ops-git-repo/ [WORKING DIR; FROM GIT]
|   - setup.py
|   - openpathsampling/
|   - [etc]
| - conda/
|   - openpathsampling/ [INSTALLED, BASED ON GIT]
|   - [etc]
```

(not actually correct directory structure, but labelled to be clearer)

The version in `conda/openpathsampling` can be the **same code** as the git version (not an outdated conda install, but the correct one). But when Python looks for the `openpathsampling` module, it finds the `ops-git-repo/openpathsampling`, because `''` (same as `pwd`, which is currently `ops-git-repo`) is the first entry in `sys.path`.

However, `version.py` is only created during the `setup.py` build/install process. So there is no `ops-git-repo/openpathsampling/version.py`, only a `conda/openpathsampling/version.py`. The module `openpathsampling.version` only exists in a different `openpathsampling` from the one we tried to import.

**Interesting side note**: Apparently there are commits that I don't see happening, which occur behind the scenes in GitHub. Because of that, the hash I see on Travis isn't listed here. For example, my commit d1a91bb above led to a hash on Travis of a7362b8. Not sure whether that a7 hash will stay alive or whether GitHub might rewrite history later, but for now it shows up (when clicked on) as a merge of the d1 hash into fa88f7e, which is our current HEAD.

Anyway, I'll do a little bit to clean this up before we merge, but the pieces are in place.

dwhswenson:
Ready for review.

Now the Travis logs will have a line telling us the hash of the commit used for the log. Version information, including git hash, is stored by `setup.py` in `version.py`. If `version.py` exists -- i.e., if OPS was installed by `setup.py` or something that uses `setup.py` -- it is available within `paths.version` after `import openpathsampling as paths`. If `version.py` is not there (e.g., OPS installed by manual modification of `$PYTHONPATH`) then there is no `paths.version`, but everything else works fine.

jhprinz:
Nice... Merging

PR #428, opened 2016-01-22
DynamicsEngine.generate_n_frames

dwhswenson:
Adds a function to `DynamicsEngine` to get multiple frames at once (`generate_n_frames`). This process is only implemented in the forward direction. If the user wants the reverse trajectory, the user can explicitly reverse -- under most such circumstances, it nerateould make more sense to use the existing infrastructure, if that is what is desired. This is intended for the many non-path-sampling examples where we generate a fixed length trajectory from an initial point.

Overriding this default implementation could (with a few other tricks) allow for much faster implementations for fixed-length trajectories on small systems.


dwhswenson:
This should be ready for review.

jhprinz:
lgtm

Issue #407, opened 2016-01-04
Allow engine-native collective variables


jhprinz:
A thought that came during discussions with Anders Lervik and Enrico Riccardi from Titus van Erp's group. I also talked with David about this but for a different reason. 

Idea is that engine have natively computed collective variables or like in the case of lammps (not sure if OpenMM can also do that) allow to specify certain user defined collective variables that are computed directly during the simulation. We should allow to turn these computed values into regular CVs that can be stored. This could bring a (small or large) speed-up. 

Especially, if we do not want or have to store all intermediate frames, but only rely on the CVs we can skip pulling the complete frame and storing. This should allow much faster simulations.

This is somewhat related to the issue of storing "empty" snapshots. 

It would also replace the now always stored energies.
```
engine = OpenMM_Engine(..., energies_as_cv=True) # True is default ?!?
kinetic = engine.kinetic_energy
potential = engine.potential_energy
storage.save(kinetic, potential)
# ... run simulations...

# print all kinetic energies
print kinetic(storage.snapshots.all)
# optional (attach a CV attribute to a snapshot) could be easily done using descriptors
snap0 = storage.snapshot[0]
print snap0.kinetic_energy
```

All that is necessary for an efficient implementation is already in place. We only need to figure out the best way to tell an engine which CVs we want computed. After that the cache does everything that is
necessary: CVs for snapshots in memory are kept. If a snapshot disappears so is its associated CV. You can store the CV anytime, but the cache that remembers values for a CV that is not stored, but where all the snapshots are stored is very limited. So if you would run a long simulation and at the end think about storing the CV some of the computed content could be lost. But that is more a user mistake since he can simple directly store the CV (not the values!) before simluation.

Another downside of this engine-side approach is that you cannot recompute missing values, this is only possible for the current approach. Still, I believe this is an important feature to have. 

Last point a PR should address is that we want also regular CVs to be computed for every frame once it is saved. This should be optional
```
cv = CV_Function(..., autosave=True)
```

Issue #311, opened 2015-09-10
Topology as part of engine?


jhprinz:
I was wondering if it makes sense to attach a topology only to an `Engine` and let each generated `Snapshot` point to the engine that made it. 

This means a snapshot is not complete (at least topology wise) without its engine! On the other hand we could have snapshots without topologies since we could think about engines that are not MD and might not use the concept of a topology. It goes in the direction of abstract snapshots with no special properties of than comprising trajectories. 

To me this seems to make sense, but I not sure.

To create Snapshots from files like `.pdb` we could have special engine that just creates snapshots, transforms them, etc.

dwhswenson:
If I recall, right now every `Snapshot` has a `topology` parameter. If I understand, you're saying to instead switch that to an `engine`, meaning that each `DynamicsEngine` must be associated with a single `topology`. Then `Snapshot.topology` can just be a shortcut for `Snapshot.engine.topology`.

All of this sounds fine to me, although you're talking about creating the ability to translate between topologies for different engines, which scales as $N^2$ in the number of engines. Fine when all the engines are very similar, but what if your snapshot needs extra information (e.g., from QM)? Then it gets messy. You would have to define every "class" of dynamics you want to do, and define a way to translate snapshots between those classes. That's still $N^2$, but a smaller $N$. On the other hand, if the goal is to keep OPS useful in general (beyond MD) then there's no way you can enumerate all such classes. I'm not sure its even possible within MD: there are too many possible "extra objects" to hold.

Also, isn't this pretty much #208/#197? Do we need a separate issue?

dwhswenson:
This will be handled in #380, right?

jhprinz:
No. I removed that to keep the PRs really separate now.

jhprinz:
Will be part of upcoming engine / snapshot restructuring.

PR #429, opened 2016-01-22
Restructuring of Snapshots

jhprinz:
This will 

- [x] precompile feature-dependant code
- [x] store snapshots more compact (no overhead for reversed objects)
- [x] fixes CVs
- [x] remove energies from snapshots (to be turned into a CV)

do be done:

- [x] better docstring generation from features
- [x] cleanup of CVs that will use compressed storage now. The code is not optimal.
- [x] implement a _renew saving_ function that allows to reuse Snapshots
- [x] check docstrings
- [x] get existing tests pass
- [x] fix notebooks

most stuff as discussed with @dwhswenson.

To me merged after #412 .

This makes the use of features much easier. And the actual implementation of Snapshot now looks like
```py
@features.base.set_features(
    features.configuration,
    features.momentum,
    features.xyz,
    features.topology  # for compatibility
)
class Snapshot(FeatureSnapshot):
    """
    A fast MDSnapshot
    """
```

The decorator takes care of everything. It will construct the constructor with correct signature. Precompile a `.copy()` and `.create_reversed()` functions, and take care of docstrings, etc...

The xyz feature just add a property, nothing else. 

A pair of forward and backward `Snapshot` will now only be stored once. Never twice. Storage takes care of getting the reversed and each feature much now, if it reverses or not. This will make storage requirements smaller. Same is true for CVs which will now consume half as much space if time-reversible.

The special saving should do the following step in one: 1. Save the snapshot 2. Do not cache it and 3. Take care of potential references in the reversed.

This also implements storing the topology with each Snapshot as a feature that can be later removed easily when switching to storing the engine.

I decided to refactor the code for loading and saving CVs while working on better caching since this is kind of the same problem.

Last things related to this:
- make energies into an optional CV (already implemented)
- engine restructuring into different folders (needs to be discussed, but simple)
- Snapshots use engine property and not topology (simple)
- Question if engines use their own snapshot stored, which can be shared 
- Split caching of objects into separate function. Should be reusable for objectstore and CVs and allow chunked loading (optional)


dwhswenson:
I haven't had a chance to review yet, but as discussed a bit in our chat today, it would be nice to have 3 levels of "copy":

* **shallowcopy**: only copies references
* **newcopy**: deep copy, creates new objects for each feature
* **copyto**: deep copy, overwrites existing destination version of each feature to avoid extra allocations. Mainly uses the `numpy.copyto` function.

The stuff in this PR should make it easy to implement those, right? Could you implement them as part of this PR?

dwhswenson:
More notes to come, but it looks to me like this doesn't include #311. Wasn't that supposed to be part of this? Or is it in there, and I don't see it?

dwhswenson:
Comment on the overall approach: I'm finding cases where I might want to create a snapshot class with a certain set of `feature`s, somewhat on the fly. It depends a bit on what exactly `features` means; one of the key points from my perspective is that a `feature` can be an instance attribute that isn't saved in storage as part of the snapshot. This can, for example, create in-engine CVs, which is related to your ideas in #407.

I think what would make this easier would be the following:

```
SnapshotClass = SnapshotFactory([feature1, feature2])
```

as opposed to

```python
@features.base.set_features(feature1, feature2)
class SnapshotClass(FeatureSnapshot):
    pass
```

I think this would be easier for a novice than dealing with decorators. Internally, you can probably re-use the decorators to implement this. Would this be an easy addition to this PR?

dwhswenson:
What happens if two features try to implement the same property? This could either happen because you have two features with the same primary name (I could have `dynq.features.coordinates` separate from `paths.features.coordinates`, or because they implement complementary aspects as properties (currently, `dynq.momenta` should implement a property called `velocities`, and `paths.velocities` should implement a property called `momenta`: using both features is a contradiction). What happens in the event of a namespace collision?

Related: how would we go about having a related property, e.g., `momenta` and `velocities`? That should be something that can be implemented as part of the feature, and then wrapped into the `Snapshot`, right?

openpathsampling/features/base.py
```diff
@@ -0,0 +1,231 @@
...
+                return func(self, *args, **kwargs)
+            else:
+                return None
+
+        return inner
+
+    return _has
+
+
+def set_features(*features):
```
dwhswenson: Could you add some more documentation in here? It is a little hard to tell what all is done by this. The fact that the actual code is *generated* makes it a bit less readable, so more explanation would help me follow it (and remember how it worked when I look at it again in 6 months)

openpathsampling/features/numpydocparse.py
```diff
@@ -0,0 +1,145 @@
+class NumpyDocParser(object):
```
dwhswenson: Docstring, with explanation, please (oh, the irony!). Do I understand correctly that this generates a docstring for snapshot classes? I see how that can be useful from the standpoint of interactive users, but does that mean anything when sphinx generates docs for web/print?

Also, the word "parser" usually implies that it *reads* that thing -- but this looks to me like it is mainly intended for writing (or maybe joining) docstrings. Not sure that "parser" is the right term for this.

openpathsampling/features/shared.py
```diff
@@ -0,0 +1,313 @@
...
+import numpy as np
+from simtk import unit as u
+
+from openpathsampling.netcdfplus import StorableObject, ObjectStore
+
+# =============================================================================
+# SIMULATION CONFIGURATION
+# =============================================================================
+
+class Configuration(StorableObject):
```
dwhswenson: Renaming suggestions:

* `Configuration` => `StaticContainer`
* `Momentum` => `KineticContainer`

Both of these are essentially container classes. I think we should name them as such. It also resolves my long-standing complaint that `Momentum` isn't the momenta. This last is getting very frustrating as I work with `dynamiq`, where everything is expressed in terms of momenta, not velocities.

openpathsampling/features/xyz.py
```diff
@@ -0,0 +1,22 @@
...
+def xyz(snapshot):
+    """
+    Returns
+    -------
+    xyz : numpy.ndarray, shape=(atoms, 3), dtype=numpy.float32
+        atomic coordinates without dimensions. Be careful.
+
+    Notes
+    -----
+    SERIOUS PROBLEM: whether .coordinates returns a u.Quantity or jut
```
dwhswenson: Is this "SERIOUS PROBLEM" in the notes still a serious problem? I think this is solved (perhaps by PRs even before this one).

openpathsampling/netcdfplus/netcdfplus.py
```diff
@@ -714,6 +718,17 @@ def create_type_delegate(self, var_type):
...
 
+        if var_type.startswith('obj.') or var_type.startswith('lazyobj.'):
+            store = getattr(self, var_type.split('.')[1])
+            base_type = store.content_class
+
+            get_iterable = lambda v: \
+                not v.base_cls is base_type if hasattr(v, 'base_cls') else hasattr(v, '__iter__')
+
+            set_iterable = lambda v: \
+                not v.base_cls is base_type if hasattr(v, 'base_cls') else hasattr(v, '__iter__')
```
dwhswenson: Why `get_iterable` and `set_iterable`? Would these be better named something like `setter_is_iterable`? Starting a name with `set` sounds like the function will set an attribute.

openpathsampling/openmm_engine.py
```diff
@@ -10,9 +11,9 @@
 
 
 class OpenMMEngine(paths.DynamicsEngine):
-    """OpenMM dynamics engine based on using an openmm system and integrator object.
+    """OpenMM dynamics engine based on using an 'simtk.openmm` system and integrator object.
```
dwhswenson: I think you mean backtick (`), not apostrophe ('), before simtk.

jhprinz:
Let's see:
#### Topology as part of engine
 No, this will be part of the engine restructuring. I prepared everything to that we can easily switch to saving the engine with a snapshot and not the topology. But the actual change will be in the next PR

#### SnapshotFactory
Good idea. It will require a `ClassName` for the `Snapshot` and an optional description
```

def snapshot_factory(name, features, description=None):
    """
    Helper to create a new Snapshot class
    """
```

and would work like this
```
mysnap = paths.snapshot.snapshot_factory('SuperSnap', [paths.features.box_vectors], 'The ultimate snapshot')
```

So far this definition is not storable! So you would have to definte the `SuperSnap` in this example before reloading it. To make this work is possible but not trivial. Features are right now only add ons to an existing snapshot class. And the feature only define same attributes, but not how to reconstruct the whole class. I think in most setups this will work since the engine will create the snapshot class and once we have the engine the snapshot class will exist too. Just to be clear on the limits of the current approach.

#### What happens if two features try to implement the same property?
Good catch. This actually was an error. Because configuration had also `xyz` registered as a property. So now there is a check for orthogonality. All features used must have disjoined sets of attribute names or an RuntimeError it thrown which lists the two competing features.

I also added a variable `required` to the features that can be used to make sure that certain attributes from other features are present, too.

#### StaticContainer and KineticContainer
I like the names, will change it once we agree on the other names. So then have `snapshot.static.` and `snapshot.kinetic.` as well as the features `static` and `kinetic`? Is that okay?

#### Serious Problem
No. I removed it. 

#### get/set iterable
Yes, you are right, will change it

#### backticks..
Yes, well, right :)

jhprinz:
I also added much more docstring to the Snapshot decorator and the NumpyDocTools (renamed it) class.

jhprinz:
One question about the copies (I thought I asked already but cannot find it)

#### shallow copy

this is clear. Create a new object `obj` and call `obj.feature1 = self.feature1` ... This is already implemented and is called just `.copy()`

#### new copy

would be a deep copy. Create a new object and fill all objects with deep copies of all features like
`obj.feature1 = copy.deepcopy(self.feature1)`

This require deepcopy but is no problem

#### copyto / copyfrom

This is similar to `__init__` ???, so you would run
```
changed_snapshot = old_snapshot.copy_from(coordinates=new_coordinates, ...)
```

I could also create a function that will just copy all content from one snapshot into another reuing existing memory if that is what you want. But after creating a duplicate what do you want to do with it? You have to change its values before saving with new ones. Wouldn't it be easier to just keep your internal numpy array, just tell the Snapshot to use this and then save it?

I assume that in some cases you have like 100 Snapshots that you keep and want to overwrite with the new data and then treat as new and save them. Correct? For this you could use a `copyto`, copy_from function. 

So now the implemented `copyto(self, target)` will copy all features and use `numpy.copyto` while `Snaphot.initialize(target, parameters...)` will reinitialize an existing snapshot and is equal to
`Snapshot.__init__(target, parameters)`. So maybe we do not need it.

dwhswenson:
#### SnapshotFactory

Question: is the example you gave with

```python
mysnap = paths.snapshot.snapshot_factory('SuperSnap', [paths.features.box_vectors], 'The ultimate snapshot')
```

actually right? Would you then create a snapshot with `mysnap(box_vectors)` or with `SuperSnap(box_vectors)`? I think the latter would make more sense. So should it be:

```python
SuperSnap = paths.snapshot.snapshot_factory('SuperSnap', [paths.features.box_vectors], 'The ultimate snapshot')
```

Repeating "SuperSnap" is a bit redundant, but it doesn't bother me too much (I assume that the `name` attribute is needed for storage to know what to build?) Only people who know enough to be contributing new engines or otherwise developing some fancier stuff would need to know about this.

> So far this definition is not storable!

Just to make sure I'm understanding correctly: the definition of the snapshot class is not storable, but the snapshots themselves are, right? So if you define a custom snapshot class in your script, you need to define it again for any analysis script. (I seem to remember something similar with certain restrictions on what functions can be stored by CVs?) That all seems fine to me. As you said, usually this is part of the engine, and it would be almost impossible to *not* include the snapshot with your engine code!

#### StaticContainer and KineticContainer

> I like the names, will change it once we agree on the other names. So then have snapshot.static. and snapshot.kinetic. as well as the features static and kinetic? Is that okay?

Maybe `snapshot.kinetic_info` and `snapshot.static_info`? My thought is that `snapshot.kinetic` could be misread as `snapshot.kinetic_energy`. Adding `_info` indicates that there's more to it than just the KE.

#### copyto / copyfrom

> This is similar to `__init__` ???

Huh -- I hadn't thought of that, but yes, that's exactly what `copyfrom` would be. Except `__init__` takes all the attributes as input, and `copyfrom` takes another snapshot. This separates those roles a little better.

I usually think of `__init__` as being closely related to `__new__`, so I only think of calling it as part of instantiation. The point of `copyfrom` is to avoid alloc'ing more memory -- but you're right, `__init__` isn't actually the part that alloc's the memory.

Note on names: I personally find `copy_to`/`copy_from` more readable (and more in keeping with Python style), but numpy uses `copyto`. So the choice is between following better Python style and following numpy's lead -- I prefer better Python, but I think a good case can be made either way.

> I could also create a function that will just copy all content from one snapshot into another reuing existing memory if that is what you want. But after creating a duplicate what do you want to do with it?

I use this for internal state in `dynamiq_engine`. When I build the engine, I allocate memory for a "working/scratch space" snapshot. When you start a trajectory, you copy to that working space. The "working" snapshot won't be saved (or even accessed outside the engine) but it has the same features as the normal snapshot.

It might be possible to do this by swapping pointers to entire snapshots (and that would also be faster), but that's also more error-prone, since the snapshot might be pointed to be other objects from outside. The `copy_from` approach takes an internal snapshot and makes it externally available.

***

BTW: more coming later. My flights yesterday got cancelled, so I'm still in Indianapolis. I'm about to head to the airport, and (I hope!) I'll be in Amsterdam tomorrow morning -- assuming I don't again spend 7 hours at the Indianapolis airport.

jhprinz:
#### SnapshotFactory

Yes, of course. The name in the function is the class name and the assignment will just allow you to call it the same in your name space. Exactly like
```
class MyClass(object):
    pass

SuperSnap = MyClass

print SuperSnap.__name__
>>> MyClass

a = MyClass()
b = SuperSnap()

print a.__class__ is b.__class__
>>> True
```
So, yes, you should write
```
SuperSnap = paths.snapshot.snapshot_factory('SuperSnap', [paths.features.box_vectors], 'The ultimate snapshot')
```

### Storing

Also yes, you can save without problems and all features are stored accordingly. But if you want to load e.g. `SuperSnap` the class itself is not reconstructed. We could in theory do that, but only using the features and not necessarily all the code that comes with it. Remember you can add lots of other functions to your Snapshot class and then use the feature decorator.

Important is that you just create the Snapshot somwhow before loading. Since it is subclassed from FeatureSnapshot and so from StorableObject it will be directly useable by the storage.

### kinetic / static

> Maybe snapshot.kinetic_info and snapshot.static_info? My thought is that snapshot.kinetic could be misread as snapshot.kinetic_energy. Adding _info indicates that there's more to it than just the KE.

I would really like a single word for it. No real reason other than aesthetics. What about `kinetics` and `statics`. I seem the point with `kinetic`, although `kinetic.energy` would read nice as would `static.coordinates` do. 

Anyhow. Once engines are refactored this name should only appear in really very few lines of code related to the feature itself.

#### copy_to / copy_from

Okay. So now I made sure that you can use just `Snapshot.__init__(target, parameters)`. We can have another name for it, if you want. But the above works fine. So this should be implemented.

`copy_from` would be the opposite (kind of and duplicate the given snapshot from external ones). I can also do that and this will use the `numpy.copyto`. We do not need it for the `__init__` since there we could do it ourselves. So I will add that, too.

#### Another bottleneck

I have identified another bottleneck and found a way around it. The way I treat the reversed snapshot. There are actually 2 ways and the faster one was implemented poorly and not properly tested.

1. have a lazy reference using `_lazy` (slow): This will treat the reversed snapshot as a lazy property and as soon as you use one lazy property you need to add a `_lazy` dict to the object and initialize it properly which takes a time. Advantage is that you can remove the reversed snapshot from memory like any other and load from disk if required. This approach also does not matter, if you also use other lazy properties like in `Configuration`. So this is useful for large Snapshots.

2. have a real reference without `_lazy` (fast): This will keep a real reference in memory (if you ever need it) as long as you remove the other snapshot as well. So it will consume a little more memory in some cases, but - if you do not have additional lazy attributes like we do in the `Configuration`, etc approach (which seems useful for large systems) - you can avoid the lazy initialization completely and this will speed up creation by a factor of 2-8.

I will do some testing and post some timings...

dwhswenson:
#### kinetic / static

`kinetics` and `statics` sound good to me -- clear enough to avoid the confusion.

> Once engines are refactored this name should only appear in really very few lines of code related to the feature itself.

And in autocompletes. That's why I think someone would be likely to get confused by thinking `snapshot.kinetic` would give the kinetic energy.

#### copy_to / copy_from

Hmm... not sure whether `Snapshot.__init__(target, params)` does what I expect of a `copy_to`. The key question is whether `__init__` does a deep copy. Do we have: 

```python
supersnap(box_vectors).box_vectors is othersupersnap.__init__(box_vectors).box_vectors
```

If that would be `True` (link the vector rather than deep copy it), then this is not the behavior I would expect from `copy_to`. 

(And now I need to put the computer away now to board the first flight of 3!)

openpathsampling/snapshot.py
```diff
@@ -117,306 +91,73 @@ def copy(self):
...
-        """
+    def create_reversed(self):
+        this = self.copy()
+        this._reversed = self
+        return this
 
-        obj = self.copy()
-        return obj.reversed
 
+class FeatureSnapshot(AbstractSnapshot):
```
dwhswenson: What is the purpose of `FeatureSnapshot`? It seems functionally the same as `AbstractSnapshot`, since all it does it call `super` for `.copy`.

openpathsampling/snapshot.py
```diff
@@ -117,306 +91,73 @@ def copy(self):
...
+    def copy(self):
+        return super(FeatureSnapshot, self).copy()
 
-# =============================================================================
-# SIMULATION SNAPSHOT (COMPLETE FRAME WITH COORDINATES AND VELOCITIES)
-# =============================================================================
 
-@lazy_loading_attributes('configuration', 'momentum', '_reversed')
-class Snapshot(AbstractSnapshot):
+def snapshot_factory(name, features, description=None):
```
dwhswenson: My gut wants the factory to be named `SnapshotFactory`. Not sure why, since I feel like this naming fits the letter of the law better. Still, it feels weird. Thoughts?

openpathsampling/snapshot.py
```diff
@@ -117,306 +91,73 @@ def copy(self):
...
-        box_vectors : periodic box vectors
-            the periodic box vectors 
-        potential_energy : simtk.unit.Quantity of units energy/mole
-            potential energy
-        kinetic_energy : simtk.unit.Quantity of units energy/mole
-            kinetic energy
-        """
-
-        super(Snapshot, self).__init__(is_reversed, reversed_copy, topology)
+    cls = type(name, (FeatureSnapshot, ), {})
```
dwhswenson: Thought: would it be ridiculous to add a `superclass` parameter to the factory function? This would default to `FeatureSnapshot`, but would allow simple overrides so that, e.g., I could use a `dynq.Snapshot` (which will be a subclass of `paths.FeatureSnapshot`) as the superclass in `dynamiq_engine`. (There might be extra methods I've implemented for `dynq.Snapshot`, or I might just want the ability to put some specific documentation relevant to the features allowed by `dynamiq_engine`.)

dwhswenson:
Okay, finally finished with the review! A few line notes were added earlier.

One last question: I'm a little confused as to the use case for `ReversibleObjectDictStore`. When would you use that vs. when would you use `ObjectDictStore`?

openpathsampling/snapshot.py
jhprinz: I thought about just having one class. The reason is that I wanted to still allow to have snapshots that do not use features (as before) the common basis is `AbstractSnapshot`. And FeatureSnapshot is meant to identify Snapshots with features. 

I guess we could get rid of it, but this way the separation is clearer.

openpathsampling/snapshot.py
jhprinz: I know what you mean. My IDE complaint because function names should be lower case with dashes. But I don't mind. You could consider it like using a constructor and hence `SnapshotFactory` would be better. We could also (maybe that solves also the above question) call this `FeatureSnapshot` and remove the extra class.

```
MDSnapshot = FeatureSnapshot(
    'MDSnapshot', 
    features, 
    description,
    super_class
)
```

openpathsampling/snapshot.py
jhprinz: Not at all. This would correspond completely to the original approach
```
@set_features(my_feature_list)
class MDSnapshot(base_snapshot_class):
    """
    description
    """
```
versus

```
MDSnapshot = FeatureSnapshot(
    'MDSnapshot',
    my_feature_list,
    base_class = base_snapshot_class,
    description = 
        """
        description
        """
)
```

I will add that. `base_class` or `super_class` ?

jhprinz:
> The key question is whether __init__ does a deep copy.

No, it does not, that would have to happen before the call of __init__. Like
```
Snapshot.__init__(old_snapshot, coordinates=new_coordinates.copy())
```

Now there is/will be a `copy_to` and a `copy_from` function. We can also add more as we need them.

> One last question: I'm a little confused as to the use case for ReversibleObjectDictStore. When would you use that vs. when would you use ObjectDictStore

Well, the ObjectDictStore (not really an ideal name actually) provides functions to store the CV, create the cache and attach it to the CV. The ReversibleObjectDict does the same but needs to set up the cache differently. In the general case there are two arrays, one for the forward and one for the backward value. In the Reversible case there is only one. This saves have the storage space. I still have not found the best solution for this, but the current version is decently fast and storage optimal. The biggest hassle is still that we have lots of missing values and lots of different caches. I would move improvements in this to a later stage.

openpathsampling/snapshot.py
dwhswenson: But isn't the use of snapshot features optional in either case? I mean, can I get the behavior of features while subclassing from `AbstractSnapshot`?

It seems like non-feature-snapshots are deprecated now... Normally we might keep deprecated aspects around a bit to give users a chance to change, but right now our only users are the people over at dynamiq-md -- and I'm pretty sure they can keep up with the changes.

jhprinz:
To be complete on the generated functions there is now:

1. `__init__(self, feature1_attr1=None, feature1_attr2=None, feature2_attr1=None, ...)`

    creates a new Snapshot and takes the attributes as given. No copying. Can be used as a class method using `Snapshot.__init__(target_snapshot, ...)`

2. `init_copy(self, feature1_attr1=None, feature1_attr2=None, feature2_attr1=None, ...)`

    same as `__init__` but will make copies of all numpy arrays. Python native types do not need to be copied. Not copied are mutable types (list, dict) and OPS objects. But this is usually not used to duplicate complex objects. More for simple snapshots and avoid object creation. This violates immutability so make sure you overwrite something that is safe!

3. `copy_to(self, target)`

     places a copy of the content of `self` into `target`. Same restrictions as `init_copy` apply. Also here immutability is violated!

4. `copy(self)`

    creates a new object and uses shallow copies of the content. Nothing is duplicated here. This is immutable safe.

5. `create_reversed(self)`

    This creates a shallow copy, except for attributes that will change when the snapshot is reversed. Numpy arries will be copied before flipping to preserve immutability.

6. the docstring :)

openpathsampling/snapshot.py
dwhswenson: Either is fine by me (although I think it should be `base_class` or `superclass` ... for some reason superclass is actually one word).

dwhswenson:
Is there a `copy_from(snapshot)` available? The point is that this is easily used regardless of the specific snapshot features, and so one bit of code can be used for many snapshot types.

(You might already have that with some `**kwargs` magic and the `init_copy` function).

jhprinz:
you can use `init_copy` which does `copy_from`. Make sure that you give all attributes or they will be casted to None.
If you want a `copy_to` that goes from snapshot to snapshot then use `copy_to` in the other direction.

openpathsampling/snapshot.py
jhprinz: Maybe we can get rid of the two then and call it `BaseSnapshot` ? That should clean things up. That would then also match `base_class = BaseSnapshot` and in python the superclasses are accessed using `__bases__` or `__base__`.

dwhswenson: I like this. Yes, definitely, that's the way to go.

jhprinz:
using `mdtraj` version 1.5.1 for now...

dwhswenson:
> Seems that ToyDynamics is faster now.

Any idea what makes the difference? I was concerned that the flexibility of the new snapshot style would slow things down, not speed them up.

jhprinz:
This is because I cleaned and improved the code so that creation and reversal is simpler and thus faster.

jhprinz:
I will check this out little more.

jhprinz:
Back on track. All tests pass...

I think we are almost ready to merge...

openpathsampling/features/numpydocparse.py
dwhswenson: (On `NumpyDocTools`): The docstring you added is good. Also, I just wanted to mention that this is a pretty cool trick. If only there was a way to get sphinx to read such dynamically-created docstrings (I assume it doesn't, at least -- we can't use the `SnapshotFactory` and get documentable default snapshots, right?)

dwhswenson:
The only unresolved issue I see is the renaming of `Configuration` and `Momentum` to `StaticContainer`/`KineticContainer` (or `StaticsContainer`, if you prefer). That could be done in a separate PR, but it should be pretty easy -- can we go ahead and do that in here?

openpathsampling/features/numpydocparse.py
jhprinz: Actually, that was the point. Sphinx will open the package and read the `__doc__` property and put that into the documentation. This was copied from my build html page
```
Snapshot([engine, configuration, momentum, ...])	The standard MDSnapshot supporting coordinate, velocities and box_vectors
```

Since I was on the engine module I have not updated the docs yet and there is something wrong with autosummary. But I think that sphinx can really use the generated docstring. 

If that is true then for sphinx there is no distinction between generated and hard-written docs as long as the docs are generated on import.

jhprinz:
Yes, I would put this into the engine restructuring. Once this is really separate we can easily change the name of the containers.

jhprinz:
Let me run a few more tests. But I think we should be good to go.

dwhswenson:
> Let me run a few more tests. But I think we should be good to go.

Should I still be expecting more tests? Otherwise, I'm ready to merge this (on the assumption that the renaming still to do happens in #434, or at least very soon.)

jhprinz:
Good to go. Please merge if you want to.

dwhswenson:
Oops! Meant to merge this as soon as I got back from the IND (where I did *not* get my residence permit this time... have to go back in a week or two). Anyway -- Merging!

PR #431, opened 2016-02-04
Add openmm platform testing script

jhprinz:
This adds a test script that should analyze your openmm installation and give suggestions on how to fix missing platforms.

Can be merged at will. Does not affect any other files.

Some things it checks:

- Check the conda installation and figure out where the dynamic libraries are
- Check if the environment variables set are existing paths and if they contain openmm files.
- Suggest how to set the environment variables appropriately.

### DIFF
--- a/openpathsampling/scripts/openmm.py
+++ b/openpathsampling/scripts/openmm.py
@@ -1,0 +1,108 @@
+import simtk.openmm as mm
+import os
+import os.path
+import re
+import conda.api
+import conda.config
+
+
+c1 = re.compile('([a-zA-Z0-9 ]*) ([\S]*): dlopen\(([\S]*), ([0-9]*)\): Library not loaded: ([\S]*)([\s]*)Referenced from: ([\S]*)([\s]*)Reason: ([a-zA-Z0-9 ]*)')
+subdirs = ['DIR', 'PLUGIN_DIR', 'LIB_PATH']
+openmm_envs = [env for env in os.environ if 'openmm' in env.lower()]
+err = []
+
+if not os.path.isdir(mm.Platform.getDefaultPluginsDirectory()):
+    err += ['The default Plugin path "{0}" does not exist.'.format(
+        mm.Platform.getDefaultPluginsDirectory()
+        )]
+
+# for env in openmm_envs:
+#     if 'dir' in env.lower() or 'path' in env.lower():
+#         path = os.environ[env]
+#         if not os.path.isdir(path):
+#             err += ['Error in environment variable "{0}": Path "{1}" does not exist!'.format(env, path)]
+
+for sub in subdirs:
+    env = 'OPENMM_' + sub
+    if env not in os.environ:
+        err += ['Needed environment variable "{0}" not set!'.format(env)]
+    else:
+        path = os.environ[env]
+
+        if not os.path.isdir(path):
+            err += ['Error in environment variable "{0}": Path "{1}" does not exist!'.format(env, path)]
+        else:
+            files = [f for f in os.listdir(path) if os.path.isfile(os.path.join(path, f))]
+            okay = False
+            for f in files:
+                if 'openmm' in os.path.join(path, f).lower():
+                    okay = True
+
+            if not okay:
+                err = ['Folder "{0}" in environment variable "{1}" does not contain files that are openmm specific. Check if the path is correct!'.format(path, env)]
+
+for s in mm.Platform.getPluginLoadFailures():
+    res = c1.search(s)
+    groups = res.groups()
+    if not os.path.isfile(groups[1]):
+        err += ['While loading a dynamic library the requested file "{0}" could not be found!'.format(groups[1])]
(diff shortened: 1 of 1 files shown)


jchodera:
Have you thought of contributing this to `OpenMM`?

dwhswenson:
As soon as it passes tests: MERGING!! :-)

My only complaint is that this would be useful to ALL users of OpenMM, not just those using OPS. It might be worth looking into putting it/a copy of it somewhere more broadly accessible.

Once this is in master, please merge it into your "weina" branch as soon as you can: I think @bolhuis is currently working off of that branch.

jhprinz:
This might help @bolhuis to figure out why his new openmm installation does not support any platform.
I recently had that problem too after reinstalling conda completely. What I did to get CUDA running again was download and install CUDA SDK 7.0 (not the newest one). Install conda and openmm with `conda install openmm`. Missing was to get the dynamic library paths correct. This is what I added to my `.bashrc`

```bash
# PATHS for OPENMM
export OPENMM_DIR="~/anaconda/pkgs/openmm-6.3.1-py27_1"
export OPENMM_PLUGIN_DIR="${OPENMM_DIR}/lib/plugins"
export OPENMM_INCLUDE_PATH="${OPENMM_DIR}/include"
export OPENMM_LIB_PATH="${OPENMM_DIR}/lib"
```

jhprinz:
@jchodera : Yes, that was the original plan. I tried to put in what I think are common mistakes. It should be cleaned up though before adding it to openmm. I will work on this a little more tomorrow. The current version assumes a conda installation.

jhprinz:
@dwhswenson : Will do. Only problem might be that when Peter pull it is could overwrite his changes of notebooks. @bolhuis should make copies of his notebooks before doing that.

bolhuis:
@jhprinz : Do I really need to install CUDA SDK? I have a macbook pro 2014 with intel iris pro GPU. This  should support OpenCL. When I check my MBPs setting with this:

  _clGetDeviceIDs(NULL, CL_DEVICE_TYPE_ALL, 0, NULL, &num_devices);
    cl_device_id* devices = calloc(sizeof(cl_device_id), num_devices);
    clGetDeviceIDs(NULL, CL_DEVICE_TYPE_ALL, num_devices, devices, NULL);

    char buf[128];
    for (i = 0; i < num_devices; i++) {
        clGetDeviceInfo(devices[i], CL_DEVICE_NAME, 128, buf, NULL);
        fprintf(stdout, "Device %s supports ", buf);

        clGetDeviceInfo(devices[i], CL_DEVICE_VERSION, 128, buf, NULL);
        fprintf(stdout, "%s\n", buf);
    }

I get the following:

Device Intel(R) Core(TM) i7-4750HQ CPU @ 2.00GHz supports OpenCL 1.2 
Device Iris Pro supports OpenCL 1.2 

This should mean that there should be 2 platforms next to the Reference platform.
However, the Openmm test installation only gives 1 platform besides the Reference.
Now, changing the dynamic library paths does not change this.

The question is thus : does the Iris Pro device need CUDA drivers or SDK, even if it is clearly OpenCL?







If so, where do I get this? Is ithsi from the

jhprinz:
@bolhuis My mistake. You only need to install CUDA SDK if your machine has an Nvidia graphics card. I kind of remembered that was the case. If not then OpenCL should work _without_ installing any drivers. That at least worked with my machine, where at some point the CUDA drivers broke (I installed SDK 7.5).

Let me see what could be the remaining issue:

1. First thing with OpenCL is that Apples drivers used to not work, so I presume you need to make sure that you run Yosemite (10.10) or El Capitain (10.11) to actually use OpenCL at all. I am not sure when they really fixed it. 

2. OpenMM should be able to access the newest libraries that ship with openmm. If your path is wrong openmm will not given an error message, but just not support the platforms. If all paths are wrong then even CPU should not work (what happened to me). Since this works for you, you probably link to old ones or some of the paths are wrong. You can try

    ```
    import simtk.openmm as mm
    print mm.Platform.getDefaultPluginsDirectory()
    ```

    that should point to your conda installation `pkgs/openmm-6.3.1-.../lib/plugins`

    If not check setting your environment variables. I am actually wondering, if these environment variables need to be set in the upcoming release. I remember them discussing about fixing the @rpath etc... 

3. After changing the environment variables you need to restart the jupyter notebook in a fresh terminal to make sure the effect is present. You can check with

    ```py
    import os
    for key, value in os.environ:
        if 'openmm' in key.lower():
            print key, ' = ', value
    ```

    if these are correctly present. 

4. Check the output of

    ```py
    import smirk.openmm as mm
    print mm.Platform.getPluginLoadFailures()
    ```

    For me it still shows errors although all 4 platforms work. But there might be a hint on what is going wrong. Maybe you can post your result here.

Other than that I have no idea what could go wrong. At least after my recent reinstall of conda I only did these steps to get it to work. Hope this will work soon!

jhprinz:
I was too quick. CUDA does not work directly. Since CUDA 7.0 is not compatible with the newest version of XCode I would have to downgrade XCode to get it to work.

PR #433, opened 2016-02-10
[WIP] Collective Variables for all objects

jhprinz:
This could become an implementation for trajectory variables. 

Currently it also add support for objects as values. 

One problem found so far

netCDF does only allow variable length of non basic types. Means variable lengths of floats are okay, but not variable lengths of matrices. There are two ways around this:

1. Always flatten matches and reconstruct when loading
2. Wrap the matrix in a storable object. Storable objects are stored as an int index and can thus be variable length. One example is Trajectory which is a variable length list of Snapshots.

I will try to implement both possibilities, but so far only 2. works. 

For use with a monodromy matrix it would mean to create a Monodromy Matrix object. You can of course create a MonodromyTrajectory class and just duplicate what a Trajectory does. This way you could reuse Monodromy matrices if possible and implement.

Or really subclass Trajectory and add the monodromy trajectory as an attribute.

This is still in development!


dwhswenson:
> netCDF does only allow variable length of non basic types. Means variable lengths of floats are okay, but not variable lengths of matrices.

Questions about this: 

* This is only a problem for trajectory CVs, right? Because each matrix of the monodromy matrix (as an example) has a fixed size, but since there is a list of `len(trajectory)` matrices associated with a *trajectory* (which has variable length), that's the source of the challenge. Am I understanding that correctly?
* Also if I understand your statement correctly, this would not be a problem for trajectory CVs with `len(trajectory)` floats (not vectors/matrices). For example, the classical action at each snapshot along the trajectory would already work, right?

I definitely agree that the idea of flattening and reconstructing would be nice to have: it has potential to be useful in lots of situations.

For the specific case of the monodromy matrix, I would need to subclass `Trajectory` anyway (to correctly handle things like joining two trajectories; reversing direction, etc), but it's nice to put as many features into the netCDF+ code as possible, rather than hacking things together that are only relevant for my specific case!

Also: does netCDF have built-in types for complex numbers? If so, it would be good to support that (if not, it's always easy to just store the real and imaginary parts separately).

jhprinz:
Yes, you got the exact problem. The reason is quite simple. The netCDF implementation work by specifying a shape like `(40,40)` (where internally instead of the `40` you define a dimension that has fixed length `40`. As a remark, dimensions can grow, but they cannot be variable!. You can define a new derive value_type from base_types that has flexible length. Like saying `str` is flexible length `char`. Or for saving trajectories I create a `var_int` to store a varying length of `int`s. I mixed the two feature into one, by allowing to set the last dimension to be `'...'` which means the last dimension is flexible. So the ints for trajectories have type `shape=('trajectories', '...'), dtype=np.int32` (in netcdf `shape` is called `dimensions` and `dtype` is `var_type`).

So, with this knowledge the problem can be expressed as: netcdf can store shapes like `(dim1, dim2, ...dimN, '...')`, but what we want for the monodromy matrix is `('trajectories', '...', 'ndim', 'ndim')` and this would require creating a complex varlength var_type which is not supported. For a single float or int this is no problem. So either 

1. create a wrapper object to that an int is enough, or
2. convert `('trajectories', '...', 'dim1', ...,  'dimN')` to `('trajectories', '...')` by flattening the last dimensions, then save and when loading use something like `.reshape((-1, dim1, ..., dimN))` to get the matrix object back. This would be a nice addition to netCDF in general and could be easily done. It requires some tricks to make this still fast. Only downside is that the full variable length object needs to be stored and loaded at once. You cannot just load single _frames_. This would require loading parts of a variable length object which is not supported.

So, in case of your very big matrix it coule really make sense to take the wrapper object approach. Still I think allowing to do 2. would be nice.

jhprinz:
Another problem occurred. There is a apparently a bug (or more) in `netCDF` so far. If you are trying to set several variable length value at once using slicing then you cannot have all of them the same length.
Kind of weird but I think it has to do with numpy casting arrays of VLEN arrays into a list of objects and treating same length arrays as one big array.

I guess there will be a way around it.

Another problem is that before we used the masking feature which means that a single float or int can take a value that is interpreted as _empty_ or unset. I used this to figure out which values have been stored and which have not yet. For VLEN variables this is tricky we could define that empty lists mean nothing but then you cannot store empty lists. 

The only real way around this is to use wrapper objects.

jhprinz:
I will also close this PR. The code to be merged will not be used since CVs have changed so much. I will copy the main texts to an issue to remind us.

PR #434, opened 2016-02-18
Restructuring of Engines and Dynamics

jhprinz:
This will rearrange all the parts involved in the generation of trajectories and snapshots into a separate module of OPS. I so far picked the name `openpathsampling.dynamics`

### Things to finish

- [x] restructuring of folders
- [x] fix imports
- [x] fix tests
- [x] fix notebooks
- [x] add error message if no dynamics engine has been imported and you are trying to load from storage.
- [x] refactor docs (at least added engines to a minimum)

### Questions?

1. How to name the package? I chose `dynamics` but maybe there is a better one
2. How to name the standerd import name in `import openpathsampling.dynamics as ???`. So far I used `dyn` and kept `toys` for the ToyDynamics Example
3. Right now loading and saving only works, if the appropriate engine has been imported. What does that mean? If you load then just by running `import openpathsampling as paths` you will no include ANY engine (which makes sense) and hence opening a storage that used a particular engine will fail. Since storage has no concept of creating engines of a particular type and the associated snapshots. This means without the line `import openpathsampling.dynamics.toy` you cannot run `storage = Storage(''mstis.nc")`. But the simple import is enough to fix this. For creating a store it is mostly no problem since you need a Snapshot of the correct type to create the storage in the first place.
4. How to deal with shared objects? Right now I have put all shared files into the dynamics folder and use the `__init__.py` there to include all the shared objects. This means that `openpathsampling.dynamics` also looks almost like an engine module itself. It just has no implementation. All the other engine module then _derive_ from dynamics. I think this is nice and the package structure reflects nicely these dependence. It is almost like subclassing modules.

### Outline

The current structure is as follows
```
engines
├── __init__.py
├── dynamics_engine.py
├── snapshot.py
├── topology.py
├── trajectory.py
├── features
│   ├── __init__.py
│   ├── base.py
│   ├── shared.py
│   ├── box_vectors.py
│   ├── coordinates.py
│   ├── ...
│   └── xyz.py
├── openmm
│   ├── __init__.py
│   ├── features
│   │   └── __init__.py
│   ├── engine.py
│   ├── snapshot.py
│   ├── tools.py
│   └── topology.py
├── toy
│   ├── __init__.py
│   ├── engine.py
│   ├── features
│   │   └── __init__.py
│   ├── integrators.py
│   ├── pes.py
│   ├── snapshot.py
│   └── topology.py
└── ...
```

So the structure of an engine module is
```
└── engine
    ├── __init__.py        # make all part available
    ├── engine.py          # logic of the engine
    ├── features           # additional features
    │   ├── __init__.py    # import features from shared ones and add new features
    │   ├── feature1.py    # implementation of new features
    │   └── ...  
    ├── snapshot.py        # implementations of new snapshots
    ├── tools.py           # implementations utility functions
    └── topology.py        # implementation of new topologies 
```

If you want to use a specific engine then import it using
```
import openpathsampling.dynamics.engine as eng

engine = eng.Engine(...)
```

the top_level `__init__.py` will take care that the module will reveal all necessary function and classes when imported. There will be two classes (usually just references) that should (must?) be present in each engine:

1. `Engine()` : This is the default engine to be used
2. `Snapshot()` : The default used snapshot so that the following lines work independently for each engine

```
engine = eng.Engine()
snapshot = eng.Snapshot()
```

although the creation is of course engine/snapshot dependent. An example for the `__init__.py` from openmm looks like

```py
from engine import OpenMMEngine as Engine
from tools import (
    empty_snapshot_from_openmm_topology,
    snapshot_from_pdb,
    snapshot_from_testsystem,
    to_openmm_topology,
    trajectory_from_mdtraj
)

import features

from snapshot import Snapshot, MDSnapshot
```

the features will always be imported to allow for
```
feat1 = eng.features.[feature1]
```

The features `__init__.py` for openmm looks like
```py
from openpathsampling.dynamics.features import *
from openpathsampling.dynamics.features.shared import Configuration, Momentum
```

note the dangerous `*` which is explicitely wanted here since OpenMM should have access to all the features. For ToyDynamics we have limited these

```py
from openpathsampling.dynamics.features import (
    base,
    coordinates,
    velocities,
    topology,
    xyz
)
```


dwhswenson:
1. In #429, wasn't topology moved to being a snapshot feature?

2. Is it really necessary that the engine and the snapshot be in separate files? It seems to me that the snapshot should be pretty close to a one-liner now; in practice, when implementing a complicated engine, I'd rather have fewer nearly-empty files. Of course, in either case, the snapshot is accessible as `engine_package.Snapshot` because `__init__.py` imports it that way.

3. Reminder to include the renaming of `Configuration` and `Momentum` to `StaticContainer`/`KineticContainer`, per discussion in https://github.com/choderalab/openpathsampling/pull/429#issuecomment-191103554 (and elsewhere).

dwhswenson:
Thoughts on your questions:

#### How to name the package?

This is the hard question. We don't have to completely settle it before 1.0: I think the roadmap to 2.0 has this splitting into a separate package, and *then* it needs a clever name.

I'd like to use something based on "engine." Too bad pyEmma already has the best logo for that.

#### How to name the standard import name in `import openpathsampling.dynamics as ???`

Requires a good answer to the previous.

I don't like `dyn`, because `dynamiq` uses `dynq`, and `dyn` and `dynq` is confusing. (the subpackage `dynamiq_engine` is based on this stuff, so using them together is unavoidable). I prefer something based on `engine`.

#### Right now loading and saving only works, if the appropriate engine has been imported

This seems reasonable. For default OPS engines (for `OpenMM`, and maybe for `ToyDynamics`), can't we just include the appropriate import statement in the `openpathsampling/__init__.py`? I always have to double-check what does and doesn't get loaded there, but there should be a way to set make it available, maybe as `paths.openmm.engine`/`paths.toys.engine`, which is good enough in my opinion.

#### How to deal with shared objects? 

> Right now I have put all shared files into the dynamics folder and use the `__init__.py` there to include all the shared objects. This means that openpathsampling.dynamics also looks almost like an engine module itself. It just has no implementation. All the other engine module then derive from dynamics. I think this is nice and the package structure reflects nicely these dependence. It is almost like subclassing modules.

Exactly. See various projects in dynamiq-md which use `dynamiq_engine` (with more of those to come). The point is that this should be reusable *without reference to anything OPS-specific*. In particular, it should have no dependence on either path movers or ensembles. Again, I see this splitting to a separate package on our way to 2.0 (along with `netcdfplus`, and maybe one or two other things). I think this sort of structure is the best way to do that: it makes it easy to add new engines, as well as to use any supported engine in a new application.

jhprinz:
> 1. In #429, wasn't topology moved to being a snapshot feature?

Well, I made it optional. So that we can easily switch from `.topology` to `.engine`. So yes, it should be a snapshot.toopology attribute for now

> 2. Is it really necessary that the engine and the snapshot be in separate files? It seems to me that the snapshot should be pretty close to a one-liner now; in practice, when implementing a complicated engine, I'd rather have fewer nearly-empty files. Of course, in either case, the snapshot is accessible as engine_package.Snapshot because __init__.py imports it that way.

No, of course not. I just wanted to make a clear point to see which part is what and how to clearly separate it per package. The `__init__.py` then combines everything. How the internal package structure looks like does not matter. 

> 3. Reminder to include the renaming of Configuration and Momentum to StaticContainer/KineticContainer, per discussion in #429 (comment) (and elsewhere).

Right. I will add that.

jhprinz:
> How to name the package?

Well, I actually meant for now. How to call the sub package / folder. Would `openpathsampling.dynamics` be fine? Then I will leave it. Otherwise `openpathsampling.engine` would also be okay.

> import name in `import openpathsampling.dynamics`

Well, I see the problem. But If I understand correctly then my `dyn` will replace yout `dynq` or do you need. Maybe I wasn't clear. In almost all cases you do not need the `openpathsampling.dynamics` subpackage, but only the concrete `engine`-specific package.

```py
import openpathsampling.dynamics.openmm as dyn
```
or
```py
import openpathsampling.dynamics.toy as dyn
```
or like you do
```py
import openpathsampling.dynamics.dynamiq as dyn(q)
```

The only object from `ops.dynamic` you will need it `Trajectory` but this is available from `openpathsampling.Trajectory`, too.

Not sure if that solves the question about the name, but might simplify the problem. Otherwise I would use the subpackage name 

```
import openpathsampling.dynamics as dynamics
```

> Loading / Saving: This seems reasonable. For default OPS engines (for OpenMM, and maybe for ToyDynamics), can't we just include the appropriate import statement in the openpathsampling/__init__.py?

Sure, as long as we do not have colliding Snapshot and Engine names. Right now I have added the most logical choice for each engine / snapshot as `Engine` and `Snapshot`

```py
import openpathsampling.dynamics.toy as toy_dyn
import openpathsampling.dynamics.openmm as omm_dyn
toy_dyn.Snapshot(..)  # will give a ToySnapshot
omm_dyn.Snapshot(..)  # will five an MDSnapshot what we used to call `Snapshot`
```

We can simply add `ToySnapshot` and `MDSnapshot` as default and then loading and saving will work out-of-the-box.

> How to deal with shared objects?

This was also meant mostly for the current subpackage (with splitting off into a ral package in the back of my head). So the concept I was envisioning is that an `engine` is a _subclassed_ but stand-alone useable package and provides all we need to run dynamics. The API should be similar and share e.g. an `Engine` and a `Snapshot` object as well as tools, topologies, etc... To make this easier we provide features and other shared objects which then comprise the final `engine.[name]` package. So you can import
```
import ops.dynamics as dyn  # loads the (abstract) base package
import ops.dynamics.openmm as dyn  # load the openmm dynamics package
import ops.dynamics.toy as dyn  # load the toy dynamics package
```

Your dynamiq engine would also be treated as a (sub-)package that (for now) uses the `openpathsampling.dynamics` submodule. Note that this submodule is already independent of the rest of OPS.

jhprinz:
#### Question

I have moved Snapshots and Trajectory to `ops.dynamics` Snapshots seemed clear and Trajectory almost. There are some specific features of Trajectory which are potentially Snapshot dependent. Like the `.md` feature to create an md trajectory. I wouuld move this to `ops.dynamics.openmm` into tools. 

To generally alleviate the access of Trajectories I think passing the feature attributes to trajectory would make sense?!?

Example
```
import openpathsampling.dynamics.openmm as dyn

s = dyn.Snapshot(coordinates=..., velocities=...)
t = Trajectory([s] * 1000) # stupid trajectory but just an example

traj_coordinates = t.coordinates  # I think that would be nice
traj_veocities = t.velocities
```

Since the feature knows its a numpy array we can make this rather fast. So far I used a _dummy_ CV which seems a little over...

jhprinz:
On the auto import. Added this for now
```
dyn = paths.openmm
```

works (almost) equal to
```
import openpathsampling.dynamics.openmm as dyn
```

where the second case tries to import twice, but it works the same. This will also make sure you can load and save.

dwhswenson:
> Well, I actually meant for now. How to call the sub package / folder. Would openpathsampling.dynamics be fine? Then I will leave it. Otherwise openpathsampling.engine would also be okay.

To me "dynamics" refers to how it is being run (Langevin dynamics? Brownian dynamics? Newtonian dynamics? Quantum dynamics? Semiclassical dynamics?) whereas "engine" refers to what is running it ("Gromacs is the engine", not "Gromacs is the dynamics"). So I prefer engine. Admittedly, this is also informed by the fact that "dynamiq" is more easily confused with "dynamics" than with "engine."

> Well, I see the problem. But If I understand correctly then my `dyn` will replace yout `dynq` or do you need.

First two lines of `dynamiq_engine_core.py` might read:

```python
import openpathsampling.dynamics as dyn
import dynamiq_engine as dynq
```

This is the problem.

> The only object from `ops.dynamic` you will need it `Trajectory` but this is available from `openpathsampling.Trajectory`, too.

This doesn't help. The whole point of splitting off the engines into a subpackage is that I won't have to import OPS when all I want are the engines.

> ```python
> import ops.dynamics as dyn  # loads the (abstract) base package
> import ops.dynamics.openmm as dyn  # load the openmm dynamics package
> import ops.dynamics.toy as dyn  # load the toy dynamics package
> ```

I really don't like this approach of using **any** engine as `dyn`. That's going to be hard to read if you work with different engines at different times. It's also really confusing if you have one calculation that uses multiple engine types (e.g., Titus's approach using QM for the transition region and MM in-state). Your version with `omm_dyn` and `toy_dyn` is better.

jhprinz:
```
import openpathsampling.dynamics as dyn
import dynamiq_engine as dynq
```

Slowly I am getting what you mean. You waht a separate package for engine stuff and then load the engine. I thought that You would only need one and in your case this is `dynamiq` engine which will use the `dynamics` (or better `engine`) package (this contains Trajectory, etc...) so you should be able to use
```
import dynamiq_engine as dynq
dynq.Trajectory(...)
```

But that requires that the engine is explicitly depended on the package. Or do you want to keep that separate and if so, what should be only engine and what dynamiq?

Okay forget the dynamics and replace all my dynamics by engine. I think that better reflects what I want.

jhprinz:
What about using `openpathsampling.engines`. We could shorten `pathsampling_engine` to `peng`...

dwhswenson:
> What about using openpathsampling.engines. We could shorten pathsampling_engine to peng...

For now, I like that. Eventually, I'd really like a good, catchy name. I just don't have any good ideas for the catchy name. This is one of the coolest and most reusable aspects of OPS -- it deserves an equally cool name.

> But that requires that the engine is explicitly depended on the package. Or do you want to keep that separate and if so, what should be only engine and what dynamiq?

I don't understand quite what you meant here.

My general picture was that we'd have something (call it `openpathsampling.engines` for now) as a package that contains abstract classes (`DynamicsEngine` being the main one). You'd subclass `DynamicsEngine`, as well as anything else you needed (specialized snapshot classes? topology classes? new features? -- not to mention all the internal stuff that involves actually making the engine work!). Then you'd import that.

I'm still a little unsure how `Trajectory` plays here. It is `p_eng.Trajectory` or is it `dynq.Trajectory` (where `dynq.Trajectory is p_eng.Trajectory`, but user doesn't need to import `openpathsampling.engines`)? Or do you make the user do the extra import? Or does it even matter -- can we get away with not directly using the `Trajectory` class? (only use it inside the engine?)

Anyway, does that do anything to answer the question you were asking, or did I misunderstand the question?

jhprinz:
Okay, then I rename to `openpathsampling.engines`. I like that too. The peng is the word children in german use to mimic the sound of shooting a gun. I thought that is quite fully for _shooting_ ...

I think we have the same picture in mind. Indeed we subclass `BaseSnapshot`, `DynmicsEngine` (or maybe `BaseEngine` `BaseTopology`, etc. whatever we need.

Then my idea was to wrap everything that belong to one _simulator package_ into one (sub-)package. Like having `engines.toy` containing `ToySnapshot`, `ToyEngine`, `ToyTopology`, etc... 

This is the only package you need. So there is a base package, that contains all the `Base...` objects and it could be loaded, but a user would never do that. A power user can import it and subclass to build derived classes and make a new simulator package. 

(Note that I sometimes mix up the word engine for the sub-package and a subclass of `DynamicsEngine`)

This way - I think - we can get by without having to import two engines at a time for production or analysis and you will only import what you need. Of course an new engine could import part from other engines. No harm done here. But it would still look like one contained package.

Now about `Trajectory` because this has a special meaning. I think, we could completely hide the constructor from the user. But it would be useful to be able to construct quickly a trajectory if needed. So that means the engine package need to import `Trajectory` from the base package `ops.engines` as a default. Hence each `engine.[name]` has a `.Trajectory` class. 

This has the benefit that we can add nice utility functions per engine to the `Trajectory` object (no additional attributes - although that could in theory also be done). An example would be the `.md` property the return an mdtraj object, which only makes sense for MDSnapshots since we need an mdtraj.Topology object. But that is not necessary. The default `Trajectory` is fine. All other utilities can be placed in `.util`.

You said that you wanted to use two different engines. Now that can be done easily with

```
import ops.engines.eng1 as e1
import ops.engines.eng2 as e2
```

There are completely independent and can be used as such BUT they cannot be stored in the same file yet. At least not efficiently and only if one Snapshot is smaller than the other. Or you have to use the joint Snapshot as a template. That already works. However, you cannot mix two different types of Snapshots in the same file. To do so is possible but requires some design decisions to be made.

1. What makes a SnapshotStore unique? I would say that this is the combination of engine and topology or more precise Snapshot class and Topology. The snapshot determines which attributes and the topology their shape. 

2. This requires each snapshot to know their type (class and engine) and hence adding engine to the attributes (I think we agreed on that already)

3. Do you want to be able to freely mix Snapshot classes in a trajectory and (related) how would you access all the snapshots when you have multiple stores. How to select the stores for a given snapshot class and topology.

...

I should not be too difficult, but that would be clearly 2.0 if you do not immediately require it.

jhprinz:
Fixing the setup.py so that the conda package will provide all the subpackages

jhprinz:
Yeah, I changed the openmmtests to `CPU` and reduced the number of steps. This time the tests ran in 15 minutes... Guess that was luck.

jhprinz:
I think this is ready to review.

dwhswenson:
I thought we had agreed to use something other than `dyn` for the `import as`?

jhprinz:
Yes, sorry, I kind of thought we will change this on the go, since some of the examples will change again. But I can do it now...

jhprinz:
Actually, no we did not agree on something. You picked up my joke, but I wasn't sure if that was serious or not :) The `peng`. I would use `eng` for now?

Another idea I liked was using `motor` although engine is more general. Perhaps for a package name.

Anyway, would `eng` be okay `peng` looks so funny :) or just `pe`?

dwhswenson:
Ooh, I like `motor` as part of the future name for the package.

I like `peng` for now -- inside jokes are good things. :-)

dwhswenson:
(I had used `p_eng` to be a little clearer of the word separation, because I didn't know the `peng` joke -- but I was fine with `p_eng`, and now that I get the joke, I prefer to drop the underscore)

jhprinz:
Btw, how common is it use motor in english? I think we use it very common for engines in cars, bikes. Also electrical engines are `Elektromotor`. We also have another word for `engine` which is `Triebwerk` but that is more for planes and literally means `Drive Works`.

Alright, so we use `peng` and `paths`.

jhprinz:
Last question. What about `CalvinistDynamics` should we turn that also into in `Engine` then? It is kind of both. It defines the type of dyanmics and it is the integrating implementation.

dwhswenson:
> Btw, how common is it use motor in english?

When referring to a car, I think "motor" and "engine" are basically used interchangeably. But for whatever reason, in my head I see "engine" as pointing from the outside of the car (hood closed), whereas "motor" need the hood up to actually see it. As if motor is more specific than engine. But I don't think there's any common-usage difference between the two.

> Last question. What about CalvinistDynamics should we turn that also into in Engine then? It is kind of both. It defines the type of dyanmics and it is the integrating implementation.

Hmm... It is basically an engine for tests -- that was the original intend. So whatever works, really. But whatever you do, it should stay in `tests`, and it shouldn't change the type of snapshot. Anything you do to it should *not* change a word of its usage in actual tests. (This is where we again get into dangerous territory: I worry about invalidating important tests.)

jhprinz:
No worries, not fiddling with Calvin ;)

jhprinz:
Done. I left `toys` for ToyDynamics and `peng` for the general case.

.gitignore
```diff
@@ -106,3 +106,6 @@ target/
 
 # IPython
 .ipynb_checkpoints
+
+# .orig
+.py.orig
```
dwhswenson: hmm... actually, can we *not* hide the `.orig` files? Having them in my untracked files reminds me to clean up after merges. There's a reason that most `.gitignore`s don't, be default, include those.

dwhswenson: If you want this behavior for you, you can change your own `gitignore_global`

jhprinz: Yeah, i was wondering why that is not in there by default. I think because git tracks these and of course ignores them. I want to see if my IDE will then gray these out. Forgot to revert that.

dwhswenson:
Other than the little thing in the gitignore, all looks good. Do you want to add a quick commit to change that, or should I merge and then change it back in #435? (which should be getting finished in the first part of next week).

More general question: instead of using the decorators to set up snapshot types as you do, could you have just used the `SnapshotFactory` instead? What would it change to do that? (I was thinking of going that route in `dynamiq_engine`.)

jhprinz:
I will fix the `.orig` flaw.

I will also change one of the implementation for the default features to use the factory so someone looking at the code can see the difference and has an example.

It is quite simple actually.

jhprinz:
Removed some redundant code in `.engines`. There is now only one `Snapshot` implementation for each type. As an example the ToySnapshot is defined using the SnapshoFactory while the openmm ones are still.

One note on inheritance for features.

If you subclass, then all features are just joined: Assume the base class has features `coor` and `velo` and your new only implements `topology` then the new snapshot will have all of them with all their properties, etc. This can lead to clashed which will be caught as a RuntimeError.

If you want to use it, let me know if this is actually what you need. 

There is actually one (minor) limitation to the use of the Factory over the decorator. If you want to add functions like the `.construct` in the `openmm.Snapshot` it is very tricky to add this when you use the factory. In this case I only have the example use the decorator.

jhprinz:
I think we are ready.

Last comment for your dynamiq engine. You should add 
```
numpy = ['attribute1', ...]
```

to your features. These are the ones that will use the `np.copyto`. It also works without, but marking these are numpy should be faster.

dwhswenson:
> One note on inheritance for features.

This seems to me like the appropriate behavior. After all, this is how inheritance should work, right?

> There is actually one (minor) limitation to the use of the Factory over the decorator.

Good point, but this is also I think exactly the desired behavior. The factory is convenient if the snapshots are only containers. If they need other behaviors, a proper subclass is a lot more convenient.

I have a few more questions about how to properly use some of this, but I'll make that into a separate issue. The code in here looks good, and I'll merge it now!

PR #435, opened 2016-02-26
Single trajectory analysis

dwhswenson:
Usually we try to define states based on some information that we get from some longer MD trajectories in the state. This PR adds an analysis class to figure out a few useful facts from initial trajectories. This also serves as a quick example of how traditional analysis can be framed in terms of path ensembles. We use `Ensemble.split` to obtain:

- [x] Continuous time spent in each state of a transition
- [x] Lifetime in a state (time from first frame in state to, but not including, first frame in another state)
- [x] Transition duration (time from leaving initial state before entering final state)
- [x] Flux through the state boundary

This allows us to do all sorts of stuff, like plot the transition segments in a different color atop the whole trajectory, or get any properties of the distribution of the associated times (mean, std dev, etc). Also includes a new function to get the frame indices for subtrajectories of a given trajectory (useful when plotting).

This was started while I was working on Octomore, but it will be broadly useful.


jhprinz:
Great idea, especially to use `ensemble.split` for other purposes.

jhprinz:
If I understand correctly then this is meant to be used also with several trajectories. Maybe we can put some of the functionality into `Ensembles` or `Trajectory`?

dwhswenson:
> If I understand correctly then this is meant to be used also with several trajectories. Maybe we can put some of the functionality into Ensembles or Trajectory?

I moved the `subtrajectory_indices` function to a method of `Trajectory`. That method takes a list of subtrajectories as input, and returns the indices of those frames as numbered within `self`. So `long_trajectory.subtrajectory_indices([subtraj1, subtraj2])` returns a list (length 2) of lists of ints. This one is really useful when, e.g., plotting a long trajectory as time vs. CV and highlighting some aspect (for example, segments identified as the transition between different states). You plot the whole trajectory normally, as just the CV (automatically using the index as the x coordinate). Then you can use `subtrajectory_indices` to obtain the correct x values to plot the CV of each subtraj, overlaying those segments in a different color. (This example, as well as ongoing development of use for the rest of this PR, can be seen in https://github.com/choderalab/octomore/pull/15)

The rest of this doesn't really belong in any of the base modules. Adding it to either `ensembles` or `trajectory` would create circular dependencies between the two (perhaps not a problem for Python to handle due to duck-typing, but bad programming).

dwhswenson:
This is ready for review.

(Still waiting on the last commit to be tested, but since that only involved changing docstrings, it should pass tests.)

jhprinz:
Great. I will review this and then keep your subtrajectory indices. Can I add the possibility to use a single trajectory without  wrapping in a list?

dwhswenson:
> Can I add the possibility to use a single trajectory without wrapping in a list?

Yeah ... I'd like to find a way other than `isinstance`, because I generally hate `isinstance`. But if we can't find another way...

jhprinz:
This looks good. One question though: 

From the constructor I would assume that this is something like 2 State Analysis since it requires a transition object. Correct? The name SingleTrajectoryAnalysis seemed misleading. 

I think you mean what analysis is done. Not sure if I have a better name like
```TransitionPropertyAnalysis```

We can leave it the way it is for now and if we reorganize the analysis package think about this again.

jhprinz:
Btw. I realized that for openmm, we can access the `dt` from the integrator object to get the real integration time. I will make a little PR to implement that.

dwhswenson:
Yeah, the name isn't totally obvious. Maybe `LongTrajectoryTransitionAnalysis` or something like that would be better. The point is that this is to take an initial long MD trajectory (or a set of them) and get some information about how to properly set up a path sampling simulation. See choderalab/octomore#15, where I'm using this.

It's generally a pretty important part of getting things set up. We've counted on people writing this themselves or using external programs to do it. My example in octomore might later get adapted into proper documentation for OPS.

> Btw. I realized that for openmm, we can access the dt from the integrator object to get the real integration time. I will make a little PR to implement that.

It might be worth making it returning `engine.time_step` or `engine.dt` a requirement of the engine API. Note in this case, however, that I'm using this by loading trajectories. There is no engine to be had. (Somewhere, Sonya has the info about what the time step was in the FAH runs....)

jhprinz:
Alright. Let's keep it for now. I wait until it passes and will merge then.

jhprinz:
Merging...

Issue #437, opened 2016-03-04
Bugs in `Ensemble.split`


dwhswenson:
## Bug 1: Unexpected behavior of overlap in `split`

Using the notation `A` to mean a frame in state A, and `X` to mean a frame not in state A, what is the expected behavior of:

```python
ensembleAXA.split(trajectoryAXAXA, overlap=1)
```

I would expect this to give 2 trajectories: frames 0,1,2 and frames 2,3,4. It does not. It only gives 0,1,2.

Minimal test:

```python
import openpathsampling as paths
from openpathsampling.tests.test_helpers import make_1d_traj

def xval(snapshot):
    return snapshot.xyz[0][0]

opX = paths.CV_Function(name="opX", f=xval)

stateA = paths.CVRangeVolume(opX, float("-inf"), 0.0)

ensembleAXA = paths.SequentialEnsemble([
    paths.AllInXEnsemble(stateA),
    paths.AllOutXEnsemble(stateA),
    paths.AllInXEnsemble(stateA)
])

trajectoryAXAXA = make_1d_traj(coordinates=[-0.5, 0.5, -0.5, 0.5, -0.5])

ensembleAXA.split(trajectoryAXAXA, overlap=1)
```

Only returns one trajectory. Correct results are obtained with `ensembleAXA.split(trajectoryAXAXA, overlap=2)`.

I think I see how to fix this one, and can do so as part of #435 if you want. I think the problem relates to the question of overshooting vs. not overshooting I mention in #95 (and which is at the heart of "pred", "over", and "hyb" suffixes in my stuff on path ensemble theory.

***

## Bug 2: `Ensemble.split(lazy=False)` errors:

```python
ensembleAXA.split(trajectoryAXAXA, overlap=1, lazy=False)
```

```pytb
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-3-2f95ecd8b92c> in <module>()
----> 1 ensembleAXA.split(trajectoryAXAXA, overlap=1, lazy=False)

/Users/dwhs/Dropbox/msm-tis/openpathsampling/ensemble.pyc in split(self, trajectory, lazy, max_length, min_length, overlap)
    432             # overloading (which is not used for Snapshots) and type()
    433             indices = self.find_valid_slices(trajectory.as_proxies(), lazy, max_length,
--> 434                                              min_length, overlap)
    435 
    436             return [paths.Trajectory(trajectory[part]) for part in indices]

[...]

PR #441, opened 2016-03-07
Fix bug in ensemble.split

jhprinz:
This adresses issue #437 and should fix it.

It also add two more features:
1. `reverse : bool` : Chose to reverse the order in which the splitting operates
2. `n_results` : int : Limit the number of results for speed up

There are also two new function as short-cuts:

```
find_first_subtrajectory(trajectory)
find_last_subtrajectory(trajectory)
```

Trajectory has a new function
```
trajectory.index_of(sub_trajectory)
```

which returns a list of indices where frames in sub_trajectory are located in traj.


jhprinz:
Ready to be reviewed.

openpathsampling/engines/trajectory.py
```diff
@@ -284,6 +284,28 @@ def next(self):
...
+        """
+        ret = []
+        for s in sub_traj:
+            if s in self:
+                ret.append(self.index(s))
+            else:
+                ret.append(None)
+
+        return ret
+
```
dwhswenson: Can we keep my implementation in #435?

dwhswenson: I'm using it -- a lot -- in the rest of the analysis. It differs by taking a list of subtrajs.

jhprinz: Sure, I forgot that you already did that. I would prefer that it can be used with a single trajectory and handle it correctly. Is that okay? I am sometimes annoyed when I need to wrap a single object in square brackets. This would require a `isinstance(traj, Trajectory)` though since both are iterable. Is it possible to merge #435 first? Or is this one required. Then I would copy your function to my PR. ...

jhprinz: Will merge after #435.

jhprinz:
Merged the two versions of subtrajectory indices and kept your name. I also remove the feature that a non-existing frame will return `None`. Now if will just fail. I think so far we don't need it. And this way we are sure that a list of ints is returned.

jhprinz:
Ready to be reviewed. Tests are also added.

dwhswenson:
Looks good (took a couple restarts to get the tests to pass).

I like the extra features -- I think those will be useful.

PR #442, opened 2016-03-08
[WIP] OPS on remote kernels

jhprinz:
This is a little experiment. I wanted to see, if OPS objects can be easily pickled and send to a remote kernel and retrieved. This could be used in potential externel OPS clients. Similar to what _Siegetank_ was doing. 

The interesting point was to reuse all the complicated caching mechanism to minimize traffic over the net. 

So, all OPS objects can be saved (but not serialized into JSON). E.g. `Trajectory` does not work, but I added that. This can be easily fixed. The more difficult problem is when you want to send

```
traj = Trajectory([snapshot1] * 10000)
```

over the net. Not that this is the common case, but it was the origin of the idea. The caching and storage code takes care to only save one snapshot, etc. 

Now, is it possible to decompose the serialization into elementary parts and speed up the transmission.
Even better would be, if on both ends there is a caching that is consistent to that when you send the
same or a similar object again. It will only send the parts that are new or are not in the cache anymore.

Now the cool part. Since storage implements more or less all the logic to split the loading and saving into the elementary objects. I created a fake storage that acts on a remote kernel. This code is actually quite short and you can send and receive any Python object to and from the kernel and it handles all the pickling and caching in the back.

Currently there is no direct use for that, but in the long term I think it is very useful.
### Example

``` py
# Create new client kernel
from jupyter_client.manager import KernelManager
km = KernelManager()
km.start_kernel()
kc = km.client()
kc.start_channels()

# Open a connection to the kernel
st = RemoteMasterStorage(kernel)

# Make a test trajectory
traj = paths.Trajectory([my_snap] * 10000)

# Push the trajectory to the kernel and give it a name
st['traj'] = traj

# run some code on the kernel
st.tell('traj2 = traj')

# try to get the trajectory back
traj2 = st['traj2']

# if that worked with all the caching than it should now, that traj == traj2
print traj is traj2
>>> True
```

I added some functions to run code on the kernel which still needs improvement. But as a proof of concept this is very nice. 

Ideas are welcome.


jhprinz:
Some more examples

```py
# assume a simple 2D setup as in the toy examples
st['engine'] = my_engine
st['template'] = template

cv = paths.CV_Function('sq', lambda x : float((x.coordinates**2).sum()))

st['cv'] = cv

# setup the engine and run
st.tell('engine.current_snapshot = template')
st.tell('traj = engine.generate_n_frames')

# get the result of the CV
cv_values = st['cv(traj)']
print cv_values
>>> [ 0.123, 0.453, ... ]
```

dwhswenson:
Funny... I was literally *just* thinking about something else that is a bit related. My main concerns are about storage. In particular, I was wondering if we had the ability yet to combine multiple storages into one. (I know we discussed this at one point, and I was trying to find that issue as you were posting this!)

For context, I was thinking about writing a little script to do a committor analysis. In terms of code, this is very easy: you just start from a given snapshot, and the trajectory stops as soon as you enter one of the states. But to get statistics, you want to run several trajectories (likely with randomized initial velocities) from the same configuration. Let's say you wanted to do 100 trajectories. (In practice, we often use this for purposes that require even fewer, but still.) Obviously, you don't want to have to run them sequentially: if you have enough nodes, you can do each of the 100 trajectories on a different node.

The question then is how easy it is to combine the storages into one storage for analysis. This relates closely to what you're talking about above -- we could have a master `CommittorRunner` class that dispatches the individual trajectories to different remotes, and then everything would be returned to the master node.

However, this approach could be very inefficient when it comes to load balancing. Let's say that one trajectory of the 100 was much longer than the others. You reserve 100 nodes using your queueing system, and most of them are doing nothing for most of the computation. This is the fundamental problem of load balancing with flexible path length trajectories, and shows up in TIS as well.

A better approach would be to treat each run separately, and have a way of combining the output files from each into unified file for analysis. Can we do this? If not, this would be a good thing to add in the context of this PR.

In general, we should try to work within the framework of existing queueing systems when thinking about parallelization. Reinventing the wheel is going to be a lot harder, and won't be allowed on all clusters (where usually you HAVE to use the pre-installed queueing system).

jhprinz:
I think you are right in most points. But this PR was not meant to do all of that. I was just curious to see, if I can efficiently pickle stuff, send it via a string and reconstruct. And this works extremely well and will very little code.

About the load balancing, I kind of disagree (with my limited experience). I see the problem and in an adaptive scheme it can very well happen that all that comes next is dependent on a single very long trajectory. This is true, but I also think that if the system and ensembles are large, the chance that really all next steps will depend on a single trajectory is very small. 

Most of our moves will only depend on some of the past so most other moves can be run anyway and if you have enough other stuff to do, also the long trajectory will have finished. If I remember correctly then we use a cut-off and say that these will be used to discover new states.

So the solution could be to design your scheme to not be too easily be dependent on a single result and not reserve 100 nodes for only 100 trajectories with different run lengths or course. In your idea all runs are independently and If you say 100 nodes and 100000 trajectories, then I assume by reserving 100 nodes your dead-time will be quite small. Or course if you spawn 100 and wait for all of them to finish then you loose too much. I have to admit that I do not have too much experience with this. But my feeling is that this problem can be solved with some tricks like looking ahead in the movetree and run next oves as much as possible as well as predict possible future moves.

In the end it probably boils down to reserving little clusters of nodes using the queing system and then use them directly. If you have to wait all the time to submit a job that can slow you down a lot. 

> In general, we should try to work within the framework of existing queueing systems when thinking about parallelization. Reinventing the wheel is going to be a lot harder, and won't be allowed on all clusters (where usually you HAVE to use the pre-installed queueing system).

Yes, that would be my first idea, too and that is exactly what I should implement for Frank I think. It is in concept very similar to how HTMD works: Your API is MD engine and cluster specific and you need to implement on how to submit jobs to the queuing system, check the status and retrieve results.

The remark about combining storages and objects is a good point. We talked about this and it would require to store something like a uuid with each object. That can easily be added. Then you need to be able to load object by uuid. You could then create a joined store that loads all uuids from all stores and fakes a single storage. This is common in distributed data and that should be possible. Using uuids with all objects will of course introduce an overhead that is bad for small systems.

In this PR it just keeps (storage-like) references to objects on both sides (client and master) which was easy to do and hence no uuids are needed. 

There are still some sutbleties like how to interprete the `.mcstep` when combining different simulations. Maybe the number gets meaning less. In theory you will get a tree like structure of PathMoveChanges, etc... Should be pretty cool.

I still need to investigate this a little more. This was just for fun :)

dwhswenson:
> About the load balancing, I kind of disagree (with my limited experience).

Last night I wrote a little game to play to test out any ideas you have on this. https://github.com/dwhswenson/parallelRETISgame

This gives you a way to play with your ideas without having the heavy expense of a real simulation. I'm still skeptical that you'll get much improvement (it only works if you're more limited by computing resources than by ensembles -- and since the general trend is toward many more nodes for computing, I don't think that's a forward-thinking problem to solve).

The current version of the game needs improvement to match reality (it currently has constant cost shooting moves -- the fact that this isn't true is actually the thing that makes it hard).

Anyway, although I'm skeptical, I hope you do find a way to make a significant improvement, both because it would be useful for TIS in general, and because then we could write a paper about how the approach was developed by doing simulations of our simulation. ;-)

(Note, by the way, that SRTIS can be parallelized to infinite nodes and still has 100% computational efficiency. This is the point -- you can't beat that. The only question is how efficient it is at exploring trajectory space.)

> So the solution could be to design your scheme to not be too easily be dependent on a single result and not reserve 100 nodes for only 100 trajectories with different run lengths or course.

You're missing the specific point I was making above. I'm doing a committor analysis. I have exactly 100 trajectories to run. They are independent. I want to analyze them together. The best way to do this is to launch 100 *separate* simulations and trust the queueing system. But then I want to analyze them together. That's where I'd rather we have something in OPS that makes this easy, instead of having to hack something together. I think the general problem is going to be common enough (again, SRTIS can be run on as many independent nodes as you want -- but the analysis should be joined together. Similarly, you can run normal RETIS with many copies, and then analyze the results together.)

> There are still some sutbleties like how to interprete the .mcstep when combining different simulations. Maybe the number gets meaning less.

It only has any meaning at all within a specific sequence (within one Markov chain).

> In theory you will get a tree like structure of PathMoveChanges, etc... Should be pretty cool.

Dangerous. You start to lose any natural support of detailed balance if you do this, since the whole concept of Markov chain Monte Carlo is that you have a *chain* of states. You can, however, average the results of multiple independent chains. That's the way to think of this.

jchodera:
Would love to chat about parallelization ideas sometime.

Some things to think about:
* Having something that automatically interacts with a batch queue system
to run every trajectory is a terrible idea. All the batch queue systems are
different, and it would be nearly impossible to keep anything up to date.
Some systems support a common DRMAA inferface, but the people at MSK who
use this continue to get burned whenever something breaks this. Also, you
lose efficiency by waiting for things to go through the queue and through
the overhead of starting things up
* Instead, a strategy where workers attach to one or more master processes
to make themselves available in a pool would be ideal. This way, users
could launch workers wherever (desktop, cluster, cloud) and spin up as many
as they can afford. There are plenty of frameworks that could be tapped to
support this in a way that is robust and requires no effort from you. (Ask
@pgrinaway, who has used many of these.) The JSON work can simply provide a
way to serialize data for transmission over the network.
* One way to parallelize is to propose not just one new proposal for any
sort of shooting step, but propose multiple proposals and pick from the
stationary distribution of all proposals plus the old trajectory. (See the
"Waste Recycling Monte Carlo" paper.) While the speedup is not large for
cases where acceptance is already high, in the low acceptance regime this
offers almost perfect speedup.
* Running independent SRMSTIS replicas does not offer perfect
parallelization except in the case that all replicas are already in
equilibrium and explore space infinitely quickly. You can always improve
this in practical cases by weakly coupling together multiple chains through
the weight updates. We are also exploring optimal ways to do this based on
stochastic approximation theory, which shows how probably asymptotically
optimal update schemes can be developed. Would love to chat about this
sometime.

jhprinz:
Just talked to Patrick and apparently there are lots of Python Frameworks out there that are useable. We should really chat about this, when we find some time. 

From the storage / OPS object structure I am almost ready to move to a UUID based approach. That means we reference all objects (snapshots, trajectories, Samples, Moves, Engines) by unique identifiers and so we can really run independent jobs, store data or transfer data between nodes and the references/dependencies between objects remain working. The JSON pickling provides a way to effectively transport and cache redundant data, but that is not necessary.

In some way this can be used very similar to what Siegetank did. Since Frank is also interested in the Adaptive Samplig stuff I will invest some time here anyway.

One thought about the problem of few very long trajectories. I think we can also use the engine / storage / parallel running of trajectories without detailed balance in replica space. Most adaptive schemes we want to do in the very near future are for exploration and state discovery and here we do not yet need correct path sampling. Also, in case we want to do only stationary properties we could run other schemes which might not suffer from the Minus Move problem.

jhprinz:
> The JSON work can simply provide a way to serialize data for transmission over the network.

Yes, that is the idea and this will work already with the upcoming UUID approach. You can simply save any object in OPS also as a JSON string (which is not as effective as the netcdf native approach), but it works and is almost identical to how Siegetank pickled its data. I.e. Numpy array will be stored with a `base64` encoding plus `shape` and `dtype`. Including the use of `simtk.units`

The current ops_kernel version will use UUID to run ops on a separate kernel and so you can mimic a parallel execution on your own machine. Never tried that, but it should work.  

The main point missing is OPS running a move scheme is parallel. I am working on that and have found a way, but it is not very elegant yet. Still, this can auto-parallelize and move tree and so a user does not have to worry about this. You only need to provide a pool of engines and they will be loaded with jobs. More on this on a separate PR.

This PR is contained in #443 and should be closed by it.

jhprinz:
> The question then is how easy it is to combine the storages into one storage for analysis. This relates closely to what you're talking about above -- we could have a master CommittorRunner class that dispatches the individual trajectories to different remotes, and then everything would be returned to the master node.

Never answered this actually. This will be possible with PR #443. Almost trivially. The question is if you want a tool to combine exisiting storages into one or you want to open a bunch of files as one virtual large file.

Both is possible but I have still no shortcuts for these.

jhprinz:
This got an update to make it work a cluster. Allegro, here we come...

dwhswenson:
Marking this for closure. It was an interesting early experiment, but (1) the PR has far too many conflicts to merge; and (2) the core goal is currently in progress using Dask as the underlying parallelization engine.

I will leave this open for comment for at least 7 days, closing no earlier than Wed 30 Dec 06:00 GMT (07:00 local).

Issue #444, opened 2016-03-11
Revisiting new `Snapshot` structure


dwhswenson:
This follows up on some stuff from #429, based on a little experience actually using that stuff in practice.

### `copy` and `deepcopy`

As discussed previously, I often want to make a copy of a snapshot where certain features (`coordinates`, `velocities`) are deep copies.  Your solution was to say "use `copy.deepcopy(snapshot)`." However, this is a problem, because some features don't make sense to deepcopy -- e.g., `Topology`. Suppose I'm doing a random walk in `Snapshot` space (two-way shooting moves are an example where we'd do this). I may only change, e.g., the velocity of one atom. So it is very convenient to say "I want a deep copy of this snapshot, and then I want to change that one part."

A similar thing happens in dynamiq, where I'm sampling initial conditions: in that case, I don't even know *a priori* which features exist.

What I still think I'd like would be a `Snapshot.copy()` function that returns a new snapshot that does a deep copy where relevant and a shallow copy otherwise. Each feature knows whether it is deep or not (topologies/masses aren't, but coordinates/velocities are) so I think that could be as easy as defaulting to deep copies, but letting a feature override with a list like `no_deepcopy = ['topology']` in `topology.py`.

### `requires` vs. `provides`?

The way things are set up now, we have two separate features for `coordinates` and `xyz`, but `xyz` requires `coordinates` (and you have the `requires` keyword set up to handle that). Could there also be a `provides`, that would allow us to combine `coordinates` and `xyz` into one file, so that any time a you say "this snapshot should have `coordinates`" you also get `xyz` for free?

This shouldn't replace the `requires` trick; instead it should augment it. For example:

```python
# coordinates.py
attributes = ['coordinates'] # settable attributes
provides = ['xyz'] # property
```

Consider how this makes the following easy:

```python
# topology.py
attributes = ['topology'] 
provides = ['masses'] 
```
```python
# velocities.py
attributes = ['velocities']
requires = ['masses'] # can come from topology or from elsewhere!
provides = ['momenta']
```

I think that would make setup of new snapshot types a lot easier. It would also mean that we could add momenta to our existing snapshots at some future date without breaking backward compatibility. (Although if we implement this, we might as well go the extra step to add momenta now.)

### One feature with multiple attributes?

A question: currently, all features follow a few rules:

* in a file called `feature_name.py`
* contain exactly one attribute; i.e., `attributes = ['feature_name']`

[...]

Issue #416, opened 2016-01-13
Sudden flip of the sign in velocities


jhprinz:
@bolhuis mentioned this in an example. That suddenly in a trajectory there is a flip of sign which should not happen.

This needs to be fixed!!!

jhprinz:
Identified the issue. Just looking for an elegant solution. This is highly non-trivial and involves
1. features
2. memory efficient caching
3. Snapshot classes
4. Paired creation of Snapshots

need to think about this a little. Try to post a solution soon.

There is one design question I need to know your opinion:
Storing reversed snapshots have a flag. Do we always want to store the same velocities in the database and determine flip from the reversed bool. Or (like in the case of ToySnapshots) do we allow that velocities and *1.0 * velocities is saved as a pair. That is nicer when looking at the storage without OPS.

Remember in MDSnapshots we store two times a reference to the same velocities and only use the flag.

dwhswenson:
It seems to me that it is pretty useful to only save a flag for reversal -- much smaller files. Only one copy of the velocities is saved, and it is flipped programmatically if the snapshot has its `is_reversed` flag set. So I think we should do that. I didn't realize the `ToySnapshot`s were saving both versions of velocity -- of course, for the 2D systems, that isn't a big hit on storage space. But for 10k atoms, it would be a problem.

As for looking at snapshots outside of OPS, I guess the question is how will people do that. If you're using these objects, but never using any kind of reversal, than the `is_reversed` flag is just an extra (unused) thing in storage. I don't think that's a problem. The only confusion would come from either (a) trying to work directly from the netCDF file without going through netCDF+'s object re-building process, of (b) trying to copy trajectories generated from a code with the `reversed` flag to a code that doesn't use it (and probably then using direct netCDF or some other way to analyze it).

(a) I don't think many people will work directly from the netcdf file -- most will probably go through the NetCDF+ interface, which means that they will work from the `Snapshot` objects (especially if we plan to split the engines/snapshots/trajectories into a reusable package later). As long as the things stored in the netCDF are well-documented, a user can write another reader for that (e.g., in C instead of Python) to re-create the trajectories correctly.

[...]

PR #445, opened 2016-03-12
[WIP] Improve handling of features

jhprinz:
Addresses issues discussed in Issue #444 .

1. `.copy()` for numpy properties
2. `provides = # list of loaded features `

Almost ready. Still missing extensive testing for edge cases like inheritance and user defined function.

Changes in this PR:
- [x] changes `attributes = []` to `variables = []`
- [x] add optional `storables = []` to override set of variables to be saved, this way also attribute-like properties can be saved
- [x] change `.copy()`, etc. to make half deep copies. Duplication numpy, but not OPS objects
- [x] use a `CodeContext` object to easily and cleaner add functions by sourcecode to snapshots
- [x] add `test_snapshot.ipynb` integration test
- [x] move `xyz` property to features instead of stand-alone feature
- [x] `variables = []` will refer only to attributes not properties, properties are detected
- [x] add possibility to use setter for properties, so attributes can be defined using properties
- [x] some cleanup in the code generation
- [x] better handling of relaying attributes / properties from snapshots to trajectories
- [x] add better comparing function for tests that includes the reversed snapshot and checks for feature equality
- [x] add some feature tests
- [x] check for correct handling of reversed snapshots (done by adding reversed testing in snapshot_compare). Addresses issue #416.


jhprinz:
This should already do it.

jhprinz:
> Suggestion, after working with it a bit: would it make sense for Snapshot.__features__ to be a namedtuple instead of a dict? That feels more natural to me in usage (snap.__features__.classes instead of snap.__features__['classes']; the question is how flexible the keys in __features__ are.

Yes, that does indeed make sense. I will change that. The dict I am using is really just a container to hold various information.

jhprinz:
Alright. Now you can access the information in `__features__` like attributes. Example
```
print MDSnapshot.__features__.variables
print Snapshot.__features__.classes
```

jhprinz:
I added a some nice pythonic way to get the reversed snapshot
```
reversed_snap = snap.reversed
# or
reversed_snap = - snap
```

jhprinz:
I will make some more tests and check that it works.

jhprinz:
I simplified the code generation and use a `with` wrapper now that takes care of the function creation.

jhprinz:
added a notebook test that checks all the basic features like subclassing, conflicting features etc... we should at some point change this to ordinary tests to include this into coverage.

But for now it is ready to be reviewed.

jhprinz:
Btw. I really like the new way to construct a function like
```py
context = CodeContext(cls, __features__)
with context.Function('copy_to)' as code:
    code += ['def copyto(self):', '    pass']
```
the `with` statement provides a `Code` object that looks like a list of strings. When the block ends, the function will be compiled and automatically added to the class. That makes the code construction much cleaner and more understandable.

dwhswenson:
A couple quick questions/suggestions, while we're working on this part of the code:

#### `do_not_save`

Can we add a `do_not_save` list to `snapshot.__features__`? For example, in semiclassical calculations the monodromy matrix is just an intermediate (that needs to be calculated during the MD), which is used to get the semiclassical prefactor, which is what we really want. Saving the monodromy matrix is, in many circumstances, unnecessary and very wasteful of disk space. Also, I think that ability to selectively choose (at `Snapshot`-class level) whether to save a given attribute, along with the ability to have arbitrary properties as `features`, should completely solve everything in #407 ("engine-native" collective variables).

I can currently get this behavior by having two versions of the same feature (one with saving, one without), but that seems a bit less elegant.

#### casting snapshots

One of the things I'm working on in `dynamiq_engine` is the idea of casting snapshot types.

The problem I'm trying to solve right now looks like this: The integrator knows that it needs a monodromy matrix (because of the type of calculation we're doing). However, the user gives a snapshot that doesn't have the monodromy matrix as one of the features. In principle, this should be fine, because every trajectory starts with monodromy==identity, so it doesn't need to be user-specified.

To make this a bit more complicated, I can't be sure exactly which other features the snapshot might have (sometimes we have electronic dofs, sometimes we don't). So what I want to do is add the monodromy feature to an existing snapshot type (which technically should change the type) and still be able to save/load. What is the recommended way to do this? Possibilities:

* Directly add it to features, modifying the snapshot type. This seems straightforward, but is it safe in practice? Does it break storage?
* Dynamically subclass the existing snapshot's with a new class that includes the monodromy matrix. This seems like it should be possible, but trickier. I tried this with stuff based on the current master (so not including this PR) and had difficulties, although I haven't had a chance to dive into what exactly the difficulty was.

Suggestions on how to go about solving that problem? Anything that should be changed as a part of this PR to make that possible/easier?

jhprinz:
added a test that checks whether snapshots do correctly create reversed snapshots and these are stored correctly. This should resolve the problem that @bolhuis had. Thought it makes sense to fix that here too.

jhprinz:
> __do_not_save__

Right. We talked about this and that is an important feature. Right now there is no way (if I am correct) to exclude a variable from being saved unless you create a property out of it. If I understand correctly you want to set the `do_not_save` indepedent of the features used. I would say that using a new feature (that is temporary) is the preferred way. You can overwrite the `do_not_save` property in __features__ 
```py
Snapshot = # from snapshot factory...
Snapshot.__features__.do_not_save[:] = ['var1', 'var2', ...]
```
Since `__features__` is a named tuple you cannot replace the content you can only change the existing list.

I propose to not name attributes to be excluded, but rather to name the ones to be stored. That forces the user to think about the storage the same he has to do when thinking about the variables type. I would call it `persistent`

```
variables = ['coordinates', 'helper', 'velocities']
persistent = ['coordinates', ''velocities']
```

An while we talk about it. Maybe we should simplify the variable creation for storage. Since the creation of space in the storage and persistent is redundant and hence a source for error. Example

```
# coordinates.py
variables = [
    Variable('coordinates', 'numpy.float32',
        dimensions=('atom', 'spatial'),
         description="coordinate of atom '{ix[1]}' in dimension '{ix[2]}' of configuration '{ix[0]}'.",
         chunksizes=(1, 'atom', 'spatial')
    ), 
    'helper'
]
persistent = ['coordinates']
...
# if we need to add a store it could look like
# statics.py
...
stores = [
    Store('kinetics', StaticContainerStore())
]
```
This way we can remove the `init_netcdf` and do it automatically. 

Just an idea. More to come...

jhprinz:
> casting snapshots

need to think about that. This is tricky and relates to the problem to store different Snapshot types in one file.

dwhswenson:
> added a test that checks whether snapshots do correctly create reversed snapshots and these are stored correctly. This should resolve the problem that bolhuis had. Thought it makes sense to fix that here too.

I think this is only half the testing solution we need for that: we should also have tests of basically a complete shooting move, and those tests should ensure that the trajectory that comes out is correct. I think I did this, but only for `CalvinistDynamics` -- we should probably find ways to make such tests with more realistic engines.

(I think this is, in fact, fixed now -- but it's the sort of thing that can easily regress, and regression would be a devastating problem, since it could invalidate just about all results from our code. So we should have lots of very strict tests to make sure we don't regress.)

dwhswenson:
> Right now there is no way (if I am correct) to exclude a variable from being saved unless you create a property out of it.

What if you just don't include the `init_netcdf`?

> I propose to not name attributes to be excluded, but rather to name the ones to be stored. That forces the user to think about the storage the same he has to do when thinking about the variables type. I would call it persistent

This seems like a good idea. This should mean that you can mark properties as `persistent`, too, right? That would be part of solving #407.

> Maybe we should simplify the variable creation for storage. Since the creation of space in the storage and persistent is redundant and hence a source for error.

While your example is cleaner, it looks like it might break independence from `netcdfplus` (or is `Variable` not a `netcdfplus`-specific thing?). If I have no interest in storage, I should still be able to use `features`. The current approach only uses `init_netcdf` when `netcdfplus` is present, right?

I know that all the engines technically inherit from stuff in `netcdfplus`, but if I really didn't want to use `netcdfplus`, I could duck-punch in a fake `netcdfplus.StorableNamedObject` which is just `object`, right? (Maybe with a few do-nothing methods). That would break all the storage features, but it shouldn't break much else -- at least, if we've done a good job of encapsulation/abstraction.

I also tend to feel a little distrustful of mixing types in `list`s. I'd rather everything be of the same type, if that would be possible.

jhprinz:
> What if you just don't include the init_netcdf?

I think that would create an error since you try to save all `variables` from that feature but there is not part in storage created. I am not 100% sure if I added fail safe that would then just do nothing. 

> This seems like a good idea. This should mean that you can mark properties as persistent, too, right? That would be part of solving #407.

Well, yes with a but. You could save properties, but you cannot load them, unless you have added getter AND setter. Then yes, I think so (although I never thought about it). In principle that is what lazy is doing, too. It add special getter and setter that save the content and handle lazy loading.

> While your example is cleaner, it looks like it might break independence from netcdfplus (or is Variable not a netcdfplus-specific thing?). If I have no interest in storage, I should still be able to use features. The current approach only uses init_netcdf when netcdfplus is present, right?

That is a part that I still need to think about. The main idea behind it is to just specify the variable type in a storage independent way. So Variable would be something that netcdfplus inherits from, not the other way. (Though I do not know where to put that). Basically it would be a variable with a type. In type-safe languages that would not be necessary, since we can check the type, which is fixed. By using `Variable` we say this is a `float` or `str` or `numpy.float32`, etc. And the storage will create appropriate containers. 

Right now this Variable will contain all that is useful and needed for netcdfplus to work. We might even just attach these Variable descriptions to each storable class and then you don't need to build a `Store` class. Much like the `VariableStore` works right now. But the information about the types is saved with the object.

So, instead of a user trying to figure out how to save something, he will describe the data to be saved and the rest does the storage...

> I also tend to feel a little distrustful of mixing types in lists. I'd rather everything be of the same type, if that would be possible.

Yes, that is not nice. You could just set the Variable definitions using a dict like
```py
types = {
    'coordinates' : Variable('numpy.float32', ('atoms', 'spatial'), ...)
}
```

There is not another problem visible that has to do with storing multiple snapshots in one file: You need several `atoms` dimensions. One per Snapshot.

Before I forget: One last _crazy_ idea: We might attach the topology to the Snapshot class. And even better `topology` and `engine`. I think I should open another issue on that idea...

dwhswenson:
> We might attach the topology to the Snapshot class. And even better topology and engine. I think I should open another issue on that idea...

#311? (I'm pretty sure `features` can basically solve my concern with translating between engines, BTW.)

A couple things about "topology" here. I don't think you want a topology as such -- what you want has nothing to do with which atoms are bonded to which, or how to interpret that in a force field (which is what the molecular mechanics topology is actually for). It's just that, in OpenMM, the number of atoms is stored in the `Topology` object (it *is* a convenient and logical place).

It seems like what you really want is the shape of these arrays, so that you can initialize the netcdf tables. For `coordinates` and `velocities`, that's defined by the number of atoms and the number of spatial dimensions. You don't need a whole topology.

How about a dictionary called `shapes`? Includes `n_atoms` and `n_spatial`. In addition, by making it a dictionary, you can add other entries. For example, I would much rather deal with `n_dofs = n_atoms * n_spatial` when dealing with the monodromy matrix, so that it is a `n_dofs` x `n_dofs` matrix, instead of a 4-dimensional tensor of `(n_atoms, n_spatial)` x `(n_atoms, n_spatial)`.

This would also enable other custom "sizes." For example, when doing nonadiabatic dynamics I need to know the number of electronic states. This is stored in `n_electronic_states`, which can be another entry in that dictionary. If you want to store the state of a Nosé-Hoover chain, you can have a `n_nhc` that you add to the `shapes` dictionary, enabling the storage of the auxiliary variables. (I'm of the opinion that this is technically required for doing path sampling with an NHC integrator, although some might argue with me.)

As part of the template-based initialization of an engine (using a snapshot that *does* include a `topology` attribute), you can always extract that dictionary from the topology. The engine can create the dictionary, and then link each snapshot it creates to that dictionary -- either as `snapshot.shapes` or as `snapshot.engine.shapes`; I think either is fine.

jhprinz:
> #311?

No, that is about having an attribute per instance not per class

> I don't think you want a topology as such

Right, that would fix it, but if we are introducing another attribute we should do it right and add all information needed to store information like the variable type and potential chunks, etc. 

My aim with topology was to keep it simple. I think usually you are the one voting for simplicity. The topology (right now) contains the information we need so just using it would be simple and do the trick.

> As part of the template-based initialization of an engine (using a snapshot that does include a topology attribute), you can always extract that dictionary from the topology. The engine can create the dictionary, and then link each snapshot it creates to that dictionary -- either as snapshot.shapes or as snapshot.engine.shapes; I think either is fine.

I am not really sure what you mean here. You say we should use the topology that contains the information and translate it and store it differently instead of using the topology?

It will also not solve the problem of using two or more different snapshot types. Currently I am thinking about using a `Structure` class which is basically containing the types of all attributes for classes that are stored in this way. This object contains exactly what you need to generate containers in the storage. 
I think you wanted the possibility to have snapshots with certain features and without it. This would allow this. Downside is the overhead of each Snapshot pointing to a structure object. What I meant before with attaching it to the class is to circumvent using an attribute `.__structure__` that nows the _shapes_ but to force the user to make separate snapshot classes. Each class gets the class specific information attached like `__structure__` and (potentially) `topology` and the instances have all the features. 

So, creating a new snapshot class means setting the shapes also. If you simulate two systems and want to store them you need to create two Snapshot classes (with different names).

Storage right now has a fixed structure per class. This is the whole problem. Instances of the same class that have a different structure/shapes cannot be stored in the same file so far. All of the discussion about engine, topology, snapshot, features, etc... is to fix that problem in the cleanest and simplest way. If python would have variable types that would (almost) fix it. This is way using the topology / structure / shapes with a class would be so nice. It is like putting strict variable types (including shapes) to the attributes. If you want different ones, you need to build a new class. These classes would of course be stored as well.

This is still working in my head. But it is more complex than I thought.

jhprinz:
Question. Where do we want to draw a line for this PR. What features do you really need ASAP and which can be post 1.0? So I can better focus.

dwhswenson:
> Where do we want to draw a line for this PR. What features do you really need ASAP and which can be post 1.0?

I think this is the question that is really important for you, so let me start there. Longer response to the longer post will come in a second post.

First, this gets back to some of the stuff I’ve said (in backchannel email) about [semantic versioning](http://www.semver.org). I’m operating under the assumption that we plan to use that. That means that we can’t make backward incompatible changes in the API after 1.0 and before 2.0.

If we stick to the rough schedule I proposed of 1.0 in first-half 2016 and 2.0 not before early 2017, then anything we want to do before 2017 needs to be possible with the 1.0 API. This is the key point: there are things I’d like to finish in the second half of this year that require this API to work.

So things fall into three categories: 1.0, 1.x, 2+. (Again, @jhprinz, this will sound familiar from private email.) Important here is that 1.x needs to be *possible without breaking backward compatibility*, which in some ways makes it the same as 1.0.

Of the things in this thread, I’d split them like this:
* better `.copy` behavior: 1.0 (ASAP). Lots of stuff is waiting on this.
* ability to save arbitrary shaped arrays (what I propose in the `shapes` dict): 1.x. This is essential for other engines such as `dynamiq_engine`, or for some integrators (like saving Nosé-Hoover chain status). I don't think it can be hacked in by other means.
* casting between snapshot types/storing multiple snapshot types: 2+. I think I can hack the necessary behavior into `dynamiq_engine` to do what I want, although I think it would be more elegant to have it as a part of the core OPS. I also think this is an important longer-term goal, especially if it also involves support for storage of multiple snapshot types. For example, there’s the approach Titus has explored of using molecular mechanics for the minus move and ab initio MD for the reactive region. I think that is a very clever idea, and I’d like us to support it -- but that can wait until 2017.
* `do_not_save`/`persistent`: 1.x or 2+. In principle, we can hack this behavior when we want it, but it’s far from elegant. If it isn’t hard to implement, it would be nice to have it available in earlier projects.
* simplified variable creation: 2+. I think we’re getting toward something nice, but, with the exception of being able to access arbitrary dimension names (as with `shapes`: to get choices other than  'atoms' and 'spatial'), I don’t think this is urgent. 

Did I miss anything? (Admittedly, wrote most of this while listening to talks about rare events at a math-focused conference in Eindhoven, so I may have overlooked something.) Anyway, I think all of these are features we'd like to have, but I hope that helps prioritize.

dwhswenson:
Detailed response on a few points:

>> #311?

> No, that is about having an attribute per instance not per class

Still, that seems like the right place to continue the discussion. Isn’t your proposal an evolution of (or mutually exclusive of) that one?

> Right, that would fix it, but if we are introducing another attribute we should do it right and add all information needed to store information like the variable type and potential chunks, etc.

In the current implementation, it would be a one-per-storage dictionary of `{string : int}`. Just JSON it. In a more complicated implementation, it might be a dict of dict, looking like `{snapshot_class_identifier : {string : int}}`. Still, one per storage, and should be easy to store once you have a way to identify which snapshot is which.

(Maybe the more complicated version should be `{(engine, snapshot_class) : {string : int}}`? The only requirement then is that, per file, you only have shape per combination of engine and snapshot class.)

> I am not really sure what you mean here. You say we should use the topology that contains the information and translate it and store it differently instead of using the topology?

I wrote "engine" when I meant "storage." Right now, you use a template snapshot to initialize storage. From that template, you look at the topology and from the topology you extract `n_spatial` and `n_atoms`, which, if I understand correctly, are kept somewhere in `storage` such that `__features__` can define shapes by giving the strings `spatial` and `atoms`.

What I want is the ability to add more things to that, or to set it directly. There are dimensions of the system other than `atoms` and `spatial`. In `dynamiq_engine` I already have electronic degrees of freedom as well as nuclear. Storing the state of extended variables from an NHC integrator would be another. Wavefunction information from an ab initio MD could be another example.

What I was saying is that this `shapes` dictionary should be the more fundamental thing. You can populate it using a topology, but a topology isn’t the same as the dictionary of shapes. Some relevant shapes are things that wouldn’t be considered part of a typical MD topology.

I do think the template is convenient, but it shouldn't be the only way.

 > Currently I am thinking about using a `Structure` class which is basically containing the types of all attributes for classes that are stored in this way.... Downside is the overhead of each Snapshot pointing to a structure object.

Why does the `Snapshot` need to point to it? Can’t the relevant features usually extract dimensions from the snapshot when needed? Put `n_atoms` and `n_spatial` as properties in the `coordinates.py` feature. There’s already an implicit requirement that `coordinates.shape == velocities.shape` (although we should add a debug-level check somewhere to ensure that).

What I mean is that, in analysis code, you never need direct access to the `Structure`. I think you only need that for setting up the storage, right? Also, in general, it isn’t clear to me whether `Structure` should be a class as such or just a `dict`, possibly with `namedtuple` values.

But snapshot features need a way to mark what their expected sizes will be, in order to be saved. That is, as far as I can tell, the only place where we need to access to this `shapes` dictionary: it is so that snapshot features can communicate meta-information to storage, although the information behind the meta-information gets set separately (by a template snapshot or by the engine or something else). Is that incorrect?

jhprinz:
Alright. Moving foward. I will leave the casting, storing multi snapshot, shapes, etc for a later PR. This resolves all the urgent problems for now and adds more tests.

jhprinz:
Ready for review. There is a more or less complete list of changes in the PR in the first post.

dwhswenson:
Looks good! The summary of changes on the first post is really useful; thanks for that. Of future things, I think `shapes` (or something to meet that need) is the most important. But this one is definitely ready to merge!

PR #448, opened 2016-03-17
Calculate flux through arbitrary interface with SingleTrajectoryAnalysis

dwhswenson:
The version of the single trajectory analysis I previously included calculates the flux out of the state. However, we’re often interested in the flux out of the state and through a given interface. This PR adds the option of adding an interface to the flux calculation. By default, the interface is equal to the state, which gives the same result as before.

- [x] Implement flux through interface
- [x] Thorough tests to include flux through interface
- [x] Clean up docstrings

This does not change anything that is currently set up, and won’t even make it so that the default `.analyze` function includes the interface. If the user wants to calculate the flux this way, the user has to ask for it explicitly.

One nice trick is that, in implementing this, I recognized that the flux calculation was very similar to a state lifetime calculation. That means they can both use the same code: `get_lifetime_segments`. For now, that is in `SingleTrajectoryAnalysis`. However, it might make sense to make it part of `Trajectory`, along with `summarize_by_volumes`. (Otherwise, perhaps `summarize_by_volumes` should be moved out of `Trajectory`?)

### DIFF
--- a/openpathsampling/analysis/single_trajectory_analysis.py
+++ b/openpathsampling/analysis/single_trajectory_analysis.py
@@ -156,7 +156,7 @@ class SingleTrajectoryAnalysis(object):
             seg[1:-1] for seg in transition_ensemble.split(trajectory)
         ]
 
-    def analyze_flux(self, trajectory, state):
+    def analyze_flux(self, trajectory, state, interface=None):
         """Analysis to obtain flux segments for given state.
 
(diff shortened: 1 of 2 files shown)


dwhswenson:
Ready for review and merge. Already passed tests.

jhprinz:
Looks good. Merging...

PR #449, opened 2016-03-18
Improve text output of move_summary

dwhswenson:
Just a quick PR to improve the output of the `move_summary` analysis. We really don’t need all the digits of each ratio. This also includes the predicted frequency of choosing each move (which means I can remove the extra cells in the analysis notebook which used to list that.).

Compare the output in `mstis_analysis.ipynb` (subsection on “Move scheme analysis”) to see the difference.


dwhswenson:

Really, just a very small change -- it's helpful for the debugging of SRTIS I'm doing, but it's also a general improvment, so let's merge it in on its own.

PR #450, opened 2016-03-21
Snapshot modifiers

dwhswenson:
New class whose subclasses return a modified version of the input snapshot. For now, only two types of modification:

* `NoModification`: just returns a copy of the snapshot
* `RandomVelocities:`: returns (Boltzmann) randomized velocities for the snapshot

These have a few uses:

1. Shooting Moves: in the general case, the shooting move includes a step where you modify the shooting point -- we skip this in 1-way shooting (which requires stochastic dynamics), but this is necessary for 2-way shooting in deterministic dynamics. The current PR won’t have implementations of the commonly used approaches in 2-way shooting, although the `RandomVelocities` could be used in an "aimless shooting" algorithm.
2. Committor Analysis: I’m about to put together a PR to run a quick committor analysis. That will require the `RandomVelocities` modification.

- [x] `SnapshotModifier` abstract class
- [x] `NoModification` and tests
- [x] `RandomVelocities` and tests
- [x] ipynb for visual inspection of `RandomVelocities` results
- [x] docstrings for all
- [x] Modifications after #445.

Note that this will be waiting for the new `copy` method in #445.


jhprinz:
This is really useful.

I few questions and ideas that I think make sense but are not required.

#### Treat Modifier class as an engine

Basically that is what it is. It generates a snapshot (from given snapshots). So why not treat it as a universal engine that can deal with most snapshots (assuming they have velocities). This would have the advantage that we can figure out, how a new snapshot was generated. This of course does only make sense outside of other engines. If that is to be used by engines themselfes it is not clear who made the snapshot. 

I think we discussed that and since each snapshot should reference an engine (which then holds that topology) that could solve the issue of having to find the engine that generated the snapshot.

This will also answer the question

#### Why use a class and not a function.

For an engine that makes sense. For sub masking you would also have to specify the reduced topology which could be useful for consistency.

dwhswenson:
> ####  Treat Modifier class as an engine

If we want it to be easy for people to add new engines, we need a clearly defined API for the engines. That should be a minimal set of functions, and it should be REQUIRED. We set these in https://github.com/choderalab/openpathsampling/issues/65#issuecomment-63080599. So, any engine must implement those functions. I don't think it makes sense to have snapshot modifiers implement those.

Furthermore, SnapshotModifiers are part of doing Monte Carlo moves. When we get around to the Monte Carlo overhaul, these will have to provide information for detailed balance. You could, for example, choose to do 2-way shooting with a SnapshotModifier that is biased to shooting along some reaction coordinate. Trying to treat it as an engine would just get really cumbersome.

I'm mixed on the idea of having the snapshot point back to this as its "creator." On the one hand, it feels like it would make sense. On the other, why is this snapshot the only one in the trajectory that doesn't point back to the engine that was used to run it? What does that mean for analysis?

> #### Why use a class and not a function.

Potentially, these can carry a lot of information. That information would have to go somewhere. Furthermore, exactly *what* information is needed depends on the modifier, so any code that calls one of these would need to have be ready for all possible call signatures for that function (or include a dictionary of parameters, in which case, you've basically hacked it into a class.) Encapsulating in the class makes sense.

The examples implemented in this PR are super-simple, and aren't representative of the kinds of things that might be done later. All modifiers have this `subset_mask`. `RandomVelocities` also has temperature (actually, `beta`). But another might select different atoms with different probabilities, and then rotate them by some randomly oriented Gaussian with a width parameter depending on which atom was chosen. That gets to be *way* to complicated to do anything other than encapsulate it in a class.

There's also nothing that says these have to be velocity-only. For example, you could certainly move a shooting point in configuration space and accept according Boltzmann weights (then doing 2-way shooting). I think that was done in one of the early TPS papers. To my knowledge, there hasn't been a whole lot of work on approaches to modify the shooting point (probably because it just doesn't matter in the kinds of systems we're typically looking at: hence, the success of one-way shooting). In ideas I have about mixing path sampling with semiclassical approaches, however, that could be really important (highly deterministic dynamics, so 2-way shooting is necessary, and time scales short enough to keep ballistic memory.).

jhprinz:
Makes sense what you say. My feeling is just that Modifiers and Engines have much in common and are (or should be) the source of Snapshots. In the context of efficient storage, casting of Snapshots to different types, referenes to topology etc. I think it is useful to know what instance generated a Snapshot since it allows nice reconstruction and interesting aspects for analysis / visualization. 

You are right in that an engine is more that a function that converts a snapshot. Still there is a strong similarity in the concept to generate snapshots. I could imagine (just to show the similarity) that a VelocityRandomizingEngine is very similar to a velocityRandomizingModifier. The engine or better the integrator is effectively a very complicated Modifier.

Let me propose another idea: What, if we say Modifier is the basis and an engine is just a subclass of Modifier. In theory I could transform every modifier into an engine by nested (repeated) application right?
If so, we would let a snapshot reference a Modifier (which is more general than an engine). This would also mean that the modifier keeps information about the topology or at least some information about the structure. 

Reversing the question: If we do not reference a modifier / engine per snapshot then where is the topology referenced for snapshots generated by a modifier? Copying the reference from the input snapshot is an option but technically wrong. Especially if you do crazy stuff.

So this is one proposal. The other one is the make Modifiers functions of engines?!? This is not so well thought through. My main point for this is your example of velocity randomization you require knowledge about the temperature/beta. This would be already present in the engine (in most cases) and so randomization seems to naturally be a function of an existing engine. You could use the referenced engine of the input snapshot and use its information (which is not in the API). I think you understand my point that some things are not clear (to me yet). 

> Why the extra effort?

My main aim is using OPS for adaptive sampling as well. And my feeling is that it is very much suited for that and can be used with little effort for these types of simulation. One of the main differences is that I would use a PathMoveScheme that does not worry about detailed balance on a PathLevel. This is the first step we are aiming at here in Berlin. Run simulations, build an MSM, pick new starting points, run again, etc... This uses paths, but not the ShootingMove it only runs the engine with fixed length. Storing the information of used modifiers we could trace back what was done in the scheme, what is the correlation between snapshots, etc. 

> What does that mean for analysis?

Yes, that is indeed the question here. My feeling is that for all path related analysis we do not need the snapshot information. For MD we need a topology object for each snapshot to use mdtraj, etc. So the topology is necessary for each snapshot.

The snapshot information would be on a seperate level if you e.g. would like to get equilibrium properties or use OPS to do ReplicaExchange (in the usual sense). I admit that I have been thinking about attaching not only the engine used, but also the previous snapshot like we do with `Sample`.
This would implement some consistency between Snapshots and Samples. Sample have previous sample and mover. Snapshots have a modifier and a previous snapshot. Analysis could then draw conclusions from the way snapshots were generated. This paragraph is very experimental and potentially the reference of snapshots is too much and a killer of performance. I might add this as a feature for my purposes though.

> Furthermore, SnapshotModifiers are part of doing Monte Carlo moves. When we get around to the Monte Carlo overhaul, these will have to provide information for detailed balance.

Okay, that is something that you planned and I know nothing about so far. I think this makes total sense, but I also have some ideas and wishes for the MonteCarlo Moves / PathMover overhaul. It should be good to talk about this soon so that we do not double efforts for this like last time :) My feeling is that we both have always an excellent concept of our own and the joint one nails it (once we find one the we both agree upon). 

Alright. This was lots of stuff and say 80% finished ideas. So do not take it too much "this is correct and how I want it". But more "this is how it fits into my picture of OPS future..."

dwhswenson:
> Let me propose another idea: What, if we say Modifier is the basis and an engine is just a subclass of Modifier.

I still don't like this idea. You’re saying that `DynamicsEngine` inherits from `SnapshotModifier`, right? That means that engines cannot be separated from the modifiers. So if you want to use the `engines` package, you also need the `SnapshotModifier`s, which are fundamentally more related to the `PathMover`s.

Why not just use contract programming? Snapshot modifiers can set a `creator` attribute in the snapshots they make without having to inherit from a common ancestor with `DynamicsEngine`. That seems to solve the fundamental problem you’re concerned about without interlocking other parts of the code. 

> The other one is the make Modifiers functions of engines?!?

I have no idea what you mean by this. If you mean `engine.modifier()`, that really doesn’t work. It makes it impossible to use multiple modifiers (used for different purposes, e.g., using different modifiers for different shooting moves). It also makes it a lot harder to reuse (again, the features of how to change the snapshot should be stored by the modifier).

I don’t think any design pattern makes more sense than making this a class. `SnapshotModifier` encapsulates information and we want polymorphism when we use it (it shouldn’t matter to a shooting mover what kind of modification we’re making, as long as the modification behaves correctly.) These are the reasons for object-oriented programming.

> One of the main differences is that I would use a PathMoveScheme that does not worry about detailed balance on a PathLevel. 

I’m unsure about some of you nomenclature here. MoveScheme and MoveStrategy are designed very specifically for TIS and TPS. In addition, the whole point of those objects is to maintain detailed balance without the user needing to thing about it. So that’s not what you want to use. I would strongly recommend using a different `PathSimulator` -- something other than `PathSampling`. `PathSampling`, by definition, is statistical sampling of path ensembles. That doesn't sound like what you're doing.

Take a look at #454, where I've done this for a committor analysis (which is takes a similar, but massively simplified, version of what you're talking about doing).

(Note: the above was drafted before a private email we exchanged. Based on that email, I’m even more certain that what you want is to have different `PathSimulator`s.)

> I admit that I have been thinking about attaching not only the engine used, but also the previous snapshot like we do with Sample.

I don't understand this idea. It sounds to me like a lot of extraneous storage with no real gain.

> This would implement some consistency between Snapshots and Samples.

What's the benefit to this? It makes it harder to work with these things, and harder to write code that uses them, and requires more storage. While it is certainly possible, I don't understand why the trade-off is a good one.

> Snapshots have a modifier and a previous snapshot.

This is not necessarily true. You can create snapshots without a previous snapshot -- I do that every time I set up a test. This idea adds storage for no purpose. There is no benefit to requiring this for every snapshot.

`SnapshotModifier` should include provenance, sure. But `SnapshotModifier` is fundamentally part of a shooting move. It is the same process that creates 2-way shooting: You select a shooting point, and then you modify it. So it can return this information in the same way we do with shooting points. The provenance is available in the move details for the snapshots where the provenance isn't obvious.

(I plan to eventually make the change that, instead of just having a `shooting_snapshot`, the details from the engine mover include `shooting_point_original` and `shooting_point_modified` -- in 1-way shooting, these point to the same snapshot, but in 2-way, they include the modification. This gives you provenance, and makes it so that analysis of shooting points doesn't care whether you did 1-way or 2-way.)

> Okay, that is something that you planned and I know nothing about so far. I think this makes total sense, but I also have some ideas and wishes for the MonteCarlo Moves / PathMover overhaul

(Discussion in private email -- my ideas are still rough, and I want to have a few parts be a little clearer to me before making a formal suggestion. You should already have an email with big overview ideas.)

dwhswenson:
This is ready for review. I expect that this will have some significant modifications when we overhaul PathMovers, but this version should work for now.

On the question of what its `engine` attribute it: this uses `.copy()` to set everything that isn't specifically changed by the modifier (for `engine`, I assume that's shallow copying?)

PR #451, opened 2016-03-22
More cleanup of Single Trajectory Analysis

dwhswenson:
Just a few more things to make `SingleTrajectoryAnalysis` easier to use in practice.

- [x] Create results object which has a few useful extra routines. Currently, we have a bunch of functions that variously return lists of trajectory segments, the lengths of those segments, or the time duration of those segments. All of this gets grouped into one class, although the old versions still work (they just call the specific class)
- [x] Switch from directly storing instance variables to returning the results (except for the `analyze` function, which runs everything). This makes it easier to use this stuff in interactive analysis.

Question: Renaming? Neither of the names of classes in `single_trajectory_analysis.py` are very good (`SingleTrajectoryAnalysis` and `TrajectorySegmentContainer`), but I don’t have any better names either. The functionality is definitely useful, though.

### DIFF
--- a/openpathsampling/__init__.py
+++ b/openpathsampling/__init__.py
@@ -11,7 +11,8 @@ from analysis.tis_analysis import (
     TISTransition, Transition, TPSTransition
 )
 from analysis.single_trajectory_analysis import (
-    SingleTrajectoryAnalysis 
+    SingleTrajectoryAnalysis,
+    TrajectorySegmentContainer
 )
 
 from collectivevariable import CV_Function, CV_MDTraj_Function, CV_MSMB_Featurizer, \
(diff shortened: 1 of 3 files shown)


dwhswenson:
Ready for review. Passes tests, and the modified files have 100% coverage.

jhprinz:
Looks good.

Maybe we should not worry about the naming too much for now. We can later to one overhaul of renaming before 1.0 release...

dwhswenson:
> Maybe we should not worry about the naming too much for now. We can later to one overhaul of renaming before 1.0 release...

Fair enough -- but please do keep these in mind for possible renaming. I've tried to come up with better names, but don't have any real winners. (We especially need to rename `SingleTrajectoryAnalysis`, since it can analyze multiple trajectories -- the name is actually wrong!)

PR #453, opened 2016-03-24
Fix input ensemble for ConditionalSequentialEnsemble

### LINKED ISSUE #452: (Minor) Bug in `ConditionalSequential.input_ensembles`
The image will show it better than I can explain it:

sr_minus_move_vis_bug

(Context: G is an innermost ensemble, M is a minus ensemble.)

The underestimation of how wide the scheme would get is something to fix later: the problem I'm writing about is that there's a bug in the way input and output ensembles are handled (apparently only in the case of the `ConditionalSequentialMover`).

I wouldn't mind if these said that they could exit as M (because of an abort during conditional sequential) and then that the `EnsembleFilter` removes that (which is what it does). But the problem that I'm concerned about is that it claims that something could enter the `ConditionalSequential` as M. If this was a normal `Sequential`, that would be true. But for `ConditionalSequential`, it is not: the input for the `ConditionalSequential` should be exactly the same as the input for its first submover. The output should be the union of all possible outputs.

dwhswenson:
Should be only the inputs of the first submover.

Resolves #452. 

New results look something like this:

fixed_cond_seq_ens

Note that the `RandomChoice` and `ConditionalSequential`s allow output from B (in case of abort, that's where you end) but the EnsembleFilter doesn't (filtering that is its job.) However, they do not allow input in B, fixing the problem.

### DIFF
--- a/openpathsampling/pathmover.py
+++ b/openpathsampling/pathmover.py
@@ -1497,6 +1497,9 @@ class ConditionalSequentialMover(SequentialMover):
     ConditionalSequentialMover only works if there is a *single* active
     sample per replica.
     """
+    def _get_in_ensembles(self):
+        return [self.submovers[0].input_ensembles]
+
     def move(self, globalstate):
         logger.debug("Starting conditional sequential move")
 


dwhswenson:
Passes tests. Ready for review and merge. (Only 3 lines added, and one was blank!)

jhprinz:
I think I understand your point, but am not sure if your solution is correct. Let me clarify:

You say that in the scheme the first `RandomChoice` will effectively not depend on previous MinusEnsembles. And this is because of the `ConditionalSequential` being wrong.

I think the first point is correct. If I understand the moves correct you will use an existing path in A extend and potentially cut the be in A again. No need of an existing Minus path.

The second point I do not understand. I would say that this is because the backward extend will always overwrite an existing path in B before it is used. Hence the reason is not the ConditionalSequential but the Extend coming before something that relies on B. 

So I would say that to fix this we would have to track in `ConditionalSequential` if some ensemble will be override and ensemble. If so that after that input from the ensemble is ignored. 
You are right that in Sequential this cannot happen since we continue even if a move is rejected. 

I think the same should be true for `PartialAccepted`

Does that make sense?

dwhswenson:
> The second point I do not understand. I would say that this is because the backward extend will always overwrite an existing path in B before it is used. Hence the reason is not the ConditionalSequential but the Extend coming before something that relies on B.

I don't understand this paragraph. What do you mean by "overwrite an existing path in B"? This is single replica. These is nothing in B until the extend.

In detail, what this move does. I'll list the ensemble status of each stage, and I think it will explain what I'm doing here.

You only get here if you have a path in ensemble `A` (single replica).

Going down the graph: 
* SRMinus: wrapper, calls its submove. Status: `A`.
* EnsembleFilter does nothing (nothing to filter yet). Status: `A`
* RandomChoice picks either forward or backward (we'll assume backward; the rest is equivalent. Status: `A`.
* CondSeq starts. Status: `A`
* EnsembleHop from `A` to `segment` ensemble (internal to minus move, so not pictured). Status: `minus` or aborted while in `A`
* BackwardExtend from `segment` to `B` (minus). Status: `B` or aborted in `segment` (or never got here)
* `SubtrajSelect` selects subtrajectory to move into `segment`. Status: `segment` or aborted in `B` (or never got here)
* Pick an ensemble (only one here) to hop to from `segment`. Status: `A` or aborted in `segment`

So I think it should be clear that the inputs of everything in the picture above are correct.

Now we start going back up the graph. What path move changes does the CondSeq see? It could be:

1. In `A` (and either successful or aborted in the first step)
2. In `segment` (and aborted)
3. In `B` (and aborted).

The whole point of the EnsembleFilter was to remove the aborted ensembles (in fact, I think that a conditional sequential shouldn't exist without an ensemble filter; it might be unecessary, but it will never be incorrect).

So I think is it reasonable to show `B` as an output until the EnsembleFilter cleans it up. The other argument would be to say that inputs to CondSeq are the inputs of its first mover, and the outputs of CondSeq are the outputs of its last mover. I don't see why you would need any kind of tracking.

> I think the same should be true for PartialAccepted

Yes, I think this is true. But I think it should be the exactly what I have here for ConditionalSequential. With PartialAccepted, it is certain that *any* intermediate output ensemble could be the final output. With ConditionalSequential, the question is "should the output only be the output of the last mover" or "should the output be the output of any mover." For CondSeq, either version could be used. If you'd prefer only the outputs of the final mover, that's another 3 lines of code -- so no problem ;-)

jhprinz:
> What do you mean by "overwrite an existing path in B"? This is single replica. These is nothing in B until the extend.

I see. What I mean is that the random choice mover does not care if there is a path in B. To clarify what I mean is simply assume that you would remove the extend from the scheme. Then It would require something to be in `B`, right? (Of course it does not make sense since we do not have a path in `segment`. But the movers don't know or care what you use as initial sampleset. 

So, as you point out. It does not matter if we would have a sample in `B` or `segment` in the initial sample set used. These will be _overridden_ in the move (that is what I mean with overriding). 

> You only get here if you have a path in ensemble A (single replica).

True, but ConditionalSequential cannot know about this fact. 

> The whole point of the EnsembleFilter was to remove the aborted ensembles (in fact, I think that a conditional sequential shouldn't exist without an ensemble filter; it might be unecessary, but it will never be incorrect).

Maybe that is the point of confusion. You might be right in that it does not make sense to have ConditionalSequential without a Filter, but currently (or the way it was) did not assume that and so I believe that the overriding strategy is the most general one.

dwhswenson:
I think I was just about to post something along similar lines. I still don't understand your use of the word "override" (or "overwrite") here: that doesn't seem to me to be the correct word. But what I realized this morning (and what I think you're writing above) is:

Assume that the same sampleset started with sampling in both A & B. By the time we get to the subtraj select, we could have two replicas in B. Which one to use would be randomly selected before doing subtraj select, and could be the one that was in the original the sample set.

A better example is the sequence shootA-repexAB as a conditional sequential. Obviously, this *does* require B to be in the input ensemble.

In that case, the original visualization was correct. However, that means that the visualization is confusing when looking at the single replica minus move, because it makes it *harder* to understand, instead of easier. That's of opposite of what a visualization should do. Any thoughts on what to do about that?

jhprinz:
> I still don't understand your use of the word "override" (or "overwrite") here

I (maybe wrongly) assumed that you have only one sample per ensemble, which I think we are doing so far. So that would mean that the extend will replace (= overwrite) the path in the `B` ensemble. If this is the case (which it might now always) then you can actually just check, if you will replace the potientially existing in `B` and henceforth ignore input from `B`. 

> By the time we get to the subtraj select, we could have two replicas in B.

Yes, but only if we do not replace the existing one.

jhprinz:
Btw. I know we discussed that a lot, but is there a way to get rid of potential multiple samples per ensemble? It makes the whole predict what movers do more complicated (as we see here).

dwhswenson:
>  I know we discussed that a lot, but is there a way to get rid of potential multiple samples per ensemble?

I don't think so. It seems necessary for AMS. Or for any replica-reservoir based approach.

jhprinz:
Then let's assume the override / replace is only replica dependent, which are unique. Would my proposed idea work then?

dwhswenson:
> Then let's assume the override / replace is only replica dependent, which are unique. Would my proposed idea work then?

I don't understand. The movers know nothing about the replicas. How does this help? The replicas don't exist from the perspective of dealing with only the movers. 

We currently *do* overwrite replicas. That's the purpose of the `apply_samples` method for a sample set; to ensure that there is only one copy of any replica in any sample set (once it becomes an active sample set, at least).

jhprinz:
Alright. Then there is no solution and we should remove the visualization. If there can be more than one replica per ensemble the current picture does not make sense.

dwhswenson:
> Alright. Then there is no solution and we should remove the visualization. If there can be more than one replica per ensemble the current picture does not make sense.

That seems a bit extreme... I think the current approach still makes sense if you stop at the canonical mover level by default.

Maybe change it to have an option `stop_at_canonical` which is `True` by default, and then users can set it `False` if they want to see the inner workings. Of course, the inner workings can be very confusing, so that's expected to by done only be more advanced users.

Under most circumstances, this visualization is actually really useful -- but this is an edge case where the details make it more confusing.

Alternately, show the inner workings, but only show the ensembles switched between until you hit the canonical mover. Actually, I think I like that the best, and it solves the problem here. You show all th structure of what is happening inside the mover, but (by default) you don't show all the ensembles involved except at more abstract levels.

jhprinz:
Maybe your right. I can add this features in the visualization overhaul that is coming...

jhprinz:
Another way I just thought about to fix this is to pass an initial sampleset or sampleset like to the visualization. This way it can figure out the exact structure after each sub-move and display it correctly.

The cut a canonical should be an option anyway.

jhprinz:
Well, the sampleset should be optional of course and not be necessary.

dwhswenson:
Starting (optionally) with an example sampleset is also reasonable (or even a list of occupied ensembles). Mainly, my opinion is that the default shouldn't be confusing, but the visualization can be helpful for more advanced users as well, so I'd hate to lose the capabilities we have.

Feel free to summarize the to-do items from this in an issue, and then we can close this without merging. I think the current behavior is actually the technically correct one, although it will be unnecessarily confusing for novices (since it was confusing for me!).

So adjusting the default behavior in a different PR, and not taking these changes, seems the correct approach.

PR #454, opened 2016-03-25
Committor calculation

dwhswenson:
Committor calculation. A subclass of `PathSimulator`. 

Setup looks like this:

```python
simulation = paths.CommittorSimulation(storage=storage,
                                       engine=engine,
                                       states=list_of_states,
                                       randomizer=randomizer,
                                       initial_snapshots=list_of_snapshots)
```
 where the `list_of_states` is a list of volumes, `randomizer` is `SnapshotModfier`, and the `list_of_snapshots` can be a list or a single snapshot.

Run it as:

```python
simulation.run(n_per_snapshot)
```

This starts by calling the `randomizer` on the `initial_snapshot`, and then runs the resulting snapshot forward and backward until a state from the `list_of_states` is reached (or until it hits the engine's maximum length). It will do this `n_per_snapshot` times for each snapshot in the `list_of_snapshots`. 

There will also be a function `analysis.committor_analysis(storage, states)` to analyze all the shooting moves based on each shooting point snapshot, looking at the resulting states of their shots. This is potentially useful in many situations other than just the standard committor analysis itself (debugging path sampling behavior, or using something like aimless shooting to get committors).

I’m still working on the details of the analysis function, but I suspect that it will return a dictionary of dictionaries: `{shooting_point_snapshot : { final_state_volume : count_integer }}`, where `None` is always one of the possible `final_state_volume`s (indicating the trajectory aborted before hitting a state, probably due to hitting the maximum length). I may also add something in there to account for the direction of the shot.

- [x] Draft committor calculations
- [x] Test committor calculation
- [x] Draft committor analysis code
- [x] Test committor analysis
- [x] Merge #450 (this requires the snapshot modifiers in that PR)
- [x] Examples for visualizing committors (as a shooting point analysis)
- [x] Docstrings

Suggestion: once merged, this will be a great place to work on implementing parallelization. The approach is pretty simple: send the next trajectory to the next available worker. This would allow us to start working with parallelization in a useful way, without having to worry about the unsolved problem of efficiently parallelizing RETIS. (Of course parallelization is not a 1.0 feature, but this is something to keep in mind.)


dwhswenson:
> This starts by calling the `randomizer` on the `initial_snapshot`, and then runs the resulting snapshot forward and backward until a state from the `list_of_states` is reached

Slight mistake in what I wrote: that should be "forward OR backward." It only does one of them each time.

There's an additional initialization parameter `direction` as well. If `direction` is `None` (default), the mover is a `RandomChoice` between forward or backward. If `direction` is positive, the mover is only forward. If negative, the mover is only backward. This gives the ability (combined with an appropriate `randomizer`) to learn a little more about the behavior at that shooting point.

Also, there's a parameter to `run` called `as_chain`. If `True`, the next snapshot is based on a modification of the most recent shooting snapshot. If `False` (default), it is based on the original input snapshot. For the existing modifiers (`NoModification` and `RandomVelocities`), this makes no difference. However, imagine a modifier that only rotates some velocity by some small angle. If not run `as_chain`, this can never explore the entire space. However, it might still tell us useful information about behavior at that shooting point (or particularly how the specific input velocities affect the shooting).

In both cases, these allow us to look at things that aren't actually the committor per se, but the same basic analysis can help us understand details of what is happening in our simulation.

dwhswenson:
Passes tests, ready for review!

To get an overview of what this does and how to use it, look over the file `committors.ipynb`.

jhprinz:
Looks very useful. Will merge. 

One question: The fallback for the computation of the hash uses a "numpy" string representation of the coordinates. This is not meant as a fallback, is it? If you are not running a discrete system it is VERY unlikely that you have two identical snapshots (as you know of course). So I assume that this is meant to compare for _only_ the same coordinates otherwise the memory address (default hash) should suffice.

Is that true? 

Since we might want to use hashing for snapshots, coordinates and velocities (all the rest should be fine) we could add a hashing function to the feature implementation to be as effective as possible. Means for toy snapshot the coordinate hash function uses the _string representation_ of `.coordinate` for MDSnapshots use the statics default hash etc...

dwhswenson:
> One question: The fallback for the computation of the hash uses a "numpy" string representation of the coordinates. This is not meant as a fallback, is it? If you are not running a discrete system it is VERY unlikely that you have two identical snapshots (as you know of course). So I assume that this is meant to compare for only the same coordinates otherwise the memory address (default hash) should suffice.

Yep, I think your description captures it pretty well. From my view, the decision to use `tostring` was based on the following use case:

You're analyzing results from a `CommittorSimulation` where you have given an initial snapshot, and you want to see the committor at that point in configuration space. In this case, the hash you want really is a hash of exactly the coordinates.

The memory address of the coordinates does not work as a hash because when snapshots are copied (as part of the `SnapshotModifier`) the numpy parts are deep-copied: this is the desired behavior, but it just means that we need a hash other than the memory address. So I went for the `tostring` (which is actually identical to the `tobytes`, if I recall correctly). It is a correctly unique hash.

> Since we might want to use hashing for snapshots, coordinates and velocities (all the rest should be fine) we could add a hashing function to the feature implementation to be as effective as possible. Means for toy snapshot the coordinate hash function uses the string representation of .coordinate for MDSnapshots use the statics default hash etc...

This is not a significant factor in the speed in practice, so I don't think we need to worry about that (yet), although I agree it could be useful in the future.

Right now, the main computational cost is in doing the `unique_trajectory`, which requires loading everything from disk. I can speed that up pretty easily, but even that just isn't a high enough priority for me to worry about it right now.

Here's the `gprof2dot` output from when I profiled the analysis:
committor

PR #457, opened 2016-03-30
Details of MD5 failues in conda install

dwhswenson:
Right now, miniconda's MD5 doesn't match the value that Travis is seeing. 

I suspect this is a bug on the part of the folks at conda. It fails for me with both [OPS](https://travis-ci.org/choderalab/openpathsampling/builds/119608319) and nightly build on [dynamiq-md](https://travis-ci.org/dynamiq-md/dynamiq_engine/builds/119605522), both failing again after re-running, so I assume this is a bug and not a fluke.

This PR just adds a little more information of "expected" and "found" after an MD5 mismatch in the miniconda setup. Potentially useful debug info that probably should make its way elsewhere.

### DIFF
--- a/devtools/ci/install.sh
+++ b/devtools/ci/install.sh
@@ -13,6 +13,8 @@ MINICONDA_MD5=$(curl -s https://repo.continuum.io/miniconda/ | grep -A3 $MINICON
 wget https://repo.continuum.io/miniconda/$MINICONDA
 if [[ $MINICONDA_MD5 != $(md5sum $MINICONDA | cut -d ' ' -f 1) ]]; then
     echo "Miniconda MD5 mismatch"
+    echo "Expected: $MINICONDA_MD5"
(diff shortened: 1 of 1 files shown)


dwhswenson:
Interesting... current expected is blank. So that will be a problem....

dwhswenson:
Looks like the most recent version of Miniconda changed the name to `Miniconda2-latest-Linux-x86_64.sh`. However, `Miniconda-latest-Linux-x86_64.sh` still pulled down an old version (`Miniconda2-3.19.0-Linux-x86_64.sh`) -- we matched the MD5 for that one.

dwhswenson:
The fact that this even gets into the proper tests means the bug is fixed. I'll re-run it'll it officially beats any stochastic errors, but this should be merged ASAP.

jhprinz:
Agreed!

Issue #456, opened 2016-03-29
Shooting snapshot not in initial trajectory?


dwhswenson:
Quick background: the analysis in #454 is intended to work with any set of MC steps that include shooting movers. The committor calculation is just one example of that; but a similar shooting point analysis can be useful in lots of cases, so I'm trying to make it work in general.

I'm getting some weird results on occasion from normal TIS calculations. It popped up while I was developing the committor analysis, but here I've given the simplest explanation I can to explain what is going wrong.

First, I run the local tests to ensure I have an updated `mstis.nc` file. Then I do the following:

```python
import openpathsampling as paths
storage = paths.AnalysisStorage("mstis.nc")

for step in storage.steps:
    try:
        # if not a shooting/extension move, this will raise an exception
        shooting_snap = step.change.canonical.trials[0].details.shooting_snapshot
    except AttributeError:
        pass # ignore the moves that aren't shooting
    except IndexError:
        pass # ignore the moves that aren't shooting
    else:
        trial_traj = step.change.canonical.trials[0].trajectory
        init_traj = step.change.canonical.trials[0].details.initial_trajectory
        if not trial_traj.is_correlated(init_traj):
            print step.change
```

For any (one-way) shooting move, the trial trajectory should be correlated with the initial trajectory. However, in a recent test, this gives the following output:

```
PathSimulatorStep : PathSampling : Step # 66 with 0 samples
 +- RandomChoice :
 |   +- RandomChoice :
 |   |   +- OneWayShooting :
 |   |   |   +- SampleMove : BackwardShootMover : False :[]
PathSimulatorStep : PathSampling : Step # 99 with 1 samples
 +- RandomChoice :
 |   +- RandomChoice :
 |   |   +- OneWayShooting :
 |   |   |   +- SampleMove : ForwardShootMover : True : 1 samples [<Sample @ 0x10f8a18d0>]
```

Okay... that's weird. So let's look at those steps:

```python
step66 = storage.steps[66]
step99 = storage.steps[99]

print step66.change.canonical.trials[0].ensemble == step99.change.canonical.trials[0].ensemble
print step99.change.canonical.trials[0].ensemble.name
```

```
True
I'face 2
```

Same ensemble, but not a special one (i.e., not MS-outer, so that's not the cause.) Let's look some more at each of those:

```python
stateA = storage.volumes["A"]
stateB = storage.volumes["B"]
stateC = storage.volumes["C"]
[...]

PR #458, opened 2016-03-31
update reversal test for snapshots

jhprinz:
This should repair the bug in #456 and make sure that reversed snapshots will not be stored twice in specific cases.

This add

1. a check that will abort saving when the reversed copy is saved
2. improve saving of lazy loaded objects across multiple files

It will not solve all potential problems. Some of them will be address once objects are references using UUIDs.


jhprinz:
That should do it. Let's see if tests pass...

dwhswenson:
Could you add an explicit nosetest for the process that you described as causing the error? That is, something the does exactly the steps you outlined in https://github.com/choderalab/openpathsampling/issues/456#issuecomment-203459562 and ensures that we get the right behavior?

That way we have a check that we don't accidentally regress into this bug again.

jhprinz:
Added testing. Should be fine after it passes tests.

dwhswenson:
Perfect! That for the fix -- makes the "committor" analysis much more broadly useful. Merging!

PR #459, opened 2016-04-04
Fix snapshot copy of `None`

dwhswenson:
Some features which are normally `numpy.array`s can sometimes be `None`, e.g., `box_vector`s. This puts a safe wrapper around the generated `copy` function that uses the `.copy()` if the value is not `None`, otherwise it just uses direct assignment (which, of course, assigns `None`).

- [x] Safe wrapper around `copy` code generation
- [x] Test to show it works in the case of `box_vectors=None`


dwhswenson:
Just a quick little change to fix a bug I found while working on other PRs. Passes tests, and ready for review!

PR #460, opened 2016-04-06
Clean up docs

dwhswenson:
There were a bunch of warnings in building docs. This cleans those up, where I could figure out what was needed.

One of the major problems was that we had listed things that aren’t actually accessible in the API (for example, abstract classes). These need to be listed in a way that they can be found. For example, we don’t import `ToyIntegrator` into `toys`. So it has to be listed for sphinx's `autosummary` as `integrators.ToyIntegrator`. When necessary, I tried to change the documentation to point to an actual class, instead of changing the API to add the class.

I couldn’t figure out exactly what’s going on with a the remaining complaints. I think it might be related to the use of double-backticks in some of the netCDF+ comments; leaving those for @jhprinz.


dwhswenson:
Ready for review.

jhprinz:
> I couldn’t figure out exactly what’s going on with a the remaining complaints. I think it might be related to the use of double-backticks in some of the netCDF+ comments

I was somehow struggling with LaTeX in .rst files and tried lots of backtick combinations :)

jhprinz:
Thanks!

PR #461, opened 2016-04-11
RuntimeWarning that should be AttributeError in Trajectory

dwhswenson:
We currently have the following:

```python
trajectory.not_an_attribute
```

```pytb
RuntimeWarning                            Traceback (most recent call last)
<ipython-input-9-4a503b36e4ce> in <module>()
----> 1 traj.not_an_attribute

/Users/dwhs/Dropbox/msm-tis/openpathsampling/engines/trajectory.py in __getattr__(self, item)
    195 
    196             else:
--> 197                 raise RuntimeWarning('Cannot access attribute "%s" from this type of snapshots')
    198 
    199         else:

RuntimeWarning: Cannot access attribute "%s" from this type of snapshots
```

Problems with this:

1. Aside from the other issues (parts 2 and 3) this forgot to fill in the `%s`
2. This is confusing to the user, because it is trying to auto-apply `Snapshot` features to a `Trajectory`. But if I was looking for an attribute of a `Trajectory` and got this error, I wouldn’t know why it was talking about `Snapshot`s. (The first time I got this was when I mistyped `trajectory.summarize_by_volumes_str`: if I wasn’t familiar with the internals here, this message would have confused me.)
3. This should be an `AttributeError`, not a `RuntimeWarning`.  If I do this with a normal list:
```python
test = [1,2,3,4]
test.not_an_attribute
```
I get
```pytb
AttributeError                            Traceback (most recent call last)
<ipython-input-2-12e045042461> in <module>()
----> 1 test.not_an_attribute

AttributeError: 'list' object has no attribute ‘not_an_attribute’
```

***

This PR fixes those three problems. New output:

```pytb
AttributeError                            Traceback (most recent call last)
<ipython-input-9-c9e248afd106> in <module>()
----> 1 trajectory.not_an_attribute

/Users/dwhs/Dropbox/msm-tis/openpathsampling/engines/trajectory.py in __getattr__(self, item)
    204                     snapshot_class.__name__
    205                 )
--> 206                 raise AttributeError(msg)
    207 
    208         else:

AttributeError: 'Trajectory' object has no attribute 'not_an_attribute'
                Cannot delegate to snapshots. 'ToySnapshot' has no attribute 'not_an_attribute'
```

#### Remaining bug

Note that there is still a bug here, but I’m not sure how to fix it and keep the feature this function supposed to add. The bug is the following:

```python
>>> import openpathsampling as paths
>>> empty_traj = paths.Trajectory([])
>>> # next should raise AttributeError
>>> empty_traj.not_an_attribute
[]
```

Because you default to thinking that any unknown attribute must be a request for a snapshot feature, you assume that an unknown attribute given to an empty trajectory must desire an empty list in response. But sometimes that should be an `AttributeError`.

### DIFF
--- a/openpathsampling/engines/trajectory.py
+++ b/openpathsampling/engines/trajectory.py
@@ -194,7 +194,16 @@ class Trajectory(list, StorableObject):
                     return out
 
             else:
-                raise RuntimeWarning('Cannot access attribute "%s" from this type of snapshots')
+                std_msg = "'{0}' object has no attribute '{1}'"
+                snap_msg = "Cannot delegate to snapshots. "
+                snap_msg += "'{2}' has no attribute '{1}'"
+                spacer = "\n                "
+                msg = (std_msg + spacer + snap_msg).format(
+                    str(self.__class__.__name__), 
+                    item,
+                    snapshot_class.__name__
+                )
+                raise AttributeError(msg)
 
         else:
             return []


dwhswenson:
Tests pass. Ready for review, and (unless there's an idea on a quick fix for what is labeled as "remaining bug"), ready to be merged.

jhprinz:
Good catch. I would merge. For the remaining. There is no fix as I see it. The reason is simple that with no snapshot in a trajectory there is no way to determine the present features in the snapshot and hence it might be a trajectory that is meant for snapshots with a `not_an_attribute` attribute. So the answer is both wrong and false. 

Means we just have to make a decision. Do we want to disallow relaying attributes to snapshots for trajectories of length zero and raising an error or (as it is now) just return the _correct_ solution for any possible attribute.

I would go for the current implementation which will possible cause less need for checks when using this relaying in other code and since the returned value is not wrong that should be fine.

So I will merge now and if we come up with a better idea we can do another PR.

dwhswenson:
> For the remaining. There is no fix as I see it.

Yeah. It is hypothetically possible to keep this behavior and catch the errors when they should be errors, but the only way I see to do that is to massively restructure `Trajectory` to know about `feature`s the same way `Snapshot`s do. That may be something to consider in the future, but my personal preference is *not* to do it right now (or before 1.0).

> I would go for the current implementation

I agree. I also think you're more likely to find yourself in that part of the code when the empty list is what you want, than when the `AttributeError` is what you want. Basically, the `AttributeError` should be raised when you mistype something and are working with writing tests (hence the zero-length trajectory) -- so it is something ***I***'m likely to do, but not something the average user will come across. On the other hand, you could have zero-length trajectories from another algorithm (some of the single trajectory analysis stuff returned zero-length trajectories, for example). In that case, you don't want to throw an error when the empty list is what you're expecting.

One possibility to consider: have a `logger.warn` when this situation arises? That will (1) give us a sense of whether this is actually happening a lot; (2) give the user who should have gotten an attribute error some message to help explain what happened. Maybe even drop it down to the level of `logger.info` if we find it too annoying as a warning.

PR #464, opened 2016-04-13
OneWayShootingMoveScheme

dwhswenson:
In a number of cases (TPS being a prime example) we frequently want to use only shooting moves. This PR create a `OneWayShootingMoveScheme` object which has the minimum necessary to do shooting.

Note that you can always add other move strategies to this, if you want. For example, if you wanted path reversal as well, you use the `.apply_strategy` function to add that. 

Basically, I’d think of it this way:

* `DefaultScheme` gives a full default for TIS calculations: RETIS, with minus, with path reversal, etc.
* `SRTISScheme` gives a full default for SRTIS: same as `Default`, but with single-replica moves (e.g., ensemble hops instead of replica exchange)
* `MoveScheme` is a completely blank slate that you need to add everything else to
* `OneWayShootingScheme` is added to provide the most common starting point (1-way shooting moves on each ensemble, with global organization according to the move type), that a user would want to build on.

This would also be the standard move scheme for TPS (we can always add a `ShiftingStrategy` if we need to later).

- [x] Draft `OneWayShootingMoveScheme`
- [x] Tests for `OneWayShootingMoveScheme`

The code is pretty simple; it basically just wraps some slightly more advanced functionality into an easy package for novices.

### DIFF
--- a/openpathsampling/__init__.py
+++ b/openpathsampling/__init__.py
@@ -1,4 +1,6 @@
-from analysis.move_scheme import MoveScheme, DefaultScheme, LockedMoveScheme
+from analysis.move_scheme import (
+    MoveScheme, DefaultScheme, LockedMoveScheme, OneWayShootingMoveScheme
+)
 from analysis.network import (
     MSTISNetwork, TransitionNetwork, MISTISNetwork, TPSNetwork
 )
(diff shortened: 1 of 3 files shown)


dwhswenson:
Passes tests. Ready for review.

jhprinz:
Lgtm.

PR #465, opened 2016-04-15
Move `engine` attribute to only `EngineMover`s and wrappers

dwhswenson:
@jhprinz and I discussed this in our Skype call from 4-5 April 2016: Basically, we said that instead of having the `engine` be an option for the `PathSampling` object, it makes more sense to tie the engine to the `PathMover`s that use it: namely, `EngineMover`s. Of course, objects that are factories for these movers (`DefaultScheme`, `ShootingStrategy`, `MinusStrategy`, etc.) will need to take the engine as an argument as well.

This prepares us for a future where multiple engines might be used as part of one simulation (e.g., Titus’s approach to use MM for the minus move and ab initio for the transition region).

- [x] Add `engine` parameter to `EngineMover`s
- [x] Add `engine` parameter to schemes and strategies
- [x] Remove `engine` parameter from `PathSampling` simulator
- [x] Thoroughly test remaining notebooks

*Why do this now?* Because it might make it possible to add multiple engine support without breaking the API, so that it can be in the 1.x series.

*But why do this* ***now****?* Because it is Friday afternoon, and I don’t feel like working on SRTIS or examples.


dwhswenson:
Updated most of the notebooks. The only ones I didn't update were things that I think we should probably delete:

* `toy_bootstrapping.ipynb`, `toy_tis.ipynb`, `toy_analysis.ipynb`: these are already listed as "attic" in the readme. Shouldn't we just delete them? There's no need for an attic with git.
* `alanine_dipeptide_old_example.ipynb`: again, this looks to be completely superseded by the newer and much nicer notebooks. Can we just delete this one?

dwhswenson:

If you look at `mstis.ipynb`, you'll see the general way that this changes setup. Most of the other changes in the ipynbs are just repeating this kind of modification.

Of course, the core changes to the code are in the `.py` files.

Big picture: `PathSampling` no longer requires an engine as input, because it is definitely (hypothetically) possible to use multiple engines within a path sampling simulation. However, most other simulation types (`Committor`, `FullBootstrap`, etc.) really do only work in the context of a single engine, so they still take an `engine` parameter. How that engine parameter is handled is now a little different (it gets assigned to the `EngineMover.engine` class variable), but otherwise behavior is the same.

jhprinz:
> Updated most of the notebooks. The only ones I didn't update were things that I think we should probably delete:

Yes, they can all be deleted. They are just cluttering up the examples directory. We can do that with the next PR.

Will merge this now.

PR #466, opened 2016-04-16
Improvements on Committor

dwhswenson:
A couple things:

1. Some tests assumed that dictionary orders were preserved in a way that they are on my computer, but aren’t guaranteed to be. This PR aims to remove those assumptions (and therefore avoid the associated random failures).
2. As mentioned in https://github.com/choderalab/openpathsampling/pull/454#issuecomment-210786426, I saw a few ways to make significant speed improvements in the shooting point analysis code. I'll try those out now. (This also gives a few extra runs on Travis to watch out for any randomly-occurring errors I might have missed.)

- [x] Fix known randomly-occurring errors in committor nosetests
- [x] Add support for faster committor calculation
- [x] Performance profiling of faster committor calculation


dwhswenson:
Slightly different input (smaller), so it isn't quite an apples-to-apples comparison with https://github.com/choderalab/openpathsampling/pull/454#issuecomment-210786426, but the following shows that (slightly) more time is spent opening the file for analysis than in actually doing the analysis. In my opinion, that's what we'd like to see: in general, the cost of opening for analysis is amortized over several analyses that are done at once, so it is okay for the analysis opening/caching to be a bit more expensive (although anything to speed it up is also welcome!)

committor

dwhswenson:
Okay, this fixes the errors that I had seen (or at least, I hope it does). If we get more errors in these in the future, I'll take a look again. It also makes the committor analysis a LOT faster.

Ready for review.

PR #467, opened 2016-04-17
Simplest TPS

dwhswenson:
This adds a little example notebook with the simplest possible TPS. Basically, @jhprinz and I keep saying that OPS makes setup of path sampling simulations really easy. But @bolhuis and @jchodera have seemed a bit skeptical, probably because most of our examples were more complicated (including either multiple states, or TIS, or extra TIS ensembles like the minus interface, or several of the above -- all of which make the process of getting a set of initial trajectories more difficult).

This PR adds a very simple (flexible path length) TPS simulation. It only involves 2 states. It shows the overall process for doing any path sampling simulations, which is:

1. Define the MD engine
2. Define the states (and interfaces, if TIS)
3. Create "network" of ensembles and "move scheme"
4. Get initial trajectories and assign them to all ensembles
5. Assemble this all into a simulation object and run the simulation.
6. Analyze.

In practice, the user has to do some work to create the MD engine, to identify appropriate states/interfaces, and to get initial trajectories. We count on other codes (e.g., OpenMM) to do the work of simplifying engine setup, and I think that getting initial trajectories in a reliable way is still a topic of research. But the rest, OPS makes as easy as possible (while allowing options). The goal is that `simple_tps.ipynb` should make that clear.

This includes #465, so it needs to wait on that before it can be merged.


jhprinz:
Very nice example. It is also very helpful to see, how we could simplify things (even more).

Would there be a way to simplify
```py
paths.SampleSet.map_trajectory_to_ensembles(trajectory, network.sampling_ensembles)
```

for TPS we know that this is the strategy to initialize from a trajectory. That would make this into a function of `PathSampling` like so

```py
simulation = paths.PathSampling(storage, scheme)
simuation.initialize_from_trajectory(trajectory)
```

dwhswenson:
> Would there be a way to simplify ...

This is a matter of trading off between making this one kind of calculation easier or making it more like other kinds of calculations. This same function is used to set up TIS. I think it is better from a pedagogical standpoint to use a function that can be reused.

The question then becomes, can we simplify it *and* have a function that can be reused? (And that's the rest of this post)

> That would make this into a function of `PathSampling` like so ...

This does not work. It should not be a method in `PathSampling`. It would not be correct for setting up TIS, and especially not for setting up SRTIS, and would be really messy for setting up multiple state or multiple interface set.

I was thinking about how to go about doing this. I think the best idea is to create two functions:

* `network.sample_set_from_trajectories(trajectory_list)`: tries to fill all its sampling ensembles with list of trajectories given, filling with the first from `trajectory_list` that satisfies the ensemble. If none satisfy the ensemble immediately, try reversing them as well. (This could later include extensions, but I think that is dangerous: it is too hard to guarantee success.)
* `scheme.check_initial_sample_set(sample_set)`: Checks what ensembles are required, but not yet filled. The tricky part here is very similar to the issues in #452 and #453: basically, it boils down to whether `mover.input_ensembles` means "`mover` requires **all** of these to be filled" or "`mover` requires **any** of these to be filled." Without a way to clarify that, this function can't be written in a way that works for both regular TIS and single replica TIS.

The first of these functions can already be written, and would replace the existing `map_trajectory_to_ensembles` function. I could write that one up this week. The second one is considerably harder, however it might be worth thinking about. I'd rather put that part on you: at least, the stuff with distinguishing between *any* and *all* `input_ensembles` being expected behavior. It also presents an interesting question for the visualization, and I think a solution to that is the best solution for #452 and #453.

Note that the second function could be called by the `PathSampling` object either during `__init__` (if we trust the initial conditions given to `__init__`) or at the start of `run`, so the check can always happen even if the user doesn't do it explicitly.

In any case, that goes to a different PR.

dwhswenson:
Passes tests. Ready for merge -- since, if you read that one little notebook, you've already pretty much reviewed it ;-)

jhprinz:
I think you are right. Maybe we could just name step 8 in the example notebook and explain the idea behind it like

Create the first sampleset that contains initial paths for all ensembles. We will use the initial trajectory and map it to all ensembles...

> and I think a solution to that is the best solution for #452 and #453.

I will look into that after I have the new visualization working properly.

Just saw that it passes. Will merge this.

jhprinz:
After the refresh it only had 2 files changed...

jchodera:
Isn't the simplest TPS fixed path length? Could we have a version of that
too?

jhprinz:
Sure, I think that requires to create a new type of Network. The TPSNetwork is variable length, but that should be only a few lines. Alternatively we can add an option to the TPSNetwork like
```py
TPSNetwork(stateA, stateB, length=100)
```

Not sure if we could take the length from the previous move. That would require a special FixedLengthPathMover, right? @dwhswenson

dwhswenson:
> Isn't the simplest TPS fixed path length?

Not in the sense of simplest to set up, which is what I'm doing here. Think in terms of the amount of information you need to set up the simulation. The correct value for fixed path length is extra information that depends on the specific system. Flexible length TPS doesn't require that.

> Sure, I think that requires to create a new type of Network.

Correct. I'll set up a `FixedLengthTPSNetwork` (and some associated `Transition` object) in an upcoming PR.

Note that there's a bit of a pain here. `TPSNetwork` is actually multiple state TPS: just that when you only have one initial and one final state, multiple state TPS reduces to the original TPS. Fixed path length multiple state TPS is algorithmically simple, but I would strongly recommend against it in practice. You can easily imagine a case with multiple states where the path length distributions are such that the longest paths from one transition are longer than the shortest escape paths from another transition.

> Not sure if we could take the length from the previous move. That would require a special FixedLengthPathMover, right?

No. `Network`s and `Transition`s are about ensembles, never about movers. We just need a special *ensemble* called a `LengthEnsemble` -- which, fortunately, was one of the first things you implemented. ;-)

PR #468, opened 2016-04-19
Fixed length TPS

dwhswenson:
Answering @jchodera’s request for fixed path length TPS. See https://github.com/choderalab/openpathsampling/pull/467#issuecomment-211494315 ff.

In the process, this should also do a little to clean up the approach currently used in TPS transitions and networks, to make it easier to re-use code.

- [x] Draft `FixedLengthTPSTransition`
- [x] Tests for `FixedLengthTPSTransition`
- [x] Draft `FixedLengthTPSNetwork`
- [x] Tests for `FixedLengthTPSNetwork`
- [x] Simple example/test putting it all together
- [x] Docstrings

### DIFF
--- a/examples/ipython/README.md
+++ b/examples/ipython/README.md
@@ -35,15 +35,22 @@ See below for a complete list of available notebooks.
 
 ## Examples
 
-#### Simplest examples
+#### Simple TPS examples
+
+Transition path sampling (TPS) is the simplest path sampling algorithm.
+These examples demostrate just how easy it is to set up a TPS simulation
+using OpenPathSampling.
 
 * [`simple_tps.ipynb`](http://github.com/choderalab/openpathsampling/blob/master/examples/ipython/simple_tps.ipynb)
-    Transition path sampling (TPS) is the simplest path sampling algorithm.
-    This example shows just how easy it is to set up a TPS simulation using
-    OPS. For a realistic system, you need to start with (a) a working MD
-    engine for your system; (b) state definitions for your system; and (c) a
-    trajectory (not necessarily physically real, but as close as possible)
-    connecting the initial and final states for your system.
+    This example does flexible-pathlength TPS. You need to start with (a) a
+    working MD engine for your system; (b) state definitions for your
+    system; and (c) a trajectory (not necessarily physically real, but as
+    close as possible) connecting the initial and final states for your
+    system.
+* [`simple_fixed_length_tps.ipynb`](http://github.com/choderalab/openpathsampling/blob/master/examples/ipython/simple_fixed_length_tps.ipynb)
+    This example is for fixed-pathlength TPS. In addition to the information
+    from the flexible-pathlength version, you also need to know an
+    appropriate pathlength to use: here we use 5000 frames.
 
 #### MSTIS example (3 parts)
 
--- a/examples/ipython/simple_fixed_length_tps.ipynb
+++ b/examples/ipython/simple_fixed_length_tps.ipynb
@@ -1,0 +1,183 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "import openpathsampling as paths"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "# always the most complicated bit\n",
+    "import openpathsampling.engines.toy as toys\n",
+    "import numpy as np\n",
+    "\n",
+    "pes = (toys.OuterWalls([1.0,1.0], [0.0,0.0]) + \n",
+    "       toys.Gaussian(-0.7, [12.0, 0.5], [-0.5, 0.0]) +\n",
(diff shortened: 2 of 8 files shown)


jchodera:
Woot! Thanks!

dwhswenson:
Ready for review.

One (relatively easy) day of work: added fixed length TPS, including thorough nosetests and a simple example. When it's that easy, I think we're doing this thing right!

A couple notes on implementation: I switched around the networks a bit to maximally reuse code. Most of what is included in the actual code of `TPSNetwork` and `FixedLengthTPSNetwork` is there to make documentation clearer (by making the reasonable call signatures more obvious). Also, there's a lot of start-to-documentation stuff (in terms of docstrings); I had already started a branch to add more docs about networks, I'll complete this stuff in that branch.

jchodera:
Looking forward to playing with this once merged.

jhprinz:
Looks fine, but I have a general question about FixedLengthTPS and more than 2 states, which apparently is possible in your PR. Do all the transitions have the same fixed length? (Think you mentioned this as a negative side of FL-TPS). If so (and I think it does not make sense the require all paths between all states to have the same length) to provide a possibility to chose lengths for each transition separately somehow? And is that okay or does it have problems with other moves present.

jchodera:
Provided your path ensembles are defined by "x_0 is in state i, x_T is in state j, and trajectory is exactly T steps long", this ABSOLUTELY makes sense.  This is exactly the same as a standard Markov state model, which is very useful to explore right now.

bolhuis:
I agree with John that Fixed length makes sense. However, did you guys realise that the shifting move also has to be implemented, to make the Fixed length move work?

jhprinz:
Interesting connection. I never thought about it this way. My problem with this approach is, that OPS will try to generate only paths between  i and j that will take exactly `n` steps. In Markov Theory we have slightly different ensembles if I am not mistaken:

```py
# TPS i -> j
Sequence([
    Length(1) & InX(stateA),
    Length(N - 2) & Out(stateA & stateB),
    Length(1) & InX(stateB)
])
```

```py
# TPS i -> j
Sequence([
    Length(1) & InX(stateA),
    Length(N - 2),
    Length(1) & InX(stateB)
])
```

So, in MSMs we do not care about what happens intermediate steps (I could be wrong about TPS though, so please correct me!)

Just rechecked the implemented code. I was wrong. FL-TPS is actually MSM with length = lagtime.
That is really useful.

I think right now there is no shifting move, but it should be easy to implement. I assume it uses an existing A->B path and cuts some at one end and extends at the other.

jchodera:
> Length(N - 2) & Out(stateA & stateB),

That Out definitely doesn't belong there.

jhprinz:
Yes, that was my initial misunderstanding and cause for confusion.

dwhswenson:
> That Out definitely doesn't belong there.

It also isn't in the code: https://github.com/choderalab/openpathsampling/pull/468/files#diff-c286e3a5e3c766764ada808595e39f07R128

(longer response following later)

dwhswenson:
(Ah, sorry: saw after rushing to computer after emails that JHP had caught this)

dwhswenson:
So: overall, let me agree with everyone (does that mean I'm also disagreeing with everyone?)

I think this is what we want, as @jchodera says. 

We need to add a shifting move, as @bolhuis says. I'll do that sometime in the next week or so -- personal things (friend's thesis defense where I'm paranimf) are slowing me down a lot this week, but, as @jhprinz says, it should be relatively easy. @bolhuis : tomorrow can we spend 5 minutes of path group meeting on how to select the length of the shift in shifting?

I also think that @jhprinz is right that it should be possible to assign different lengths to different transitions. I think we both take a lot of pride in designing OPS to do things that might hypothetically be useful, even if we don't have an example of it today -- and that is one such thing.

Unfortunately, that's a little hard to manage within one `network` object, while still keeping it as simple as possible for the common case of only one transition, or all transitions of the same length. But that's where the distinction between `network`s and `transition`s come in handy. You can create a custom `network` that includes `FixedLengthTPSTransition`s of different lengths. I'm hoping that the docs I'm assembling in #471 will help explain how to make a custom `network`. At some point, we should add classes/structure to simplify creating custom networks; however (for reasons of time), I'd prefer if that waits until after 1.0.

jchodera:
> I also think that @jhprinz is right that it should be possible to assign different lengths to different transitions. I think we both take a lot of pride in designing OPS to do things that might hypothetically be useful, even if we don't have an example of it today -- and that is one such thing.

Sure, it would be interesting to explore flexible-trajectory-length multistate TIS with absorbing core states, but the original idea was to explore "plain vanilla" TPS with fixed path lengths and regions denoting the two (or more) states.

jhprinz:
> Sure, it would be interesting to explore flexible-trajectory-length multistate TIS with absorbing core states, but the original idea was to explore "plain vanilla" TPS with fixed path lengths and regions denoting the two (or more) states.

In the context of mimicking MSMs we need the fixed length for all states. Of course this only makes sense if we also use the MSM state definitions.

jhprinz:
Merging...

jchodera:
> In the context of mimicking MSMs we need the fixed length for all states.
Of course this only makes sense if we also use the MSM state definitions.

Yes!  But more traditional two-state TPS doesn't require a partition of
unity, and that is still a useful case to have a tutorial / example of as
well.

PR #469, opened 2016-04-20
Better way to set up initial trajectories

dwhswenson:
As discussed starting with https://github.com/choderalab/openpathsampling/pull/467#issuecomment-211276954, there may be easier ways to set up initial conditions.

The best place to put this in the move scheme. That is the place where we have knowledge of exactly *which* ensembles are actually in use. (The `network` includes unused ensembles.)

The main function here is:

```python
MoveScheme.initial_conditions_from_trajectories(self, trajectories, sampleset=None)
```

If you give a SampleSet as the `sampleset` parameter, you will only fill ensembles that aren’t already in that sample set. The `trajectories` parameter can be a list of trajectories or just a single trajectory.

One you have a list of ensembles to fill, for each ensemble, the code tries to find the ***first*** trajectory from the list that satisfies the ensemble. If it does not find a satisfactory trajectory in the list, it goes through the list again, reversing each trajectory. 

In the future, we might add an `extension` option, which would then check if it can use `can_append` on each trajectory, and then try to extend the trajectory forward, followed by a similar backward attempt. This isn’t implemented yet, but could be without too much difficulty.

No errors are thrown if not all ensembles are filled. It is expected that this might just be used as a first setup. So there's another function, `MoveScheme.check_initial_conditions(sample_set)`, which returns `True` if and only if the given sample set is an appropriate set of initial conditions for the move scheme.

To make all of this work, there's a function in the `MoveScheme` called `list_initial_ensembles()`. This function returns a list of lists of ensembles. This result should be read as "one member of each sublist must be in the initial sample set". This makes it easy to either do the two common approaches (all-replica TIS: each ensemble is its own list; single-replica: one sublist with all ensembles) or many things in between. (Ensembles not in sublist are turned into length-1 sublists.) This gives us a way that SRTIS can override the default behavior and recognize that a single sample for *any* of its root-level ensembles is acceptable. That will be implemented as part of the SRTIS branch.

- [x] Implement `initial_conditions_from_trajectories()`
- [x] Tests for `initial_conditions_from_trajectories()`
- [x] Implement `check_initial_conditions()`
- [x] Tests for `check_initial_conditions()`
- [x] Docstrings


dwhswenson:
Ready for review. Includes usage in the `simple_tps.ipynb`. Not sure that it is always the best way to set it up: if you already have ensemble information, that can be useful (using the `translate_ensembles` trick in one of the toy system notebooks).

But this should be a bit easier than before.

jhprinz:
I think it is really useful for a user that start with OPS to concentrate on understanding the other basics first and then have to deal with a smarter way of setting up the initial SampleSet. 

It is also a nice shortcut for tyring things out.

PR #470, opened 2016-04-21
Better test for shooting

### LINKED ISSUE #416: Sudden flip of the sign in velocities

dwhswenson:
Building off of ideas from the tests I wrote in #454, I’ve written tests for our shooting movers that should catch it if we ever regress on #416 (which relates to these “kinked” trajectories we were seeing when the reversal of the backwards shooting snapshot was done incorrectly.)

@jhprinz fixed the problem a while ago (#429 or, at worst, #445), but I didn’t want to close the issue until we had a more thorough test. This is that test.

Closes #416.

This also includes some fixes to actually allow multiple engines, following on #465. I removed the way engines made themselves default automatically on creation. The place to set the default engine is inside a `PathSimulator`, or directly by using `EngineMover.default_engine`. The implicit defaults were causing problems in tests that involved more than one engine.

### DIFF
--- a/openpathsampling/engines/dynamics_engine.py
+++ b/openpathsampling/engines/dynamics_engine.py
@@ -82,7 +82,8 @@ class DynamicsEngine(StorableNamedObject):
         self._check_options(options)
 
         # as default set a newly generated engine as the default engine
-        self.set_as_default()
+        # self.set_as_default()
+        # REMOVED because this breaks the ability to have multiple engines
 
     def _check_options(self, options = None):
         """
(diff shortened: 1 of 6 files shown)


dwhswenson:
Ready for review. Main point is that it checks that shooting works with a real engine (instead of just `CalvinistDynamics`). I think (hope) that should give us some extra safety against regressing on #416.

jhprinz:
Nice work.

PR #471, opened 2016-04-22
Docs on Networks

dwhswenson:
Mostly, better docs about Networks, and stuff to include them in the docs. But there is one new trick worth mentioning that I added:

Our `.. markdown::` directive only *imports* markdown text. But most of the markdown stuff I’ve written for the docs is intended to be single-page-per-file. Dumping them all onto one page creates a very long page, and implies connections between sections that don’t exist. See existing version of guides/ to see a  case where dumping all of this together just makes it more confusing.

So I tossed together a little `premake.py` script that we need to run before we `make` the docs. Currently, it just creates an rst wrapper file for the specified md files. Note that the wrappers are generated during the 

Another option would be to actually use pandoc to convert the md files to rst files. This would just involve making a fancier version of this `premake.py` script. In the long run, that might be a good idea (with some cleverness, we might be able to post-process that and get links to class names working in the md-based docs).

Will require #468 to be merged first.


jhprinz:
Nice. If I understand then this will generate a file with the `.. markdown::` directives. We could just create a new directive that will do that. That should be cleaner from the maintenance perspective but do the exact same thing. I think we might do that later when we have time to clean up. I would have to think about how to do that properly.

dwhswenson:
>  If I understand then this will generate a file with the `.. markdown::` directives. 

Yep, that's correct. (Plus, of course, this PR contains miscellaneous updates to the actual documentation material.)

> We could just create a new directive that will do that.

That might be a better way to do it; in any case, the goals I had were the following:

* use Markdown basically as if it was RST  for the purposes of writing docs.
* don't require an extra file per Markdown doc in the repo (i.e., don't store `my_doc.rst`, which is just a wrapper for `md_doc.md`: instead, generate it when needed)

As long as we accomplish that, then I'm  happy.

That said, as much as I prefer Markdown's syntax where it differs from RST, there are some things (notably links) that make me think I might have to switch to RST.

More docs to be added soon to the PR; it'll probably be ready in the next couple days.

dwhswenson:
Ready for review. Just docs, so it should be pretty easy (you've already seen the only programmatic thing in the `premake` script discussed above).

Overview of a few changes:
* removed the old `examples/ipython/transitions_and_networks.ipynb`, which was just Markdown anyway, and replaced with `docs/guides/transitions_and_networks.rst`. Also expanded that documentation.
* new `network.rst` for API documentation
* re-added `ensemble_frames.png`, which got deleted at some point (I think by accident)
* other minor fixes (copyright date, license, etc.)

jhprinz:
lgtm

PR #473, opened 2016-04-26
strict_can_append

### LINKED ISSUE #472: Faster analysis: new `can_append`-like functions
I was looking into why the analysis was running so slowly, and I found an important reason. The basic issue is that `can_append` does the right thing for path sampling, but not for analysis -- especially with sequential ensembles. Explanation:

Take the following (familiar) ensemble:
```python
ensemble = paths.SequentialEnsemble([
    paths.LengthEnsemble(1) & paths.AllInXEnsemble(stateA),
    paths.AllOutXEnsemble(stateA | stateB),
    paths.LengthEnsemble(1) & paths.AllInXEnsemble(stateA | stateB)
])
```

To make it simple, let's assume 1D; `stateA` is `x < -1.0`, and `stateB` is `x > 1.0`. Now look at the following trajectory:

```python
trajectory = make_1D_traj([-0.1, 0.0, 0.1, 0.2])
```

So we are always outside the states. No subtrajectory of `trajectory` is in `ensemble`, so `ensemble.split(trajectory)` should return empty.

What happens when we try `ensemble.split(trajectory)`? This uses `can_append` to try find subtrajectories that fit the ensemble. The key thing, and the challenge here, is that `ensemble.can_append(make_1D_traj([0.0])) == True`, because we define `can_append` as "this trajectory can be a subtrajectory of a trajectory in the ensemble." That's the right definition of `can_append` to use for a lot of things (for example, it makes 2-way shooting more sane), but it screws up our analysis. To see why, let's step through what happens with `ensemble.split(trajectory)` (where I'll just use the list of x-coordinates as a stand-in for the trajectory):

```
ensemble.can_append([-0.1]) == True
ensemble.can_append([-0.1, 0.0]) == True
ensemble.can_append([-0.1, 0.0, 0.1]) == True
ensemble.can_append([-0.1, 0.0, 0.1, 0.2]) == True
# no more frames, so check the ensemble
ensemble([-0.1, 0.0, 0.1, 0.2]) == False
# step to next initial frame
ensemble.can_append([0.0]) == True
ensemble.can_append([0.0, 0.1]) == True
ensemble.can_append([0.0, 0.1, 0.2]) == True
ensemble([0.0, 0.1, 0.2]) == False
# etc
```

The scaling is therefore quadratic in general. This is why our analysis is so slow.

Again, I think the definition of `can_append` is what we want for it. So what we need are new ensemble functions. I propose calling them either `can_prefix`/`can_suffix` or `strict_can_append`/`strict_can_prepend`. The point is that they will only return true if the given trajectory can be the beginning (resp. ending) of a trajectory.

For most ensembles, they are exactly `can_append`/`can_prepend` : this is true for all the `VolumeEnsemble`s, as well as for the `LengthEnsemble`. Combinations behave in exactly the same way as the combinations of `can_append`/`can_prepend` as well. The only major difference I've noticed is in the `SequentialEnsemble`, where the function would be like its `can_append`, but with the that allows you to start with frames matching ensembles other than the first ensemble removed.

Using these, the above step through would be:

```
ensemble.strict_can_append([-0.1]) == False
# because (Length(1) & In(A)).can_append([-0.1]) == False
# step to next initial frame
ensemble.strict_can_append([0.0]) == False
ensemble.strict_can_append([0.1]) == False
ensemble.strict_can_append([0.2]) == False
# now split returns the empty list and we're done: linear scaling
```

I'll plan to add these functions: any preference on their names?

dwhswenson:
Resolves #472. See that issue for a detailed description of the problem. In discussion below, I use the terminology for the `can_append` analog; the modifications for `can_prepend` are what you would expect.

Since the problem is that `can_append` accepts trajectories that are *any* subtrajectory, not just the start of a trajectory in the ensemble, the solution is to have new functions that require that the given trajectory actually be the *start* of a trajectory in the ensemble.

Several name options I’ve considered for these functions:

* `strict_can_append` / `strict_can_prepend`
* `can_prefix` / `can_suffix`
* `can_be_prefix` / `can_be_suffix`

Currently, I’m using `strict_*`, because that seemed most natural to me. But if there’s a preference for a different name, that can be changed.

For many of the “atomic” ensembles we use to assemble more complicated ensembles (e.g., `VolumeEnsemble`s, `LengthEnsemble`s) this is exactly the same as `can_append`. For most other ensembles (`Wrapped`, `Combination`, etc), it behaves just like `can_append`, but with `strict_can_append` in place of `can_append`. However, it does require a separate function for the `SequentialEnsemble`. This new function will disallow the possibility of starting from anything other than the first subensemble of the sequence.

- [x] Implement `strict_can_append`/`strict_can_prepend` for ensembles other than sequential
- [x] Tests for `strict_can_*` in ensembles other than sequential
- [x] Implement `SequentialEnsemble.strict_can_*`
- [x] Tests for `SequentialEnsemble.strict_can_*`
- [x] Check speed improvement in analysis
- [x] Check coverage in `ensemble.py`


dwhswenson:
Going to do further checks, but just looking at a simple test, this change means that instead `find_valid_slices` calling its ensemble's `can_append` 5402 times, it calls the `strict_can_append` 299 times. Later in the call graph, this translates to `AllInXEnsemble.can_append` being called 631 times instead of 31838.

In other words, massive speed improvements for analysis.

dwhswenson:
Tested on the analysis notebook where I’d noticed that this was going very badly. Previous timings:

```
CPU times: user 8h 55min 59s, sys: 6min 55s, total: 9h 2min 54s
Wall time: 8h 58min 50s
```

New timings:

```
CPU times: user 1min 55s, sys: 876 ms, total: 1min 56s
Wall time: 1min 56s
```

So yeah, that’s a little bit faster. Worth the coding effort.

Note that these speed-ups are really only relevant when doing analysis using `ensemble.split`. It won't have an effect in the generation of trajectories, because the sequential ensemble caching mechanism should have had the same effect already.

jhprinz:
Phantastic. This seems really to be a huge improvement. I realized in the alanine example that this step was surprisingly slow (not as slow that it would have been a problem though) and this was only a single call to split.

dwhswenson:
Now that it passes tests (and even increases coverage!) I'll call this ready for review.

How big of an improvement really depends on how many frames (and how much the trajectory is like the worst-case), since this makes an algorithm that scales as $N^2$ worst-case into one that scales as $N$. The example I showed had trajectories of something like 10000 frames, so that makes a huge difference!

This also includes some cleanup stuff, and notes for further cleanup, in `ensemble.py`. We’ll need to complete the cleanup before release.

I'm still having frequent problems passing `alanine.ipynb` due to timing out. Specifically, cell 18 (generating the first trajectory) often takes a very long time. I tried in on my (5-year-old) laptop, and it ran quickly with no problems. When I checked the call graph from `gprof2dot`, I didn't see anything that looked particularly slow, so my guess is that the problem might be just that the processor resources on Travis are pretty weak. (I don't think my old laptop had GPU acceleration, but that might also account for it.)

jhprinz:
Seems to have been more work than I thought. Very good. Merging...

jhprinz:
Yes, `alanine.ipynb` not sure. it always runs on my machine without problems. We could just make it much smaller or use smaller interfaces. This examples might need an overhaul anyway.

PR #474, opened 2016-05-11
[WIP] Create storages with example system setups

jhprinz:
### What this is

I find myself constantly needing the code to to create example systems. Instead of copying the code cells and hoping that I did not forget anything this will just store a template and the corrresonding engine in a storage file. 
Currently it will only generate the mstis toy setup (not the MISTIS) and the alanine. You can then load a setup simply by 

```py
setup = Storage('setup_xyz.nc', mode='r')
template = setup.template
# or
template = setup.tag['template']
engine = setup.tag['engine']
# if necessary like for OpenMM
engine.initialize()
```

For alanine I have also added psi and psi which seemed to be useful.

If we split of the set up of the system we could even split of this part in the examples.
```
mstis_system.ipynb # only setup the system and engine
mstis_prepare.ipynb # generate initial pathways
mstis_generation.ipynb # here actual OPS starts
mstis_analysis.ipynb # analyze
```

### DIFF
--- a/examples/ipython/create_example_data.ipynb
+++ b/examples/ipython/create_example_data.ipynb
@@ -1,0 +1,283 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Create example system to be loaded later for examples"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### A 3 well test potential in 2D with ToyDynamics and a Langevin Integrator"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "import openpathsampling as paths\n",
+    "import numpy as np\n",
+    "import openpathsampling.engines.toy as toys"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "# Toy_PES supports adding/subtracting various PESs. \n",
+    "# The OuterWalls PES type gives an x^6+y^6 boundary to the system.\n",
+    "pes = (\n",
+    "    toys.OuterWalls([1.0, 1.0], [0.0, 0.0]) +\n",
+    "    toys.Gaussian(-0.7, [12.0, 12.0], [0.0, 0.4]) +\n",
+    "    toys.Gaussian(-0.7, [12.0, 12.0], [-0.5, -0.5]) +\n",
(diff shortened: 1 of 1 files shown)


dwhswenson:
> If we split of the set up of the system we could even split of this part in the examples.

What's the need for a separate "setup-only" file? Everything from that is in the first file you actually make. I'm pretty sure that's how the MSTIS examples already work, right? `mstis.ipynb` loads from the output of `mstis_bootstrap.ipynb`. This sequence *has* to be run in order, otherwise it doesn't run. So the existence of that file is guaranteed. (The choice to recreate the network was intentional to show how you can remap sample sets between sets of ensembles which contain equivalent ensembles -- that's something that is likely to come up often enough that the easy tool for it should show up in the examples.)

> I find myself constantly needing the code to to create example systems.

Why? Where? Again, if it is part of a sequence (as both `alanine_dipeptide` and `mstis` are), the fact that the files have to be run in order guarantees that you can just reload the system from the earlier file. Doesn't that solve the problem?

***

The point of the examples is to teach, not to be an optimal implementation. I think that you're thinking like a programmer here, and not like a teacher. I agree that, in real life, it makes sense to have a file containing stuff used in common across a whole project. But that requires a level of abstract thinking that we shouldn't require of our novice users.

I'm especially opposed to storing the phi and psi variables in alanine dipeptide. Making these ***is part of the learning process***. The goal in the examples is to teach, not to just show off what we can do. If important things get repeated, that's fine. I'm less opposed to having a separate "engine creation" notebook, except that should be a separate notebook for each example. If a student is preparing to do something using only toy dynamics, why should they have to read through a notebook that has a bunch of stuff about setting up an OpenMM system?

jhprinz:
> Doesn't that solve the problem?

No, I want a _some tutorial file that might change in the future_ independent solution. Using the `trajectory.nc` file is what I used so far and but I thought this way it is clearer and better separated.

I agree that we should not necessarily use this in the examples that are meant to explain how to setup a system. But in other examples like in talks, I could be useful to have a short yet executable example.

It is also VERY useful in cases where we do not need initial trajectories, like MSMs. This is where it all started.

> I think that you're thinking like a programmer here, and not like a teacher.

Well, no. First, I way balancing that difficulty in setting up the system as being a confusing to teaching OPS. I remember you always saying "look this is the most complicated part". And I agree. So you could just start with the examples system (magically) teach all the stuff that you actually want to teach and later sneak in the _annoying_ setting up of the system. I would think that is very much thought like a teacher. 

But all of this was not meant to spawn a huge discussion. I only wanted to have a shortcut in OPS to start with the common systems without relying on tutorial stuff. That's it. Like openmmtools is extremely useful for openmm users and so this would be useful for me (and I guess other people that want to dive in without having to setup a system).

jhprinz:
Will just do it locally not being part of the package

Issue #475, opened 2016-05-13
Problem in test_netcdfplus.ipynb


dwhswenson:
I'm seeing regular test failures of `test_netcdfplus.ipynb` cell 73. Re-running often fixes. An example that failed (and I didn't re-run): https://travis-ci.org/choderalab/openpathsampling/builds/129109156, which is from a PR that is entirely documentation changes.

I suspect this is because that cell outputs a dictionary, and it expects the exact same string (meaning that the dictionary print order is uniquely defined, which isn't true). @jhprinz, could you fix this?

dwhswenson:
Close by #477.

dwhswenson:
Nope, that didn't seem to fix it. See https://travis-ci.org/choderalab/openpathsampling/builds/130191736.

jhprinz:
Wierd. I am not able to reproduce the error... We look into that...

dwhswenson:
Can ipynbtest output the mismatched text? That would go a long way toward debugging the problem.

jhprinz:
Yes, trying right now...

jhprinz:
I think this can be closed. `ipynbtest` was updated and should not report the error now.

dwhswenson:
Agreed.

PR #477, opened 2016-05-14
Fix dict order in `test_netcdfplus.ipynb`

jhprinz:
This should fix the problem with the test in cell 73 and closes issue #475.


dwhswenson:
Thanks for the quick fix. This looks to me like it will fix it, and it passes tests -- are you ready for me to merge it?

jhprinz:
Yes, please do. If it really was the ordering in the dict. This should be the fix.

PR #478, opened 2016-05-14
Ensemble cache for frame-by-frame ensembles

dwhswenson:
This will fix a significant bug in the ensemble code. This boils down to the statement that ensembles should act on trajectories, not on frames. To explain the problem, let's start with some terminology:

***Definition*** *(frame-by-frame ensemble)*: If a trajectory $(x_0, …, x_L)$ is in a *frame-by-frame* ensemble, then all you need is to look at is $x_{L+1}$ to know whether $(x_0, …, x_{L+1})$ is in the ensemble.

For example, consider `AllOutXEnsemble(stateA)`. If we know the value for the previous trajectory, we only need to look at the next frame, without considering the whole trajectory. If all frames $x_i$ with $i<=L$ were outside the state, and frame $x_{L+1}$ is outside the state, then obviously all frames with $i<=L+1$ are outside the state.

However, this is **not** true for combinations of frame-by-frame ensembles. This is the bug that this PR will fix. For example, consider `AllOutXEnsemble(stateA) | AllOutXEnsemble(stateB)`, and consider a trajectory `0A00B`, where `A` indicates a frame in `stateA`, `B` indicates a frame in `stateB`, and `0` indicates a frame in neither.

While the subtrajectory `0A00` is in `AllOut(B)` (and therefore `AllOut(A) | AllOut(B)`) and the subtrajectory `00B` is in `AllOut(A)`, the full trajectory is in neither `AllOut(A)` nor `AllOut(B)`, and therefore is not in their union, `AllOut(A) | AllOut(B)`.

The current code fails this test if built during a trajectory using `trusted=True`, which essentially tries to take a frame-by-frame approach to the trajectory. (Note: if not using `trusted=True`, our code gets the right answer -- but then building a trajectory scales quadratically with the length of the trajectory instead of linearly.) With `trusted=True`, the code assumes that the combination of frame-by-frame ensembles is also frame-by-frame. But a combination of frame-by-frame ensembles is not necessarily frame-by-frame. In the `0A00B` trajectory, each frame individually is in `AllOut(A) | AllOut(B)` -- again, this goes back to my old mantra that "ensembles are about trajectories, not frames."

The approach I’m using to fix this is to assign a cache to every frame-by-frame ensemble. Assuming it is still using the same trajectory (we already have ensemble cache to check that), we will cache the result of the previous subtrajectory. Our frame-by-frame ensembles have a behavior similar to the poisoned frame in PET: for some, once they are false they are always false (`AllIn`, `AllOut`) and for others, once they are true they are always true (`PartIn`, `PartOut`). This means that if we cache a result for trajectory $(x_0, …, x_L)$, we can combine that with a result based on only the frame $(x_{L+1})$ to get a result for the whole trajectory $(x_0, .., x_{L+1})$. For the `All*` ensembles, the combination is a logical `and`. For the `Part*` ensembles, it is a logical `or`.

In either case, the combination should then return the correct result. Consider the following, where each column corresponds to a frame, and the rows tell you what various ensembles should give as you build the trajectory from leftmost column to rightmost:

```
     traj 0A00B
  AllOutA TFFFF
  AllOutB TTTTF
    Union TTTTF
Intersect TFFFF
  Current TTTTT (union; trusted=True)
```

The caches of `AllOutA` and `AllOutB` should give the values above once the current frame has been evaluated.

- [x] Draft initial test to reproduce problem
- [x] Add support for ensemble cache to AllInX/AllOutX `__call__`
- [x] More complete tests for AllInX/AllOutX `__call__`
- [x] Support/tests for AllInX/AllOutX `can_*` methods
- [x] Fill in coverage gaps

### DIFF
--- a/devtools/ci/ipythontests.sh
+++ b/devtools/ci/ipythontests.sh
@@ -27,7 +27,7 @@ ipynbtest.py --strict "test_snapshot.ipynb" || testfail=1
 date
 ipynbtest.py "tutorial_storage.ipynb" || testfail=1
 date
-ipynbtest.py --strict "test_netcdfplus.ipynb" || testfail=1
+ipynbtest.py "test_netcdfplus.ipynb" || testfail=1
 date
 ipynbtest.py --strict "test_cv.ipynb" || testfail=1
 date
--- a/openpathsampling/ensemble.py
+++ b/openpathsampling/ensemble.py
@@ -1612,7 +1612,17 @@ class AllInXEnsemble(VolumeEnsemble):
         frame_num = -(cache.direction + 1) / 2  # 1 -> -1; -1 -> 0
         reset = cache.check(trajectory)
         if reset:
-            cache.contents['previous'] = None
+            if len(trajectory) < 2:
+                cache.contents['previous'] = None
+            else:
+                # NOTE: is it possible that we'd reset a cache more than
+                # once in a single trajectory? that could mean that this
+                # starts to scale quadratically
+                # I think a necessary condition for that to be a problem is
+                # that this ensemble shows up in both a Union and an
+                # Intersection ...
+                cache.contents['previous'] = self(trajectory[:-1],
+                                                  trusted=False)
         cached_val = cache.contents['previous']
         if cached_val == True or cached_val is None:
             # need to check this frame (no prev traj, or prev traj is True)
@@ -1666,7 +1676,7 @@ class AllOutXEnsemble(AllInXEnsemble):
     '''    
     @property
     def _volume(self):
-        return ~ self.volume
+        return ~self.volume
     
     def __str__(self):
         return 'x[t] in {0} for all t'.format(self._volume)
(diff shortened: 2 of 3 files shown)


dwhswenson:
Decided to leave out support for `Part*XEnsemble`s with this, because (1) those don't currently make use of the `trusted` flag, and so they don't have the bug; (2) the `trusted` flag mainly needed for `can_append`, and isn't as useful when doing `__call__`; (3) I couldn't think of any specific situations where you could actually use this (since `engine.generate` runs until the given function returns `False`, not until it returns `True`.)

That said, if we ever want to add a cache to `Part*XEnsemble`s, it should be pretty easy to base it off the code in the `All*XEnsemble`s.

dwhswenson:
Passes tests, ready for review.

This also turns off the `--strict` option in `test_netcdfplus.ipynb`. That's intended to be a temporary measure until #479 fixes problems with that.

jhprinz:
I have just uploaded an improved version of ipynb-test that should fix the problem.

jhprinz:
Great. That looks like a rather simple solution. Nice. I expected it do have more lines. Not sure if I completely understood what you did. But since the test works I will merge it.

PR #479, opened 2016-05-14
fix test_netcdfplus

jhprinz:
Needs some work. I am not able to reproduce the error...


jhprinz:
I identified the issue and uploaded a better ipynb version that takes care of it. Seems that in some cases the textual output stream of stdout might be split into multiple cells, while in stored files this is always a single cell. I assume that has to do with when the stdout is flushed and this is also the reason why it works sometimes.

Just checking, if the new ipynb-test works.

jhprinz:
I think this is fixed. At least I did not see this error happening again. Let's close this for now. If another error appears I will look at this again. So far I think I caught the source of the problems, although I cannot guarantee that this will catch all future problems.

jhprinz:
The PR will add the `--strict` again and streamline a few cells in the `test_netcdfplus.ipynb`, nothing else. The fix is in some changes in the `ipynbtest.py`.

dwhswenson:
Seems to be working again -- good by me!

PR #480, opened 2016-05-15
[WIP] find testnetcdf bug

jhprinz:
just to see what is happening on travis. Not to be merged!


jhprinz:
Okay, this has to do with the way the cell produced output. Somehow there are two outputs stored and only one generated or the other way round.

So, its either a bug in ipynbtest or a change in jupiter notebook between my and the current one that travis uses.

jhprinz:
This is not needed anymore.

PR #481, opened 2016-05-17
Add `allow_self_transitions` to `TPSNetwork`

dwhswenson:
Right now, if we create a `TPSNetwork` with `from_states_all_to_all`, we get literally *all* the possible transitions, including A->A. Typically, we don’t want the self-transitions to be allowed.

This PR adds an `allow_self_transitions` flag to the creation of `TPSNetwork` and `FixedLengthTPSNetwork`. This flag will default to `False`, in which case there are no A->A transitions.

This makes a rather significant change in the way MSTPS is handled internally: instead of saying that there is only one sequential ensemble to check (leave any state; go to any state), now it is the union of a sequential ensemble for every state (leave that state; go to any other state). This will have some negative effects on speed, especially when building the trajectory backwards. We'll see scaling something like $O(t * N_S)$, where $t$ is the length of the trajectory and $N_S$ is the number of states.

However, this is important in order to have multiple state TPS work correctly.

- [x] Add `allow_self_transitions`
- [x] Tests for `allow_self_transitions=True`
- [x] Tests for `allow_self_transitions=False`
- [x] Update docstrings

### DIFF
--- a/openpathsampling/analysis/network.py
+++ b/openpathsampling/analysis/network.py
@@ -82,6 +82,8 @@ class GeneralizedTPSNetwork(TransitionNetwork):
         acceptable initial states
     final_states : list of :class:`.Volume`
         acceptable final states
+    allow_self_transitions : bool
+        whether self-transitions (A->A) are allowed; default is False
 
     Attributes
     ----------
--- a/openpathsampling/analysis/network.py
+++ b/openpathsampling/analysis/network.py
@@ -194,8 +194,10 @@ class GeneralizedTPSNetwork(TransitionNetwork):
 
 
     @classmethod
-    def from_states_all_to_all(cls, states, **kwargs):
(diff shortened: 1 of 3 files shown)


dwhswenson:
Ready for review; passes tests.

Note: There are three ways to make a TPSNetwork:

* `TPSNetwork.__init__(initial_states, final_states)`
* `TPSNetwork.from_states_all_to_all(states)`
* `TPSNetwork.from_state_pairs([pairs_of_states])`

The `allow_self_transitions` only applies to the first two. If you explicitly give the pair `(stateA, stateA)` in the list of allowed state pairs in the third version, then we trust that you actually meant to do that. The first two try to guess what you implicitly meant, and in that case, we assume that you don't actually want A->A transitions unless you request them directly with the `allow_self_transitions` flag.

jhprinz:
looks good

Issue #483, opened 2016-05-19
Box_Vectors not set in openmm engine


jhprinz:
I call this a serious but super easy to fix bug and did that already in the uuid PR #443.

When assigning the `current_snapshot` for openmm then the box_vectors are not set. I am actually not sure what openmm is simulating right now, but I assume that we do not using periodic boundaries. This is a serious bug, but easy to be fixed. That might be a reason that we get some NaNs in the Alanine simulations.

dwhswenson:
Closed by #484.

PR #484, opened 2016-05-19
fix setting box_vectors

jhprinz:
Ready for review and merge

This closes issue #483 .

Also adds a test.

### DIFF
--- a/openpathsampling/engines/openmm/engine.py
+++ b/openpathsampling/engines/openmm/engine.py
@@ -218,12 +218,18 @@ class OpenMMEngine(DynamicsEngine):
 
     @current_snapshot.setter
     def current_snapshot(self, snapshot):
-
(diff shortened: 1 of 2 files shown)


dwhswenson:
Good catch -- definitely an important bug to fix! I'm not sure how this would be causing the NaNs (seems to me like, with no periodic box, you'd be less likely to have atoms get too close, which is what usually causes NaNs), but since this definitely needs to be fixed, we can hope it fixes other problems too!

Waiting for it to actually pass all tests, then I'll merge.

BTW: @jhprinz -- you keep using the syntax "closes issue #whatever". Adding the word "issue" in there breaks the GitHub magic auto-close. It should be "closes #whatever". See allowed syntax at: https://help.github.com/articles/closing-issues-via-commit-messages/

PR #485, opened 2016-05-19
Add `file_size` and `file_size_str` to netcdfplus

dwhswenson:
This just gives a shortcut to obtain the size of a file associated with a netcdfplus storage. It also gives a property with string output to convert that to a human-readable number in kB, MB, or GB (technically kiB, MiB, or GiB).

This is useful when doing analysis in an ipynb -- it is nice to get an overview of how big the file is before running analysis that might take a while.

Should be ready to merge immediately on passing tests.


jhprinz:
Good.

PR #486, opened 2016-05-19
Fix bug in move scheme frequency calculation

dwhswenson:
Fixes problem that move schemes included the EmptyMove (used to set initial conditions) when calculating the frequency at which a move was actually used. This gave $N/(N+1)$ for $N$ moves when only one move was possible: so instead of the only move possible being run 100/100 = 100% of the time, it was run 100/101 = 99.01% of the time (gradually getter closer to 100% with more trials, but still, looks like nonsense)

Ready to merge as soon as it passes tests.


jhprinz:
Good catch.

PR #491, opened 2016-06-07
Switch engine storage from scheme to mover

dwhswenson:
At some point in making the alanine dipeptide examples, I put the storage of the engine in as part of a move scheme. However, this doesn't really make sense -- it needs to be stored with the `EngineMover`, whether it is part of the scheme or not. And it is only passed to the scheme as a convenience function.

This was causing problems for OPSPiggybacker. This PR fixes those problems, and it seems that it also runs (short versions of) the AD TPS examples correctly too. Should be ready for review/merge as soon as it passes tests.


jhprinz:
Good

PR #495, opened 2016-06-16
DirectSimulation for rate and flux

dwhswenson:
This PR adds a simulator class for direct MD calculations of the flux through an interface or of the rate. This is useful because MISTIS requires a separate flux calculation. Although we can calculate the flux from an existing trajectory (see `single_trajectory_analysis` and #435), we often want to calculate the flux without saving every frame (especially with toy models). That is possible with this PR.

- [x] Direct MD calculation of flux and rates
- [x] Tests for rates
- [x] Test with Calvinist dynamics
- [x] Tests for flux
- [x] Tests for storage
- [x] Docstrings

### DIFF
--- a/examples/ipython/mistis_flux.ipynb
+++ b/examples/ipython/mistis_flux.ipynb
@@ -1,0 +1,296 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "%matplotlib inline\n",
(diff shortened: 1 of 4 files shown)


dwhswenson:
Passes tests. Ready for review.

Mostly straightforward, and exactly what it says in the top post. Also includes an ipynb to get the flux for our MISTIS toy example.

There is one clever trick: I came up with a way to mock up a 2-dimensional system using only one dimension. See `testDirectSimulation.test_flux_from_calvinist_dynamics` for the trick and its description.

jhprinz:
lg;wm

PR #497, opened 2016-06-22
Tests and cleanup of bootstrapping

dwhswenson:
When I started writing the section of the paper on the toy model MISTIS and the bootstrapping, I found that a few details that were a bit confusing. I think this may have led to some confusion in the stuff that Abel was doing. To make sure that the behavior in the code matches what I write in the paper, I’m writing tests for the bootstrapping.

This may also lead to some changes in the behavior of the `FullBoostrapping`, and possibly in the `MISTISNetwork`. I think I’ve occasionally seen bad initial conditions in the MISTIS ipynb test, causing annoying failures. I’m hoping that this PR might fix that. At the very least, it will give a nice bump to our test coverage, and ensure that our paper matches our code!

- [x] Tests for the `first_traj_ensemble`
- [x] Tests for the sampling ensembles
- [x] Change output of `FullBootstrap` & `Bootstrap` to file/stream based (so we can write to `/dev/null` or `stdout` depending on what we want)
- [x] Tests when extra interfaces exist
- [x] Tests when forbidden states exist
- [x] Tests when bootstrapping fails from too many attempts
- [x] Add numpydoc docstrings to FullBootstrapping

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -3,6 +3,7 @@ omit =
     */python?.?/*
     */site-packages/nose/*
     */openpathsampling/tests/*
+    */mdtraj/*
 exclude_lines = 
     pragma: no cover
     def __repr__
--- a/openpathsampling/analysis/network.py
+++ b/openpathsampling/analysis/network.py
@@ -555,7 +555,7 @@ class MISTISNetwork(TISNetwork):
         unnamed_states = [s for s in list_all_states if not s.is_named]
         name_index = 0
         for state in unnamed_states:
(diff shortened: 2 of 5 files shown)


dwhswenson:
Okay, I don't think I'm going to change anything else in this (for this PR). I'll make some minor modifications to the `MISTISNetwork` in a separate PR.

This is ready for review. There's a little cleanup, but mostly new tests (and a big jump in test coverage!)

jhprinz:
Lg;wm

PR #498, opened 2016-06-23
Fix `storage_mem_test`

### LINKED ISSUE #447: More ipynb cleanup
This follows on #446. There were a few things I wasn't sure how to fix from some of @jhprinz's notebooks. Rather than do something that might not match the original intent of the notebook, I'm leaving these for @jhprinz.

NOTE: you should only work on this *after* merging #446. I've already made a few changes to these notebooks to give partial fixes. There are just a few cells left where I don't know what you want.

Two notebooks have problems:

#### `storage_mem_test.ipynb`

The problem is in cell 15. It looks like all values that you calculated in `dd2` are zero. All other cells run fine.

#### `tutorial_visualization.ipynb`

* First problem: cell 5. Toy snapshots no longer have `configuration`s. Switching `configuration` to `coordinates` doesn't work, since of course coordinates don't have the `idx` attribute.
* Second problem: cell 8. Possibly related? But this seems to be choking inside of the visualization code.

Again, the rest of it runs without complaint.

jhprinz:
Resolves #447.

There was a bug in activation the observation of storable objects so the later analysis failed in the notebook. This fixes it. Since the notebook is not run as a test it should pass.

To be merged after the storage version PR #397 

Contains #397 so if you merge this one. The storage version will be merged as well. Only change in this one is fixing the notebook and a few lines in `netcdfplus/base.py`. Maybe that is faster than to wait that #397 passes alanine eventually.


jhprinz:
Ready for review and merge. Note that #397 will also be merge with this one!

dwhswenson:
Was the necessary stuff from this included in #397? (i.e., can this be closed?)

jhprinz:
Yes, this should not be merged and closed!

PR #499, opened 2016-06-23
Fix bug in storing CVs

### LINKED ISSUE #489: Netcdfplus doesn't store functions for CV_Functions in tests outside of OPS
I don't think this is supposed to be the desired behavior.

Consider the following file, to be run as a test with `nosetests`:

```python
# test_file_locations.py
from openpathsampling.tests.test_helpers import make_1d_traj
import os

class testFileLocations(object):
    def setup(self):
        if os.path.isfile("myfile.nc"):
            os.remove("myfile.nc")

    def teardown(self):
        self.setup()

    def test_something(self):
        template = make_1d_traj([0.0])[0]
        cv = paths.CV_Function("x", lambda snap : snap.coordinates[0][0])
        storage = paths.Storage("myfile.nc", "w", template)
        storage.save(cv)
```

Put it in `openpathsampling/openpathsampling/tests/`, run `nosetests test_file_locations`, and everthing is fine.

Put it in `~/testing` and do the same, and you get an error:

```
RuntimeError: Locally defined classes are not storable yet
```

The problem is that you miss an important category of function. You account for non-class callables that are local (assuming the module is `__main__`) or what you call `loaded`, meaning the module includes `openpathsampling`. 

What you miss is the possibility of using OPS code as the basis for other packages (so module doesn't include OPS) and running tests on those packages (so the module at runtime isn't `__main__`).

In principle, if we can store `lambda` functions when they're in an OPS directory, then we should be able to store them outside of OPS.

A couple more details: I added a little debug logging before the failure. The debug code:

```python
        logger.debug("Failing for callable " + str(c) + " from module " +
                     str(f_module) + ". Is a class? " +str(is_class))
```

gave me the output

```
openpathsampling.netcdfplus.dictify: DEBUG: Failing for callable <function <lambda> at 0x1114af9b0> from module ops_piggybacker.tests.wtf. Is a class? False
```

This also explains why we never saw it before: previously, we had `openpathsampling.tests.test_name` as the module, so it qualified as "`loaded`".

jhprinz:
Resolves #489 

Actually this reduces security a little and removes the check for locality. I was not able to reproduce the error so I cannot test if it works now.


dwhswenson:
It should be trivial to reproduce the error. Make a directory *not* in the OPS path, like `$HOME/my_test_directory`. Paste the example from #489 into a file called `test_file_locations.py` in that directory. Type `nosetests`. You should get the error.

jhprinz:
Thanks will try

jhprinz:
Needed to copy some code. But now it seems to work. I just removed the check for locality. Worst thing that can happen is that it will fail for an unknown reason. But blocking it to run working versions does not make sense. I will check if you can load the stored version correctly though

jhprinz:
Works. Nice. That was easy. I assume the local stuff had some security and legacy reason. I actually do not remember why I checked for that. Does not seem to make sense.

jhprinz:
Restarting...

jhprinz:
I remember why I did the local check. Actually the solution was to first try the fall_back save solution and if that is not working try the hard way using marshal. That does the trick.

The original local check was to store the package and the function if it was not local. Now it tries to figure out the package (like mdtraj) and save only the function name. If not a known package then use saving by bytecode.

dwhswenson:
Remarking again: this is assigned to me for review, but currently has conflicts with master. @jhprinz : can you fix conflicts? Then I'll review it!

jhprinz:
Sorry, did not see that. Was a conflict in notebooks that appeared merging other PRs.

dwhswenson:
No worries. Also, the merge makes it easier to review, since it previously had other PRs mixed in the diff.

It's not clear to me that the test you added would identify this problem. (In fact, I'm not sure that there is an easy test for it, since it would need to be tested outside the test suite.) However, it looks like this should basically fix it -- and you did test it on an external file, right?

A couple line notes coming: should be quick fixes.

openpathsampling/netcdfplus/dictify.py
```diff
@@ -240,78 +240,71 @@ def callable_to_dict(c):
...
+            global_vars = ObjectJSON._find_var(c, opcode.opmap['LOAD_GLOBAL'])
+            import_vars = ObjectJSON._find_var(c, opcode.opmap['IMPORT_NAME'])
+
+            builtins = dir(__builtin__)
+
+            global_vars = list(set([var for var in global_vars if var not in builtins]))
+            import_vars = list(set(import_vars))
+
+            if len(global_vars) > 0:
+                err = 'The function you try to save relies on globally set vari00ables' + \
```
dwhswenson: typo with the 00 in `vari00ables` (trying to reset the kernel? :-P)

openpathsampling/netcdfplus/dictify.py
```diff
@@ -240,78 +240,71 @@ def callable_to_dict(c):
...
+                                   if module not in ObjectJSON.allowed_imports]
+
+            if len(not_allowed_modules) > 0:
+                err = 'The function you try to save requires the following modules to ' + \
+                      '\nbe installed: ' + str(not_allowed_modules) + ' which are not marked as safe!'
+                err += '\nYou can change the list of safe modules in "CV_function._allowed_modules"'
+                err += '\nYou can also include the import startement in your function like'
+                err += '\n' + '\n'.join(['import ' + v for v in not_allowed_modules])
+
+                print err
```
dwhswenson: probably don't need to print this error as well as raising it -- raising it should do. (debug code?)

jhprinz: Correct. This was meant as an explanation on how to fix the issue as to be able to fix it. We can either add it to the logger or add the text to the exception. Ideas?

openpathsampling/netcdfplus/dictify.py
```diff
@@ -240,78 +240,71 @@ def callable_to_dict(c):
...
+
+                raise RuntimeError('Cannot store function! Not allowed modules used.\n' + err)
```
dwhswenson: can we change `Not allowed modules` to `Unsafe modules`? "Not allowed" sounds weird as an adjective; could be "disallowed" or "forbidden", but that implies an intentional prohibition instead of an omitted permission. Since the error text uses "list of safe modules", I think "unsafe" is a good term here.

openpathsampling/netcdfplus/dictify.py
dwhswenson: Don't you already add the text the to exception? That works by me.

openpathsampling/netcdfplus/dictify.py
dwhswenson: (also, in that note, I just mean in the text of the RuntimeError. The variable name is clear enough, based on the code defining it!)

openpathsampling/netcdfplus/dictify.py
```diff
@@ -240,78 +240,71 @@ def callable_to_dict(c):
...
+                # print [obj for obj in global_vars]
+
+            not_allowed_modules = [module for module in import_vars
+                                   if module not in ObjectJSON.allowed_imports]
+
+            if len(not_allowed_modules) > 0:
+                err = 'The function you try to save requires the following modules to ' + \
+                      '\nbe installed: ' + str(not_allowed_modules) + ' which are not marked as safe!'
+                err += '\nYou can change the list of safe modules in "CV_function._allowed_modules"'
+                err += '\nYou can also include the import startement in your function like'
```
dwhswenson: typo: `statement` not `startement`

openpathsampling/netcdfplus/dictify.py
jhprinz: > (also, in that note, I just mean in the text of the RuntimeError. The variable name is clear enough, based on the code defining it!)

sure, I forgot to remove the `print err`

> can we change Not allowed modules to Unsafe modules?

done.

openpathsampling/netcdfplus/dictify.py
jhprinz: :)

jhprinz:
ready to review

dwhswenson:
Looks good! I'll have to keep in mind that we have the word-wrap now... might be useful in the future to also give it an option to indent a certain number of spaces.

Merging!

PR #500, opened 2016-06-24
Visualization update for MovePathTree

### LINKED ISSUE #452: (Minor) Bug in `ConditionalSequential.input_ensembles`

### LINKED ISSUE #494: Add initial sampleset to MoveTree Visualization
this will replace #453. We mainly agreed that the current visualization can be confusing because of the way we treat moves.

#### Example

A move will take a sample from ensemble `A` and create from it a sample in ensemble `X`. After that we will have a reversed move that will take the sample in `X` and create one in `A`. One would assume that the combined move should have input and output in `A`, but it will actually have input `A` and output `A` and `X`.

It gets even more complicated, if we have multiple samples per ensemble. Assume that initially there was something in `X` then the second move would chose between two and effectively depend on `X` so an input from `X` is necessary.

This is technically correct, but not what is really happening. Hence it is confusing to see and could alarm a user where everything is right.

One way to solve this is to add an initial sampleset or just a list of occupied ensembles before the mover is run. That should solve the ambiguity.

jhprinz:
Resolves #453. Resolves #452. Resolves #494.

To be merged after #400.

This will add some features to pathmovers that allows us to inspect how a mover will shift samples between ensembles. 

#### Example

A Shooter will "remove" a sample from the ensemble it is set to and replace it with one from the same ensemble. For a simplest this looks like "replace a sample in ensemble X". This can be expressed by

```
X -> X
```

A RepEx move will switch to samples two two different ensembles, e.g. X and Y. Expressed like this

```
X -> Y
Y -> X
```

An Extension will take one from ensemble X entend and put it in ensemble Y.

```
X -> Y
```

If you the extension twice you get

```
X -> Y
X -> Y
# simpler
X -> Y, 2
```

Consider a RepEx followed by a shooting `X -> X`. Then we naively get

```
X -> Y
Y -> X
# followed by
X- >X
# reduces effectively to
X -> Y
Y -> X
```

Problem is that this is only true, if we have exactly one sample in `X`. If there are two, we have a second possibility in case after the repex the shooter does not chose the same replica, but another one.

```
X -> Y
Y -> X
X -> X
```

#### What is it?

This is what an `InOut` will represent. A list of tuples with a multiplicity. This should be mostly hidden from the user and it only meant for inspection and sanity or consistency checks. Since the concatenation of moves can have several possibilities a `InOutSet` is returned containing all possibilities.

The `InOut[Set]` objects will take care of all the ambiguity.

#### What else?

The second object introduces to make this work is `ReplicaState`. This is the stripped version of a sample set. It is basically a `Counter` or `dict(Ensemble : multiplicity)` and represents now many samples are in a particular ensemble. The objects can be created from a `SampleSet` using `ReplicaState.from_sampleset`.

What is that useful for. Basically as a Mover moves a concrete SampleSet. A `InOutSet` moves a `ReplicaStateSet`. So all possibilities that a mover might act upon a set of samples will be taken into account will all possibilities a mover is called with.

This allows to inspect a whole `PathMove`Tree given a specific initial sampleset.

#### What questions does that answer?

1. Given an initial sampleset you can ask what potential simplest will be used at each sub mover, multiplicity (like in AMS) included.
2. This resolves #453 and #452 because it knows that when the MinusMove is called there is no sample in the helper ensemble, Thus allowing for correct move trees.
3. Same as 2. but for output ensembles. 
4. You can check a mover for consistency (which should not be necessary when using MoveSchemes!). That means you can check if a Mover will leave the number of samples per ensemble unchanged or not, which can then lead to failures over several iterations. 
5. You can check what the minimal necessary samples per ensemble is so that the mover can be run without problems. This should not be necessary when using MoveScheme, as these provide the initial list of samples.

#### Future?

As discussed with @dwhswenson we will at some point use graph-based methods to solve these kind of problems. For know this does the job, but that will also mean that we might use this experimentally and mark it as (might change in the future)


jhprinz:
Once it passes ready for review. Since this is experimental and we quickly need plots I will keep tests to the tutorial for now. I used a large notebook to check for all kinds of cases and analysis but think this would need other tests.

I tested the common cases that include all the movers we use now. What is not covered is adding and removing replicas in moves like for bootstrapping. Actually, adding should work fine, but removal of replicas which currently is not supported by mover.apply could be tricky.

jhprinz:
Updated. Now there are basically two files to be reviewed.

dwhswenson:
#### Simple additions: a few docstring details

The in-out stuff is a lot of lines for a small thing. What you have there looks like it will work (although see below for a few specific questions). However, we might later find a more general way of doing that small thing, or might find a way (once PathMovers are DAGs) to get some of the functionality from another package.

As such, could you add some markers in the class docstrings here to help us later identify that these are all tied together? I'm thinking something like just a line saying that, e.g., `ReplicaStateSet` is used for determining ensemble input-output flow for PathMovers. To me, the name alone doesn't make that obvious, and I can imagine being confused by it at a later date.

Similarly, could you add the other related classes in a "See also" section? Again, this will help us maintain the code in the future.

#### Larger question: handling optional ensembles

This seems to work with what we have now, but I'm still not sure how it will work with single replica TIS (which was where the problem arose in the first place). Shouldn't this use the format I developed for `scheme.list_initial_ensembles`? SRTIS requires "one ensemble of the group `(A, B, C)`" for initial conditions. In a more general case, you could have a scheme that requires "ensemble `A`, and one from the group `(B, C)`." Is there a way to give that input to this?

jhprinz:
> Simple additions: a few docstring details

Done. Should be easy to replace or update it later

> Larger question: handling optional ensembles

Simple answer. The current implementation cannot handle conditional ensemble selection, like in `RandomAllowedChoiceMover`. All it can do is evaluate all possible changes in a mover, but not simplify it conditioned on specific initial conditions. it makes sense to implement this, but I need to figure out if that is difficult or simple. In anyway I would not add it to this PR and see if I can do that next week or so. Adding the capabilities will not affect the API and only make this feature less _experimental_ until we figure out how to do it with a graph-based approach.

dwhswenson:
Sounds good. Anything else underway? If not, I'll go ahead and merge.

jhprinz:
A little reformatting of the `PathMover.py` file and little more docs. I realized there are potentially some functions to be removed or simplified. Mostly legacy. But we can fix that later...

When it passes ready to review and merge.

openpathsampling/pathmover.py
```diff
@@ -877,6 +890,16 @@ def legal_sample_set(globalstate, ensembles=None, replicas='all'):
         self.replicas and the parameter ensembles. If ensembles is None, we
         use self.ensembles. If you want all ensembles allowed, pass
         ensembles='all'.
+
+        Parameters
+        ----------
+        globalstate : `openpathsampling.SampleSet`
```
dwhswenson: don't worry about changing it before merge, but I think we want to use the syntax ``:class:`.SampleSet` ``. This leads to Sphinx to link correctly. I don't think we need to explicitly label the namespace if it is in the same namespace (makes readability harder, and doesn't add anything for Sphinx).

See: http://www.sphinx-doc.org/en/stable/markup/inline.html#xref-syntax
Example in use: http://openpathsampling.org/latest/api/generated/openpathsampling.pathsimulator.FullBootstrapping.html#openpathsampling.pathsimulator.FullBootstrapping (except the `TISTransition` parameter)

dwhswenson:
This all looks good. As mentioned in line-note, we should make sure we have link-creating syntax for our Sphinx/numpydocs docstrings, but that can be handled as part of general documentation cleanup.

jhprinz:
My idea was to use `:any:` as the default. This will figure out the most reasonable object itself. This is new in Sphinx 1.5. (or so). Had not seen this before. We should check that before we fix all by hand.

PR #504, opened 2016-06-25
New histogram

dwhswenson:
New histogram code to handle sparse histograms. This no longer uses numpy's histogram (which I’ve always hated) and instead uses a sparse histogram approach (which generalizes nicely to arbitrary dimensions). 

There is one subtle difference in the behavior. The question is about how the edges of the bins are included. In this new current version, all bins are left-inclusive, right-exclusive: `left_edge <= x < right_edge`. In `np.histogram`, this is true for all bins *except the rightmost, which includes both edges*. The numpy approach makes sense for things with actual absolute bounds, and especially with absolute bounds that you’re likely to reach (e.g., scores on a test: group scores by 5%, and 0% should be included in 0%-5%, and 100% should be included in 95%-100%). However, it makes less sense when you’re not likely to exactly include an upper bound. These are typically our cases. Furthermore, the approach I use means that we can combine histograms of related data without needing to know the full range of values. (Example: comparing path length distributions from different calculations, where there *is* no strict upper bound.)

Another minor change: this means that the upper bound of the `bin_range` is no longer used. Previously, we would have cut out data; now we actually have all the data. (Later in this PR, or in a follow-up to it, I’ll probably replace that input with `left_bin_edge` and `bin_width` inputs.)

The sparse histogram will be useful for the path density plot and for free energy plots based on the reweighted path ensemble (as well as some non-path-sampling stuff that I originally wrote similar code for).

The overall setup is that we now have two histogram classes: `SparseHistogram` which contains shared code and acts as the base class, and `Histogram` which is only for 1D (and contains 1D-specific things like cumulative histograms).

- [x] Switch original histogram to being counter-based (to make it more similar to sparse histogram internals)
- [x] Split original histogram reusable code into `SparseHistogram` class
- [x] Re-run analysis notebooks to be sure all still works
- [x] Add details for new class, `SparseHistogram`
- [x] Tests for `SparseHistogram`
- [x] Better docstrings


dwhswenson:
Ready for review and merge.

jhprinz:
looks good...

PR #505, opened 2016-06-27
Make `details` implicit in PathMoveChange

dwhswenson:
This makes it so that the information contained in the `details` object of a PMC can be accessed directly; i.e., `change.some_attribute` will return the value of `change.details.some_attribute` if such exists.

The `details` object is necessary for storage, but I think it can be confusing to the user as to which things are inherently stored in the change itself and which are in its `details` attribute. This aims to avoid that confusion.

Just a quick change; should be ready to merge at any time.


jhprinz:
looks good. Waiting for it to pass.

PR #506, opened 2016-06-28
Path density histogram

dwhswenson:
This PR will implement path density plots. The easiest way to think of path density plots is in terms of what a single path contributes to them: each bin that is traversed by the path adds a count of 1 to the (multidimensional) histogram. This means: (1) it doesn’t matter how long a single path stays in one bin, or how many times it enters that bin; or equivalently (2) important areas of the histogram don’t represent long staying times (stable states), rather, they mean that lots of paths go through there (possible bottlenecks). Path density plots also interpolate (linearly) to include any bins that are skipped between frames.

Historically, path densities have been calculated in a 2D projection. This implementation, based on the `SparseHistogram` code from #504, works in an arbitrary dimensional projection. 

This includes:
* `PathHistogram` : a variant of `SparseHistogram` designed to think in terms of trajectories (after they’ve been mapped to CVs). This allows you to interpolate to include cells between frames that were saved (linear interpolation, initialization keyword `interpolate`), to ignore the *number* of times a given histogram cell is visited and instead only count whether the cell was visited by a path (initialization keyword `per_traj`), and to assign a specific weight to the entire trajectory (keyword `weight` in `add_trajectory` method).
* `PathDensity`: a wrapper to create a path density plot. Basically a version of the `PathHistogram` that can take a list of normal OPS trajectories, map each to the “reduced” trajectory of CVs that the `PathHistogram` requires, and calculate the resulting histogram based on the `per_traj=True` version of the `PathHistogram`.

Note that the `PathHistogram` object will later be reused, with the `per_traj` normalization set to `False` and including the per-trajectory weights, in order to calculate free energy plots from the reweighted path ensemble.

An overview of the capabilities of the `PathHistogram` object are shown in `tutorial_path_histogram.ipynb`. It’s one of those “picture worth a thousand words” things. I think it is a lot easier to see what the options do by looking at that, than it is to explain them in words.

- [x] Implement `PathHistogram` as a `SparseHistogam`
- [x] Add interpolation to `PathHistogram`
- [x] Fix interpolation for the (literal) corner case
- [x] Add `per_traj` normalization to `PathHistogram`
- [x] `tutorial_path_histogram.ipynb`
- [x] nosetests for `PathHistogram`
- [x] `PathDensity`
- [x] nosetests for `PathDensity`
- [x] docstrings

Note: in the not-to-distant future, it might be nice to write a little function to simplify plotting of these. These use data structures that are very similar to the ones for my contact map code, and @sroet has written some matplotlib code to plot those very quickly (going through a pandas frame is fine for small systems, but really sucks once your grid is large). That code should be reusable for these.

### DIFF
--- a/examples/ipython/tutorial_path_histogram.ipynb
+++ b/examples/ipython/tutorial_path_histogram.ipynb
@@ -1,0 +1,415 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Tutorial: Path Histogram\n",
+    "\n",
+    "This gives a little overview of the option and behavior of the `PathHistogram` object. The `PathHistogram` is used for path density plots and free energy plots. It extends the basic `SparseHistogram` code by allowing for interpolation between bins, and by allowing one to normalize on a per-trajectory basis."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "import openpathsampling as paths\n",
+    "from openpathsampling.analysis.path_histogram import PathHistogram"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "%matplotlib inline\n",
+    "import matplotlib.pyplot as plt\n",
+    "import scipy\n",
+    "import pandas as pd\n",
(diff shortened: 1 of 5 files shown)


jhprinz:
wow, 9 subpoints ...

dwhswenson:
> wow, 9 subpoints ...

Yep ... although some of them are just to call attention to a few subtle things I made sure to do correctly (the literal corner case is one: never happens in practice, but the code is correct).

And also, managed to finish all the missing bits in one evening of work. Passes tests; ready for review!

jhprinz:
Nice. I think the will be very useful in the future.

PR #508, opened 2016-06-29
Switch analysis to being step-based, instead of storage-based

dwhswenson:
This makes it so that the main analysis functions to get TIS rates (and related: crossing probabilities, flux from minus move, and conditional transition probability) are based on a list of steps, instead of storage.

The storage requirement predated the existence of MC steps unifying the active sample set and the path move change. This is a much better way of doing it. Among others, it makes it much easier to implement error analysis (not a target for 1.0, but something that shouldn’t be hard for a user to do, or for us to implement in 1.x).

- [x] Switch network/transition analysis to step-based (rate, crossing prob, conditional transition prob, flux)
- [x] Switch move scheme analysis to step-based (acceptance rate, selection frequency)
- [x] Switch replica exchange network analysis to step-based (mixing matrix, flow, etc.)


dwhswenson:

I've noticed some problems arising from insufficient sampling during tests. I think we want to keep these errors (we don't want to give the user bad numbers). Mainly, once the file format is settled, we'll use a single, well-converged files as a test for the analysis.

jhprinz:
> I've noticed some problems arising from insufficient sampling during tests. I think we want to keep these errors (we don't want to give the user bad numbers). Mainly, once the file format is settled, we'll use a single, well-converged files as a test for the analysis.

Yes, I would like not to fail these. Using a file that is converged is good as long as it is not too big.

PR #509, opened 2016-06-30
MISTIS option: `strict_sampling`

dwhswenson:
Adds a `strict_sampling` option when creating a `MISTISNetwork`.

The default (`strict_sampling=False`) MISTIS network samples transitions from the initial state to *any* final state. This is sometimes a source of confusion (although we think it is the better choice to make). If you want to reject paths that go to any state other than the defined final state, then use `strict_sampling=True`.

- [x] Implement `strict_sampling`
- [x] Tests for MISTIS network
- [x] Docstrings


dwhswenson:
Passes tests; ready for review!

jhprinz:
I remember, the thing that confuse Abel (and me)...

jhprinz:
looks, good. Not a big change, but very useful.

PR #510, opened 2016-07-01
Fix bug in storage string in `netcdfplus`

dwhswenson:
Really stupid typo of mine. It always reported as "GB" even when it meant "MB", "kB" or even just "B".

7-character fix for a bug @bolhuis caught. This should be ready to merge as soon as it passes.


jhprinz:
I thought I had fixed that in some other PR, but apparently not :) Good catch!

PR #511, opened 2016-07-01
Path density for alanine dipeptide TPS example

dwhswenson:
Fixes some problems in the path density histogram and adds a plot to the example notebook

- [x] Fix path histogram for sequential points in the same cell
- [x] Test for sequential points in same cell in path histogram
- [x] Simplify plotting of path density histograms
- [x] Add PathDensityHistogram to `alanine_dipeptide_tps_analysis.ipynb`
- [x] Tests and docstrings for plotting


dwhswenson:
Ready for review and merge! Left the outputs on the `AD_tps_analysis` clean, rather than cause more errors. But you can see the path densities in earlier versions, and I'll be adding one to the paper soon.

jhprinz:
The notebooks look good. Not sure if I completely got the HistogramPlotter, but seems to work as expected. Will merge.

PR #512, opened 2016-07-03
Simplify setup of MISTIS example

dwhswenson:
In the process of writing up the MISTIS example, it felt like the setup was hard to explain concisely. This PR makes the setup easier, and therefore easier to explain in the paper. Much of it can be used for MSTIS, as well.

- [x] Full bootstrapping can either take extra interfaces (volumes) as before, or extra ensembles
- [x] Give `initial_conditions_from_trajectories` an `avoid_reuse` flag (default `True`; current is equiv to `False`)
- [x] New `populate_minus_ensemble_from_set` to take a sample set to possibly extend from; check for prepend as well
- [x] Tests for new behavior
- [x] Docstrings for new stuff


dwhswenson:
Passed tests; ready for review! The compare before/after for assigning trajectories to ensembles in `mistis_setup.ipynb` (starting cell 17) to see where this makes a difference.

jhprinz:
Looks useful... Will merge and then fix the missing visualization.

PR #514, opened 2016-07-06
Easily handle assigning flux to network for analysis

dwhswenson:
Again, writing the example in the paper shows where there’s room for improvement. This PR makes it easier to use an external flux calculation for the MISTIS rate. The problem is that there needs to be a flux for every transition. The flux from a state and through an interface is, of course, uniquely identified by the tuple `(state, interface)`. The network receives a dictionary of `(state, interface): flux`, and assigns the flux to the appropriate transitions, using the new `network.set_fluxes(flux_dictionary)` function. 

This takes much less deep understanding on the part of the user than the previous approach. The code for the function is short, but to do the correct thing would require advanced understanding.

- [x] Add method to easily set flux for network analysis
- [x] Tests for setting flux for analysis
- [x] Use in MISTIS example


dwhswenson:
Passes. I'll call this part ready for review/merge, but I noticed that there was some strange behavior with moves running to max length allowed by the engine (unrelated to the goal of this PR, but worth knowing about). I'll check on that soon.

jhprinz:
Looks good. It seems to be a rather special use case.

PR #515, opened 2016-07-06
Fix versioning in `storage.py`

dwhswenson:
I want to play around a little bit with improving the stuff from #397. The hope is that, instead of having the version number hard-coded in `storage.py`, we can get Travis tests to manage the version install correctly.

Also, a question with this: @jhprinz, do we want to go ahead and bump up to 0.9, as @jchodera suggested? I'd say we're close enough to deserve that!


dwhswenson:
That seems to have fixed it. Actually, the strange bit is that adding a line that I don't think should have made a difference is what seemed to actually fixed it (the change in https://github.com/openpathsampling/openpathsampling/pull/515/commits/09afe5c583b38693b025a731b82c1e0a67c471cd, which should NOT have done it). But changes I made to `__init__.py` actually *should* fix for certain install patterns that otherwise should fail. (Namely, using `$PYTHONPATH` directly, which is a sin, but one that foolish people frequently commit.)

Question remains about bumping version to 0.9. @jhprinz, if you think we should do that, I'll make it part of this PR! Otherwise, this is ready for review.

jhprinz:
> Question remains about bumping version to 0.9. @jhprinz, if you think we should do that, I'll make it part of this PR! Otherwise, this is ready for review.

Yes, we should do that. Do you want to open a PR?

dwhswenson:
I can make it part of this one. :-)

dwhswenson:
Well, that passed -- and it says it installed 0.9.0! Do you notice anywhere that I might have missed a version? (did a grep of .py and .yaml for the current version number)

dwhswenson:
Also, @jhprinz, I think you'll see why the changes I made to `__init__` might have made a difference. But if you have any idea how 09afe5c passed, please let me know. After pushing that, I looked at the code again and said, "nevermind, that won't fix it" ... and then it did!

Definitely ready for review!

jhprinz:
Not sure if that explains it. But I am using `short_version` for comparison. So potentially some wrong `version.py` being imported. Still, looks good and will merge...

Issue #507, opened 2016-06-28
Error in alanine_dipeptide_mstis_bootstrap.ipynb


bolhuis:
When running this notebook error occurs when creating inter core trajectories at step/cell 40:

```pytb
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-40-79f854d66f52> in <module>()
      7         try:
      8             count += 1
----> 9             tt = engine_high.generate_forward(tt[-1].reversed, core2core)
     10             storage.save(tt)
     11 

/Users/bolhuis/git/openpathsampling/openpathsampling/engines/dynamics_engine.pyc in generate_forward(self, snapshot, ensemble)
    236         """
    237 
--> 238         return self.generate(snapshot, ensemble.can_append, direction=+1)
    239 
    240     def generate_backward(self, snapshot, ensemble):

/Users/bolhuis/git/openpathsampling/openpathsampling/engines/dynamics_engine.pyc in generate(self, snapshot, running, direction)
    311 
    312             # Do integrator x steps
--> 313             snapshot = self.generate_next_frame()
    314             frame += 1
    315             if frame % log_freq == 0:

/Users/bolhuis/git/openpathsampling/openpathsampling/engines/openmm/engine.pyc in generate_next_frame(self)
    241         self.simulation.step(self.nsteps_per_frame)
    242         self._current_snapshot = None
--> 243         return self.current_snapshot
    244 
    245     def minimize(self):

/Users/bolhuis/git/openpathsampling/openpathsampling/engines/dynamics_engine.pyc in __getattr__(self, item)
    164     def __getattr__(self, item):
    165         # default is to look for an option and return it's value
--> 166         return self.options[item]
    167 
    168     @property

KeyError: 'current_snapshot'
```

This was running after I did a conda update --all

dwhswenson:
FWIW, I just verified that I get the same error when I run this notebook. Labelling this as "bug".

jhprinz:
I will look into this. Note that there is a very strange thing with accessing attributes defined by properties in python:

If you access an _computed_ attribute that is defined using `@property` and a function then, if there is an error in the execution of the property function python by default returns a KeyError on the property and not the real error that occurred!

So you (or me) have to look inside the `current_snapshot` property funtion of the engine used and see why it cannot return the current snapshot. I presume there is something going wrong with OpenMM. Not sure

dwhswenson:
[...]

PR #516, opened 2016-07-08
New DynamicsEngine.__getattr__ 

dwhswenson:
See discussion in #507. This will not solve that issue, but it should make it easier to debug.

A few questions:

* In the event that the re-run works, do we `return result` or raise the `AttributeError`. Currently, I'm raising the error under the assumption that if something went bad, we can't trust the results.
* Do we give a special message for non-property, in-dictionary errors? I'm not entirely sure how to create a test for that scenario; currently it would give the same as an actual does-not-exist.

### DIFF
--- a/openpathsampling/engines/dynamics_engine.py
+++ b/openpathsampling/engines/dynamics_engine.py
@@ -173,7 +173,8 @@ class DynamicsEngine(StorableNamedObject):
                     # return result  # miraculously fixed
                     raise AttributeError(
                         "Unknown problem occurred in property" + 
-                        str(p.fget.func_name)
+                        str(p.fget.func_name) + ": Second attempt returned"
+                        + str(result)
                     )
             # for now, items in dict that fail with AttributeError will just
             # give the default message; to change, add something here like:
--- a/openpathsampling/engines/dynamics_engine.py
+++ b/openpathsampling/engines/dynamics_engine.py
@@ -75,10 +75,6 @@ class DynamicsEngine(StorableNamedObject):
         # Better would be a link to the topology directly. This is needed to create
         # mdtraj.Trajectory() objects
 
-        # TODO: Remove this and put the logic outside of the engine. The engine in trajectory is only
-        # used to get the solute indices which should depend on the topology anyway
-        # Trajectory.engine = self
-
         self._check_options(options)
 
         # as default set a newly generated engine as the default engine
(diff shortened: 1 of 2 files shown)


dwhswenson:

As stated above, there are two special cases where I don't think the behavior is totally obvious. Let me explain each in a little more detail.

* The approach is to calculate the property a second time, after it raised the `AttributeError`, and raise the specific internal error. But what happens if the second time, no error is raised? I think the two choices are (a) pretend no error was raised and return the result from the second time; (b) raise an error anyway, basically saying "something weird happened in this property, but I don't know what". I went with (b), based on the feeling that any result we get is no longer trustworthy (and this indicates some weird internal change of state while calling the getter, which is pretty suspicious).

* We check whether the `item` is in `__class__.__dict__` to see if it might be a property. What if `item` is in `__class__.__dict__`, but isn't a property? Well, to start, I have no idea how this raised an attribute error in the first place. I can't come up with a test that makes this situation possible. As such, I'm not sure what to do. The two choices here are (a) raise a special error for this case; (b) treat it as the normal case and check against the possibility that something of that name exists in `options`. I went with (b), mainly because I thought there might be some weird case where that was wanted.

Anyway, it is easy to change the behavior of either of those two special cases. I think distrusting the weird self-correcting result is the probably the right thing to do in the first one, but I really have no idea for the second.

jhprinz:
Nice solution. I am not sure whether the repeat execution and hope that the result is the best solution. I was supposed to be to get the real error, but after an error weird things can happen. I would always raise an error and return the working result in the message. Otherwise you might never realize there is an error occurring somewhere. That is safer.

> What if item is in __class__.__dict__, but isn't a property?

Well, I think that cannot happen unless you are calling a descriptor like property (at least not that I know of).  

> If the class also defines __getattr__(), the latter will not be called unless __getattribute__() either calls it explicitly or raises an AttributeError.

So, it can only happen, if the attribute implements a descriptor. We might check for existance of `__get__` instead of checking for a property. Would that answer your question?

dwhswenson:
>  I would always raise an error and return the working result in the message.

Added the successful result to the error message (already was raising error: I agree -- if things went bad once, we can't trust them any more).

> So, it can only happen, if the attribute implements a descriptor. We might check for existance of `__get__` instead of checking for a property. Would that answer your question?

Well, mainly I was trying to think of a case where this would occur. Outside of really sadistic stuff (`item` isn't in `__class__.__dict__` at initialization, but something earlier in the attribute lookup process adds it to the dictionary before raising an `AttributeError`), I just can't imagine how this could happen. It sounds like you agree that we should treat that case (which shouldn't be realistically possible) in the same way as if it wasn't in the dictionary at all, right? Basically, to give the default message.

Anyway, if that's agreed, then this should be ready to merge once it passes tests.

jhprinz:
Good enough for me. Agreed and will merge...

PR #517, opened 2016-07-11
New Storage implementation

### LINKED ISSUE #98: Splitting of netCDF4 files

### LINKED ISSUE #311: Topology as part of engine?

### LINKED ISSUE #444: Revisiting new `Snapshot` structure

### LINKED ISSUE #523: MdTrajCV / PyemmaCV need Topology object to be created
#### Problem

A CV that wants to use mdtraj (or pyemma) as the object from which to compute values need a mdtraj.Topology object to work. Currently the CV will check the first frame of the list of snapshots for  the topology object. An use this one. This is bad 

1. it has overhead from finding the topology object
2. a CV definition only makes sense with the correct CV
3. In rare cases this can break using shallow stored snapshots. Imagine you have all CVs precomputed and want to continue. Then you extend an existing trajectory and in theory need to compute on the added snapshots. The mdtraj CV however requests the first frame to look for the topology which might be shallow and hence empty. 

The easy and probably best way out is to just create a CV with the Topology. If you use always the same Topology then in storage it will only be saved once. So no problem here and the CV will not have to look for the appropriate Topology

```
cv = CV_MdTraj_Function('name', md.SomeFunction, engine.topology, ...)
```

jhprinz:
Closes #503. Closes #443. Closes #98.  Closes #311. Closes #444. Closes #523.

This is the final large PR that will implement UUID support. I am sorry for the large PR but all the storage stuff, CVs, handling of snapshots and features was so intermingled that there was no way to do that step by step. To review this I will write a little guide of the changes and how they will affect the changes.

#### Storage in general

- [x] A Storage is initialized without the need of a template. You can open it and store other things, like an engine, volumes, ensembles, without the need of a template snapshot. (Note, to create a disk cache of CVs you need a template to determine the shape of the CV automatically. By default the first stored snapshot `snapshots[0]` is used), so this works now `st = Storage('new_file.nc', 'w')
- [x] Improved caching of PathMoveChanges. This makes AnalysisCache for PMCs about 3 times faster when loading.
- [x] storage can be open in two modes, with and without UUID support. UUID means that _all_ objects will carry a 16byte UUID that contains the system and time openpathsampling was imported and a running number of the object created. This should be unique enough to identify all objects. You can run with UUID for toy systems to improve performance. But this is only significant for very small systems and toy dynamics.
- [x] UUID support is default, to disable it use `use_uuid=False` when creating a `Storage`
- [x] removed clone and clone_empty the first can be replaced by `cp` and the second should just open a new file and save what you want.

#### Storage of `Snapshot`

- [x] snapshots are stored by a general list of `Snapshot` holders. 
- [x] attach an arbitrary number of `Snapshot` types, so multi Snapshot support is there
- [x] autocreate a snapshot placeholder
- [x] Multi snapshot support includes empty snapshots. These are stored with a UUID or index, but no information like coordinates. This is nice to split files into trajectories and rest and reduce file sizes. Combined with `allow_partial=False` this can save a lot of diskspace if you want.
- [x] There is no `.template` anymore. The `.template` can be set by storing `st.tag['template'] = snapshot`
- [x] Improves the handling of proxy object. They now allow better comparison. Effectively two proxies pointing to the same object are now tested equal and hashing is fixed as well, so searching a proxy in a cache will find objects point to the same object. E.g. You create two proxies that point to snapshot 10. Call `cv(proxy1)` will cache it and `cv(proxy2)` will actually use the cache. This improves performance a lot.
- [x] The `st.snapshots` will auto-create one store for the first snapshots that is to be saved. If you want to store a second type which is different you need to manually add it using `cvs.add_type(template)`. This is default. You can override this behavior using `cvs._treat_missing_snapshot_type` which can be `['create', 'ignore', 'fail', 'single']`. Single is the described default - only one store is created automatically. `create` will create always a new one. `fail` will always fail with an exception and `ignore` will not create a store, not save the snapshot, but save it as an empty shell, so you can still save the CV values without the snapshots
- [x] Prevent storing of snapshots. You can set `st.snapshots.only_mention=True` and then snapshots are saved only as empty shells. CVs can be cached, but you cannot compute new CVs.

#### Engines

- [x] engines carry a SnapshotDescriptor that holds the Snapshot class, like MDSnapshot, ToySnapshot, etc... and a dict of dimensions like {`atom`: 1600, 'spatial':3} to create a suitable store and save snapshots. 
- [x] If an engine has a topology is up to the engine, it is not required
- [x] There exist a number of `NoEngine` that carry only the descriptor plus some necessary information, e.g. `TopologyEngine` carries the descriptor and a topology object. These `NoEngine`s are used to create snapshots from files without needing a real engine
- [x] Engines do not need a template. They can have arbitrary parameters. The `DynamicsEngine` requires a descriptor attribute and so all engine need to provide a `descriptor` attribute from which storage figures out how to store snapshots.

#### Snapshots

- [x] Snapshots reference an engine and not a topology! This is important since you need always an engine. You can use the `NoEngine`s to build snapshots from files, etc. This will usually change the order in which to setup an engine
- [x] There is a `SnapshotDescriptor` class that is immutable, i.e. a frozen set that contains the class and the dimensions needed to store a snapshot. Since it is a frozen set you can can simply compare two descriptors using `==. This also allows hashing and using it as a dict
- [x] Functions in analysis and visualization that use snapshots now work (almost exclusively) with proxies if possible. That way most function can be run, even if the actual snapshots are not stored.

#### CVs

- [x] rewriting code of ChainDict and CollectiveVariable
- [x] removed lots of functions not used anymore
- [x] Improve ChainDict and fix a bug that made `cv(snapshot)` about 10 times slower than `cv[snapshot]`. Sounds weird but one additional function call was too much. This is faster now.
- [x] CVs are now initialized only using option relevant for the CV, not storage.
- [x] There are options for disk cache handling that are needed if a CV is stored. These can be set using the `cv.with_diskcache(...)` function. Or after saving a CV using either `storage.cvs.add_diskcache(...)`
- [x] A CV can only be stored for snapshots that are mentioned in the storage. This is so avoid to save all sorts of intermediate snapshots that are never stored and hence never analyzed. Since snapshots can hold empty Snapshot objects you can save CVs without the actual Snapshot and run analysis without ever needing to load a single snapshot.
- [x] A CV diskcache is not handled inside the `snapshots` class for performance reasons.
- [x] A CV diskcache has option `chunksize`: the block size in which CV values are loaded into the cache. This greatly increases performance and make CVs much faster available even without precaching
- [x] A CV diskcache has option `allow_partial`: If false the CV is basically saved as a time_symmetric snapshot.feature. This is the most common case and this is fastest and you do not have to worry about syncing a cache. Everytime a snapshot is stored the CVs are autocompleted and stored as well. This is default. If `allow_partial` is `True` then the CV is saved as they are computed and synced. So it depends a lot on the order in which you compute CVs and save the snapshots
- [x] For `allow_partial=True` exists 2 functions to fill the diskcache. (Note that for complete this was not necessary). `storage.cvs.sync(cv)` will add all snapshots

### Misc

- [x] Some improvements on caching for speedup
- [x] Netcdf and Storage have been fitted into the 80 char width
- [x] Simplify iterators for some OPS objects

### Tests

- [x] Extensive testing of loading and saving CVs
- [x] Extensive testing of loadng and saving snapshots

### Examples

- [x] Update examples to new storage

## Speed

This is almost ready. I hope I caught everything that we discussed and it works as you like. Performace is about the same as before without UUID and adds about 50% for the small toy examples. So 500 steps rate matrix computation `3s` and with UUID `4.0 - 4.5s`. This is mostly slow by accessing all the steps and all the PMCs etc. CVs are not super fast, but with UUIDs almost as good as before.

## Questions

What to do about this and do you agree.

1. Remove `.template` and leave it up to the user to save a template. Or have a placeholder for that.
2. Default is with UUID support (which is slower)
2. CV storage: The default case is `time_symmetric` and `complete`. Is that okay?
3. non-symmetric CVs need to be stored with `allow_partial`. This makes them little slower, but I think that is okay. Having more options is confusing

### SnapshotStorage

There is a global list of UUIDs for all snapshots that is shared by CV diskcache and snapshots. This list also contains a pointer to a snapshot store the way we had it before. This way you can store multiple snapshots and mix these also in trajectories. You can also mix Empty Snapshots and multiple ones in Trajectories. Only if you access a snapshot directly it will be loaded. 

### CV diskcache

This works as a store. You can actually access the diskcache like a normal store and request objects. Only difference is that this store only accepts int indices and not snapshots. The index nedded is the one the snapshot has in the storage. This keeps it reasonably fast and you have to store the UUID -> index table only once.

### CV caching

A CV has a fast cache for objects and proxies in memory, if this does not work it tries the disk cache which uses chunks. If this is missing but in storage the chunk will be loaded an the values is returned. If this also fails the function is evoked to compute the values. Computed values will be passed back to the memory cache

### Comments on reviewing

I now that the internals are very complicated and probably very time consuming to be reviewed. So maybe you should suggest more tests and we want to pass and make sure that works. The files that have significantly changed are the stuff from `storage` and `netcdfplus`, but these affect all sorts of tests and notebookds as well. Let me know if I can help.


jhprinz:
Phew, this was a lot of work...

jhprinz:
Added a simple example on how to split into traj and rest and then analyze only the rest.

jhprinz:
That should be it...

jhprinz:
??? Coverage decreased by 10 %??? When I run it locally it increases by 2%...

So, this is ready, but fails due to coverage and I do not know why...

dwhswenson:
Run it again. That looks wrong to me. Still looks like a 2% decrease, but not 10%:

https://travis-ci.org/openpathsampling/openpathsampling/builds/144277430#L1884

Currently reads:

```
TOTAL                                                  10704   2874    73%
```

jhprinz:
I think I found it. I had used `--cover-package openpathsampling` because locally it started testing all sorts of other stuff. Apparently I should not do that here...

dwhswenson:
Wait. Did you add visualize.py to the `__init__`? Coveralls now sees that, and msm.py, and clustering.py. It sees them all as new files. See https://coveralls.io/jobs/16265516, click on "All"

dwhswenson:
> I had used --cover-package openpathsampling because locally it started testing all sorts of other stuff.

Aha... yes, that would explain it. You should use the .coveragerc that is located in the main directory. I symlink that into my tests directory in my local installs.

jhprinz:
little more cleanups and I removed the remote capabilities and will add this in a separate PR.

jhprinz:
I added by accident the fix for using topologies also here. I relates to what @dwhswenson and me discussed yesterday briefly about performance issue when using CVs that rely on mdtraj.Topology objects.

dwhswenson:
Given the size of the PR, there’s no way that I can actually review these changes to the code in anything resembling a reasonable amount of time. So I’m just going to review what the top post says was done, and trust that the code actually implements that (I feel confident trusting you on that!)

Most of this just boils down to “needs more explanation” -- some questions to be answered in this PR, some examples needed in the notebooks/tests to show how things work.

### storage in general

All of this sounds very good to me. I like that it is easier to initial storage, and that simpler initialization made the `clone`/`clone_empty` irrelevant. UUIDs are a good idea, and faster PMC loading is great!

### storage of `Snapshot`

Again, I like most of this. I think we need more documentation giving examples of how this works (example notebooks, also nosetests for the expected behaviors: some of those seem covered, some not).

At the very least, please write example notebooks for some of the options (`allow_partial`, `only_mention`) before merge. That makes documenting it easier.

Can you clarify the difference between `_treat_missing_snapshot_type` as `create` vs. `single`? It is unclear to me what `single` does, especially if the snapshot types have different dimensions or different features.

As mentioned in person, I think `only_mention` could have a more explicit name. Maybe `save_as_proxy` or `only_save_proxy`, which is what I understand it to mean.

### Engines

I assume it is also easy to create a `SnapshotDescriptor` from a template snapshot? (since this is effectively what you were doing before) Or from a `topology` object for the engine?

### Snapshots

All of this sounds good to me.

### CVs

> Improve ChainDict and fix a bug that made cv(snapshot) about 10 times slower thancv[snapshot]. Sounds weird but one additional function call was too much. This is faster now.

Whoa! Nice!

Can you explain what you mean by `diskcache`? How is this different from the cache we had before? (Wait… you do that later in the post. Never mind. Although it is still a little confusing. Does this just mean that CVs have an extra cache layer? First check memory cache, then check this "disk cache,” then check saved file? Why is the disk cache called disk cache?)

`allow_partial` might take a better name. Would this be equivalent to `allow_without_reverse_partner`? (that’s a bit too long of a name, but more clear.)

### Misc

Woohoo! I can read it on my 80 char terminal without going crazy!

### Tests

Soon, it would be good to move the tests from ipynb to nosetests.

## (Your) Questions

1. *Remove template*: yes, agreed
2. *Default UUID*: yes, agreed
3. *Default CV is `time_symmetric` and `complete`*: Disagree. Default behavior should be what is always correct. Smart users will figure out how to make it faster; dumb users wouldn’t figure out that they made a mistake. Let’s prevent the dumb users from having something to blame on us.
4. *non-symmetric CVs must have `allow_partial`*: Agree. I don’t think there’s a clean way around this.

jhprinz:
> At the very least, please write example notebooks for some of the options (allow_partial, only_mention) before merge. That makes documenting it easier.

Will do. Probably end of the week...

> Can you clarify the difference between _treat_missing_snapshot_type as create vs. single? It is unclear to me what single does, especially if the snapshot types have different dimensions or different features.

Should also put this into the documentation. In brief here:

The storage can automatically handle different snapshots. A unique _storable_ snapshot type is described by a `SnapshotDescriptor`. This is basically a dictionary containing 

1. dimensions like `{ 'atom' : 1651, 'spatial':3 }. The ones necessary to create the netCDF arrays in the store and
2. the snapshot class which knows all the features.

So, if you want to store snapshots that differ in one of these, they need to be initialized in the storage to contain the specific netCDF variables to hold the data. 

Now to answer the question. If you create a storage it has no container for snapshots. You can add these by running 

```
storage.snapshots.add_type(snapshot_descriptor or template)
```

instead of doing it by hand there is a default setting what should happen when you try to store a snapshot _type_ that is not yet storable. You can either use

1. `single`: This wil create space for the first snapshot you save. So if only one type exists this will automatically allow storing without any `add_type`. If you want to store a second type it will throw an error. Since auto creation will only create a single one. If you had created one by hand then it will also not create a second one.

2. `create`: This will *always* create a new container for every new type saved.

3. `ignore`: Will not create a new container but store the snapshot only by reference (which only works for UUIDs). This means you can save CVs for that snapshots and store trajectories, but the coordinates, velocities, etc will not be available

4. `fail`: This will also never create a container and throw an error when you try to save a new snapshot type.

Default is `single` which I think makes sense.

> Can you explain what you mean by diskcache? 

Yes. This was a difficult part. To simplify the chaindict, etc I use the following _fallback_ scheme
for a CV

```
input 
  v 
fast_memory_cache (weak)
  v
[ disk_memory_cache > load from disk_store if necessary (in chunks) ] optional
  v
compute using function
  v
None
```

So, try first simple memory cache that references CVs for snapshots or proxies that exist still in memory. If that does not work you check the disk_memory_cache if it exists. This is only the case if you habe stored stuff on disk. If an element is not in the disk_cache it tries to load the chunk that contains that value. If that does not work either or the disk_cache does not exist it will run the function to compute the value. If that fails (if there is no function) it will return None.

Syncing with the diskcache works wither automatic, when you use _complete_ mode and you save a snapshot or you need to run either

1. `cv.sync(store)` to add all values in the fast_memory_cache that are missing to the disk_store
2. `cv.complete(store)` to store all missing values in the disk_store for all snapshots that are reference in the store. This might require reloading lots of snapshots and should be slow, but can be useful.

> allow_partial might take a better name. Would this be equivalent to allow_without_reverse_partner? (that’s a bit too long of a name, but more clear.)

Actually, `allow_partial=False` refers to the mode of storing and 

1. requires no reverse partner or to be symmetric w.r.t. time-reversal and 
2. that it will be auto-completed when a snapshot is stored.

`allow_partial=True` means

1. a value for each snapshot will be stored, so `s` and `s.reversed` are counted as two different snapshots. 
2. values are stored in the order as they come and
3. are not auto-completed. You need to use `sync` or `complete`


> Default CV is time_symmetric and complete: Disagree. Default behavior should be what is always correct. Smart users will figure out how to make it faster; dumb users wouldn’t figure out that they made a mistake. Let’s prevent the dumb users from having something to blame on us.

Well, okay, then I propose that we use `complete` if time_symmetric is True, and default is `time_symmetric=False` but a user only needs to set `time_symmetric` if he also sets `allow_partial/complete` this will be overridden. We might also make time_symmetric a requirement for CVs to force the user to set this.

jhprinz:
i will also fix the error in travis ASAP.

jhprinz:
tests should run now... test are coming...

dwhswenson:
First, checking how close this is to done -- I see you've added some examples. Have you made the changes about defaults with `time_symmetric`?

> Well, okay, then I propose that we use `complete` if time_symmetric is True, and default is `time_symmetric=False` but a user only needs to set `time_symmetric` if he also sets `allow_partial/complete` this will be overridden. 

This sounds good to me.

> We might also make time_symmetric a requirement for CVs to force the user to set this.

Not sure what you mean here ... as in it should be a required argument when making the CV?

Another option (although I don't like adding lots of extra classes) would be to make a `CoordinateFunctionCV`. This is overwhelmingly the most likely thing to be used (for custom CVs, at least) and will always be `time_symmetric`. All this complication about possibly having a CV not be `time_symmetric` only happens if you use something other than coordinates in your CV function, and that's pretty rare. It has been done -- the very first TIS paper defined states in terms of coordinates and velocities (well, IIRC, kinetic energies: so that takes us back to velocity-dependent but time-symmetric). But in that case, it was done because it was a small system running NVE, and the recrossing probability is quite high.

That should be as easy as just writing a subclass of the main `CV_Function` which doesn't have `time_symmetric` as an `__init__` option (and instead passes `time_symmetric=True` to `super`).

jhprinz:
> Not sure what you mean here ... as in it should be a required argument when making the CV?

Yes, that was the idea. I used your idea to have a special `CoordinateFunctionCV` and `CoordinateGeneratorCV`. Btw we should rename the CV classes to `NameCV` and not `CV_Name`.

dwhswenson:
> `CoordinateGeneratorCV`

What does this do? (differently from `Function` version)

> Btw we should rename the CV classes to `NameCV` and not `CV_Name`.

Works for me. I see the addition has already been made on #527.

jhprinz:
> What does this do? (differently from Function version)

Nothing, just uses `cv_time_symmetric=True` automatically, should be the one to be used, when you write a function that only depends on coordinates... It will also handle setting up the default storage type. I think it is easier to establish using this class than saying. We recommend not using the default but always set `cv_time_symmetric=True`. There is also a `CV_CoordinateGenerator` version for class instances like it is done with `MSMBuilder` or `PyEMMA`.

jhprinz:
So, fixed the tests and notebooks to reflect the changes. Default is now `time_symmeric=False` but you can use `CV_CoordinateFunction` which uses `time_symmetric=True`.

Anything missing?

dwhswenson:
> There is also a `CV_CoordinateGenerator` version for class instances like it is done with `MSMBuilder` or `PyEMMA`.

Ah, this was the source of my question. Wasn't sure what the "generator" part meant.

Your responses cleared up all of my questions. As said before, I'm not going to review in detail. 

I hope you'll make some of your ipynb tests of storage and related into nosetests, so we can count them in coverage (even with the nosetests you did include, this PR is still dropping coverage by 0.6%: that's on the order of 65 semantic lines of code missing coverage). But that's for later.

The only other concern I have is about the name of `allow_partial`. But I'll add that to #527.

With that: MERGING! (Finally!!)

PR #518, opened 2016-07-11
[WIP] A simple MSM engine

jhprinz:
Useful to simulate OPS given an MSM matrix. There is a notebook that explains it. 

Still experimental


jchodera:
Awesome, thanks!

jhprinz:
Little updates. It works in memory now and has some helpers that make it faster...

dwhswenson:
Adding a comment from our Skype call yesterday: this PR will be split into two parts, one that really does the simple MSM engine, and which will be part of the 1.1 "engines" release, and one that does the fancier OOM stuff, which is more experimental and will be for a later release.

PR #521, opened 2016-07-15
Update Alanine MSTIS example

jhprinz:
This updates the Alanine MSTIS example for the paper. I made the following changes.

- Remove `alpha` and `beta`. This only uses the 6 states `A`-`F`
- Added some parameters to `initial_conditions_from_trajectories`, see below
- New `Ensemble.split` now uses an iterator instead of `n_results`. This makes it easier to keep looking for subtrajectories while doing analysis on the subtrajectories.
- New `DynamicsEngine.extend_forward(trajectory, ensemble)` and `DynamicsEngine.extend_backward(trajectory, ensemble)` which will simply extending an existing trajectory into a new ensemble.
- moved the alanine plots into separate class in `alatools.py`
- New `MSTISNetwork.generate_inter_state_samples(snapshot, engine)` that will create samples that bounce between states until all states are hit. The set of samples will be returned as a list of samples.
- `PathSampling` will not duplicate the initial sampleset when initialized. This would prevent a restart because the connection to the old simulation is broken.
- Alanine bootstrapping does not use bootstrapping anymore. It uses `MSTISNetwork.generate_inter_state_samples` with the mstis network and then runs `SampleSet.generate_from_samples` on the `mstis.all_ensemble` with the high temperature engine. This will fill all ensembles and return the initial sampleset. 
- added a generator to generate snapshots, `engine.iter_generate` that will yield the current trajectory in regular intervals. Useful if you generate a long trajectory and want to keep track of progress.

The example already runs using UUIDs. Question is how many samples we want to generate. I think starting with 5000? 1000 are about 1GB.


jhprinz:
@dwhswenson @bolhuis: Trying to fix the `MSOuterInterface` but I am not sure how this fits into Weinas existing example. Did she also remove her outermost interface and used it for the outer? I don't think so. And If not, can I use the same lambda as the outermost interface? If not, how to I get another lambda? I could just take the outermost `+5` or so?

My feeling is that this is quite complicated. How difficult is it to use the proposed state hopping?

We can of course run it without the MSOuter...

dwhswenson:
MS-outer was developed for replica exchange. My understanding is that Weina's first alanine dipeptide paper (2011, the one with Kristen Marino on it) did not include replica exchange. Weina's second AD paper (2013) was single-replica, and used the state-swap.

To get the state swap stuff working would take roughly a week of work (requires new movers, and new handling of move strategies). Also, I would want to check a few things with @bolhuis, who is on holiday until August 15. So it probably wouldn't get done until August 19 or so.

dwhswenson:
> And If not, can I use the same lambda as the outermost interface? If not, how to I get another lambda? I could just take the outermost +5 or so?

You can do it however you want. You can put another copy of the interface at the same place as the previous. You can put it a little further out. All that matters is getting decent replica exchange into the MS outer ensemble and across to other states. Where to place it is a matter of optimization.

jhprinz:
Thanks, I added the `+5` to shift it a little outwards, we will see what happens.

jhprinz:
A quick summary of the changes to the `initial_conditions_from_trajectories`.

- I move most of the logic to `SampleSet` so that you can in general check a simplest for working with an ensemble definition like reported from schemes. `MoveScheme` will use these functions
- `Ensemble`s have three functions to generate a sample from trajectories 

    1. `get_sample_from_trajectories` : just look in trajectories for a valid sample
    2. `split_sample_from_trajectories` : split trajectories into valid samples and pick one
    3. `extend_sample_from_trajectories` : try to extend suitable subtrajectories into a valid sample, like `populate_minus_interface` did

    There are fallback implementations, but you can override these in a class. This makes most sense for extending and currently this works for `MinusInterface` and `TIS`.

- Extend has two modes. `minimal` and `complex` Complex uses a complex sub ensemble currently `A-X-A` like the segment ensemble and `minimal` uses a simple crossing `A-X`. Both of them are useful or even necessary to create the initial samples from a long trajectory like in our case. The last state visited will not have a segment ensemble but a `A-X` one to extend from.

- All the three sample generating function have specific options. Some important are `used_trajectories` which specify trajectories to be treated as already used and 
- `reuse_strategy` which specified how to act if the same trajectory is picked. `once` will not allow using a trajectory twice`, `avoid` will first look for alternatives but then use the used ones. `all` will ignore used ones. `once-symmetric` will also ignore the reversed copy` and `avoid-symmetric` will also avoid the reversed copy of a trajectory. This is the default.
- `unique` (maybe a better name) allows to pick the means to pick a sample if several would be possible. Choices are `first` picks the first possible, `shortest` picks the shortest of all `median` pick the median length wise `longest` picks the longest. For TIS I think `shortest` would be suitable to pick more likely ones, for Minus `median` makes sense to pick a more likely one.
- `scheme.initial_conditions_from_trajectories` will run a `strategy` that basically consist of a list of function calls to be done on all ensembles to get new samples. The default is 

    1. run `get_sample_from_trajectories`
    2. run `split_sample_from_trajectories`

    with default options. there are two other strategies to be used `extend-complex` and `extend-minimal`. You can give a list of strategies like `['get', 'split', 'extend-complex']. If you use a tuple then the second part contains options to be used. An example for MSTIS and using a long trajectory is:

    ```py
    strategies = [
        # 1. split and pick shortest and exclude for MinusInterfaceEnsembles
        ('split', {'unique': 'shortest', 'exclude': paths.MinusInterfaceEnsemble}),  
        # 2. split and pick median length
        ('split', {'unique': 'median'}),
        # 3. extend-complex (implemented for minus with segments A-X-A)
        'extend-complex',
        # 4. extend-minimal (implemented for minus and tis with crossings A-X)
        'extend-minimal']
    ```

- the `initial_con...` will run one strategy after another. It will use all trajectories fw and bw.

- the main idea is that you can design a suitable strategy, but you can also use the specific functions yourself. 


I think we should merge this and can later discuss, if we want the magic function in the example with longer explanation or do it the hard way or a mix of both.

jhprinz:
I won't finish this. The cleaning up of the sample generation will need more notebook updates. I finish this tomorrow...

jhprinz:
This should work now, but conda seems to be down. Packages cannot be installed somehow... Will retry later.

dwhswenson:
> conda seems to be down

Yeah, I've seen a lot of that recently. Down or very, very slow to transfer.

jhprinz:
last comment, the `populate_minus` is replaced by the general `extend_sample_from_trajectories` which works pratically the same. `minus_replica=` is named `replica=` and you pass `trajectories=` instead of a simplest, etc.

jhprinz:
Wrong button ...

jhprinz:
This keeps on failing at mistis. No idea how to fix that. Have been restarting PRs all day...
Is it possible that there is something weird going on? There should be lots of overlap.

I also need to understand why the max CV is necessary and not the ensemble itself. Anyway... More tomorrow.

dwhswenson:
> This keeps on failing at mistis. No idea how to fix that.

You're timing out. That's the problem currently.

jhprinz:
I added 2 cells like
```
mstis_calc.run(200)
```
so that it will not time out, but make some more steps. That might solve this in my case.

jhprinz:
This worked. Thanks @dwhswenson . You can have a look now. Basically improves ways to generate samples and adds some convenience to engines like the iterator and a `extend_forward`, `extend_backward` to extend an existing trajectory into an ensemble.

openpathsampling/tests/testensemble.py
```diff
@@ -2325,8 +2325,8 @@ def test_populate_minus_ensemble_from_set(self):
         predestined_snaps = [trajB[-1]]+ttraj['upper_out_in']
         predestined_traj = [s.xyz[0][0] for s in predestined_snaps]
         engine = CalvinistDynamics(predestined_traj)
-        sample = self.minus_nl2.populate_minus_ensemble_from_set(
-            samples=sset, minus_replica_id=-1, engine=engine
+        sample = self.minus_nl2.extend_sample_from_trajectories(
+            sset, replica=-1, engine=engine
```
dwhswenson: change the name of this test when you change the name of the function

jhprinz: good point...

dwhswenson:
When you change the tests in `testMoveScheme. test_initial_conditions_from_trajectory`, you're changing the intent of the function. When doing `avoid_reuse=False` it is supposed to return the *first* from the list that satisfies the ensemble, and never in the reversed form. Can you make a way to call it so that the original test results do not change?

jhprinz:
> When you change the tests in testMoveScheme. test_initial_conditions_from_trajectory, you're changing the intent of the function. When doing avoid_reuse=False it is supposed to return the first from the list that satisfies the ensemble, and never in the reversed form. Can you make a way to call it so that the original test results do not change?

Not totally sure if I understand your concern. Do you want an option to not sort the input list of trajectories and automatically use the reversed one? That is what the function currently does. 
I can easily add that.

dwhswenson:
> Not totally sure if I understand your concern.

The results I intended were:

1. Take the first trajectory that is in the list that satisfies the ensemble. If avoiding re-use, take the first that hasn't already been taken.

2. Only if none satisfy the ensemble, take the first that, reversed, satisfies the ensemble. If avoiding re-use, take the first that hasn't already been used.

That's what the tests checked. Changes in the tests indicate to me that this is no longer the case. In particular, line 537 changes `traj2` (first that satisfies) to `traj3`. And lines 555-559 say that you can't even know whether `traj3` or its reverse will be used.

jhprinz:
Correct. I changed the behavior of the function and changed the tests accordingly. The _debatable_ strategy is now:

1. Sort the list of input trajectories by shortest first, then
2. create a list of trajectories as [shortest, shortest.reversed, 2nd shortest, 2nd shortest reversed, ...]
3. Use this as the list in the rest

This behavior is checked. I am not totally sure if we should force this sorting, etc. Of leave it optional. It makes sense in lots of cases, but probably not always. 

If you agree, I add something like `pre_process=['sort', 'reversed']` (need to think about it) to make this behaviour optional. This way the user can always set this the way he wants to.

dwhswenson:
Yeah, my intent is that the user may have given the trajectories in a specific order. This allows you to very easily fill new initial conditions (with equivalent but non-identical ensembles) using an old sample set. This is also why it was extremely important to have the ensembles in a canonical order.

jhprinz:
That totally makes sense. I have added an option to change the initial operations on the trajectories given. This is mostly for convenience an things the user can also do himself. Default is the add the reversed trajectories to the ones given in the same order. You can (optional) sort by shortest first, or not to the adding of mirrored trajectories. That will reproduce your original function. I added much more tests to make sure all of these option do as they claim.

jhprinz:
btw. Did the coverage disappear?

dwhswenson:
> btw. Did the coverage disappear?

??? What do you mean?

dwhswenson:
I think the `initial_conditions_from_trajectories` stuff looks reasonable now. It's a little hard to tell, since you completely rewrote those tests while also changing the code. But I trust that there's now a way to do what I originally intended that function to do.

Other points, just going off the top-post:

> * New `DynamicsEngine.extend_forward(trajectory, ensemble)` and `DynamicsEngine.extend_backward(trajectory, ensemble)` which will simply extending an existing trajectory into a new ensemble.

Are these really necessary? Keep in mind that `engines` should be usable without `ensemble`s (thinking post-2.0 split). Also, these don't do what you say. They run until `can_append`/`can_prepend` fails. The resulting trajectory is not necessarily "in the `ensemble`." I think making the user explicitly state that they want `ensemble.can_append` will remind them of that.

> * New `MSTISNetwork.generate_inter_state_samples(snapshot, engine)` that will create samples that bounce between states until all states are hit. The set of samples will be returned as a list of samples.
> * Alanine bootstrapping does not use bootstrapping anymore. It uses `MSTISNetwork.generate_inter_state_samples` with the mstis network and then runs `SampleSet.generate_from_samples` on the `mstis.all_ensemble` with the high temperature engine. This will fill all ensembles and return the initial sampleset.

These two aren't reflected in the current PR. Also, I don't think that `generate_inter_state_samples` should be a method of MSTIS. Make it a separate function, if you're going to make it. Otherwise we end up with dozens of methods for generating initial conditions attached to each network. (`MSTIS.generate_samples_from_metadynamics`, `MSTIS.generate_samples_from_neb`, etc., etc.) Having all of that show up in the documentation of each network will make the documentation very hard to understand. If we keep them aside, maybe in a module for `obtain_initial_conditions`, it will be easier on users.

> * added a generator to generate snapshots, engine.iter_generate that will yield the current trajectory in regular intervals. Useful if you generate a long trajectory and want to keep track of progress.

Will this also make it possible to store along the way? I was thinking about the possibility that a combination of large system, long trajectory, and small RAM could cause problems (e.g., minus moves).

jhprinz:
> But I trust that there's now a way to do what I originally intended that function to do.

It is not obious, but your original tests are still in there, plus more for the other common cases. To do that I turned your test into a function.

jhprinz:
> Are these really necessary? Keep in mind that engines should be usable without ensembles (thinking post-2.0 split). Also, these don't do what you say. They run until can_append/can_prepend fails. The resulting trajectory is not necessarily "in the ensemble." I think making the user explicitly state that they want ensemble.can_append will remind them of that.

Valid objection. They are not necessary, but are very useful to write cleaner code and avoid using the Prefix/Suffix...Ensemble, although it does the same. 

Correct, engines should not know about ensembles. I thought this is a nice short-cut to keep, but it is only used once. I will remove all the functions in engine that use ensembles and add that functionality as another way to generate a sample from a trajectory. Basically it is `extend_sample_from_trajectory` in the simplest possible way. 

> These two aren't reflected in the current PR.

These were old code that I did not use, but might be interesting. I can move it to a tools file where we collect useful code and decide later if we delete it. It is indeed not used. It will generate trajectories that bounce between the states and avoid running minus. I think that is even faster than running one long trajectory so it could be part of a more specific initial generation process. Or be an example on another way to get initial samples.

>Will this also make it possible to store along the way? I was thinking about the possibility that a combination of large system, long trajectory, and small RAM could cause problems (e.g., minus moves).

Yes, that is possible. Although it requires you to be careful about what you store. The trajectory object will remain the same to keep the ensemble caching working. If you would store an incomplete trajectory it would be remembered as stored and you cannot store it again.

What you have to do it something like this:
```py
#save snapshots of incomplete trajectory
for idx, snap in enumerate(income_traj.iter_proxies()):
    # you could loop only over new ones
    storage.snapshots.save(snap)
    if isinstance(snap, paths.BaseSnapshot):
        # replace by proxy to release memory
        income_traj[idx] = storage.snapshots.proxy(snap)
```

We could writea little function to store only the snapshots.

jhprinz:
ready to merge (once it passes tests)

jhprinz:
After that we can run a long simulation, then reduce it and run analysis for the paper.

dwhswenson:
> I can move it to a tools file where we collect useful code and decide later if we delete it. It is indeed not used. It will generate trajectories that bounce between the states and avoid running minus.

I think it is a good idea to have these in a separate file. Things like this are also what @bolhuis has suggested as targets for the ESDW in Traunkirchen, so having it in a partly completed state where it can be cleaned up/improved there is an option to consider.

> ready to merge (once it passes tests)

I still need to properly review the changes to `ensemble.py`: they are significant enough that they don't show in GitHub, so I need to do a local checkout. I assume most of them are related to the extension parts of `initial_conditions_from_trajectories`? (essentially, generalizing what was previously `populate_minus_ensemble`?)

dwhswenson:
It turns out that the reason I couldn't view `ensemble.py` in GitHub is because your IDE is aggressively rewriting code to fit its desired coding style. Please turn this feature off. It makes it very hard to review the code: there are 317 changes in `ensemble.py`; only a very small fraction of them are actually related to meaningful differences, but I have to look through all 317 to see what has changed.

It also makes our code worse/causes problems we'll have to fix later. Here are a few of the problems:

* You're renaming lots of variables. That is annoying, and can make things less clear. We typically refer to states/volumes with capital letters, so renaming `inA` to `in_a` makes the variable name less clear. `in_A` would be fine, but I don't see why you need to rename all variables anyway.
* Your IDE is wrapping lines in docstrings in such a way that they break the docs. For example, you wrapped the docstring of `TISEnsemble` as follows:
```
    Attributes
    ----------
    initial_states : `openpathsampling.volume.Volume` or
    list of `openpathsampling.volume.Volume`
        Volume(s) that only the first or last frame may be in
    final_states : `openpathsampling.volume.Volume` or
    list of `openpathsampling.volume.Volume`
        Volume(s) that only the last frame may be in
```

This makes the docs look like this:
image

* I don't like this `replica_sign` thing. I see that you're trying to reproduce what I'd done before, and I liked giving replicas that start in the minus ensembles a negative replica ID. But honestly, that's not worth the extra overhead it creates when you generalize and simplify. The replica ID is just a number; the value of that number doesn't have any real significance. You can figure out the initial ensemble for each replica with `{s.replica: s.ensemble for s in storage.steps[0].active}`, so trying to define that with a number isn't important.
* `MinusEnsemble.extendable_sub_ensembles`: Please don't just copy-paste from `__init__`!!! You're redefining `self._segment_ensemble` every time you call this property!
* I'm still a little uncertain about this whole `extendable_subensembles` thing. I'll merge it (after you fix these things), but I don't see how this can ever be a truly general approach. It already has to be redefined for every ensemble, and a custom ensemble would need a whole custom ***class*** (i.e., not just an instance of an arbitrary `SequentialEnsemble`) to make it work.

jhprinz:
> I don't like this replica_sign thing

Agreed, me neither. Will remove it...

> but I don't see how this can ever be a truly general approach.

Why do you try? This is not meant to be the most general possible something. It merely means that in for some movers it makes sense to define feasible sub ensemble to crop to and then try to extend. I think this is very general and it does not have to be for all ensembles. It is also not required to implement these. I think for now this is the most elegant _compromise considering available time and what we want to achieve_  to handle this and if we come up with a better one we can implement it or change in 1.5 or 2.0

> Your IDE is wrapping lines in docstrings

No, that was me :) trying to enforce your 80 character limit and I thought that this is the way in .rst files to correctly wrap long first lines. But I think there needs to be a wrapping char like a backslash or so. I never checked and wanted to sort this out when polishing docs.

jhprinz:
> MinusEnsemble.extendable_sub_ensembles: Please don't just copy-paste from __init__!!! You're redefining self._segment_ensemble every time you call this property!

There was a purpose to it (if you see the commented code). I initially wanted to make all sub ensembles with less loops to be subensembles, too. If you want a `A-X-A-X-A-X-A` then you could first search for `A-X-A-X-A` and then for the segment `A-X-A`. For the current implementation you are right, I could and probably should do that. But implying that I just copy code... :)

jhprinz:
> MinusEnsemble.extendable_sub_ensembles: Please don't just copy-paste from init!!! You're redefining self._segment_ensemble every time you call this property!

Now I understand. I missed removing that one line. Why don't you just say so? I got totally confused and looking for something else :)

jhprinz:
All done, except the renaming? I was not sure. Do you want to rename these back? And if so `in_A` or `inA`. I can do that and am fine with both.

dwhswenson:
> Now I understand. I missed removing that one line. Why don't you just say so? 

Sorry -- it was the combination of that line, and the fact that most of the rest of that section of code was unused, that frustrated me. Now that it is commented out as a "for future" thing, it makes more sense.

> Do you want to rename these back? And if so `in_A` or `inA`. I can do that and am fine with both.

Please do -- I think `in_A` is best. The capital matters; adding the underscore is more PEP8-like, so better.

jhprinz:
There was still a special case in a test I did not account for where the actual order in which ensembles are filled mattered in the `initial_conditions...` test

jhprinz:
This is ready, once it passes. I reverted the `in_AX/out_AX/leave_AX` back

jhprinz:
pinging @dwhswenson to note that this should be ready.

dwhswenson:
All looks good. Re-running tests to get a pass (on principle), then I'll merge.

Are we stable enough on file format to create a (reasonably) converged file to use as tests, instead of using the output from our quick integration tests? I don't see anything else in my near-future plans that will affect that. @jhprinz?

dwhswenson:
Oh, actually, looking at output -- one thing still to change in here.

It appears that there's a bunch of output coming from  `sample.py` line 549 (code at https://github.com/openpathsampling/openpathsampling/pull/521/files#diff-dd0830ca47c09b861d699a1bf6b49d8dR549). Maybe this could be info-level logging? Or add an option to `refresh_output`: `ipynb_only=False`, set it `True` for this case? Just so there's not a bunch of text in the output of the nosetests.

dwhswenson:
This looks good to me now. You're done making those last modifications? (That is, I can merge it now?)

jhprinz:
Yes, ready to merge. I updated the output function as you suggest, perhapsthe way to determine if it is running in ipynb. But it works.

dwhswenson:
Merging!

PR #528, opened 2016-07-31
InterfaceSet

dwhswenson:
[JHP and I already discussed this, and he's seen most of this text in an email. But I'm putting it here for transparency]

### Problem

* Information about the location ($\lambda_i$) of the interface can't get into our analysis. This makes a lot of things hard. For example, I rely on a hack-y approach to identify the outermost interface lambda value when calculating the rate (problems that Abel had), and I'm finding the similar issues in trying to set up a bias for SRTIS.
* Note that not all interfaces setups will have a valid version of $\lambda_i$: if both max and min values change, there isn't a clear $\lambda_i$ associated with the interface. So there should be a way to fail gracefully if trying to do something based on that.

### Proposed solution

* `InterfaceSet` object contains volumes, lambdas, and CV
* `interface_set.lambdas` is `None` if no clear $\lambda_i$ can be associated: anything that tries to use a specific lambda value should know enough to check that the return isn't `None`, and either have a fallback (e.g., the current version of calculating the outermost interface can be a fallback, but `interface_set.get_lambda(interface_set[-1])` is much more direct. For the SRTIS stuff, if you don't have the `interface_set.lambdas`, you'll have to manually create the bias (instead of doing it automatically if you do have the lambdas).
* Shortcuts for the common examples, so the `VolumeInterfaceSet` command just replaces the current use of `paths.VolumeFactory.CVRangeVolumeSet` (literally with no change, even in arguments). However, the user can directly build the `InterfaceSet` object as well, just as he or she could have made a custom list of volumes. Getting all the features would require adding the CV, and the lambda values for each interface.
* `InterfaceSet` acts like a list of volumes, but also carries some additional metadata for context

### Bonus

* Removes need for explicit inclusion of order parameter when building network objects (need InterfaceSet instead of list of volumes).

- [x] Draft InterfaceSet objects
- [x] Tests for InterfaceSet objects
- [x] Docstrings
- [x] Replace volume lists with InterfaceSets in tests/examples
- [x] Replace network input with interface sets


dwhswenson:

I did not update the AD MSTIS examples, mainly because I didn't want to conflict with #521. It should be easy to fix. All you need to do:

* instead of using a list of volumes for the interfaces, use a `VolumeInterfaceSet`. This just takes the CV, the minimum edge (either 0.0 or -inf for that example), and the list of interface lambda values.
* don't include the CV when defining the network

I'll also update the paper soon to reflect these changes (possibly after the follow-up PR to this, which I'll open tonight or tomorrow, about the MS-outer interface stuff).

jhprinz:
Looks good. Is the way we discussed it already. Will merge and update the other PRs accordingly

PR #530, opened 2016-08-02
New handling of MS-outer interface

dwhswenson:
As discussed with @jhprinz, I’m modifying the way that we create the MS outer interface. Previously, we automatically created it. But that led to some weird behavior (see, for example, the way it was unclear whether the last interface listed is MS-outer or not in the MISTIS example.) Plus, we think we’re going to move away from using MS-outer interface in the future (better to use state-swap moves; same decorrelation with less sampling effort).

The basic idea extends #528. Now we define the MS-outer in a separate step, as

```python
ms_outer = MSOuterInterface(interface_sets, volumes, lambdas)
```

(with lambdas optional) or, if the `interface_sets` have a `.new_interface(lambda_i)` function:

```python
ms_outer = MSOuterInterface.from_lambdas(interface_sets_lambdas)
```

where `interface_sets_lambdas` is a dictionary of the interface set to the lambda value for its part of the MS-outer ensemble. (I think this second version will be more commonly used; it reads cleaner.)

Then, instead of being built immediately from the input interfaces for the `network`, each of `MSTISNetwork` and `MISTISNetwork` will have an additional option (after the transition info tuple they start with now) called `ms_outers`, which is a list of these `MSOuterInterface` objects (MISTIS can have multiple MS-outer interfaces).

Overall, significance is this:
* The way we used to set up outer interfaces was easiest if you assume you’ll want an outer interface. I no longer think that assumption is a good one.
* The new approach is much more explicit about what you’re setting up, and I think it will be more clear to users (instead of magically taking one of their interfaces and turning it into an MS-outer, they have to define the MS-outer explicitly).

- [x] Create the `MSOuterTISInterface` object
- [x] Tests for `MSOuterTISInterface`
- [x] Docstrings
- [x] Use `MSOuterTISInterface` in networks with current input
- [x] Update `MSTISNetwork` inputs
- [x] Update `MISTISNetwork` inputs
- [x] Code cleanup after changes


dwhswenson:
This is done and ready for review. Passes tests.

Again, I didn't change the AD MSTIS example; @jhprinz, you'll have to do that. The example in the `mstis.ipynb` should show you roughly what it should look like now.

The significant change here is that networks now take an `ms_outers` parameter, which can be a list of these new `MSOuterTISInterface` objects. This makes set-up *with* an MS-outer a little harder, but it makes setup *without* an MS-outer more logical, and makes the setup actually match what I expect a user would expect.

jhprinz:
Looks good. I would prefer a simpler way, but it reflects what is necessary to know to construct the MSOuterInterface so let's keep it until we use the other proposed way.

dwhswenson:
> Looks good. I would prefer a simpler way, but it reflects what is necessary to know to construct the MSOuterInterface

On the other hand, I think that they MSOuter will become a thing of the past. We're including it now for compatibility with existing work. State swap is a better approach. So the good news is, leaving the MSOuter *out* of the whole thing is a lot easier now. :-)

PR #535, opened 2016-08-03
Rename CV

jhprinz:
Rename all the CVs to end in `CV` rather than start with it. This is clearer and conforms to the pythonic names of classes


jhprinz:
Should be ready now once all tests pass.

PR #541, opened 2016-08-04
New WHAM code

dwhswenson:
The previous WHAM code didn’t correctly account for the possibility of having a different number of samples in each ensemble, which can be very important with SRTIS. Plus, the old code was poorly documented, had no unit tests, and was hard to maintain (all of which was because it was my quick Python port of @bolhuis’s perl script).

New version, based heavily on `pandas`. 

* Works correctly with different counts in each interface
* Shorter code
* More readable
* Clear connections between code and literature (references pointing to relevant lines in Frenkel & Smit)
* Unit tests (used to be one of the least-covered files; now at 90+%)
* Designed to be refactored into a more generic container to prepare our data for other histogram joining methods (e.g., MBAR, TRAM)
* Probably faster, too, although I didn’t formally check (but I did remove a `math.exp` from an inner loop, and I switched lots of loops to either pandas/numpy functions or list comprehensions).

I think it is a bit of an improvement!

- [x] New pandas-based WHAM code
- [x] Clear comment-based explanations connecting with F&S
- [x] Docstrings
- [x] Nosetests for toy data
- [x] Test that it fails on insufficient data input


jchodera:
This code will need to be overhauled at some point to compute everything in log space. It's just going to die horribly when the ratio of partition functions becomes too large, since you're exponentiating things all over the place.

dwhswenson:
> This code will need to be overhauled at some point to compute everything in log space. It's just going to die horribly when the ratio of partition functions becomes too large, since you're exponentiating things all over the place.

Doesn't the use of a cutoff (which also removes noisy data) fix this -- at least in combination with binary weights? At a given $Q$ (bin number), we only have contributions from a few histograms. This means that $w_k^Q = 0$ for any histogram $k$ that doesn't have significant population in bin $Q$. So we don't have to worry about discrepancies in $Z_k$ from $\sum_k W_k^Q M_k / Z_k$, because $W_k^Q = 0$ except where the $Z_k$ are within a few orders of magnitude.

I haven't done the math to rewrite this in log space, but keep in mind that our weights can be actually 0, not just exponentially small. Would that have consequences in a log space version?

dwhswenson:
Passes tests, ready for review.

I'll also mention that this now has a much better error message if there isn't enough overlap to combine the histograms. See the `check_cleaned_overlaps` method.

jhprinz:
I think this looks fine. We can sort out the John's argument later. We might just use pymbar at some point. Trying to fix these kind of problems twice should not be necessary.

PR #545, opened 2016-08-05
[WIP] Enable simtk.units in CVDefinedVolume

jhprinz:
It adds `simtk.unit` and `simtk.openmm` to the allowed packages and rewrite the 
`CVDefinedVolume` and `PeriodicCVDefinedVolume` that they work also with simtk.unit

Problem was that simtk.unit does not support `%` so I had to rewrite it.

Should pass tests. Ready for review and merge

### DIFF
--- a/openpathsampling/volume.py
+++ b/openpathsampling/volume.py
@@ -248,8 +248,8 @@ class CVDefinedVolume(Volume):
         '''
         super(CVDefinedVolume, self).__init__()
         self.collectivevariable = collectivevariable
-        self.lambda_min = float(lambda_min)
-        self.lambda_max = float(lambda_max)
+        self.lambda_min = lambda_min.__float__()
+        self.lambda_max = lambda_max.__float__()
 
     # Typically, the logical combinations are only done once. Because of
     # this, it is worth passing these through a check to speed up the logic.
@@ -359,7 +359,7 @@ class CVDefinedVolume(Volume):
             return super(CVDefinedVolume, self).__sub__(other)
 
     def __call__(self, snapshot):
-        l = float(self.collectivevariable(snapshot))
+        l = self.collectivevariable(snapshot).__float__()
         return l >= self.lambda_min and l <= self.lambda_max
 
     def __str__(self):
--- a/openpathsampling/volume.py
+++ b/openpathsampling/volume.py
@@ -396,8 +396,10 @@ class PeriodicCVDefinedVolume(CVDefinedVolume):
     """
 
     _excluded_attr = ['wrap']
-    def __init__(self, collectivevariable, lambda_min = 0.0, lambda_max = 1.0,
-                                       period_min = None, period_max = None):
+
+    def __init__(
+            self, collectivevariable, lambda_min=0.0, lambda_max=1.0,
+            period_min=None, period_max=None):
         super(PeriodicCVDefinedVolume, self).__init__(collectivevariable,
                                                     lambda_min, lambda_max)        
         self.period_min = period_min
@@ -419,7 +421,16 @@ class PeriodicCVDefinedVolume(CVDefinedVolume):
 
     def do_wrap(self, value):
         """Wraps `value` into the periodic domain."""
+
+        # this looks strange and mimics the modulo operation `%` while
+        # being folly compatible for simtk numbers and plain python as well
+        # ints and floats. I think we usually have floats, but the tests
+        # from dwhswenson use ints and it should work with both
         val = value - self._period_shift
+
+        # little trick to check for positivity without knowing the the units
+        # or if it actually has units
+
         if val > val * 0:
             return value - int(val / self._period_len) * self._period_len
         else:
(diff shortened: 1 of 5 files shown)


jhprinz:
Wait, this get insanely complicated... `NaN` and simtk. `%` and simtk. Ahhh...

jhprinz:
Okay, that should do it...

jhprinz:
You can now use CVs that output Quantities. Make sure that your ranges also use compatible units or you get an error.

openpathsampling/volume.py
```diff
@@ -406,7 +419,16 @@ def __init__(self, collectivevariable, lambda_min = 0.0, lambda_max = 1.0,
...
+        val = value - self._period_shift
+        if val > val * 0:
+            return value - int(val / self._period_len) * self._period_len
+        else:
+            wrapped = value + int((self._period_len - val) / self._period_len) \
+                * self._period_len
+            if wrapped >= self._period_len:
+                wrapped -= self._period_len
+
+            return wrapped
```
dwhswenson: can you add a comment here to remind future-us why you did this? I get why, from your comments in the PR, but I'm pretty sure I'd look at this later and go "WTF?"

(As easy as "equivalent to: $OLD_STUFF; needed because unit don't like %")

openpathsampling/volume.py
```diff
@@ -359,8 +366,14 @@ def __sub__(self, other):
...
     def __call__(self, snapshot):
-        l = float(self.collectivevariable(snapshot))
-        return l >= self.lambda_min and l <= self.lambda_max
+        l = self.collectivevariable(snapshot).__float__()
+        if self.lambda_min != float('nan') and self.lambda_min > l:
+            return False
+
+        if self.lambda_max != float('nan') and self.lambda_max < l:
+            return False
+
```
dwhswenson: why would `lambda_min` or `lambda_max` ever be NaN?

jhprinz: it should be `inf` actually, but you designed some tests that use `nan`. That's why :)

dwhswenson: Really? How were those tests passing before? And what tests? I don't see them in `testvolume.py`

jhprinz: Not sure. It reported a string `nan` not having a `__float__` property so I just added that. Could be that it is actually only in the constructor and not the call test. Anyway I changed it to `inf` and `-inf`

dwhswenson: > Anyway I changed it to `inf` and `-inf`

Still seeing the same things here... and the point is that it shouldn't need the nan-test here. Actually, it should be even simpler than the original, since Python understands `lambda_min <= l <= lambda_max`.

jhprinz:
Fixed the two issues. I left in a check for `float('inf')` to allow a user to use `inf` without specifying a unit. Hope this will pass now

jhprinz:
Tests pass. Ready for review again :)

dwhswenson:
I still don't understand why there's an explicit check against `float("±inf")`. After all, `x < float("inf")` for any `x`. (That's kind of the point, right?)

openpathsampling/volume.py
```diff
@@ -419,7 +422,16 @@ def __init__(self, collectivevariable, lambda_min = 0.0, lambda_max = 1.0,
 
     def do_wrap(self, value):
         """Wraps `value` into the periodic domain."""
+
+        # this looks strange and mimics the modulo operation `%` while
+        # being folly compatible for simtk numbers and plain python as well
```
dwhswenson: "fully compatible" not "folly compatible" (you might say that simtk numbers not supporting % is folly, though!)

openpathsampling/volume.py
```diff
@@ -419,7 +422,16 @@ def __init__(self, collectivevariable, lambda_min = 0.0, lambda_max = 1.0,
...
+        # ints and floats. I think we usually have floats, but the tests
+        # from dwhswenson use ints and it should work with both
```
dwhswenson: If you used degrees, you're likely to give `int`s as input. Of course, maybe the better way is to make sure they get cast to float at some point.

jhprinz:
This is a simtk.unit Because you cannot compare 
```
3 * unit.nanometers < float('inf')
```

What would work is

```
3 * unit.nanometers < float('inf') * u.meters
```

To allow to just use `min_lambda='inf'` without knowing the unit that is used. I added the
check for infinity. We can argue that we remove it and a user needs to write `min_lambda=`floor('inf') * unit` instead, but we already allow to use a string.

dwhswenson:
Can you add a comment for that as well? Any time there's code that looks pointless but is necessary for some obscure reason (ugh... I hate units), it would be nice to have a comment explaining the reason.

Add that, fix comment typo (in line note above), and this should be ready to go!

jhprinz:
> ugh... I hate units

True, that is making a lot of stuff complicated coding wise...

Done... Should pass tests since I only added comments

PR #550, opened 2016-08-08
Fix running `snapshot == proxy` bug

### LINKED ISSUE #546: Bug in `shared_configurations`
I'm having a little trouble tracking down the exact conditions that cause this, but I'm finding a bug in `shared_configurations` based on the fact that `Trajectory.iter_proxies()` can return "Iterator() over list of `Snapshot` or `LoaderProxy`", as its docs describe.

Here's my understanding of the problem. I call `shared_configurations` (well, actually, I start at `shared_trajectory`). It calls `as_proxies`, which in turn calls `iter_proxies`.

Now, `shared_configurations` calls both `self.as_proxies()` and `other.as_proxies()`, and then does the set intersection of them. However, I'm seeing a case where `self` is returning snapshots, and `other` is returning proxies. This, of course, means the intersection is empty.

Changing `shared_configurations` to do:

```python
        if not time_reversal:
            print "self:"
            for s in set(self.as_proxies()): print s
            print "other:"
            for s in set(other.as_proxies()): print s
            print
            return set(self.as_proxies()) & set(other.as_proxies())
```

I get output:
```
self:
<openpathsampling.engines.toy.snapshot.ToySnapshot object at 0x10f8dcc10>
<openpathsampling.engines.toy.snapshot.ToySnapshot object at 0x10f8ae050>
<openpathsampling.engines.toy.snapshot.ToySnapshot object at 0x10f8dcd10>
(etc)
other:
<openpathsampling.netcdfplus.proxy.LoaderProxy object at 0x10f8e90f0>
<openpathsampling.netcdfplus.proxy.LoaderProxy object at 0x10f8e9370>
<openpathsampling.netcdfplus.proxy.LoaderProxy object at 0x10f8e91e0>
(etc)
```

I haven't made a minimal working example to cause this, but it is happening in https://github.com/dwhswenson/OPSPiggybacker/pull/10.

jhprinz:
Closes #546.

It should fix the problem that using a mix of proxies and real objects are not comparable.

The fix works, but always using UUIDs. Every OPS objects has a UUID, which kind of replaces the memory address for comparison. If two objects have the same UUID they are considered the same and the uuid is also used as hash. If you use UUID storage then these uuids are stored with the object and reloaded, so you can compare and merge across files. If you store without uuids (which should be removed at some point) then after saving, closing, reopen and load you get the same object, but with a different uuid. 

What is new now is that when you load a proxy from a store without UUIDs it will remember that for this index a specific UUID was used. So if you then load the real object it will have the same UUID as the proxy, even whenusing non-uuid storages.

There is a little gain of speed for not storing the uuids, but overall I would like to remove this in the future. It is a hassle to keep both ways working.


jhprinz:
This ready for review in case the tests will pass...

dwhswenson:
This look good to me. I checked it out locally, and can confirm that it solves the problem in #546. Merging!

Issue #549, opened 2016-08-08
WHAM keeps failing with insufficient overlap


jhprinz:
I keep getting errors from WHAM about

```
RuntimeError ("Insufficient overlap to combine histograms.
```

not always but quite often.

I think this is due to too small bin sizes, is that correct? If so, could we just make the bins a little larger in the examples?

dwhswenson:
> I think this is due to too small bin sizes, is that correct?

No, it is due to insufficient sampling. This error means that there is no overlap in the histograms. Changing the bin sizes will only fix it if you make the bins larger than the space between interfaces.

dwhswenson:
One thing that might be worth changing: are we still using the *square* of the distance as an order parameter? I originally set it up that way, thinking that we'd save computational cost by skipping the `sqrt`. However, our overhead is so much more than that `sqrt`, so it might be worth including it in the order parameter.

I don't think it will fix this problem (you still won't have overlap between interfaces), but it would probably make more intuitive sense to work in a linear space instead a quadratic one.

jhprinz:
I changed the binsize in one example from `0.02` to `0.05` and then it worked. Not sure why then.
Can that make sense?

dwhswenson:
Maybe, although partly by luck and partly by doing bad science. The interfaces are at 0.00, 0.09, 0.16, 0.25. And your bins edges at 0.0, 0.05, 0.10, etc. So probably you're saying that a result of 0.06 crosses the interface at 0.09. In other words, your bins are too wide to be meaningful.

Really, we should switch to a linear order parameter. Then we need actually sufficient sampling.

jhprinz:
Got it. Good point. Can you change that?

dwhswenson:
Yeah -- right now working on the SRTIS stuff, but I can add this to my list for the day.

jhprinz:
Thanks. I have enough stuff to do. Need to give a presentation tomorrow in the group seminar...

jchodera:
Your shouldn't be allowed to select bins that clash with the interface
definitions, should you?

jhprinz:
Actually, maybe @dwhswenson can clarify if we actually need the smaller intermediate bins.
I think these do not contribute to the computation of the reweighting factors. it is merely for
plotting and getting the nice probability to reach a certain CV value, is it?

dwhswenson:
You definitely shouldn't do it. "Shouldn't be allowed to" is a more difficult proposition.

[...]

PR #553, opened 2016-08-08
Use interface lambda hinting in WHAM DF cleaning

dwhswenson:
The problems in #549 were two-fold: both were sampling, but for different reasons. First, sometimes interfaces didn't overlap because no path from the inner interface crossed the outer interface. Second, if 100% of paths in an outer interface went well past the interface boundary, the automatic identification of the interface boundary failed.

Both can be solved by more sampling. The second one, however, could also use the new `InterfaceSet` object to hint what the interface values are.

That's what this code implements. If the interface lambdas are defined in the interface set, then we use that information. If not, we fall back on the old approach.

This should help tests pass some, although the first problem remains.

### DIFF
--- a/openpathsampling/analysis/tis_analysis.py
+++ b/openpathsampling/analysis/tis_analysis.py
@@ -389,7 +389,9 @@ class TISTransition(Transition):
                 self.histograms['max_lambda'].values(),
                 fcn="reverse_cumulative"
             ).sort_index(axis=1)
-            wham = WHAM()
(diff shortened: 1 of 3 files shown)


dwhswenson:
Ready for review!

jhprinz:
I guess we will need to discuss the actual creation of the table used for WHAM. It seems very complicated somehow.

But this looks good, so I will merge it now.

PR #555, opened 2016-08-09
Slightly prettier version of ugly SRTIS (`SRTISBiasFromNetwork`)

### LINKED ISSUE #260: Single Replica TIS

### LINKED ISSUE #525: Include dummy moves in counting fraction of move attempts?
A question, relevant for SRTIS:

***When counting the fraction of moves of a given type, should we include null (dummy) moves?***

*Context*: One check that the move scheme is behaving correctly is to examine the fraction of moves attempted with a given `PathMover`. So we figure out the number of moves for that mover, divide it by the total attempts, and compare the result to an expected number (for single replica, expected is calculated with the assumption of uniform sampling of the ensembles). We report this along with the acceptance ratio in the `scheme.move_summary` function. The question here is what should be the "total attempts": with or without dummy moves? SRTIS has dummy moves in order to satisfy detailed balance.

As an example, assume a simulation had 3 dummy moves, 2 shooting moves, and 1 replica exchange. Do we say that replica exchange was attempted 1/3 of the time, or 1/6 of the time?

*Advantage for not including the dummy move in the calculation (example answer is 1/3)*: Numbers should more closely match what we should get out of the multiple-replica scheme that the single replica scheme is based on.

*Advantage for including the dummy move in the calculation (example answer is 1/6)*: This normalizes based on the number of steps given to the `run` function. It’s also a more clear check of what the code actually does (includes the detailed balance corrections made by the dummy moves).

From a coding standpoint, including the dummy moves in the calculation may require modifying the `IdentityPathMover` to allow an option as to whether or not it should be included in this count. This is because it is used for both the initial conditions (which are never included in the count) and for these dummy moves, which would be included. I'd do this by adding an option `counts_as_attempt` to its `__init__`. Default would be not to include (`False`). I think that’s the easiest fix for it; probably only a few lines.

Thoughts? (I'd particularly like to hear from @bolhuis on this.)

dwhswenson:
Still using the ugly move scheme, but this makes it very easy to set up fixed bias SRTIS from a previously analyzed `TISNetwork` object.

- [x] Easy set-up of TIS bias from a network object as input
- [x] Tests for SRTISBiasFromNetwork
- [x] Docstrings
- [x] Fix counting of dummy moves (see #525)

This finishes fixed-bias SRTIS. Resolves #260. Resolves #525.


dwhswenson:
Passed tests. Ready for review.

jhprinz:
Looks good. I do not completely understand everything, but I trust in the tests to reflect correct execution.

Issue #551, opened 2016-08-08
Ensure we are preventing people from doing stupid things with WHAM


jchodera:
From #549, it seems that we may have accidentally made it incredibly easy to get erroneous results from WHAM. We need to revisit our approach to make it easy to do the right thing and hard to do something different.

Tagging @pgbolhuis

dwhswenson:
That's @bolhuis [no pg on github :-)]

Also, to clarify: this has nothing to do with WHAM specifically. It's just the histogram generation procedure.

bolhuis:
Hi,

Hailing from the tranquil region of Vosges, here is my input.

This issue is alluding me a bit. Why is there insufficient overlap? Basically this is only possible when there is an interface ensemble/ replica that does not connect to the next interface, I.e when there is no path that reaches the next interface. For retis this is only true when there is at least a pair that has not a single exchange. 

Is this the case? It should be easy to check,not? 

In all other cases the wham should work. However, there is the issue of the 5% cut off,  related to the minimum required number of entries in the bin. Are you applying this pruning step? This could lead to problems. Solution: The check for sufficient overlap could remove the pruning step if there is no overlap, but there should.

Another option is to decouple the crossing histogram from wham. Indeed the crossing probability is the basic observable coming from Tis and acts as a diagnostic as well.
If the wham or Mbar is done on the discrete interface histograms, aka as the path numbers, the check is trivial, and Mbar can be directly applied. The crossing prob hist can be kept track of separately, with arbitrary resolution, determined by the user.
This should limit the potential for user errors. 

Is this feasible?
What do you think?

Peter

jchodera:
The way this would work with MBAR is that the interface definitions would
be used in an iterative scheme to find the normalizing constants, and then
you can compute whatever expectations you want (such as unbiased crossing
probabilities with any order parameter) afterwards.

jhprinz:
Let me clarify my point of view a bit, because I think David got it not completely right. My frustration was that it kept failing and I spend a whole day just waiting and restarting to get the tests pass eventually. 

My general confusion was, that this did not seem to happen before. I think what we need to happen is that at least one sample drawn from, say, `A0` will be part of `A1` and one sample drawn from `A1` will be part of `A2`. 

[...]

PR #556, opened 2016-08-10
ipynb initial conditions hack

dwhswenson:
PR I promised in https://github.com/openpathsampling/openpathsampling/issues/551#issuecomment-238674341

We've been having trouble with analysis failing because the test-generated data is insufficient.

I think the best solution is to generate a real dataset that gets tested in the analysis, but @jhprinz seemed pretty frustrated with the current situation, so I put together this hack. 

There are three ways that insufficient sampling could cause the rate calculation to fail. This makes it so our examples won't show two of the three:

1. **No overlap between ensembles in the same interface.** This is now avoided by putting in an intial condition that forces overlap. There's still the possibility of falling under the cutoff threshold, but I don't think that was the main problem we had.
2. **Unable to determine outermost interface lambda value.** This happened if 100% of trajectories in the outermost interface went to the other state (so its crossing probability was 100% for all lambdas). This has been avoided by using the interface set lambda value (when that is available).
3. **No minus moves done; unable to determine flux.** This one remains a potential problem, but should be rarer than the other two were.

This also makes the examples harder to understand, and demonstrates things that I would consider bad practice instead of best practice. But it makes the tests pass.

There are still situations where the *setup* fails, but this should reduce the likelihood that the analysis does.


dwhswenson:
Passes tests (on first try.) Ready for review. Main changes:

* use interface set to determine outermost lambda, when possible
* use initial conditions defined by setting all initial trajectories to crossing the outermost of the set
* reduce steps to 300 in each test

jhprinz:
Thanks. I will merge this now.

PR #561, opened 2016-08-12
Rename .md -> .mdtraj

jhprinz:
As the title says...


dwhswenson:
One thing to consider, on the question of function or property:

```python
>>> trajectory.mdtraj is trajectory.mdtraj
False
```

This seems weird to me. It seems much clearer if it reads

```python
>>> trajectory.to_mdtraj() is trajectory.to_mdtraj()
False
```

That makes sense: it returns a different object every time.

The question is if we'll trade the clarity for convenience. But we should keep in mind that it is a potential point of confusion.

jchodera:
> trajectory.to_mdtraj() is trajectory.to_mdtraj()

`to_mdtraj()` or `asMDTraj()` are both good. 

Is there a standard way that `MDTraj` names conversion functions?

jhprinz:
> `>>> trajectory.mdtraj is trajectory.mdtraj`

We can cache that with a weak reference. So as long as the mdtraj objects exists it will return the same object. Just saying we can have this work properly.

I understand the concern, but then we would have to fix all the other properties as well of which we have several. I personally like the idea of a property, because it is short and it looks nicer to me. 

It is a change of representation which is usually done with a `to` or `as` function I believe like numpy to python list `tolist()`. I guess that could be confusing, but we intuitively used properties a lot for these kind of things, so I am not sure, if that is necessarily confusing. If we attach `as` or `to` it should be a function though.

So, which one do we want?

```py
trajectory.mdtraj.save_pdb('...')
trajectory.asMDTraj().save_pdb('...')
trajectory.toMDTraj().save_pdb('...')
trajectory.as_mdtraj().save_pdb('...')
trajectory.to_mdtraj().save_pdb('...')
```

jhprinz:
I can quickly make the change but we should agree

dwhswenson:
I think I personally prefer `to_mdtraj()`, followed by `as_mdtraj()`, followed by `tomdtraj()`. CamelCase functions are C++: OpenMM's Python API has them because they really *are* C++ functions, wrapped with SWIG.

`to_mdtraj()` mirrors MDTraj's `to_openmm()` for topologies. `tomdtraj()` mirrors numpy (and pandas) `tolist()`, although I find it harder to read without the underscore.

jchodera:
I'm totally good with those suggestions!

dwhswenson:
This is ready to merge, right @jhprinz?

jhprinz:
Yes, I picked `to_mdtraj()` now, which is pythonic and similar in mdtraj.

jhprinz:
Yes, since it passes. Go ahead.

PR #562, opened 2016-08-15
Fix InOut Slowness

### LINKED ISSUE #557: Really bad scaling in `in_out` for `SequentialEnsemble` prevents visualization
In #539, I mentioned that I couldn't get the repex-shoot-repex move to visualize. The problem, it turns out, isn't a bug so much as an implementation that is incredibly slow for sequences that include multiple ensembles.

Run the following:

```python
# FullEnsemble as a placeholder
from openpathsampling.ensemble import FullEnsemble
ensA = FullEnsemble()
ensB = FullEnsemble()

def repex_shoot_repex_sequence(n_repeats):
    sequence = []
    for i in range(n_repeats):
        part_i = [paths.ReplicaExchangeMover(ensA, ensB),
                  paths.OneWayShootingMover(ensA, paths.UniformSelector(), engine=None),
                  paths.OneWayShootingMover(ensB, paths.UniformSelector(), engine=None),
                  paths.ReplicaExchangeMover(ensA, ensB)]
        sequence.extend(part_i)
    return paths.SequentialMover(sequence)

import time

for i in range(4):
    mover = repex_shoot_repex_sequence(i)
    start = time.time()
    inout = mover.in_out
    stop = time.time()
    print i, stop-start
```

I got the following output:

```
0 5.00679016113e-06
1 0.00130391120911
2 9.96101999283
```

It never got to the last output, because I killed it after almost an hour. Memory usage for that little script spiked up to 5GB at points (although it mostly hung out in the 200-400MB range).

(Also, having them be different versions of the same ensemble isn't the problem: I did that to simplify the code above; the original test script created a whole network and used two ensembles from that.)

jhprinz:
Closes #557 .

This was not a slow implementation issue (it is still slow in some cases), but it was a bug. Turns out I was looping wrong at some place which apparently was not caugth in my testing. Now it is much faster, running the swap five times will still take 7 seconds. But that is much faster than before.

Also the result before was not exact. Now it is... The fix is rather small, so don't laugh...


dwhswenson:
Haha -- amazing how easy it is for a typo of `1` instead of `2` to cause huge problems!

We should still look into speeding that up even further, but if I can visualize the move scheme in less than a minute, we can call it good enough for now.

Merging!

PR #563, opened 2016-08-15
[WIP] an experiment to treat all Empty and Full Ensembles as singletons.

jhprinz:
This changes the `__hash__` and `==` so that Full and Empty Ensembles/Volumes are identical. Will improve speed at some points in the code.

### DIFF
--- a/openpathsampling/ensemble.py
+++ b/openpathsampling/ensemble.py
@@ -194,7 +194,7 @@ class Ensemble(StorableNamedObject):
         if self is other:
             return True
 
-        return NotImplemented
+        return str(self) == str(other)
 
     @abc.abstractmethod
     def __call__(self, trajectory, trusted=None):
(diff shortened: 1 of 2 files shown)


dwhswenson:
> Will improve speed at some points in the code.

Is this actually a bottleneck anywhere?

jhprinz:
> Is this actually a bottleneck anywhere?

No, but the fix for the hash is fast and when you place two `EmptyEnsemble()` in a `set` one will disappear. The comparison using `str()` does already work.

jhprinz:
The idea was that this would work:

```py
>>> a = paths.FullEnsemble()
>>> b = paths.FullEnsemble()

>>> a == b
True
>>> s = set(a,b)
>>> print len(s)
1
```

But, I think this will cause trouble with proxies and no time to fix this. Was a nice idea though. I will close this for now. Not so important.

PR #564, opened 2016-08-15
Custom Mover / Strategy example: RepEx-Shoot-RepEx

dwhswenson:
Example for paper 2.

Currently I'm just doing this on a very simple 2-state 2D system, and I may still tweak the system a little bit. However, it might be nicer to do this with the "Geisslerene" dimer in WCA system. Isn't that an openmmtools test system?

Conclusion appears to be that this fancy move scheme doesn't gain us much, if anything. Oh well. Still amazing how easy the test was to set up and try!

Leaving it as WIP for now, since I'm still playing around with it, and since I might switch to a more interesting underlying model.


jhprinz:
> Still amazing how easy the test was to set up and try!

Yes!!!!!

dwhswenson:
Ready for review. The main change is in the new [`custom_strategy_repex_shoot_repex.ipynb`](https://github.com/dwhswenson/openpathsampling/blob/f059dc3e36a074bb9cfffc03886f2e2e966f7bdb/examples/ipython/custom_strategy_repex_shoot_repex.ipynb).

We might switch to a different underlying model system (the one I chose is pretty pitiful, and the interfaces were completely arbitrary). However, the basic output will be used in an example of how to create a custom path mover and move strategy, and then measure whether it enhances replica mobility. (This specific approach shows no real improvement for this model.) This will be an example in paper 2.

jhprinz:
looks good.

PR #568, opened 2016-08-17
Reorganize examples (part 1)

### LINKED ISSUE #490: Reorganize `examples` directory
As discussed a little in #482, we are all in agreement that the `examples` directory needs to be reorganized. Let’s figure out how to restructure it.

First point: does everyone agree with getting rid of the extra layer of `examples/ipython` and moving all the ipynb examples into the main `examples` directory? (All of our examples are ipynbs, the only other things in there will be initial information needed for startup, e.g., PDBs.)

Second: the structure within that directory. I gave a partial suggestion before, based on the idea that we would organize by system (toy, alanine dipeptide), then by method (TPS, TIS). @jchodera felt that it was better to organize by method, then by system.

I’ll put a proposal in the first comment below, and I’ll reserve this top post for description of the consensus.

## Consensus / Summary

* *do we move everything to `examples`?* (Agreed yes)
* *what overall structure do we decide on?* (Basically as below)

dwhswenson:
Steps toward #490.

This one will include the toy models and the AD MSTIS example. Other notebooks will move in subsequent PRs.

- [x] create READMEs for all subdirectories (as listed in #490)
- [x] move toy MSTIS notebooks to subdirectory
- [x] move toy MISTIS notebooks to subdirectory
- [x] get tests working again
- [x] move AD MSTIS notebooks to subdirectory
- [x] clean up READMEs to point to correct files
- [x] rename sequential files to include numbers

May also include adding a little extra documentation to the notebooks here and there.


dwhswenson:
For the sake of sanity, this is going to become "part 1". Moving all at once isn't going to happen.

@jhprinz : Anything you have going that involves changes to the toy engine examples? If not, I'll just put those into this PR. Let me know if it is also safe to move around the alanine dipeptide MSTIS stuff -- I know you're about to do the big run of those; I can move them before or after that, according to your preference.

jhprinz:
Please move all you want. I will run the example on a cluster anyway and will need a little script instead of a notebook.

dwhswenson:
Passes tests; ready for review. This only completes the move of:

* alanine dipeptide MSTIS
* toy model MSTIS
* toy model MISTIS

Others will follow in later PRs, but this is a good start. READMEs for other directories are already in place.

PR #569, opened 2016-08-19
Making InOut faster (with hidden options for more inspection)

### LINKED ISSUE #557: Really bad scaling in `in_out` for `SequentialEnsemble` prevents visualization

jhprinz:
Closes #557 .

This will improve spee for the example that David hat and related things. All stuff that made it slow, especially handling cases with multiple samples per ensemble are disable by default to keep computations fast.


jhprinz:
It is ready. Although it could use more testing.

dwhswenson:
This looks good. I like the idea of moving the InOut stuff to a separate file -- that works even better than my previous suggestion to mark related classes with "see also" docstring sections.

I assume (from comments in #557) that you testing this against the repex-shoot-repex example? Assuming that is fast (or fast enough, meaning less than a minute), this should close the issue!

Merging!

jhprinz:
Well, yes. As long as it runs with the example you gave me that was fine. It is my test, if you wish

Issue #65, opened 2014-11-11
Address confusion due to similarity of `Simulation` and `Simulator` names


dwhswenson:
We have two objects, `Simulator` and `Simulation`, which have very similar names and very different purposes. I think that will be terribly confusing for users. In addition, we have a `Calculation` class whose purpose might also be confused with the other two.

Here is my description of what each object does (@jhprinz, let me know if I'm missing any aspects):
- `Simulation`: interacts directly with the underlying MD engine. It just transfers state information (positions, velocities, etc.) back and forth between MD engine and our code.
- `Simulator`: this one is a bit harder to describe. It mainly involves two functions: (1) setting up the system (including setting up our storage and setting up topologies, force fields, etc. for the MD engine); (2) the `generate` function, which runs the MD and stops when we reach a stopping condition.
- `Calculation`: think of this as the `main()` for whatever kind of sampling we're doing: it's the part that distinguishes the overall sampling algorithm, e.g., making TIS different from SATIS or AMS.

Originally the `Simulator`/`Simulation` divide was because we needed our own class (`Simulator`) to interact with OpenMM's `Simulation`. I think we can clean that up now; whether we also need to change the `Calculation` name is up to you.

Here's my suggestion for refactoring:

Let specific `Simulation`s (e.g., `OpenMMSimulation`) inherit from `Simulator` (which might be renamed to `Simulation`). For cases like `OpenMMSimulation`, which currently inherits from `openmm.app.Simulation`, we could use multiple inheritance, or we could make the `openmm.app.Simulation` into an instance variable for the `OpenMMSimulation`. So there would be a superclass with the code to set up the `storage` and for the generic `generate` function, and the subclasses would support system set-up and dynamics state information for OpenMM, or for my ToyDynamics, or (perhaps one day) for Gromacs.

From a user perspective, this would mean that instead of having to build both a `Simulator` and a `Simulation`, you just need a single object which combines information from both. In fact, the final object is kind of an OpenTIS wrapper around the MD engine's objects, providing a few extra functions and the ability to translate to and from our data structures.

If this seems like a good idea, I can implement it. Any thoughts?

PR #570, opened 2016-08-19
[WIP] Handling of `NaN` and other errors in engine

jhprinz:
An engine might produce `NaN` or other errors that might be catchable and being fixed by a suitable retry. This PR adds _mutable_ not stored settings on how to deal with it
```
    Attributes
    ----------
    on_nan : str
        set the behaviour of the engine when `NaN` is detected. Possible is
        1.  `fail` will raise an exception
        2.  `ignore` will completely ignore it and continue
        3.  `retry` will rerun the trajectory (default)
    retries_when_nan : int, default: 2
        the number of retries (if chosen) before an exception is raised
    on_error : str
        set the behaviour of the engine when an exception happens. Possible is
        1.  `fail` will raise an exception
        2.  `ignore` will completely ignore it and continue
        3.  `retry` will rerun the trajectory (default)
    retries_when_error : int, default: 2
        the number of retries (if chosen) before an exception is raised
    on_max_length : str
        set the behaviour if the trajectory length is n_frames_max. Possible is
        1.  `fail` will raise an exception
        2.  `ignore` will completely ignore the limit and continue
        3.  `stop` will stop and return the max length trajectory (default)
        4.  `retry` will rerun the trajectory
    retries_when_max_length : int, default: 0
        the number of retries (if chosen) before an exception is raised
    on_retry : str or callable
        the behaviour when a try is started. Since you have already generated
        some trajectory you might not restart completely. Possibilities are
        1.  `full` will restart completely and use the initial frames (default)
        2.  `50%` will cut the existing in half but keeping at least the initial
        3.  `remove_interval` will remove as many frames as the `interval`
        4.  a callable will be used as a function to generate the new from the
            old trajectories, e.g. `lambda t: t[:10]` would restart with the
            first 10 frames
```

When the engine fails or is stopped by the user, it will always shut down gracefully and return an exception giving the reason. All exceptions along the way are also capture and printed in the log file.


dwhswenson:
Questions on this (looking only at top-post, no code):

How does this affect the Monte Carlo? What @jchodera wants is not exactly an `ignore` but a `reject`: the move is rejected. The approach I thought of for this was: (1) raise a custom exception (maybe `EngineNaNError`); (2) the MC has a choice of either catching the error and returning `NaNErrorMoveChange` (which is rejected) or not catching the error.

Importantly, this is *not* a retry. This considers the NaN move as a valid trial, but rejects it. However, you should be able to easily check your simulation for a large number of moves rejected because of NaN.

Note that you *can't* just assume that a NaN-containing frame will always reject. Consider:

```python
>>> x = paths.FunctionCV("x", lambda s: s.xyz[0][0])
>>> vol = paths.CVDefinedVolumes(x, 1.0, float("inf"))
>>> ensemble = paths.AllOutXEnsemble(vol)
>>> # okay, we don't have a `trajectory2D` function, but you can imagine it
>>> trajectory = trajectory2D([[[0.0, 0.0]], 
>>>                            [[0.1, float("nan")]]])
>>> ensemble(trajectory)
True
```

So if you're sampling that ensemble, you might accept that trajectory. Right?

jhprinz:
I see. I can add that possibility, too. A problem I discovered is that OpenMM will fail with a `ValueError('Particle is nan')` when using CPU and run fine using CUDA but return a snapshot with NaNs. And crash once in a while (although I think that is MacOS two graphics card problem related)

Then there is still the question if you should continue until it really breaks or if you detect NaN. 

> Importantly, this is not a retry. 

This is clear. I just remembered John wanting to auto retry. This is different from rejecting and then in the scheme doing a retry.

So, should I remove all the retry stuff? Is this unintended?

dwhswenson:
> So, should I remove all the retry stuff? Is this unintended?

On that I would say ask @jchodera. I don't have a problem with it existing as an option if you've already implemented it, but if you haven't implemented it, it might not be worth implementing.

jhprinz:
Almost done. I left the retry as an option which is disabled. So the engine passes a NaNError to the EngineMover that generates a SampleNaNError which will be caught by the SampleMover/MC an return a RejectedMoveChange. I still need to test this somehow and make the capture optional.

jchodera:
I think "retry" is an important option! This is essentially a scheme that
excludes trajectories that lead to NaN. I think that is the preferred
behavior.

That's weird about the CUDA platform. My understanding was that it would
check for NaN every time the pair list was updated, every 100 steps or so.
It's possible that you are bringing back snapshots more frequently than
that so just might get a snapshot with NaNs.

jhprinz:
Okay, I implemented both. I actually like the idea that a trajectory that contains NaNs will be considered a rejected one. Not sure of the implications on the dynamics, but wouldn't this mean that under the given dynamics (including integrator) trajectories that go 'NaN' are always excluded and we do not want these anyway. Of course, dynamics that could lead to NaN are not really what we want, but for bootstrapping this should work. 

I would assume that just running more steps in a PathSimulator and doing the retry will be more or less equivalent in many cases. The only difference is, that if you do not care about the dynamics so much, using retries that do not completely retry the full step, but cut off the `NaN` trajectory and continue could be faster.

The default is, that the Engine throws a `EngineNanError` (there is also now a `EngineMaxLengthError` but that is not the default). The Mover will catch this error and effectively create a RejectedMoveChange that contains the last valid trajectory but will always be rejected, even if it would be a valid sample. I think that cannot happen since the engine should have stopped by then. But to make sure, it will be rejected. This way `NaN` will not wreck a simulation.

Now that I think of it. Could it be that this will oversample trajectories _in the vicinity_ of crashing ones?

jchodera:
Thanks, JHP!

Some quick comments:

> Okay, I implemented both. I actually like the idea that a trajectory that
> contains NaNs will be considered a rejected one. Not sure of the
> implications on the dynamics, but wouldn't this mean that under the given
> dynamics (including integrator) trajectories that go 'NaN' are always
> excluded and we do not want these anyway.
>
Both "retry" and "reject" lead to excluding trajectories with NaNs, right?
"reject" means that we accumulate statistics on those rejections, "retry"
does not?

> Of course, dynamics that could lead to NaN are not really what we want,
> but for bootstrapping this should work.
>
I've tried to be very clear: There is no way to completely avoid NaNs.
Using single-precision very much increases the probability of running into
NaNs.

> I would assume that just running more steps in a PathSimulator and doing
> the retry will be more or less equivalent in many cases. The only
> difference is, that if you do not care about the dynamics so much, using
> retries that do not completely retry the full step, but cut off the NaN
> trajectory and continue could be faster.
>
> The default is, that the Engine throws a EngineNanError (there is also
> now a EngineMaxLengthError but that is not the default). The Mover will
> catch this error and effectively create a RejectedMoveChange that contains
> the last valid trajectory but will always be rejected, even if it would be
> a valid sample. I think that cannot happen since the engine should have
> stopped by then. But to make sure, it will be rejected. This way NaN will
> not wreck a simulation.
>
> Now that I think of it. Could it be that this will oversample trajectories *in
> the vicinity* of crashing ones?
>
Excellent point. That may lead to altered statistics in the ratio between
different classes of trajectories in the vicinity of classes that produce
NaNs. I guess that means "retry" is preferred?

dwhswenson:
#### Question for clarification

> The Mover will catch this error and effectively create a RejectedMoveChange that contains the last valid trajectory but will always be rejected, even if it would be a valid sample

I haven't looked at the code yet, but when you say "effectively create a RejectedMoveChange," what exactly do you mean? Mainly, is there a quick and easy way to check for moves rejected because of `nan`? I would think that the easiest ways to do that would be to either add a `reason` to rejected move change (which could later include other explanations like "not in ensemble", or "failed acceptance probability test"), or to have a subclass of `RejectedMoveChange` specific to `nan`-rejected. But the main thing is that it could be really important to know that one of your replicas/ensembles is getting rejected a lot due to NaN.

#### NaNs in general

> I've tried to be very clear: There is no way to completely avoid NaNs.

@jchodera : We heard you, and while I trust that you're right, I still don't understand why this is so much of a problem that it becomes inevitable that every practical run has a NaN (which seems to be the problem you're having). Can you point to a resource that explains this?

The only time I've encountered dynamics that had this problem was a case where the potential involved taking the square root of a number that would become negative with a large step size. In that case, NaNs were easily avoided by using an adaptive time step integrator. (I guess I had another version of that where even the adaptive time step didn't help, but that was basically because the conserved Hamiltonian had a value on the order of 10^-400, which can't be represented by a `double`.)

#### Retry vs. reject

>> Could it be that this will oversample trajectories *in the vicinity* of crashing ones?

> Excellent point. That may lead to altered statistics in the ratio between
different classes of trajectories in the vicinity of classes that produce
NaNs. I guess that means "retry" is preferred?

This runs counter to my intuition. "Retrying" seems to me like it violates detailed balance. I don't see how retrying a trajectory because it went NaN differs from retrying because a move was rejected, which of course violates detailed balance.

Quick experiment, using a "configurational" sampler instead of a path sampler, but I think the argument is the same:

```python
import numpy as np

def V(x):
    return -np.sqrt(-(x-0.5)*(x+0.5))

# no efforts at subtlety or elegance in this sampling routine
def run_monte_carlo(n_steps, beta=2.0, max_dx=0.05, initial_position=0.0, reject_nan=False):
    position = initial_position
    history = []
    n_nan_trials = 0
    for step in range(n_steps):
        old_weight = np.exp(-beta * V(position))
        new_weight = float("nan")
        while np.isnan(new_weight):
            trial = position + 2.0*max_dx*(np.random.rand() - 0.5)
            new_weight = np.exp(-beta * V(trial))
            if np.isnan(new_weight):
                n_nan_trials += 1
                if reject_nan:
                    new_weight = 0.0
        acceptance_probability = new_weight / old_weight
        if np.random.rand() <= acceptance_probability:
            position = trial
        history.append(position)
    return (history, n_nan_trials)
```

Run it with:

```python
(reject, n_nan_reject) = run_monte_carlo(1000000, reject_nan=True)
(retry, n_nan_retry) = run_monte_carlo(1000000, reject_nan=False)
```

A quick check shows us that about 1.43% of trials are NaN in "retry" mode, and 1.38% in "reject" mode. Not really a significant difference (probably due to a 1D nature of this system).

Plotting:

```python
# http://www.wolframalpha.com/input/?i=Integral%5BExp%5B-2.0*Sqrt%5B-(x-0.5)*(x%2B0.5)%5D%5D,+x,+-0.5,+0.5%5D
norm_expV = 0.468451

plt.hist(reject, bins=51, alpha=0.5, normed=True, label='Reject')
plt.hist(retry, bins=51, alpha=0.5, normed=True, label='Retry')
plt.plot(x, np.exp(-2.0*V(x))*norm_expV, label='Exact')
plt.plot(x, np.exp(-2.0*V(x))*norm_expV * 0.96, 'm.', label='Exact*0.96')
plt.plot(x, np.exp(-2.0*V(x))*norm_expV - 0.05, 'm-', label='Exact-0.05')
plt.xlim(-0.6, 0.8)
plt.ylim(0.0, 1.6)
plt.legend()
```

gives us:

image

You're correct that `reject` samples *more* near the NaN region, but it also samples more correctly. I'm not sure why the exact was so much higher: please feel free to check over my quickly-hacked code to make sure I didn't do anything stupid. But I tried to correct for the difference with ad-hoc shift and ad-hoc scaling of the exact result. In all cases, `retry` looks low.

jchodera:
> @jchodera : We heard you, and while I trust that you're right, I still don't understand why this is so much of a problem that it becomes inevitable that every practical run has a NaN (which seems to be the problem you're having). Can you point to a resource that explains this?

There's no good reference that I know of, though this would be a fun paper to write.

The issue is with single-precision forces and weak thermostats. With double-precision forces, or with strong thermostats, this is much less of a problem. As you get to higher forces (which is often the case with forced or driven barrier crossing), the numerical stability may become poor, and more velocity or random number choices may lead to trajectories that quickly blow up. We see this all the time when using NCMC (in which we drive the potential or coordinates), and have switched to using Metropolized integrators like GHMC that can reject NaNs as a consequence.

> The only time I've encountered dynamics that had this problem was a case where the potential involved taking the square root of a number that would become negative with a large step size. In that case, NaNs were easily avoided by using an adaptive time step integrator. (I guess I had another version of that where even the adaptive time step didn't help, but that was basically because the conserved Hamiltonian had a value on the order of 10^-400, which can't be represented by a double.)

Oof, there are a *lot* of problems with adaptive timestep integrators for molecular simulation. They are generally not symplectic or even measure-preserving, so have a whole host of problems when it comes to anything involving stationary distributions.

Thanks so much for the concrete 1D example! There are a bunch of ways to handle excluded regions in Monte Carlo: The most common is using proposals that omit the forbidden region, in which case we need to then correct for the fact that the probability of ending up in the allowed region is increased, and this correction has to go in the Metropolis-Hastings acceptance criteria in `P(old | new) / P(new | old)`. This explains why the "retry" sampler---which effectively excludes proposals to the forbidden region without correcting the acceptance criteria---gives very erroneous results.

So @dwhswenson is totally right---"retry" will give incorrect statistics, while "reject" should give correct statistics.

I imagine the difference between the "exact" statistics and your histogram would mostly disappear if you changed
```python
plt.plot(x, np.exp(-2.0*V(x))*norm_expV, label='Exact')
```
to
```python
plt.plot(x, np.exp(-2.0*V(x)) / np.exp(-2.0*V(x)).sum(), label='Exact')
```
or, even better, numerically integrated over the histogram bins to get the unnormalized probability and then normalized.

Thanks, guys!

jhprinz:
> I haven't looked at the code yet, but when you say "effectively create a RejectedMoveChange," what exactly do you mean?

Okay, currently (still sorting it out) it will force to return a `RejectedPathMoveChange`. I added a `{ 'error': 'nan'}` to the details in the Sample as well as the PMC. But i like the idea of using subclasses. I think it is easier to remove these from analysis than having to load all the details to check. Having an additional attribute would also be fine. i just think the first is more elegant. We can still add the concrete reason to the details.

> This runs counter to my intuition. "Retrying" seems to me like it violates detailed balance. I don't see how retrying a trajectory because it went NaN differs from retrying because a move was rejected, which of course violates detailed balance.

I did not mean to say that retrying is better. Sorry for the wrong implication. I just thought that both ways will count wrong, but now I think it was actually only retrying that will do that. 

> So @dwhswenson is totally right---"retry" will give incorrect statistics, while "reject" should give correct statistics.

Yes, I think so. I never thought about the problem of proposing impossible moves, but ii acts as if the energy is infinite. As long as proposals are symmetric you should be fine. And into `nan` is zero and from `nan` is also zero, so we're good. And the little test shows it.

Btw. The little analysis is phantastic! Not sure how you patched that so quickly together, but it is pretty cool.

Well, this I do not understand. If I understood David correctly then he said that both will not work correctly but in his example the reject works fine. Is that correct? I would have assumed that both fail but the reject seems not to overcount. How is that possible? Retrying surely fails DB so that is out. I am confused, but I found the reason for the not fitting exact solution. I think you had a sign missing. The 0.46... is the solution, if you flip the sign in the exp. In the other case the integral is `2.24395` and so the normalization is `0.445643`. 

Okay, so rejecting should be the default because it is the right thing todo. We might keep the retry, becuase in cases like bootstrapping it might be faster. Does that make sense?

jchodera:
> Okay, so rejecting should be the default because it is the right thing todo. We might keep the retry, becuase in cases like bootstrapping it might be faster. Does that make sense?

Sounds great to me!

dwhswenson:
@jchodera : Thanks for the explanation. I'm still surprised that it is *such* a huge problem, but it helps to at least see what the problem is.

> I think you had a sign missing. The 0.46... is the solution, if you flip the sign in the exp. In the other case the integral is 2.24395 and so the normalization is 0.445643.

Nice catch! Yep -- I figured I'd done something stupid. Also, note that the correction of 0.44../0.46.. gives a number pretty close to the scaling by 0.96 that I got by eye, so that part of the plot should be pretty close to what the *correct* exact value would have been.

> If I understood David correctly then he said that both will not work correctly but in his example the reject works fine. Is that correct? I would have assumed that both fail but the reject seems not to overcount.

I had meant that I thought reject works, from a detailed balance standpoint, though I'm a bit disquieted that we can't avoid NaN trajectories. I think this example helps make it clear why it works: as you get closer to the boundary, the slope of the potential goes to infinity. So you can sort-of pretend (not in a mathematically rigorous sense, but just conceptually) that the area with a "NaN" potential is because there's an infinite wall, and the potential is actually infinite outside the domain. In this case, the probably of accepting a move there is, of course, 0. You can try, but you have to reject.

(The other thing I was saying is that, depending which degree of freedom gets the NaN, there could be cases where the NaN-containing trajectory satisfies the ensemble, and is therefore accepted. So we need to have the engine say "this trajectory contains NaNs and can't be trusted" -- which is exactly what you're doing.)

> I added a `{ 'error': 'nan'}` to the details in the Sample as well as the PMC. But i like the idea of using subclasses. I think it is easier to remove these from analysis than having to load all the details to check.

I think either approach is fine. To me, the trade-off is:

* subclass: easier to check
* attribute: easier to generalize

The only thing is I would suggest using `rejection_reason`, or something like that, for the attribute name, instead of `error`. My thought is that many (hopefully most) rejection reasons would not be errors!

***

Question on this: Is there any possibility that the NaNs are being hidden somehow when running in an ipynb? I've seen this error occur in a pure-python script on alanine dipeptide TPS, but I haven't gotten it to occur in an ipynb yet.

jchodera:
> Question on this: Is there any possibility that the NaNs are being hidden
somehow when running in an ipynb? I've seen this error occur in a
pure-python script on alanine dipeptide TPS, but I haven't gotten it to
occur in an ipynb yet.

Are you running on the same simtk.openmm.Platform within the ipynb as when
running from the command line?

dwhswenson:
> Are you running on the same simtk.openmm.Platform within the ipynb as when
running from the command line?

Should be -- I'm reloading the same OPS engine from a storage file.

(I would check, but the notebook version is currently at about step 5000 of 10000)

jhprinz:
Actually the behavior of the engine has been changed. You cannot specify the type anymore since this ended up in confusing all of us that you cannot just load and a restart would mean to recreate a new engine and make a new scheme, which makes comparison hard, etc...

So, if you want a specific engine you need to run
```
engine.initialize('platform')
```
now, the default initialize will pick the fastest possible. I remember having problems at some point with the way ipython imports environment variables, so it might be that ipynb uses CPU while python actually can use OpenCL in your case. I so far had `nan` only when using graphics cars, CPU always raise a `Particle has nan coordinates` (something) error.

dwhswenson:
Wrote a long thing checking engine types, but: nevermind... it finally NaN'd for the first time after 5400 steps. (After having done 10k flexible shooting steps before, and 1k fixed shooting steps.)

jhprinz:
THis is practically done. I have not made an example, but there are tests that ensure the correct error is thrown:

Some small questions:

1. I added all the options for when and how to fail to the `.options[]` where we also pu the `n_frames_max` already. This will be stored! But I would keep this as keep the default values. Not that the stores value should be immutable. Is that a good choice? Or do we want to keep them as attributes (which I had before). Also, is the storing of the current setting okay or confusing?

2. I added an option `EngineMover.reject_max_length = False` If set to `True` then if a max length exception is thrown (which requires `Engine.options['on_max_length'] = 'fail'`) then this will also result in a rejection. Not sure, if we want to use this, but it would be possible now.

You can merge when it passes and you think its okay. I can add a little example notebook later, which is what is still missing.

dwhswenson:
> I added all the options for when and how to fail to the .options[] where we also pu the n_frames_max already. This will be stored! But I would keep this as keep the default values. Not that the stores value should be immutable.

I'm not entirely clear what you're saying here. I think they should be *either* stored and effectively immutable, or not stored (in which case mutable is fine). But if the stored value isn't the value of that is actually used, I think that has potential to cause a lot of confusion. As for which of those two options is better, I have no idea -- both have advantages and disadvantages.

> I added an option EngineMover.reject_max_length = False If set to True then if a max length exception is thrown (which requires Engine.options['on_max_length'] = 'fail') then this will also result in a rejection. Not sure, if we want to use this, but it would be possible now.

I think that rejecting immediately upon an attempt to exceed the max length is what we want. So if the trajectory is already 5000 frames long, (`max_length==5000`), then the next time you try to take a step, it raises an exception (which is caught and mapped into a particular `MoveChange`.

Other options should be possible, but having it occur as an exception makes it easy to flag as "abnormal behavior." If you get a lot of it, then there's a general problem with your simulation.

The only downside is that you need to then *not* used max length the same as fixed TPS length. (In general, though, max length should be seen as a failsafe to prevent you from wasting CPU hours if everything else goes wrong.)

***

I'll start looking over this code, although more explanation of what you meant in the first quoted paragraph above would help.

openpathsampling/movechange.py
```diff
@@ -365,6 +365,26 @@ def _get_results(self):
...
+
+    This will return no samples also as its result, hence it is
+    rejected.
+    """
+    pass
+
+
+class RejectedMaxLengthSampleMoveChange(RejectedSampleMoveChange):
+    """
+    Represents an rejected SamplePMC because of occurance of NaN
```
dwhswenson: please correct the docstring here: this one is max length, not NaN.

Question about both of these: they return empty `result`, but do they return empty `trials`? I think that it is worth trying to return the (invalid) trial if possible.

dwhswenson:
Also, if you didn't actually make any changes to code in the AD TPS examples, could you return those to the version on `master`? Otherwise I have to keep fixing conflicts.

openpathsampling/movechange.py
jhprinz: Thanks for the catch. Yes, they return the last valid sub trajectory without the broken frame. Reason is mostly because some platforms do not even return the broken snapshot, while some return one inclusing nan.

jhprinz:
> Also, if you didn't actually make any changes to code in the AD TPS examples, could you return those to the version on master? Otherwise I have to keep fixing conflicts.

Will do. These should not be affected.

Quick question: Some of the platforms will throw an exception when nan appears, like `CPU`. So you cannot continue, even if you selected `ignore`. Therefore I think continuing with nans should not be allowed and thus. `ignore` should not be an option. Same for `error`. You could just ignore this frame and do a retry with the last coordinates, but that is weird and will probably lead to the same error again.

dwhswenson:
> Quick question: Some of the platforms will throw an exception when nan appears, like CPU. So you cannot continue, even if you selected ignore.

Is it not possible to catch the exception and convert it to something that makes sense within our framework?

jhprinz:
> Question about both of these: they return empty result, but do they return empty trials? I think that it is worth trying to return the (invalid) trial if possible.

Also the error contain the last working trajectory as `.last_trajectory`. You can do

```py
try:
    traj = engine.generate(...)
except dyn.EngineNanError as e:
    print 'last trajectory was of length %d' % len(e.last_trajectory)
```

jhprinz:
> Is it not possible to catch the exception and convert it to something that makes sense within our framework?

You can continue moves, etc, but you cannot keep on running this trajectory. You have this one frame that was caught and all you have is the last trajectory. If you continue you have to leave out this frame and retry. In some cases you really get a faulty frame you can append...

jchodera:
> Quick question: Some of the platforms will throw an exception when nan appears, like CPU. So you cannot continue, even if you selected ignore. Therefore I think continuing with nans should not be allowed and thus. ignore should not be an option. Same for error. You could just ignore this frame and do a retry with the last coordinates, but that is weird and will probably lead to the same error again.

Agreed---the only reasonable options are (1) let the exception through, (2) reject the trajectory and back up to the last accepted sample without NaNs, or (3) retry from the last good frame (which we have established may have an incorrect acceptance probability).

dwhswenson:
> You can continue moves, etc, but you cannot keep on running this trajectory. 

Ah, yes, I see. And yeah, once the trajectory is contaminated with `NaN`s, there's no way to continue running it (even if you can, a `NaN` in one coordinate will infect others and soon you're trying to run dynamics on a big list of nothing but `NaN`)

jhprinz:
> I'm not entirely clear what you're saying here. I think they should be either stored and effectively immutable, or not stored (in which case mutable is fine). But if the stored value isn't the value of that is actually used, I think that has potential to cause a lot of confusion. As for which of those two options is better, I have no idea -- both have advantages and disadvantages.

It is the same that you change `engine.options['n_max_length'], after you created the engine. It is saved and reloaded, but you can (of course) change it. 

Also not sure, what is best. You probably do not want to change this afterwards. But to change them to create a new engine, is that reasonable? I would almost prefer that. This would mean, that you change the `.options` dict to a NamedTuple.

dwhswenson:
> But to change them to create a new engine, is that reasonable? I would almost prefer that. This would mean, that you change the `.options` dict to a NamedTuple.

I also think this sounds good. It's weird to have storage misrepresent things, especially things you're likely to reload and modify later like an engine. And since you've already created the really convenient way of making a new engine with only a few modified options....

jhprinz:
> Agreed---the only reasonable options are (1) let the exception through, (2) reject the trajectory and back up to the last accepted sample without NaNs, or (3) retry from the last good frame (which we have established may have an incorrect acceptance probability).

Okay, then we have

```
    on_nan : str
        set the behaviour of the engine when `NaN` is detected. Possible is
        1.  `fail` will raise an exception `EngineNaNError`, user can catch it (like movers do) or let it fail
        2.  `retry` will rerun the trajectory in engine.generate, that causes incorrect acceptance

on_error : str
        set the behaviour of the engine when an exception happens. Possible is
        1.  `fail` will raise an exception `EngineError`, user can catch it or let it fail
        2.  `retry` will rerun the trajectory in engine.generate, that causes incorrect acceptance

on_max_length : str
        set the behaviour if the trajectory length is `n_frames_max`. If `n_frames_max == 0` this will be ignored and nothing happens. Possible is
        1.  `fail` will raise an exception `EngineMaxLengthError`, user can catich it
        2.  `stop` will stop and return the max length trajectory (default)
        3.  `retry` will rerun the trajectory in engine.generate, that causes incorrect acceptance
```

fail is default and the error contains the last good trajectory for whatever use.

maxlength will fail if `len(traj) > n_max_frames` but return a traj of `len(n_frames_max)` as last working copy so that no generated trajectory is longer than n_frames_max.

Did I miss something?

dwhswenson:
List of approaches sounds good to me.

I think the phrase `that causes incorrect acceptance` is a little weird (`that` refers to what exactly? the `retry` approach, I think, but I read it as the trajectory). I think you mean `does not satisfy detailed balance`, right?

Is there a reason not to make the `fail` option default for all 3?

jhprinz:
> Is there a reason not to make the fail option default for all 3?

No, that would be my preferred choice, but currently we do `stop`

dwhswenson:
> No, that would be my preferred choice, but currently we do `stop`

Well, that's only because we didn't have the option to raise an error before, right? ;-) I'm fully in support of changing that default!

jhprinz:
> I'm fully in support of changing that default!

Good. Will do that. And then it is ready to merge. Once it passes tests.

jhprinz:
While I am at it. To unify the options of an engine, would it be okay to have the integrator as a regular attribute, like openmm has? We have the topology this way and it seems more consistent that way.

jhprinz:
I think I would make a separate PR, because this affects all ToyEngines, everywhere. So not now...

dwhswenson:
> While I am at it. To unify the options of an engine, would it be okay to have the integrator as a regular attribute, like openmm has? We have the topology this way and it seems more consistent that way.

`DynamicsEngine` shouldn't have it (and shouldn't have `topology` as an attribute, either). I'm not sure why `ToyEngine` still has a `topology`. When I originally did that, it was just to shoehorn it into something that looked like what you'd done for OpenMM -- several parts of the code *required* a `topology` attribute to get, e.g., `n_atoms`. `ToyEngine` *should* have a `potential_energy_surface` (or just `potential`) object, which should contain the information equivalent to a topology. I think the major improvements you made about the engine, snapshot features, etc., should make that really easy clean up.

I think the way it should work is this: `DynamicsEngine` has a minimal set of initialization parameters (which are then attributes). Each subclass should have a different set that makes sense for that subclass. `integrator` makes sense for both `ToyEngine` and `OpenMMEngine`, but maybe there would be some other engine where it was redundant. `topology` makes sense for `OpenMMEngine`, but does not make sense for `ToyEngine` (which should have a `potential_energy_surface` instead). It's okay to add extra parameters to the initialization of subclasses, since we don't expect them to be perfectly interchangeable.

See also #65.

dwhswenson:
> Good. Will do that. And then it is ready to merge. Once it passes tests.

... you haven't pushed yet... waiting eagerly! :-) (Still at the office; would love to run the AD example overnight)

jhprinz:
> I think the way it should work is this: DynamicsEngine has a minimal set of initialization parameters (which are then attributes). Each subclass should have a different set that makes sense for that subclass. integrator makes sense for both ToyEngine and OpenMMEngine, but maybe there would be some other engine where it was redundant. topology makes sense for OpenMMEngine, but does not make sense for ToyEngine (which should have a potential_energy_surface instead). It's okay to add extra parameters to the initialization of subclasses, since we don't expect them to be perfectly interchangeable.

Alright. What I thought and wanted. So we clean this up shortly after this. DynamicsEngine now only has stuff about the number of steps, maximum frames and all the error handling which makes sense for all of them. These are options. All the really specific stuff are parameters on the constructor.

Will need a few more minutes. Need to change to named tuple .

dwhswenson:
> Alright. What I thought and wanted. So we clean this up shortly after this.

Excellent. One question: will this mean I have to re-run the AD examples again? It sounds like this will break backward compatibility. (That's fine, and I absolute *must* start these runs ASAP so I can have the long fixed TPS run done in time for my talk next week. Just checking the situation.)

> Will need a few more minutes.

No worries. I'm working on visualizations of how we use ensemble to do a lifetime calculation (rate for long MD) for that talk. Plus, the custom strategy is running its last version, and has 12000 seconds left. ;-)

jhprinz:
Actually. The namedtuple stuff is kind of weird. Let's postpone that to the engine cleanup. It is kind of messy changing all the `options['xyz']` to `options.xyz`

dwhswenson:
> The namedtuple stuff is kind of weird. Let's postpone that to the engine cleanup.

Sure... it's also already immutable-enough, if we make the habit of accessing it as `engine.foo` instead of `engine.options['foo']`. (Because `engine.foo = bar` doesn't work, right?)

dwhswenson:
(Although I like the extra certainty in setup from a namedtuple; removed the need for some of the "check allowed options" code you wrote)

jhprinz:
nosetests work. Now the real ones...

jhprinz:
>  (Because engine.foo = bar doesn't work, right?)

Yes, it does not.

jhprinz:
That's it. Ready to be merged.

Later engine cleanup will be separate PR.

dwhswenson:
Great! I've been reviewing recent changes as you made them, so I'll merge this as soon as it passes tests. Then I'll start giving it some thorough stress-tests on the AD system starting tonight, so pretty soon we should be certain that it fixed the NaN problem! @jchodera : once this is merged, it would really help if you can check that this solved your NaN problems, too!

PR #571, opened 2016-08-23
Rename states in AD TPS examples

dwhswenson:
"Also, Dude: Alpha and beta are not the preferred nomenclature. C7-equatorial and alpha_R, please!"


dwhswenson:
(As the wise Walter Sobchak would have told us.)

This will also include paper-ready versions of the AD TPS examples. That's why it is still WIP.

dwhswenson:
I'm going to call this ready for review.

There's still more clean-up to do in order to make the exact figures I want for the TPS section of the paper, but the main calculation has been run. This PR includes some more rearrangement of examples (moving the AD TPS into its directory). In addition, I split the AD TPS analysis into three files: one for fixed-length, one for flex-length, and one that combines the two (and requires that you run both).

And of course, as the title says, it corrects the name of the states.

I'd like to merge this so that I can put some more effort into the documentation PR that has the file-moving in this as a requirement.

jhprinz:
looks good.

Issue #566, opened 2016-08-16
Remove unused/redundant code


jhprinz:
When reviewing the coverage I realized that we have some functions that are actually not used anymore or redundant. I think we should remove this to avoid confusion. I will start compiling a list of things I think we can safely remove and once we agree we can make a PR and remove the code. 

#### Where to use `#pragma: nocover
- [x] Check all the fail saves (e.g. `RuntimeError` as well)?
- [x] A function call like `if xyz is None: return None`?
- [x] Empty functions and abstract ones
- [x] Set operations that return `True` or `False`, e.g. `FullEnsemble.can_append ... return True`

#### Code to be removed

- [x] `Trajectory.n_frames` (we use `__len__`)
- [x] `Trajectory.prepend` (not used in regular python lists)
- [x] `Trajectory.n_atoms` use `traj[0].n_atoms`, `Trajectory.n_atoms` should return the number of atoms for each snapshot in the trajectory by passing to snapshot
- [x] `Trajectory.n_spatial` use `traj[0].n_spatial`, `Trajectory.n_spatial` should return the number of spatial dimensions for each snapshot in the trajectory by passing to snapshot
- [x] `SampleSet.__add__` I think we use `.apply_samples`
- [x] `Sample.__call__` returns the trajectory? Is there a reason for it
- [x] `NetCDFPlus.KeyDelegate` Not used anywhere.
- [x] `NetCDFPlus.clone_store` Not used anymore
- [x] `Storage.template`. `Storage.topology`, `Storage.n_atoms`, `Storage.n_spatial` since we can have mutiple snapshots this is not meaningful and can easily be replaced by `storage.snapshots[0].topology`, etc
- [x] `Ensemble._lencheck` Not used. Was originally to speed up testing for a trajectory being part of an ensemble by using length ranges.
- [x] `EnsembleCombination._continue_fnc` not used anymore
- [x] `Ensemble.use_shortcircuit` we strictly use `True` and so we could remove it
- [x] `SymmetricDifferenceEnsemble` and `RelativeComplementEnsemble`
- [x] `ChainDict.Transform` Not used anymore
- [x] `Topology.subset` We agreed to remove that
- [x] `SampleSet.__eq__` replace by `set(self.samples) == set(other.samples)`

#### Things to leave the way they are

- [x] `AppendedNameEnsemble` Is this still needed? Not sure.
[...]

PR #572, opened 2016-08-23
Remove unused part I

jhprinz:
Removes unused and legacy parts we agreed upon. See #566


jhprinz:
I have done all the things besides test cleanup in #566. I added a message to the `iter_extendable_slices` function that it is experimental and not properly tested. We can do that later, if we want. Otherwise this is ready to be merged when it passes.

jhprinz:
great. Ready to merge. I removes lot of stuff and by that increases coverage. I needed to change some copy to replace by equivalent functions.

openpathsampling/engines/trajectory.py
```diff
@@ -99,18 +99,6 @@ def reversed(self):
...
-        """
-        Prepend a snapshot
-
-        Just convenience method to replace insert(0, snapshot)
-        """
-        self.insert(0, snapshot)
-        # And a generation of scientist-programmers who grew up learning
-        # "OPS trajectories are just Python lists" scream in pain when they
-        # find this after googling "python list.prepend not working".
-        # (Blame JHP. This was his doing.) ;)
```
dwhswenson: Alas, the end of one of my favorite comments in our code.

Still, for the better.

openpathsampling/sample.py
```diff
@@ -119,14 +121,15 @@ def __setitem__(self, key, value):
...
     def __eq__(self, other):
-        if len(self.samples) == len(other.samples):
-            for samp1 in self.samples:
-                if samp1 not in other.samples:
-                    return False
-
-            return True
-        else:
-            return False
+        return set(self.samples) == set(other.samples)
```
dwhswenson: can you add a length check to this as well? you could have duplicated samples in a sample set, then I would argue that these wouldn't be equal

dwhswenson:
Could you clear outputs/check out clean versions of AD TPS notebooks? I'd like the stuff I'm running now to have a chance at being the last versions. Unless there's new code from yours that is actually necessary (most of it looks like temporary debug that isn't needed for example purposes)

dwhswenson:
A couple small suggestions, otherwise this looks good! Nice to bump up 0.5% in coverage for the code equivalent of picking things up from the living room floor. :-)

openpathsampling/engines/trajectory.py
jhprinz: I might add some other place for you in the future :)

openpathsampling/sample.py
jhprinz: Important, I checked the code again and it seems that we agreed before that one sample can only be present once. If you apply samples then the one with the same replica number will be replaced. So, at least by applying you cannot have the two samples with the same replica number present. Just to make sure I will change it to using a `collections.Counter` which will also compare the numbers of times a specific sample is present.

jhprinz:
If it passes, it is ready.

jhprinz:
> Nice to bump up 0.5% in coverage

Yes, I like this way of writing tests...

openpathsampling/sample.py
dwhswenson: `collections.Counter` is even better than a length check. My concern is that there are also `append` and `extend` methods for `SampleSet`, which could create a `SampleSet` where you'd get false equality from the set-based version. (Didn't there used to be an `apply` method which did the cleanup to make sure there was only one of each? There should be such a thing, but I don't see it in the code now.)

dwhswenson:
Merging -- please note my question in the [line note above](https://github.com/openpathsampling/openpathsampling/pull/572#discussion_r77439248). That's not to be done in this PR, but we might need to add that behavior.

PR #573, opened 2016-08-23
Fix OpenMM Precision Bug

jhprinz:
There was a bug (presumeably) by some merge faults so that the openmm_platform_properties were saved with an engine, but not used in the construction of the actual OpenMM Instance.

This will fix this. On my machine now OpenCL does not work with mixed precision. This seems to be the cause of all the `NaN`s.

@jchodera This is an important fix for you, too, if you use OpenCL!!!!


dwhswenson:
Took a couple re-runs to pass tests, but merging now! @jhprinz, thanks for figuring this out last night, after hours of text messages as we tried to figure out what the problem might be!

dwhswenson:
Fingers crossed that this fixes everything!

PR #574, opened 2016-08-24
Move css to separate file

jhprinz:


jhprinz:
I move the CSS for visualization to a `vis.css` file in a `css` folder. That cleans up the code a lot. We can also come up with different styles :) now.

jhprinz:
We need to decide, what to include at some point and edit the `manifest.in` file.

jhprinz:
I got this to work, we can now really move files to any place. We just should decide how to do that. Could be part of 0.92 or so...

dwhswenson:
Just to see if I'm understanding: you're saying that we should solve #593 before merging this in for a release, right?

That is, the *ability* to move files is already done (and in this PR), but we haven't selected a *standard location* for them, and should before this is merged?

(Or if not, should this then be reviewed now and included in 0.9.1? I'm fine either way.)

jhprinz:
> That is, the ability to move files is already done (and in this PR), but we haven't selected a standard location for them, and should before this is merged?

Yes, I use `openpathsampling/css` for the css files but we should agree where to put these files. I just wanted to stress that I figured out how to make the files appear in the package and how to access them later.

jhprinz:
REady to be review and merged.

PR #575, opened 2016-08-24
Simple store creation

### LINKED ISSUE #554: Weird storage situation: Add warning if saving without designated store?
When I made the `MSOuterTISInterface` object, I forgot to add a store for it.

What's weird is that it seemed to store just fine. In fact, it even reloaded fine. However, it is filled with equivalent (but different) objects, such that the dictionaries in it, etc, don't work.

Maybe this is a weird effect of UUID?

To see the weirdness, run this after running the MSTIS sequence:

``` python

storage = paths.AnalysisStorage("mstis.nc")
network = storage.networks[0]

print "Old outers:", network.ms_outer_objects

outer = network.ms_outer_objects[0]
new_outer = paths.MSOuterTISInterface(outer.interface_sets,
                                      outer.volumes,
                                      outer.lambdas)
print "New outer, built from old:", new_outer

print "\nNow let's save it:"
new_store = paths.Storage("test.nc", "w")
new_store.save(new_outer)
```

My output:

``` pytb
Old outers: [<openpathsampling.high_level.ms_outer_interface.MSOuterTISInterface object at 0x11d143490>]
New outer, built from old: <openpathsampling.high_level.ms_outer_interface.MSOuterTISInterface object at 0x11985fd90>

Now let's save it:
  File "test_weird_store.py", line 18, in <module>
    new_store.save(new_outer)
  File "/Users/dwhs/Dropbox/msm-tis/openpathsampling/netcdfplus/netcdfplus.py", line 632, in save
    obj.__class__.__name__)
RuntimeWarning: Objects of type 'MSOuterTISInterface' cannot be stored!
```

I noticed this because, even though the MSOuter the reloaded version had different interface sets in them (making dictionaries fail). It's really cool that it stored it (but where? I couldn't find the store containing it!), but this is the sort of thing where a warning on that kind of save might prevent some debugging time.

jhprinz:
Resolves #554 . 

With this update you can create `store`s for `netcdf+` simply by doing

``` py
st = Storage(name, 'w')
st.add_class_json_store('mystore', object_cls).set_caching(True)
```

after that you can normally store and load objects

``` py
st.mystore.save(my_obj)
print len(st.mystore)
```


jhprinz:
See `test_netcdfplus.ipynb` for a little demonstration.

jhprinz:
Ready, when it passes.

dwhswenson:
The code looks fine, but a couple questions before I merge:

* How will this affect storage split across multiple files with the new UUID tricks? I can imagine that one file has the store added, and another doesn't; could that cause problems? (or do all objects of a type always go to the same file? although I thought we weren't doing that so snapshots/trajectories/samples could be combined from multiple files)
* Would it be possible (maybe in the future) to have some clever fallback that tries to do this if saving an object that has no registered store? (Or is that completely impossible?)

jhprinz:
> How will this affect storage split across multiple files with the new UUID tricks? I can imagine that one file has the store added, and another doesn't; could that cause problems? (or do all objects of a type always go to the same file? although I thought we weren't doing that so snapshots/trajectories/samples could be combined from multiple files)

Good point. I did not think about this, but currently the splitting works using fallbacks. So, if you add a store to one file and this file is the fallback for another. Then you cannot access your new store from the other file using the fallback. This would require to add the store to the _main_ file also which can then delegate. Example: If your main file contains all the new objects and you save snapshots in another place you are fine, because by default all files contain a snapshot store.

> Would it be possible (maybe in the future) to have some clever fallback that tries to do this if saving an object that has no registered store? (Or is that completely impossible?)

You can easily add a default behavior that does exactly this. But a object does not know about it storage mechanism (where there could be many). If you want a real store, then it also needs a name. Which we would have to add to the base_class. Something like `_json_store_name `... That would be okay. I mentioned before that currently the Base(Super)class takes this place.

```py
class MyObject(StorableObject):
    _json_store_name = 'coolstuff'

c = MyObject()

storage.save(c)  
# this creates the `.coolstuff` json based store if not taken
# and implicitely call storage.add_json_class_store(c.base_cls)

print len(storage.coolstuff)
```

I would have a flag to switch this behavior off. Default is on. I would only work, if
the `_json_store_name` is set in the base cls.

jhprinz:
This should work as explained above. You can now very simply create new storages, if you want to. I put this onhold because it will affect a tiny bit saving performance and I am not sure to what extend yet. Basically the overhead is to always check, if we might create a store because it could be possible. I think this is not noticeable, but it could be.

dwhswenson:
Should this be reviewed now or not? I take the "onhold" label to mean that it isn't intended to move forward any time soon.

dwhswenson:
@jhprinz : What's the current status with this? As I'm creating new objects that do different kinds of analyses, it would be really useful to be able to easily add a new store.

>  I put this onhold because it will affect a tiny bit saving performance and I am not sure to what extend yet. Basically the overhead is to always check, if we might create a store because it could be possible. I think this is not noticeable, but it could be.

I'm not sure if I fully understand what you mean here, but this seems like it is an ideal place for a `try:/except:`. The `try` covers the common case that the store exists; the `except` catches when it isn't. Slow if the exception happens (which will occur only once), but otherwise very fast.

dwhswenson:
Marking this for closure; not really planning on making any more major changes to netcdfplus, as it will be replaced by SimStore. I will leave this open for comment for at least 7 days, closing no earlier than Wed 30 Dec 06:00 GMT (07:00 local).

PR #579, opened 2016-09-12
OpenMM masses snapshot feature

dwhswenson:
Currently, the `RandomVelocities` snapshot modifier can't be used with OpenMM, because we don't have a standard way of accessing masses. After this PR, both OpenMM snapshots and toy snapshots will have a `.masses` property, which should return the right thing.

There's a bit of strangeness on the OpenMM side because we try to get the mass from the OpenMM `system` object, and use the `topology` as a fall-back. This is a result of having different engines (e.g., the OpenMM `FileEngine`) create these snapshots. If the mass stored in `system` does not match the `topology`, things could get weird.

- [x] Draft OpenMM masses
- [x] Tests for OpenMM masses
- [x] Update `RandomVelocities` modifier to use `snap.masses`
- [x] Test `RandomVelocities` with both OpenMM and Toy engines
- [x] Add nosetest to show that running a step works after randomizing velocities


dwhswenson:
Passes tests, ready for review! At this point, the committor analysis should also work with OpenMM -- previously, it failed because we didn't have an easy way to get the masses.

This also adds some commented-out `strip_units` functions. Not sure if we want these or not, but I put them in for now. I had a lot of frustration over the fact that OpenMM (`simtk.unit`) doesn't actually treat `amu` as a true unit of mass. It is mass/mole. This is why the `masses` feature that I wrote gives you the option of `masses_per_mole` or just `masses`. I assume that when someone asks for `snapshot.masses` as part of a calculation (e.g., to set the velocities according to the Boltzmann distribution), they want the actual mass, so `masses` returns the actual mass. This differs from OpenMM, where `masses` returns mass/mole. (I expect OpenMM does that so that energy units come out as energy/mole, which is useful, but it does make other calculations annoying and incompatible with the way other engines would work.)

jchodera:
> I had a lot of frustration over the fact that OpenMM (simtk.unit) doesn't actually treat amu as a true unit of mass. It is mass/mole. This is why the masses feature that I wrote gives you the option of masses_per_mole or just masses. I assume that when someone asks for snapshot.masses as part of a calculation (e.g., to set the velocities according to the Boltzmann distribution), they want the actual mass, so masses returns the actual mass. This differs from OpenMM, where masses returns mass/mole. (I expect OpenMM does that so that energy units come out as energy/mole, which is useful, but it does make other calculations annoying and incompatible with the way other engines would work.)

I'm confused at the problem here. Isn't this confusion rectified simply by using an appropriate value for the Boltzmann constant or the ideal gas constant? For example, if you use
```python
from simtk import unit
# Define appropriate form of Boltzmann constant
kB = unit.BOLTZMANN_CONSTANT_kB * unit.AVOGADRO_CONSTANT_NA
# These things work fine for sampling from the Maxwell-Boltzmann distribution when masses are in amu
kT = kB * temperature
beta = 1.0 / kT
```
This is a really common idiom when using OpenMM. It seems a bit silly to go crazy with `strip_units` functions just because particle masses are defined in `amu`

dwhswenson:
> This is a really common idiom when using OpenMM. It seems a bit silly to go crazy with `strip_units` functions just because particle masses are defined in `amu`

Current implementation doesn't include `strip_units` -- they're commented out.

The problem is that if you write, as I did, a function that requires mass to be in units of actual mass, OpenMM doesn't work. So what I did do differently is to say that `snapshot.masses` *always returns actual mass*. This specific case dealt with $k_B$, but I could have just as easily been looking at something based on momentum (say I wanted to look use this in some quantum dynamics approach, where classical approximations of operators are typically based on momenta). In that case, redefining $k_B$ gives me nothing.

`snapshot.masses_per_mole` returns the standard OpenMM masses in per-mole units.

jhprinz:
Looks good. Does that mean we can implement the two-way shooting now?

dwhswenson:
Two-way requires a bit more than this, but two-way + this PR should make "aimless shooting" pretty easy.

Two-way is on my list of upcoming "modules" to implement, hopefully to be officially recognized by my grant. I'm hoping that we'll have things in place so that "aimless" can be implemented at the ESDW in Austria.

Issue #580, opened 2016-09-13
Snapshot copying and `StaticsContainer`/`KineticsContainer`


dwhswenson:
In working on #579, I've come across what I think is a bug in the behavior of `snapshot.copy()` when using OpenMM snapshots.

Short version: based on our (long!) discussions in #429 and #444, I thought that `snapshot.copy()` was supposed to essentially be a deep copy for data (shallow for other things, like topology/engine). Importantly, it is a deep copy for numpy data, like in the toy engine's velocities and coordinates. However, it doesn't seem to work this way for the OpenMM snapshot, where the coordinates and velocities are wrapped in the `statics` or `kinetics`.

Quick example:

```python
import numpy as np
import openmmtools as omt
import openpathsampling.engines.toy as toy_engine
import openpathsampling.engines.openmm as omm_engine
import simtk.unit as u

# test a toy snapshot
init_coord = np.array([1.0, 2.0])
init_vel = np.array([3.0, 4.0])
toy_snap = toy_engine.Snapshot(coordinates=init_coord, velocities=init_vel)
toy_copy = toy_snap.copy()
toy_copy.velocities[1] = 1
print toy_snap.velocities, toy_copy.velocities
# [ 3.  4.] [ 3.  1.]

# test an openmm snapshot
sys = omt.testsystems.AlanineDipeptideVacuum()
omm_snap = omm_engine.snapshot_from_testsystem(sys)
# initial velocity
print omm_snap.velocities[0]
# [ 0.  0.  0.] nm/ps

omm_copy = omm_snap.copy()
omm_copy.velocities[0] = [1.0, 2.0, 3.0] * u.nanometer / u.picosecond
print omm_snap.velocities[0], omm_copy.velocities[0]
# [ 1.  2.  3.] nm/ps [ 1.  2.  3.] nm/ps

print hex(id(omm_snap.kinetics)), hex(id(omm_copy.kinetics))
# 0x11104e510 0x11104e510
```

In general, we need a solution for this to do 2-way shooting. It needs to work with either the toy engine or OpenMM, so it can't *require* `statics` or `kinetics` objects. The things to balance here are:

* we don't want to always make new `statics`/`kinetics` when we make a copy of the snapshot
* we want the option to tell the snapshot to make a deep copy of container attributes like `statics` or `kinetics`
* we want this to be independent of the details of the snapshot: it shouldn't care whether we have container objects or direct numpy arrays (i.e., OpenMM vs. Toy snapshots)

[...]

PR #581, opened 2016-09-14
Implement `Copy_with_replacements`

jhprinz:
Addresses #580. 

As discussed with @dwhswenson . Ready to be reviewed


dwhswenson:
Thanks for the quick implementation! Took me a second to see why `is_reversed` would always be `False`, but yes, that's definitely the right behavior. All of this looks good -- waiting for it to pass tests, and then will merge. (Which might be tomorrow if I go to bed before tests pass.)

dwhswenson:
Side note: the `quiet_loggers` thing is exactly what I've been planning to do as part of a refactoring of some of our testing stuff.

dwhswenson:
Finally passed tests! Merging!

PR #588, opened 2016-09-19
Reorganize `examples` (part 2)

dwhswenson:
This will include as much as I can other than the AD TPS examples, which are waiting on #571.

(Already includes fixes for a couple errors I noticed after @jhprinz merged #568.)

- [x] Move everything except the AD TPS examples out of the ipython directory
- [x] Restructure ipynb tests to be easier to run locally (and match Travis runs)
- [x] Switch AD input to `resources` directory
- [x] Add new analysis testing
- [x] Update READMEs.

This does almost all of #490: next steps will be done in #571.


jchodera:
The new `examples/` directory structure looks SO MUCH BETTER. Thanks!

dwhswenson:
Passes tests, ready for review! Note that this includes the fix in #597, so that should also be reviewed.

Includes a new trick for testing (perhaps avoiding the common stochastic failures!)

Instead of analyzing the file we generate every time, I `curl` down a 1000-step run I did before. At the top of the analysis notebook is a little logic that checks whether such a file exists; if so, use that file for analysis. Otherwise, use the one generated by the example sequence.

This way users will still analyze their own data, but we can pass the tests without having to run longer simulations. Both MSTIS and MISTIS now run 100 steps (still worth doing to avoid errors in sampling), but analysis should no longer require luck to pass.

One thing: these files are currently hosted in my personal Dropbox account. I'd like that to change soon. A few options:

1. Put them in this repo. The files shouldn't change until version 2, so it won't add a lot to the repo.
2. Put them in their own repo.
3. Put them directly on the openpathsampling.org website.

Personally, I'd prefer the third of those options, but I want to run it by @jchodera first.

jchodera:
> Includes a new trick for testing (perhaps avoiding the common stochastic failures!)
> Instead of analyzing the file we generate every time, I curl down a 1000-step run I did before. At the top of the analysis notebook is a little logic that checks whether such a file exists; if so, use that file for analysis. Otherwise, use the one generated by the example sequence.

I use travis caches for this. Example [here](https://github.com/choderalab/alchemy/blob/master/.travis.yml#L16-L18), with cache-detection code [here](https://github.com/choderalab/alchemy/blob/master/alchemy/tests/test_alchemy.py#L520-L532).

> One thing: these files are currently hosted in my personal Dropbox account. I'd like that to change soon. A few options:
> Put them in this repo. The files shouldn't change until version 2, so it won't add a lot to the repo.
> Put them in their own repo.
> Put them directly on the openpathsampling.org website.

We can put them in the amazon bucket if they're not too big, but that also charges directly to my credit card, so I'd prefer the travis cache if it isn't too hard.

dwhswenson:
> but that also charges directly to my credit card,

Good point, which I had overlooked. I'll definitely look into the Travis cache (which effectively puts it on Travis's S3 account, but I suspect they have better deals on transfers than you do).

The current files are ~100MB. Not huge, but when we expect over 1000 transfers a year (our current test pace is greater than that), it does add up.

There also might be a possibility of hosting it through my current grant's repository, which is supposed to include examples and test data, as well as software. If I have trouble getting Travis caches to work, I'll look into that.

For now, it's fine in my Dropbox; only problem is that other people can't easily replace it if that's needed (but could easily re-run the example and change the `curl` URLs, so it's not like my access to that is *so* irreplaceable).

jchodera:
100GB is only $9/year, but it sounds like travis caching is the best option if it's not hard to manage.

jhprinz:
Excellent. Can you revert the `pyqt` thing. You posted that it is fix. Otherwise it seems fine.

dwhswenson:
To clarify -- pyqt still needs to be pinned to 4. Just not 4.11.4-py27_3. I can switch to the less restrictive pin, but I hope we won't need any pin at all soon.

Do you want the less restrictive pin in this PR?

jhprinz:
No, lets keep this for now. We can fix it later...

jhprinz:
Merging...

PR #589, opened 2016-09-20
Snapshot features: `n_degrees_of_freedom` and `instantaneous_temperature`

dwhswenson:
Note: these are not intended to be features that are required of every engine. However, they could be very useful.

In particular, I'm pretty sure we want something like `n_degrees_of_freedom`: that shows up in enough other calculations, and will be very much engine dependent. Not having this leads to all sorts of ugliness when I wanted to check instantaneous temperature in the AD TPS example: a lot of code that is OpenMM-specific.

Whether we directly include `instantaneous_temperature` is a little less vital, in my opinion. A competent user can handle that little bit of code for their engine. For now, I've added it for consideration.

- [x] Draft `n_degrees_of_freedom` and `instantaneous_temperature` for OpenMM
- [x] Draft `n_degrees_of_freedom` and `instantaneous_temperature` for Toy engine
- [x] Tests for `n_degrees_of_freedom` and `instantaneous_temperature` for OpenMM
- [x] Tests for `n_degrees_of_freedom` and `instantaneous_temperature` for Toy engine
- [x] Add docstrings

### DIFF
--- a/openpathsampling/engines/openmm/engine.py
+++ b/openpathsampling/engines/openmm/engine.py
@@ -374,7 +374,11 @@ class OpenMMEngine(DynamicsEngine):
             self.simulation.context.setVelocities(snapshot.velocities)
 
             # After the updates cache the new snapshot
-            self._current_snapshot = snapshot
+            if snapshot.engine is self:
+                # no need for copy if this snap is from this engine
+                self._current_snapshot = snapshot
+            else:
+                self._current_snapshot = self._build_current_snapshot()
 
     def generate_next_frame(self):
         self.simulation.step(self.n_steps_per_frame)
--- a/openpathsampling/engines/openmm/features/__init__.py
+++ b/openpathsampling/engines/openmm/features/__init__.py
@@ -1,3 +1,4 @@
 from openpathsampling.engines.features import *
 from openpathsampling.engines.features.shared import StaticContainer, KineticContainer
 import masses
+import instantaneous_temperature
--- a/openpathsampling/engines/openmm/features/instantaneous_temperature.py
+++ b/openpathsampling/engines/openmm/features/instantaneous_temperature.py
@@ -3,6 +3,13 @@ import simtk.unit as u
 
 @property
 def n_degrees_of_freedom(snapshot):
+    """
+    Returns
+    -------
+    n_degrees_of_freedom: int
+        number of degrees of freedom in this system (after accounting for
+        constraints)
+    """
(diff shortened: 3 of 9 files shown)


dwhswenson:
Came across a problem with this: Need @jhprinz's advice on how to handle it best with respect to proper storage.

Problem:

```python
import openpathsampling.engines.openmm as omm_engine
import openmmtools as omt

test_system = omt.testsystems.AlanineDipeptideVacuum()
template = omm_engine.snapshot_from_testsystem(test_system)
engine = omm_engine.Engine(
    topology=template.topology,
    system=test_system.system,
    integrator=omt.integrators.VVVRIntegrator()
)

engine.current_snapshot = template

print type(template.engine)
# <class 'openpathsampling.engines.openmm.tools.OpenMMToolsTestsystemEngine'>
print type(engine.current_snapshot)
# <class 'openpathsampling.engines.openmm.tools.OpenMMToolsTestsystemEngine'>

snapshot = engine.generate(engine.current_snapshot, [lambda x, foo: len(x) <= 1])[0]
print type(engine.current_snapshot.engine)
# <class 'openpathsampling.engines.openmm.engine.OpenMMEngine'>
```

I think we want `current_snapshot` to always return an a snapshot with the `OpenMMEngine` as `engine`. Otherwise, accessing these attributes before we've run the simulation (e.g., first frame) will crash, because `OpenMMToolsTestsystemEngine` has no `simulation` object to find the 

I see two approaches to fix this. Either make sure that we have the right engine when we ***set*** `current_snapshot`, or make sure we have the right engine when we ***get*** `current_snapshot`. The implementation I'm playing with now is based on giving the correct `engine` when we ***set*** it. It involves adding an extra `if` statement in the `setter`:

```python
            # After the updates cache the new snapshot
            if snapshot.engine is self:
                # no need for copy if this snap is from this engine
                self._current_snapshot = snapshot
            else:
                self._current_snapshot = self._build_current_snapshot()
```

This way the thing in `self._current_snapshot` always comes from this engine.

jhprinz:
This approach should work and I also prefer to directly alter the snapshot once you set it. For storage this should not matter at all since the stored engine is completely independent of the current snapshot. 

To make sure: You want this so that for sure each snapshot from the engine you have access to all attributes that snapshots from the engine should have. I think that makes sense. Go ahead and add this in this PR if you want.

dwhswenson:
Passes tests, ready for review!

jhprinz:
Looks good.

PR #595, opened 2016-09-22
[WIP] Fix for the 5000 max length problem

jhprinz:
A quick fix


jhprinz:
Well, it passed, but that does not mean, that it is fixed...

dwhswenson:
I don't understand how this could possibly fix it. Is one of the following statements not true?

1. The input to the function that causes the problem is a single trajectory.
2. The change you made affects how that function handles input consisting of many trajectories.

jhprinz:
Exactly my point...

Some obeservations so far. The function only gets an A->X->A path and it will really try to extend it. All the sorting, splitting will do nothing just pass on the path to be extended FW and BW.

This suggests that the error is somwhere else. Or we had much longer trajectories before but they were not detected or the limit was ignored.

dwhswenson:
My best guess is that we're seeing a spontaneous transition during the extension. But I've seen it too frequently for that to feel right.

jhprinz:
@dwhswenson Is this still a problem? Do we want to put in onhold for now?

dwhswenson:
Yes, I still see this on occasion. Is anything in this PR actually useful (and not already done in another PR?)

I feel like this should be an issue, not a PR -- we have a problem, but haven't identified the source of it yet, and there's no specific plan for how to fix it. For now, we get a stochastic error on some tests (1/5 maybe?).

jhprinz:
Agreed. Will move to an Issue

Issue #585, opened 2016-09-17
Undo pinning of pyqt to version 4


dwhswenson:
Once `conda` ~~or `matplotlib`~~ fixes the compatibility issues between `conda`-installed `matplotlib` and `pyqt` 5, we should undo #584, which is intended as a temporary fix to get the build working. See also #583.

Creating this issue so we remember to come back to it in the future. On hold until upstream makes the fix.

dwhswenson:
Relevant issue to watch: https://github.com/ContinuumIO/anaconda-issues/issues/1068

dwhswenson:
Resolved by #655.

PR #597, opened 2016-09-22
Pin pyqt 4.11.4-py27_3

dwhswenson:
As the folks at `conda` are working to make it so we can unpin `pyqt` overall, things got temporarily a little worse. Now we need to pin to a specific build.

From the response I got at https://github.com/ContinuumIO/anaconda-issues/issues/1090, it sounds like it'll be fixed soon. But this should handle it in the meantime.

### DIFF
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -28,7 +28,7 @@ requirements:
     - mdtraj
     - svgwrite
     - networkx
-    - pyqt 4*
+    - pyqt 4.11.4 py27_3
     - matplotlib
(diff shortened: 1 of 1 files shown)


dwhswenson:
Nevermind: the people at ContinuumIO just fixed it: https://github.com/ContinuumIO/anaconda-issues/issues/1090#issuecomment-249202001

And builds of #589 show that it's working for us. However, the same fix does show up in other PRs, so we can partially unpin (to `4*`) in those PRs, or just wait for the larger problem in #585 to be solved.

jhprinz:
Yes!

PR #600, opened 2016-09-26
Two-Way Shooting

dwhswenson:
Two-way shooting: requires a snapshot modifier in addition to a shooting point selector.

I think there can be a difference between forward-first shooting and backward-first (depends on the specific ensemble; there is no difference for standard TPS and TIS). Because of this, this is implemented much like one-way shooting: `ForwardFirstTwoWayShootingMover`, `BackwardFirstTwoWayShootingMover`, and a random choice mover `TwoWayShootingMover`.

Also has a move strategy for two way shooting.

- [x] `ForwardFirstTwoWayShooting`
- [x] `BackwardFirstTwoWayShooting` and `TwoWayShooting`
- [x] `TwoWayShootingStrategy`
- [x] Tests


dwhswenson:
Well, once I got working on it, this was easier than expected. Ready for review. Even though it was originally targeted for 1.x, we kinda need it for Traunkirchen, so I guess it'll be part of 1.0!

jhprinz:
looks good. Will merge. Did not look too complicated

PR #601, opened 2016-09-29
Committor improvements

dwhswenson:
As part of a collaboration we're spinning up, I'm running some committor calculations. In the process of that, I found a number of things that could be improved in the committor code.

Only thing that modifies the API is that I changed the output of `committor_histogram` to match the output of `numpy.histogram` and `numpy.histogram2d` (if the `committor_histogram` is a 2D variant). I'd call that more of a bug fix than an API break (and we *are* still 0.9.x, so we can break the API!)

Everything else just makes it easier to work with:

- [x] make `ShootingPointAnalysis.committor_histogram` match numpy's behavior
- [x] separate out the ability to add to the `ShootingPointAnalysis` one step at a time
- [x] make it possible to create a `ShootingPointAnalysis` object without it doing analysis on initialization by giving it `steps=None`
- [x] add class method `ShootingPointAnalysis.from_individual_runs` which takes a list of `(shooting_point, final_state)` tuples and creates the shooting point analysis object
- [x] Add output to `CommittorSimulation` to see how much more of the simulation is left to run.
- [x] Tests
- [x] Docstrings

I'll keep this as WIP for the time being, as I continue starting this new project. Not sure if there is much else I'll need to change, though.


dwhswenson:
I think I've done all that I'll need to with this. Ready for review.

jhprinz:
Looks good. Merging..

PR #606, opened 2016-10-04
Reorganize analysis

### LINKED ISSUE #492: Reorganize `analysis` directory
The `analysis` directory not really all analysis any more. Some objects that were originally designed for after-simulation analysis also became simulation-setup tools (transitions, networks, move scheme).

This issue is mainly to suggest moving files around. I think it should look something like this:

```
openpathsampling/numerics/
|- histogram.py
|- lookup_function.py
|- wham.py
openpathsampling/analysis/
|- replica_networks.py
|- tools.py
|- single_trajectory_analysis.py
|- shooting_point_analysis.py
openpathsampling/???
|- networks.py
|- transitions.py
|- move_scheme.py
|- move_strategy.py
```

One somewhat significant change: splitting `tis_analysis.py` into `transitions.py` (which is mostly what it is) and `analysis/tools.py`.

As for the `???`, this could either be at the root `openpathsampling` layer, or in some subdirectory. I’d prefer a subdirectory, though I’m not sure the name for it. The point is that these modules provide very specific tools for TIS and TPS: they’re mainly relevant if you’re doing those types of path sampling. I think of them as sort of an `app` layer, so that might be a good directory name. I’m open to other suggestions.

There’s also a (seemingly very incomplete) `msm.py` that @jhprinz added there. Where would that belong?

Anyway, before doing this reorganization, I wanted to get feedback from others on what a better structure would be.

dwhswenson:
Resolves #492.

- [x] move things to `numerics/`
- [x] move things to `high_level/`
- [x] split `tis_analysis` into `analysis/tools.py` and `high_level/transitions.py`


jhprinz:
lgtm

PR #608, opened 2016-10-05
Workaround for bug in `mdtraj.Topology.from_dataframe`

jhprinz:
I think this works.

Turns out the bottleneck is `mdtraj.Topology.from_dataframe` which is very slow, but I think this will fix it.

It is kind of an ugly fix. It will increase the residue numbers artificially with multiples of 10000. Then use `.from_dataframe` and revert the changes afterwards. This results in the same structure reading a `.pdb` file would give.


jhprinz:
Have a look.

dwhswenson:
Yeah, it ain't pretty, but it'll do the job.

I'll merge. I did some reading on the PDB file format, and it looks like a correct PDB *must* have the atoms in the order implied by SEQRES the residue order. So users can't do something crazy like jumble the order of the PDB entries.

http://www.wwpdb.org/documentation/file-format-content/format33/sect9.html

But really, can't everyone just start moving to a better file format?

PR #609, opened 2016-10-05
Update Vis Docs

### LINKED ISSUE #337: Visualization questions and cleanup suggestions

jhprinz:
Resolves #337 .

Just an update on what visualization can do. No code changes so merge at any time!


dwhswenson:
This definitely looks like it'll be the sort of thing I was hoping for.

Could you add a version after running it? This will go straight into the documentation, but that's only useful if there's a version that has been run.

Also, did the `join_blocks` option (what I like to call the "boxcars" version, because the trajectory looks like a long train) get renamed to `horizontal_gap`? This is one that I like to use, especially for example code, so I want to know how to make it work!

jhprinz:
Last thing: I looked into the wrong step numbers and turns out:

Step Zero / the initial step has `.change = EmptyMoveChange()`. An EmptyMoveChange is technically rejected for it does not replace samples and hence the initial step is rejected. 
Since its `.change.samples` property is empty it is interpreted as if the initial change did not create any samples and this is what is used to determine the (generating) step. Thus the `*` and not number 0. This also causes the initial step to be show as a rejected step, which is weird since this step does not have an accepting part.

I propose to use an `InitialSampleSetMoveChange` that has all the samples in it that are used and makes this consistent. I can tweak visualization to show old files correctly. Suggestions?

I have also added a new column for the legend to show the range of steps where this sample was in the active set. Where the `step` legend is actually the step in which the sample was produced. An example what confused me first and led to this option was that I used `steps[0:100]` and the last shown step number was `76`. This is because the sample was made in `76` but is active in `76-99`. 

 I still need to look into the question, whether `accepted=True/False` for ReplicaStepList returns the correct list, but other than that I am done.

jhprinz:
> I propose to use an InitialSampleSetMoveChange that has all the samples in it that are used and makes this consistent. I can tweak visualization to show old files correctly. Suggestions?

Or we just use the `AcceptedSampleMoveChange` without a Mover. This will also give the correct result.

dwhswenson:
> Or we just use the `AcceptedSampleMoveChange` without a Mover

This seems reasonable, but if you do this, double check that we're still getting the correct results out of `scheme.move_summary`. I think we will (because it [looks like the summary functions ignore changes with `mover=None`](https://github.com/openpathsampling/openpathsampling/blob/master/openpathsampling/high_level/move_scheme.py#L741)). Just be sure we don't get a weird extra count in there.

Changing the code that I linked above from `move_scheme.py` will have weird effects on SRTIS, so if you change that, you need to check the behavior of SRTIS analysis as well. Preferably, don't change it.

jhprinz:
> Just be sure we don't get a weird extra count in there.

Yes, that should be fine. I did not need to change any of that analysis code. The mover of the initial change is `None` as it should be. And if you really restart there will be no additional step. Hence the counting will not change.

dwhswenson:
@jhprinz: When you have a chance, could you merge `upstream/master` into this (`visdocs`), to remove the commits from the `restart` branch from the diffs to be reviewed here? Thanks!

openpathsampling/snapshot_modifier.py
```diff
@@ -37,6 +38,7 @@ class SnapshotModifier(StorableNamedObject):
     __metaclass__ = abc.ABCMeta
 
     def __init__(self, subset_mask=None):
+        super(SnapshotModifier, self).__init__()
```
jhprinz: Actually I think this was missing. It is necessary for storable objects. In rare cases you might
not do it. Just to point @dwhswenson to a potential mistake.

PR #610, opened 2016-10-06
Fix storage for TPS Network

dwhswenson:
There was a bug in the way TPS networks were stored. This fixes that bug, but in a way that still allows us to open older files. Support for the older files is marked as deprecated, to be removed in 2.0.


jhprinz:
Looks good. Can this be merged?

dwhswenson:
Yep -- hadn't assigned it because I forgot to check after the tests ran.

PR #615, opened 2016-10-10
Fix resaving bug in TPS storage

dwhswenson:
Something I missed in #610. There was a problem when we re-saved a network that had been saved in the old way. This fixes the problem.

Should be immediately ready for review.


jhprinz:
Looks good. Will merge.

PR #616, opened 2016-10-10
Convenience functions for restarting from move scheme

### LINKED ISSUE #611: Inconsistency in initializing `PathSimulators`
This proposes a little change in how to setup a PathSimulator, it will not break the object structure created by PathSimulator objects!
##### Origin of the problem

When ploting the evolution of replicas I realized that the initial step is kind of weird: Currently, when creating a new PathSimulator, its internal step is set to zero and the given sampleset it used as is. The samples are not changed! This was altered from making copy_reset to allow continuing without the PathSimulator introucing the initial step everytime you restart. This would double the samples and hence cause a bias. For the very first initial step this would be necessary. otherwise the samples still point to their parents and in the plots I have in the vis tutorial you can see a first sample that has no step. This is the parent of the one we actually want to use and was stored and shown because we did not reset the very first initial sample when setting up the pathSimulator.

Now the idea: When we want to continue a simulation we load the last step and create a new PathSimulator with the cycle number of the last step (or use the length of steps in the storage). Take the sampleset from the last step. Put all in the PathSimulator and then run to produce new steps.

What if we just created a PathSimulator that does not know about a step or the sample_set but takes an initial `(MC)Step` object, which itself contains these information (and much more). We also start an engine with snapshots and then get new snapshots. THis way we seed the PathSimulator with a step and get new steps. This seems consistent.
##### Example

``` py

# first setup

# this object is even immutable now and could be loaded!
calc = paths.PathSampling(
    move_scheme=storage.schemes[0]
)

# so save it
storage.save(calc)

# set the storage to be used
calc.storage = storage

# set the initial conditions from the step
# which will automatically reset the samples and create a sample_set
calc.current_step = paths.InitialSampleSetStep(samples)

# this knows exactly the mccycle from the step object
mstis_calc.run_until(2000)

# restart by loading

calc = storage.calculations.load('mysampling')

# this might be called implicitly when loading
calc.storage = storage

# set the initial conditions from the step
calc.current_step = storage.steps.last.active,

# this knows exactly the mccycle from the step object
mstis_calc.run_until(10000)

```

This will be compatible with the current approach where we do not save the PathSimulators but we would need to change almost every notebook. This is a hassle, but doable.
##### What about special PathSimulators

Well, besides the current step and the current_sampleset there is might be other information needed to continue, but this needs to be contained in the `MCStep` otherwise you have a state in the PathMover and you cannot continue. Unless, you would know exactly what the last state was and had it saved and reconstruct it by hand. A little like we do now: We take the last mccycle number the last sample_set and the continue.
##### What is the meaning of the mccycle if I change PathSimulators

Well, this is my current missing point. Say you start bootstrapping (well it does not return a step yet, but this is special). Say it does and you have one step bootstrapping and then continue with PathSimulation. Would you then start at `mccyle=0` or `mccycle=1`. That should be up to the user. Usually we treat steps from different simulators separate. Having the PathSimulation object storable is also useful because the step can know about it and we can then sort steps by Simulator used.

``` py
steps = storage.steps.by_pathsimulator(calc)
```
##### What if a PathSimulator does not use steps like Bootstrapping

Well, then we either 
1. say this needs to be initialized differently and takes an initial sample_set to start or continue from
2. we do not care and say this is just special and not a real MCStep generating Simulator
3. We rewrite BootStrapping to also use MCStep
##### Bottomline

I think this is a really reasonable change, but I can understand if we do not want to do it now. It will take a day to make the changes I assume. If not in 1.0 then I guess we have to wait until 2.0 because of some API changes.

jhprinz:
Resolves #611. 

At least partially for now. We agreed that the PathSimulator is not really to be stored, but to be considered the main loop of the simulation and hence a users choice everytime you setup a simulation. It can still be saved, but it does not contain the current status. 

So, if you want to restart you need to select the initial step to continue from. This implements a few conveniences to do that simpler, while being compatible with the existing code.
- [x] implement `.restart_at_step(step, storage=None)` for loaded simulators
- [x] implement class method `from_step(step, storage, initialize=True)` to create a new PathSampling object from a existing `step`. Initialize, if true, will reset the samples and add a new initial step that is detached from the given step.
- [x] add `initialize=True` to `__init__` with same functionality
- [x] save initial step on creation with storage, otherwise there is a `.save_initial_step` function.


jhprinz:
This adds two convenience ways to restart when loading a `PathSampling` object. It is already part of the vis update so you can merge this too, if you want.

dwhswenson:
@jhprinz: Conflicts here (in `restart`) now. I think this is my next to review, though, right?

jhprinz:
@dwhswenson Yes, fixed it. Hope it works now...

jhprinz:
After that the vis update. Which I updated in another PR a little more...

PR #618, opened 2016-10-19
Support for constraints in `RandomVelocities`

dwhswenson:
Previously, the `RandomVelocities` snapshot modifier did not account for constraints. This could cause problems in committor simulations (or aimless shooting, when we add that).

This PR adds `OpenMMEngine.apply_constraints`. This does not require that other engines (e.g., `Toy`) include an `apply_constraints` method. That keeps the simplest engines simple, in cases where there's no such thing as constraints.
- [x] Add `apply_constraints` function to OpenMM engine
- [x] Use `snapshot.engine.apply_constraints(snapshot)` in `RandomVelocities`, within `try: except:`
- [x] Tests
- [x] Update docs/docstrings


dwhswenson:
Tests pass; docstrings done. Ready for review.

jhprinz:
very nice. Will merge.

Issue #620, opened 2016-10-24
Change in storing policy for referenced objects


jhprinz:
This is born out of convenience for restart and trying to reflect what a user might want to happen, when he saves an object.
#### The problem

Some objects, e.g. `Sample`, reference other "older" samples which are part of their history. This is useful to build path trees but when you save an object to a new storage the current _completeness_ policy will trigger storing all related samples, which will trigger storing all their related samples and so on. This will lead to completely re-save a full simulation and not just the single sample.

In this case what the user wants is to store only this last sample and not everything.
#### How to solve this

There are a few ways to get around this:
1. Make a copy of the objet that does not have the reference to the other objects. This is the simplest yet most technical solution. You need to know which objects can have this issue and need to decide where to make the _cuts_. For a sample this would be `.previous` but there could be more. So this requires a user to know about these problems and fix them manually.
2. There is a helper function for a sample set to do that for all samples alike. `sampleset.copy_without_parents()`
3. We change the policy on delegation of saving for some attributes. For samples this would mean that saving a sample the object referenced in `.previous` will be stored only with its UUID but the referenced object is not stored. This will cause what the _above_ user wanted. But it will not cause the chain of storing events which might have been expected. I could imagine some cases where it will not stuff but a user might think he had stored it. In the current setup I am pretty sure that we the exact behaviour we would like to have. But: If you then save a sample you need to save `sample.previous` by hand provided that it is not saved already (which it will be in the production case).

sroet:
Would it be a possibility to define the storage behaviour of option 3 when opening a storage? 
For example if a storage is opened with ```fallback=True``` allow for another option like:
```
only_save_uuid_from_fallback=True/False
```

jhprinz:
Yes, that would be possible and seems like a good idea. You can even change this behaviour on the fly since it does not affect the storage itself, only what is stored. 

If I understand you you want to exclude things that are not yet stored in the fallback storage, i.e. only store, if you cannot load the object in any way. That seems like a good default, when you have a fallback attached, just use it. In case you do want our current behaviour there could be an option to really store everything despite it being already in the fallback.

I would say we we do not need option 3. then and the fallback based is much simpler and cleaner for the user. Thanks.

sroet:
> If I understand you you want to exclude things that are not yet stored in the fallback storage, i.e. only store, if you cannot load the object in any way. 

Yes this is indeed what I would like to use.

jhprinz:
Is this solved now?

[...]

PR #621, opened 2016-10-26
Exclude object in fallback from storage

jhprinz:
As discussed in #620 this will implement the default exclusion of objects that are already stored in
the fallback. It can be turned of by

``` py
storage.exclude_from_fallback = False  # default is `True`
```

Same can be switched on for proxy objects that are neither stored in the fallback or the storage itself.

``` py
storage.exclude_proxy_from_other = True  # default is `False`
```

This way you can simple continue with a second file and not worry about storing objects again. This comes at a little performance overhead for checking in the fallback but that should be minimal.


jhprinz:
@sroet Could you have a look at the example notebook in this PR and see, if that is what you want or need?

sroet:
It look like It works the way I expected it to work. However, would this work for fallback 'chains' (eg. a tps run of 100 steps with a new storage every 25 steps where the new storage has the previous as fallback)?
Looking at code it only checks if ```store.fallback == obj._store```(line 879 in  openpathsampling/netcdfplus/objects.py )

 This would be False if I'm looking for something that is actually in the fallback of the fallback (of the fallback ... etc.), right? Please correct me if I misinterpreted that line of code

jhprinz:
Yes, correct. Good catch. I thought that would already work by recursion, but I am not using the correct functions. Fixed this already, but am checking a little more. Why some things are not as fast as it should be.

dwhswenson:
@jhprinz : what's the status on this? @sroet is starting to need it! :-)

jhprinz:
Just updated it... Ready for review. I think there could be more testing though... Got ahead if you think this is okay. The bigger update will come soon, but this makes it possible what @sroet needs.

dwhswenson:
This looks good, but I'm not entirely sure I understand `exclude_proxy_from_other`. Can you explain that a little more? I think this comes into play for the second case I describe below; I'm not sure when you would want to turn that behavior off.

Here's the behavior I would expect by default; does this represent the way it works?

* if `fallback` is set, and you try to save an object that exists in the fallback chain, then the object isn't saved
* if `fallback` is not set, you would re-save all objects to the new file

The first example is if I'm splitting my simulation over multiple files sequentially (due to limitations on disk space or wall time).

The second example is when I'm setting up a new simulation based on an old one (e.g., copying snapshots from a TPS run to be used for a committor analysis). 

I'll merge this, but some clarification would be helpful!

jhprinz:
`exclude_proxy_from_other` was an idea that I had. A proxy object must reference a store from where you can load the object. So, if you store a proxy this is usually excluded, but only if the store that is referenced is the save as the one you want to store it in.

So, you can now tell what is supposed to happen in this case. Do you want to load the proxy object and re-save it or just keep the reference and assume that the user will take care of it.

Note that this case is only used when the object is also not in the fallback. So you have this reasons not to store an object

1. If the `__uuid__` exists in the current store
2. If the `__uuid__` exists in the fallback and `exclude_from_fallback=True`
3. If the `__uuid__` is marked using `store.remember(obj)`
4. If the object is a proxy and the store reference is the current store
5. If the object is a proxy and `exclude_proxy_from_other=True`

PR #623, opened 2016-11-01
little update to PathTree visualisations

### LINKED ISSUE #622: PathTree Bug(s)
When I try to generate a path tree from a TPS simulation with the code:
``` python
tree = ops_vis.PathTree(
    steps[:],
    ops_vis.ReplicaEvolution(replica=0, accepted=True
    )
)
tree.options.css['scale_x']=0.5/len(steps)
tree.options.css['width'] = '200%'
SVG(tree.svg())
```
where ```steps``` is a list of mc steps from multiple save files(the tps simulation is broken across multiple save files). It only shows the step numbers of the last file and an asterisk for all others, as seen in:

tree

While running
``` python
step = steps[20]
print step.mccycle
```
gives:
```
20
```
Also, could it be possible to scale the vertical connectors independently?
Because as you can see, above the highlighted try: the width of my trajectories are actually smaller than the connectors. This scaling issue is probably because my original trajectory was taken with a 10x higher save frequency (resulting in 74000 frames in comparison to around 100 for the short tries). It is actually cut off in the above image, this is the full image (without the width scaling):
full_tree

Lastly, the automatic scaling (without both the width and scale_x option), results in an unreadable image (I first thought I got no output at all):
wrong_tree

jhprinz:
This resolves #622 .

A few additions that @sroet found to be useful or being a bug fix.

- `SampleSet.__contains__` can be tested with replica_ID `int`, ensemble `Ensemble` or a sample `Sample` that is contained.
- some code cleanup in `visualization.py`
- Convenience functions to add custom css styles to your PathTree. The tutorial notebook has already been updated (my fault, wrong PR)
- Added some more CSS classes to better selected specific elements to be changed
- The function that generates the list of samples for ReplicaEvolution now only uses samples that appear in the steps provided. It will not scan longer.
- There should also been an update in the previous PR that takes care of the issue of `*` for steps in other files.


jhprinz:
Ready to be reviewed.

PR #625, opened 2016-11-05
ToyEngine updates

dwhswenson:
This fixes one significant bug in `ToyEngine`, and also adds lots of docstrings.

The bug is that the `n_steps_per_frame` parameter was being ignored. Now it is used. I switched our notebooks to take `n_steps_per_frame=1`, since that's what they were actually doing before.

Other than that, I added docstrings for toy engine stuff, and made sure that it shows up in the API docs.

There are also a couple little updates based on discussions with ESDW participants: improved the install instructions, and added jupyter to our requirements (apparently, we currently *do* require it, based on some imports of IPython that aren't checked by a `try: / except:` -- it doesn't cause problems for tests because our tests require jupyter.)


dwhswenson:
Passed tests, so it is ready for review.

jhprinz:
Very good. Will merge...

PR #626, opened 2016-11-06
CV File external engine

dwhswenson:
This is a little "engine" that allows us use the OPS analysis tools for an engine that just outputs some reduced set of CVs. Obviously, it would be very hard to run sampling with this (you would need a way to get from the reduced space of your CVs into the full space for your starting frames; OPS will only know how to write the CVs), but that's not the goal.

The goal here was to make it easy to analyze output from a simulation without needing to have an OPS version of the CV. (The idea came out a question that Gerhard Hummer asked me at the Lorentz Center workshop.)

NOTE: This is based on the ExternalEngine, and so it requires that #144 be merged first. Therefore, I'm marking it "onhold" and with a 1.x milestone.


openpathsampling/engines/cv_file/cv_file_engine.py
```diff
@@ -0,0 +1,90 @@
...
+        options['n_spatial'] = n_spatial
+
+        template = np.array([[0.0]*len(cv_names)])
+        super(CVFileEngine, self).__init__(options=options,
+                                           template=template)
+
+        # explanation for the strange i=i is because lambdas are annoying
+        # when defined in loops: http://stackoverflow.com/questions/938429
+        cvs = [paths.FunctionCV(self.cv_names[i],
+                                lambda s, i=i: s.xyz.flatten()[i])
```
jhprinz: Question: Just to clarify -- each column corresponds to a new CV right? There might be cases (especially for performance) where you want to use several columns together as one CV.

openpathsampling/engines/cv_file/cv_file_engine.py
```diff
@@ -0,0 +1,90 @@
...
+               for i in range(len(self.cv_names))]
+        self.cv = {cv.name: cv for cv in cvs}
+
+        self.shape = (n_atoms, n_spatial)
+        self.velocities = np.zeros(self.shape)  # HACK
```
jhprinz: This could be a good use-case for a `ConfigurationSnapshot` called `Configuration` :) Just an idea.

PR #627, opened 2016-11-07
[WIP] Update ChunkSizes for faster load/save

jhprinz:
This should give a load/save speedup.


jhprinz:
I will close this. It is part of #636 .

PR #629, opened 2016-11-08
Cleanup of `Details` objects

dwhswenson:
We had a lot of old cruft (pre-dating the `MoveChange` object) that was stored as part of various `Details` objects. Following some email discussion with @jhprinz, this PR does the following:

- [x] Remove all default values from `Details` (`trials`, `results`, `inputs` from `MoveDetails`; `selection_probability` from `SampleDetails`): these were unused/redundant with `MoveChange`
- [x] Rename Metropolis-related details: `total_acceptance`=>`metropolis_acceptance`; `random_value`=>`metropolis_random`
- [x] Add `input_samples` to `SampleMoveChange`
- [x] Check that everything is being stored correctly in new results.

Note that this does not (yet) end the distinction between `MoveDetails` and `SampleDetails`, although now the objects are essentially equivalent (and are just renamed subclasses of `Details`.  @jhprinz, if you think I should change that here, I can easily.

### DIFF
--- a/openpathsampling/movechange.py
+++ b/openpathsampling/movechange.py
@@ -304,7 +304,7 @@ class SampleMoveChange(MoveChange):
     This is the most common MoveChange and all other moves use this
     as leaves and on the lowest level consist only of `SampleMoveChange`
     """
-    def __init__(self, samples, mover=None, details=None):
+    def __init__(self, samples, mover=None, details=None, input_samples=None):
         """
         Parameters
         ----------
@@ -327,6 +327,7 @@ class SampleMoveChange(MoveChange):
             samples = [samples]
 
         self.samples = samples
+        self.input_samples = input_samples
 
     def _get_results(self):
         return []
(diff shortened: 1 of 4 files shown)


dwhswenson:

`SampleMovers` must also provide `input_samples` to their move change. This is done automatically in the Metropolis part of `SampleMover`. Other movers give changes with `input_samples == []`. 

Lots of old and unnecessary stuff removed. I don't know whether that will make our files noticeably smaller/faster, but it certainly won't hurt!

Nothing in this breaks the ability to read old files, although old files and new files would require different analysis, since some names have changed (for obscure analyses that we've never done. We aren't 1.0 yet, we can change this!)

openpathsampling/movechange.py
```diff
@@ -46,6 +47,12 @@ def __init__(self, subchanges=None, samples=None, mover=None, details=None):
             self.samples = []
         else:
             self.samples = samples
+
+        if input_samples is None:
+            self.input_samples = list()
```
jhprinz: Just a note because I read this recently. For some reason `[]` is faster than `dict()`. I would have guessed that
the interpreter treats it the same, but in theory you could alter the behaviour of dict and hence you need to do a function call while `[]` is really syntactic and will always give an empty list. My test showed a factor of 2x.

jhprinz:
Lgtm...

PR #633, opened 2016-11-09
PathMover update - remove `Sample.details`

jhprinz:
This removes the `.details` property from `Sample`. The necessary information is now stored in the `MoveChange`.

### DIFF
--- a/openpathsampling/pathmover.py
+++ b/openpathsampling/pathmover.py
@@ -2326,9 +2326,8 @@ class MoveDetails(Details):
 class SampleDetails(Details):
     """Details of a sample
 
-    Notes
-    -----
-    Deprecated. Will not present in future versions
+    .. note:: Deprecated in OpenPathSampling 0.9.3
+          `SampleDetails` will be removed in OPS 2.0.0
     """
 
     def __init__(self, **kwargs):
--- a/openpathsampling/pathmover.py
+++ b/openpathsampling/pathmover.py
@@ -1080,7 +1080,7 @@ class SubtrajectorySelectMover(SampleMover):
     If there are no subtrajectories which satisfy the subensemble, this
     returns the zero-length trajectory.
 
-    Parameters
+    Attributes
     ----------
     ensemble : openpathsampling.Ensemble
         the set of allows samples to chose from
@@ -1156,6 +1156,18 @@ class RandomSubtrajectorySelectMover(SubtrajectorySelectMover):
     If there are no subtrajectories which satisfy the subensemble, this
     returns the zero-length trajectory.
 
+    Attributes
+    ----------
+    ensemble : openpathsampling.Ensemble
+        the set of allows samples to chose from
+    sub_ensemble : openpathsampling.Ensemble
+        the subensemble to be searched for
+    n_l : int or None
+        the number of subtrajectories that need to be found. If
+        `None` every number of subtrajectories > 0 is okay.
+        Otherwise the move is only accepted if exactly n_l subtrajectories
+        are found.
+
     """
 
     def _choose(self, trajectory_list):
@@ -1168,6 +1180,19 @@ class FirstSubtrajectorySelectMover(SubtrajectorySelectMover):
 
     If there are no subtrajectories which satisfy the ensemble, this returns
     the zero-length trajectory.
+
+    Attributes
+    ----------
+    ensemble : openpathsampling.Ensemble
+        the set of allows samples to chose from
+    sub_ensemble : openpathsampling.Ensemble
+        the subensemble to be searched for
+    n_l : int or None
+        the number of subtrajectories that need to be found. If
+        `None` every number of subtrajectories > 0 is okay.
+        Otherwise the move is only accepted if exactly n_l subtrajectories
+        are found.
+
     """
 
     def _choose(self, trajectory_list):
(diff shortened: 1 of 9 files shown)


jhprinz:
At a later point we need to also remove 

1. `Sample.parent` and 
2. rename `Sample.bias` to `Sample.acceptance_probability`. 

The first requires more work for PathTreeVisualization and I will do that at a later state not necessarily for 1.0  we can do the transition secretely at anytime.

The second is simple but requires some backward compatibility for storage.

openpathsampling/pathmover.py
```diff
@@ -1503,12 +1502,6 @@ class RandomAllowedChoiceMover(RandomChoiceMover):
     from sub movers that actually can succeed because they have samples in all
     required input_ensembles
 
-    Attributes
-    ----------
-    movers : list of PathMover
-        the PathMovers to choose from
-    weights : list of floats
-        the relative weight of each PathMover (does not need to be normalized)
```
dwhswenson: I noticed that you removed these inherited attributes from the docstring. Could you double-check that Sphinx is including them when you build the docs, and also that they should up with IPython's `?` trick? I seem to remember at least one of those not respecting docstring inheritance.

Better would be to find a way to get docstrings for inherited attributes to show up in those, though.

openpathsampling/pathmover.py
```diff
@@ -2338,8 +2322,13 @@ def __init__(self, **kwargs):
         super(MoveDetails, self).__init__(**kwargs)
 
 
+# leave this for potential backwards compatibility
 class SampleDetails(Details):
     """Details of a sample
+
+    Notes
+    -----
+    Deprecated. Will not present in future versions
```
dwhswenson: I was looking through some numpydoc format stuff earlier, and there's a specific way to make deprecation warnings:

```rst
.. note:: Deprecated in NumPy 1.6.0
          `ndobj_old` will be removed in NumPy 2.0.0, it is replaced by
          `ndobj_new` because the latter works also with array subclasses.
```

I'd say we're deprecating it in 0.9.2 and removing in ... 2.0? or even 1.0?

dwhswenson:
Is this ready for review? No WIP tag, passes tests, and looks pretty complete -- but not assigned to me yet. Anyway, the line notes I just added will probably be most of my review anyway.

openpathsampling/pathmover.py
```diff
@@ -601,7 +606,14 @@ def move(self, sample_set):
             )
 
         # 4. accept/reject
-        accepted, details = self._accept(trials)
+        accepted, acceptance_details = self._accept(trials)
+
+        # update details
+        kwargs = {}
+        kwargs.update(call_details)
+        kwargs.update(acceptance_details)
```
dwhswenson: Side note: 2.0 will treat the acceptance as a separate mover, meaning that it will remove the risk of collision in this update.

(Until then, contributors just need to know that `metropolis_acceptance` and `metropolis_random` are not allowed as names of details in their movers -- I don't think it'll be a problem).

openpathsampling/tests/testopenmmengine.py
```diff
@@ -212,7 +212,8 @@ def test_nan_rejected(self):
         change = mover.move(init_samp)
 
         assert (isinstance(change, paths.RejectedNaNSampleMoveChange))
-        assert_equal(change.samples[0].details.stopping_reason, 'nan')
+        print change.details.rejection_reason
```
dwhswenson: remove this print statement from tests

openpathsampling/pathmover.py
dwhswenson: for now, I'm fine to either remove or leave, whichever you prefer. Might be better to put them back in, though (you'll see the changes easily in this PR's diff), so that these docstrings are in place for Traunkirchen.

jhprinz: help does not. `__doc__` is either inherited from super or the new is taken. Sphinx will do this if not turned off

openpathsampling/pathmover.py
jhprinz: Thanks. Yes, changed it to be deprecated now and be removed in 2.0. We are not using it for analysis anymore and it is not needed. But if you remove it, loading of an old file will fail.

openpathsampling/tests/testopenmmengine.py
jhprinz: done

dwhswenson:
Just waiting for tests to pass. However, you might consider whether it is easy to return those docstrings now. If it's easy, please go ahead and do it. If it'll be a pain, don't bother, but please don't remove any more repeated docstrings until we have docstring inheritance solved.

jhprinz:
Done.

dwhswenson:
excellent, thanks! will merge as soon as tests pass

openpathsampling/sample.py
```diff
@@ -746,23 +748,20 @@ class Sample(StorableObject):
...
-        the Monte Carlo step number associated with this Sample
     """
 
     def __init__(self,
                  replica=None,
                  trajectory=None,
                  ensemble=None,
                  bias=1.0,
-                 details=None,
+                 # details=None,
```
dwhswenson: I think this (or one of the similar lines) is causes failures now. If you're going to keep the old files working, you'll need to have a way to let it set details even if it isn't used.

dwhswenson:
That fixed it! Merging.

PR #634, opened 2016-11-10
Fixes for test output

dwhswenson:
Two very minor fixes:

* #633 left a print statement (used for debugging, I assume?) in `pathmover.py`. Removed that.
* numpy 1.11 will start throwing RuntimeWarnings if `true_divide` has a NaN or divides by zero (I was *using* `true_divide` to get the desired behavior that `NaN/0.0` gives `NaN`!) Conda is still on 1.10, but I've upgraded to 1.11 and see this warning when I run tests locally.

Should be ready for review immediately; just a matter of waiting to pass tests.


jhprinz:
good.

PR #636, opened 2016-11-13
Storage -> Only with UUID

jhprinz:
Solves #620.

This fixes `Storage` to use only with uuid. This means that some tpy example might run a little slower, but it simplified the code A LOT! And makes it much better maintainable. It has also some major speed improvements (not sure how these turn out in a real system).

So far it is completely backward compatible, but I would like to change the file format a little to make it faster for dealing with UUIDs.

Main idea for that is to just use long instead of creating string representations and `UUID` objects. It now is really just a long integer and thus much faster to compare and change and store. Since this part is a significant part in loading objects it should load much faster.

It makes the following changes:

- [x] removal of `use_uuid` or `referenced_by_uuid`
- [x] fixes tests, notebooks are run `use_uuid=True`
- [x] store the UUID internally as 16byte - long and not as a UUID object anymore
- [x] storage is left as is (it stores UUID as a string representation, which is an moderately expensive transformation 
- [x] remove decorator for auto creation of delayed loading. This were only 4 variables and using the decorator will destroy the signature for the help of `Sample`, `MCStep` and `SampleSet` which is bad for a user (we can write a better decorator later and this is fixed in Py 3.4+ anyway)
- [x] some cleanup in the loading functions
- [x] use `VariableStore` in the correct order to speedup object creation with `args` and not `kwargs`
- [x] rearrange some code in multiple files for better maintenance. There is now a `stores` folder that has the code related to a `store`
- [x] update the `cache_all` functions for faster analysis loading
- [x] increase chunk sizes


jhprinz:
Note that tests are slower because of the much larger chunk sizes now. Thus creation and opening will take a little longer and lead to slower tests. This should have the reversed effect when loading large files.

jhprinz:
If we agree to make this transition we can put it into 1.0. It was mainly done when I was preparing for the talk at ESDW next week.

jhprinz:
Correction, we are not using the non-uuid storage anymore in the examples (as far as I can tell), so the transition might be less problematic than expected.

jhprinz:
I think this is ready to be reviewed. This boils mainly down to accepting that we always use `uuids` and that these are internally stored as a`long` int and not a formatted `str`.

This affected a lot of files, but most of them only some docs, some cleanup, etc. Nothing really serious.

Note that this is not backward compatible with files that are not using `uuid` yet. 

Let me know what you think, when all tests pass.

dwhswenson:
This PR is mixed with #627 (all commits from #627 except some recent merges are included in this). Should I be reviewing both?

openpathsampling/netcdfplus/cache.py
```diff
@@ -374,7 +374,7 @@ class LRUChunkLoadingCache(Cache):
 
     """
 
-    def __init__(self, chunksize=100, max_chunks=100, variable=None):
+    def __init__(self, chunksize=256, max_chunks=8192, variable=None):
```
dwhswenson: Just curious: how much optimization did you do with these chunking numbers? on what kind of system? It might be really nice to have some benchmarks of this.

openpathsampling/sample.py
```diff
@@ -756,17 +760,24 @@ class Sample(StorableObject):
...
     def __init__(self,
                  replica=None,
                  trajectory=None,
                  ensemble=None,
                  bias=1.0,
-                 details=None,
                  parent=None,
-                 mover=None
+                 mover=None,
+                 details=None
```
dwhswenson: What is the current status on `Sample.mover` and `Sample.details`? Are these both deprecated? I think they were only used in the path tree visualization, and if that has switched to using the `MoveChange` instead, then they can be deprecated (can't remove yet because of backward compatibility).

openpathsampling/storage/stores/collectivevariable.py
```diff
@@ -150,18 +150,4 @@ def cache_all(self):
...
-        Parameters
-        ----------
-        cv : :obj:`openpathsampling.CollectiveVariable`
-            the CV from which you want the cache to be filled
-
-        """
-        store = self.storage.snapshots.cv_list.get(cv)
-        if store is not None:
-            store.fill_cache()
+        pass
```
dwhswenson: ??? empty function? Is this now obsolete, but kept in for API compatibility? If so, please add deprecation note.

openpathsampling/tools.py
```diff
@@ -34,18 +34,36 @@ def in_ipynb():
...
+    if refresh:
+        if is_ipynb:
             IPython.display.clear_output(wait=True)
+        else:
+            if last_output is not None:
+                lines = len(last_output.split('\n'))
+                CURSOR_UP_ONE = '\x1b[1A'
+                ERASE_LINE = '\x1b[2K'
+
+                output_stream.write((CURSOR_UP_ONE + ERASE_LINE) * (lines - 1))
```
dwhswenson: what is this going to do if I write to a file?

openpathsampling/tools.py
```diff
@@ -34,18 +34,36 @@ def in_ipynb():
...
+
+            last_output = output_str
+    else:
+        last_output = None
 
-        output_stream.write(output_str)
-        sys.stdout.flush()
+    output_stream.write(output_str)
+    sys.stdout.flush()
```
dwhswenson: no; `output_stream.flush()`

jhprinz:
> I spent a while trying to figure out why you removed a bunch of stuff until I saw that the same text was in a new file.

You are right. This happend because I kept on making changes to get the alanine example fixed ASAP and at the end I stripped of the stuff that is only `hal` related. Sorry, if it looks so confusing. It was not meant to be that way.

openpathsampling/high_level/move_scheme.py
```diff
@@ -721,6 +722,10 @@ def move_summary(self, steps, movers=None, output=sys.stdout, depth=0):
         depth : integer or None
             depth of submovers to show: if integer, shows that many
             submovers for each move; if None, shows all submovers
+        scheme_copies : lits of :class:`MoveScheme`
```
jhprinz: I agree, this is actually a point I wanted to discuss separately. I (still) need it to join data generated from two MoveSchemes. I think I need your insight for a better solution. I will remove it for now to be sure.

openpathsampling/high_level/transition.py
```diff
@@ -331,10 +334,10 @@ def total_crossing_probability(self, steps=None, method="wham", force=False):
             # wham.load_from_dataframe(df)
             # wham.clean_leading_ones()
             tcp = wham.wham_bam_histogram(df).to_dict()
-        elif method == "mbar":
-            pass
+        # elif method == "mbar":
+        #     pass
         else:
-            raise ValueError("Only supported methods are 'wham' and 'mbar'.  "
+            raise ValueError("Only supported methods are 'wham'.  "
```
jhprinz: :) Damn you grammar...

openpathsampling/high_level/transition.py
```diff
@@ -74,6 +76,7 @@ def _tps_ensemble(self, stateA, stateB):
             paths.AllInXEnsemble(stateB) & paths.LengthEnsemble(1)
         ])
 
+    # TODO: Can this kwargs be removed? It is not used
```
dwhswenson: (Clarification: it is needed in the network, but not in the transition)

openpathsampling/netcdfplus/cache.py
jhprinz: There is no testing. I tried to use number of base 2. I do not have time at this point to run an optimisation. I am happy that I find time to work at all on this. This is based on caching about 2M values and chunk size of 256 to match number I used for the other objects.

I think we should name the optimization 1.X

openpathsampling/tools.py
jhprinz: Ah yes, I forgot about this. I used it to keep the output short and mimic the behaviour of overwriting the output in a cell. If you write to a file it will contain the control characters and if you run `cat ...` it will _reply_ the output. In an editor you just have the control characters.

I like this, but we might add an additional optional flag.

openpathsampling/sample.py
jhprinz: Correction. I already fixed this and we should be able to remove the `details=None` without problems.

openpathsampling/high_level/transition.py
jhprinz: Okay, thanks. That answers it.

openpathsampling/high_level/move_scheme.py
dwhswenson: Mainly, my thoughts go like this: A lot of the high-level objects were originally designed to group data for analysis. While writing the analysis scripts, I realized that these would also make it really easy to set up a simulation. So analysis got combined with setup.

That's what should change. The setup side of all of these should stay the same, but there should be another object that takes a `MoveScheme` or a `TransitionNetwork` and actually provides the analysis. Then the `MoveScheme` does the setup, and the `MoveSchemeAnalyzer` does the analysis. The difficulties you're having would then be trivially solved with `MoveSchemeAnalyzer(schemes=[scheme1, scheme2])`.

I think I'll keep the current behavior available as a convenience (and for backward compatibility). For example, is it really nice to write `network.rate_matrix(steps)`, but it really should be `TISNetworkAnalyzer(network).rate_matrix(steps)`. (By adding an `analyzer` object to the network, it should be pretty easy to maintain the conveniences.) This would also make it easier to have multiple methods to calculate the rate (not just questions of WHAM/MBAR, but also path-type analysis, which is what we should be using.) This could be `1.1` or `1.0`; depends on how quickly we write the paper ;-)

openpathsampling/netcdfplus/cache.py
dwhswenson: > I think we should name the optimization 1.X

Agreed -- I'm very curious as to what numbers will work best here, and whether we can write helpers to estimate optimal chunk sizes for a given use case (asking information from the user about their machine, and looking at the system to be studied, for example).

openpathsampling/high_level/move_scheme.py
jhprinz: I think that is an excellent choice. So we can keep the two things separated as we do with visualisation (which is also some kind of analysis). So let's keep this for now.

openpathsampling/storage/stores/collectivevariable.py
jhprinz: Sorry, you actually found a _bug_ or not optimal setup. I fixed this and now it is much faster...

openpathsampling/sample.py
dwhswenson: > Correction. I already fixed this and we should be able to remove the `details=None` without problems.

Will it still load the old files (the ones in my Dropbox, used for testing)? Although maybe we should re-run them soon anyway, and call anything files older than 0.9.2 incompatible after some future point.

openpathsampling/tools.py
dwhswenson: > I like this, but we might add an additional optional flag.

Would this work?

```python
if output_stream is sys.stdout:
    CURSOR_UP_ONE = ...
    ERASE_LINE = ...
else:
    CURSOR_UP_ONE = ""
    ...
```

openpathsampling/sample.py
jhprinz: Yes, it will ignore any existing details.

jhprinz: And I am using your file for testing.

openpathsampling/tools.py
jhprinz: I guess we should answer for us, if we want this to mimic the behaviour of an output cell? My (still existing) problem with the approach is that if we wanted to do just that, we have no way of knowing, if we now run in a next cell and hence keep the last result or replace the last result.

Does that make sense? Example

```
# cell 1
refresh_output('step1.1')
refresh_output('step1.2')

#cell 2
refresh_output('step2.1')
```

should look in the console like
```
step 1.2
step 2.1
```

So we might put this functionality into a new function that needs a `new_cell` command to make this clear. This one could of course also work in a notebook.

dwhswenson: If this gets any further into having state, then the `refresh_output` method needs to become a class. (Not feeling great about the global in there already.)

jhprinz: Yes, I don't like the global as well, but this is more a one time check, if we run in a notebook. I actually though about an output class, that works like this:

```py
o = Output(options)
o.refresh('step 1.1')
o.refresh('step 1.2')
o.next()  # or new_cell or new_step
o.refresh('step 2.2')
o.write('-- there was an error --')
```

We could also have it be more flexible like adding a progress bar, etc... But that should be a new feature and issue.

jhprinz:
#627 contains only the changes of chunksizes and is contained in this one.

jhprinz:
@dwhswenson I think I fixed the little issues still there when we talked. It also has the updated alanine mstis setup with the fixed distance function.

Let me know if something is still missing...

dwhswenson:
Not sure... I can't dive into to too far right now, but my suspicion is in with `bash -x` on the `after_success.sh` script. Maybe writing to `stderr` (as the `-x` debugging does) is seen as indicating an error?

Note that this seems to be occurring in the nightly, too. It is not a problem for OPSPiggybacker, though (which does not have this sort of `after_success` script.)

Is there any reason we can't move the `after_success.sh` into an [`after_success:`](https://docs.travis-ci.com/user/deployment/custom/) or [`deploy:`](https://docs.travis-ci.com/user/deployment/script/) step in the `.travis.yml`?

openpathsampling/high_level/transition.py
dwhswenson: @jhprinz : Were you going to remove the `kwargs` in this? As long as you add the TODO, and we agree it should go, then you might as well make the change here!

openpathsampling/tools.py
dwhswenson: Anything further on this? As I suggested above, I think you can just check if `output_stream is sys.stdout`, and modify the behavior according to that. At least, that fixes the problem of if someone wants to write to a file instead of stdout (stdout is default).

jhprinz:
- removed the `kwargs` in transition.py
- made the update function in `tools.py` only work if output is stdout.

At some point we might turn this into a class that has better output capabilities

dwhswenson:
Looks good now! Merging!

PR #638, opened 2016-11-21
Method to avoid refreshing ipynb cells from wrapped simulators

dwhswenson:
This should fix a problem that @sroet brought to my attention at the ESDW today. The basic problem is this:

Some path simulators will wrap other path simulators (e.g., for some iterative process). The problem that @sroet was having was that, even if the output of the wrapped simulator was sent to `dev/null`, the `refresh` command was emptying the cell, deleting desired output from the outer simulator.

This adds a new attribute for `PathSimulator`s called `allow_refresh`, which defaults to `True`. When a simulator calls the `refresh_output` command, the `refresh` attribute is set to `self.allow_refresh`. Combining `inner_sim.allow_refresh = False` with `inner_sim.output_stream = os.devnull`, this can completely silence the output behavior of the inner simulator.

@sroet: Could you check out this branch and test that it fixes your problem? We should check that it fixed the issue before merging.

### DIFF
--- a/openpathsampling/pathsimulator.py
+++ b/openpathsampling/pathsimulator.py
@@ -71,6 +71,10 @@ class PathSimulator(StorableNamedObject):
     output_stream : file
         Subclasses should write output to this, allowing a standard way to
         redirect any output.
+    allow_refresh : bool
+        Whether to allow the output to refresh an ipynb cell; default True.
+        This is likely to be overridden when a pathsimulator is wrapped in
+        another simulation.
     """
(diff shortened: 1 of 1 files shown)


sroet:
@dwhswenson This indeed fixes my problem

dwhswenson:
Fixes the problem and passes tests. @jhprinz : this is ready for review!

jhprinz:
If it works. This can be merged. Hard to tell from here.

PR #639, opened 2016-11-22
Details on engine exceptions

dwhswenson:
Previously, the `SampleMover` would overwrite the details from engine exceptions, meaning that, e.g., for a shot that hit max length, you couldn't figure out the shooting point. This fixes that.


jhprinz:
Looks good. Is this ready?

dwhswenson:
Yep -- got distracted while waiting for tests to finish. Merge at will!

PR #640, opened 2016-11-22
Failsafe for `segment_id` in `to_dict` of mdtraj topology

jhprinz:
I had problems that in some cases the `mdtraj.Topology.Atom` object does not have a `segment_id` property. In this case the segment_id will be empty. This fixes this, but in the long run we should work on getter a better pickling for the topology.


jhprinz:
This fixes it for me. Ready to be reviewed when it passes

dwhswenson:
Looks fine to merge. Question: is this an MDTraj 1.7.x vs. 1.8 change? If not, what leads to this happening?

PR #645, opened 2016-12-01
Shortcut: `ops_load_trajectory`

dwhswenson:
We've talked about adding this before. It's basically an OPS wrapper on MDTraj's `load` method. After MDTraj loads the file, we convert the MDTraj trajectory to an OPS trajectory.

I've locally tested with Gromacs file including a `.gro` for the `top` keyword, and it works. At some point, we'll want to add formal unit tests on that.

(I got tired of every analysis-based example starting with what was basically this little one-line function.)

Would like to put this in 0.9.2, so please review soon.


dwhswenson:
Passes tests. Reviewing this should really take no more than a minute or two (it's 2 new lines of code, plus smoke tests for those two lines.)

jhprinz:
Lgtm.

PR #647, opened 2016-12-01
AD committor example

dwhswenson:
Another example of using the committor analysis. Wanted to put this into the official OPS docs. It's based on an (abandoned) project with some machine learning people, so a few parts of it are done in a slightly odd way (treating each shot as an "experiment").

The original work from this was in [a gist](https://gist.github.com/dwhswenson/e1578a40ef67aae66da15e7b065c8f7c). I'd rather put it in the OPS docs and link to it here as an example for my E-CAM module about the committor analysis.

No code, only docs, so should be easy to merge. Can be merged before or after the 0.9.2 release.

### DIFF
--- a/docs/examples/miscellaneous/committors.rst
+++ b/docs/examples/miscellaneous/committors.rst
@@ -1,10 +1,12 @@
+.. basic_committor::
+
 Basic Committor/Shooting Point Analysis
 =======================================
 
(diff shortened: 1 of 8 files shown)


jhprinz:
this looks fine. I will merge it.

PR #648, opened 2016-12-02
Remove UUID from `print details`

### LINKED ISSUE #635: Printing UUID in details
Currently, when we `print change.details`, the output includes the UUID for the `details` object. We should probably exclude that. I also notice that the code says it will ignore ensembles -- I assume that's because the `ensemble.__str__` is so long. Perhaps we should instead print the `repr` for ensembles?

Let's leave this as is until after Traunkirchen (because I don't want to have to change my slides!) but it should be a really easy fix later.

dwhswenson:
This resolves #635. Makes two special types of objects when calling `Details.__str__`:

* **keys** where you don't print at all (currently only `__uuid__`, but can be extended)
* **types** where you use `repr` instead of `str` (currently only `Ensemble`, but can be extended)


jhprinz:
I think I would exclude all entries starting with a `_` by default. Question is, if you add an option to show otherwise hidden keys. Maybe just state in the docs that a key does not have to be private.

dwhswenson:
The only one that was being printed before was the UUID. Keep in mind, this is the `Details` object. We don't add much to it that we don't specifically intend to save there. However, since the UUID gets added to *everything*....

jhprinz:
Exactly, everytime we change the `StorableObject` we need to change the details object as well. That is, what I had in mind. Not print things that are supposed to be hidden. But we can merge this and change it again, when this becomes necessary.

PR #649, opened 2016-12-03
Fix OpenMM instantaneous temperature

dwhswenson:
There was a bug in the `instantaneous_temperature` feature for OpenMM engine snapshots. It did not set context with desired snapshot, so the temperature that returns was always the from the last snapshot the engine had generated. Fix, plus regression test, in this PR.

Should be ready for review as soon as it passes tests.


jhprinz:
Looks good.

PR #650, opened 2016-12-10
Velocity Direction Modifiers for 2-way shooting

dwhswenson:
This PR includes a way to modify the velocities of a snapshot, as described in the early work on 2-way shooting. Note that these techniques can get very complicated if the system includes constraints. To avoid that, this snapshot modifier checks that the number of degrees of freedom is as expected (meaning the engine's snapshots must implement the `n_degrees_of_freedom` snapshot feature, otherwise this raises an error.)

It includes an abstract class `GeneralizedDirectionModifier`, and two concrete subclasses: `VelocityDirectionModifer` (which modifies all velocities) and `SingleAtomVelocityDirectionModifier` (randomly picks a single atom to change its velocity.) Both of these have the ability to restrict the atoms which they change to a subset of the full system (as do all `SnapshotModifier`s.)

The basic approach is to modify the velocities according to a Gaussian in each direction, and then rescale them to put them on the sphere with the magnitude of the original velocity. This satisfies detailed balance (in the absence of contraints). It also means that the speed of each atom stays the same; only the direction changes.

- [x] Implement `GeneralizedDirectionModifier`
- [x] Tests for `GeneralizedDirectionModifier` on toy snapshots
- [x] Tests for `GeneralizedDirectionModifier` on OpenMM snapshots
- [x] Implement `VelocityDirectionModifer` and `SingleAtomVelocityDirectionModifier`
- [x] Tests for concrete modifier subclasses
- [x] Include corrections for linear momentum and kinetic energy
- [x] Docstrings

### DIFF
--- a/openpathsampling/__init__.py
+++ b/openpathsampling/__init__.py
@@ -134,7 +134,10 @@ from sample import Sample, SampleSet
 from shooting import ShootingPointSelector, UniformSelector, \
     GaussianBiasSelector, FirstFrameSelector, FinalFrameSelector
 
-from snapshot_modifier import NoModification, RandomVelocities
+from snapshot_modifier import (
+    NoModification, RandomVelocities, VelocityDirectionModifier,
+    SingleAtomVelocityDirectionModifier
+)
 
 from storage.storage import Storage, AnalysisStorage
 
--- a/openpathsampling/engines/openmm/features/masses.py
+++ b/openpathsampling/engines/openmm/features/masses.py
@@ -1,4 +1,5 @@
 import simtk.unit as u
+import numpy as np
 
 @property
 def masses_per_mole(snapshot):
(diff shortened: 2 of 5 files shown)


dwhswenson:
This passes tests and is ready for review. It provides a (limited) ability to modify the shooting point for two-way shooting. If the engine is using constraints, this code refuses to run (because it's really hard to get these approaches to satisfy detailed balance in the presence of constraints).

This is more a proof-of-concept (to complete the two-way shooting E-CAM module) than something I would really suggest using. In most cases, it probably makes more sense to either use one-way shooting, or to use full velocity randomization (possibly on a subset of the atoms), which we already had the ability to do. On the other hand, this "change the direction of some atoms" picture is the standard way people first learn about path sampling, so we might as well have it.

jhprinz:
lgtm...

PR #652, opened 2016-12-14
Fix bug in acceptance probability calculation for 2-way shooting

dwhswenson:
This fixes a little bug I found when calculating the acceptance probability in 2-way shooting. The issue is this:

In 1-way shooting, when we calculate the acceptance probability (i.e., the length of the trajectory, minus the two endpoints that we don't shoot from) we *always* have a full trajectory. It might be A-to-A instead of A-to-B, but it isn't a half-trajectory.

In 2-way shooting, on the other hand, it could happen that your forward shot lands in state A (in a TPS, A->B ensemble). That means that there's no way for the backward shot to ever give something that satisfies the ensemble, so, intelligently, we quit there and the trial is only the forward partial trajectory.

Of course, this won't be accepted, because it doesn't satisfy the ensemble. However, we were checking the acceptance probability for it anyway. And it *is* possible that you shot from the first non-state frame, and went right back into the state, meaning that your trial trajectory is length 2: shooting point and frame in state.

This means that when you say "the odds of selecting that point from the new trajectory are 1/(L-2)", you divide by zero. Program crashes.

The fix is easy: if you don't satisfy the ensemble, we don't even bother checking acceptance probability. This does mean we'll check the ensemble twice for engine movers; at some point we might be able to speed that up some. (Only matters for toy systems, and even then, probably mainly for the minus move.)

### DIFF
--- a/openpathsampling/pathmover.py
+++ b/openpathsampling/pathmover.py
@@ -551,7 +551,13 @@ class SampleMover(PathMover):
             'metropolis_random': rand
         }
 
-        logger.info("Trial was " + ("accepted" if accepted else "rejected"))
+        if accepted:
+            result_str = "accepted"
+        else:
+            result_str = ("rejected. Acceptance probabilty "
+                          + str(probability))
+
+        logger.info("Trial was " + result_str)
 
         return accepted, details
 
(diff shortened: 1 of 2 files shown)


dwhswenson:
Apparently the first fix attempt broke committors for some reason. So instead I changed it so that  `sum_bias > 0.0`, otherwise the probability (`f/sum_bias`) is 0.

Also added a little additional information to the logger output telling whether a move has been accepted or rejected (gives the acceptance probability; if that is 0.0, then it was rejected before the path length was tried.) Might still add more info to that later, or better yet, log some info elsewhere on things like path length. @sroet was asking for stuff like this.

dwhswenson:

PR #654, opened 2016-12-28
Updates to flux calc in TrajectoryTransitionAnalysis

dwhswenson:
A little bit of improvement to make a direct calculation of the flux (not just the dictionary that is then used to calculate the flux) 

* `analyze_flux` function now works with multiple trajectories as input
* new `flux` function calculates the actual numeric flux (previously, only had the `analyze_flux` method, which returns a dict of the "in" and "out" frames, from which the flux can be calculated)

Also fixes a stupid error in the `analyze` method where I'd flipped "A" and "B". Finally, adds a note for future improvement in `ensemble.py` (unrelated; just something that I noticed could be made faster with a little more algorithmic effort).


dwhswenson:
Tests complete and passed. This is ready for review.

PR #658, opened 2017-01-19
Channel analysis

dwhswenson:
This adds an analysis object to study how well multiple channels are sampled in path sampling simulations. This could be useful in several contexts: 

* Similar things have been done before to see how well the various final states are sampled in MSTIS. There should be plenty of switching between various final states, which can be seen as different "channels"
* Some TPS simulations can sample different channels (mechanisms) for the same physical transition. This will make it possible to study that as well.

The approach implemented here uses `Ensemble`s (and specifically the `Ensemble.split` function) to characterize channels, as I have previously done in the AD TPS example. The ensembles are the user input. Each path can be characterized as being in one, multiple, or no channels. Then `ChannelAnalysis` object can calculate switching matrices or lifetimes (in MC step) spent in each channel.

### DIFF
--- a/openpathsampling/__init__.py
+++ b/openpathsampling/__init__.py
@@ -28,8 +28,8 @@ except ImportError:  # pragma: no cover
     if not ops_setup.preferences['released']:
         version.full_version += ".dev-" + version.git_version[:7]
     isrelease = str(ops_setup.preferences['released'])
-        
-        
+
+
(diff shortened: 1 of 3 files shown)


dwhswenson:

jhprinz:
I think that looks fine. Will merge as it just adds a new feature.

PR #659, opened 2017-01-25
DirectSimulation result storage

dwhswenson:
`DirectSimulation` keeps its results in the `DirectSimulation` object, using `@property` methods to calculate several quantities from a few simplified internal structures.

Previously, there was no way to save those simplified internal structures. This adds two methods, an `@property` called `results` and a method called `load_results`, which put those internal structures into a `dict` and load from that `dict`.

I chose to do it this way, rather than by modifying the `to_dict`/`from_dict`, because this allows the user to store the simulation object before the results are calculated.

One question: `load_results` is currently a separate method. The other option would be to make it the setter of the `results` property. I'm not sure which makes more sense:

```python
sim = DirectSimulation(...)
sim.load_results(results)
# which also allows
sim = DirectSimulation(...).load_results(results)
```

or

```python
sim = DirectSimulation(...)
sim.results = results
```

I think I prefer the `load_results` version, but either would work. I would rather not do both, under the principle of "there should be *one* obvious way to do it."

### DIFF
--- a/openpathsampling/pathsimulator.py
+++ b/openpathsampling/pathsimulator.py
@@ -165,7 +165,7 @@ class BootstrapPromotionMove(SubPathMover):
         # Bootstrapping sets numeric replica IDs. If the user wants it done
         # differently, the user can change it.
         self._ensemble_dict = {ens : rep for rep, ens in enumerate(ensembles) }
-        
+
         # Create all possible hoppers so we do not have to recreate these
         # every time which will result in more efficient storage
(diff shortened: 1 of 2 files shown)


dwhswenson:
Tests pass; ready for review.

jhprinz:
Concerning the question of loading or using a setter. I like this better in this case. Setting a dict seems weird kind of. So this okay. I will merge this.

PR #662, opened 2017-02-09
Add CVs for every object

jhprinz:
This adds the feature to use CVs for every object. To be precise, it replaces CVs by some general feature of NetCDF+ and now CVs are just a special case of these general `Attribute` object.

One example is an Trajectory attribute.

 ```py
max_cv = FunctionAttribute('max_cv', Trajectory, lambda t, cv: max(cv(t)), cv=cv)

# print all maxima of `cv` for all trajectories
print max_cv(storage.trajetories)
```

pretty easy now. You can of course _annotate_ all sorts of objects like this. Storage and caching works exactly like for `CV`s. Only difference is that a normal `Attribute` does not have a `reversible` property. This only makes sense for `Snapshot`s that have this quality.

The main reason to push this was that I could not analyse the long Alanine trajectory. It just took too long. One way to solve this would have been to increase the cache sizes, but this new approach makes it possible to cache the _max_lambda_ values efficiently and hence speed up the rate computation a lot.

This still would require the analysis to use this or as @dwhswenson suggested to make this attributes part of a network or whatever object is responsible for the computation of the max_lambda in the analysis.

I think to really get more speed ups, we also need a way to cache information about the _minus moves_. This should be close to the annotated trajectories.

#### Other changes

The generalisation of CVs also made some changes to the package structure.

1. `chaindict.py` is now part of `netcdfplus` which I think is a good place. it was always weird to be in ops
2. I also tried to split all the generic parts and the snapshot specific ones and the general parts are now in `netcdfplus` while the rest is in `storage`

#### Note

This will make a change in the Storage so that you can still read new and old files, but you cannot append old style files. At least I disabled that by default although it should be possible to append these files as long as no new attributes/cvs are added

#### Todo

- [x] fix notebooks
- [x] add some more testing
- [x] add docs


dwhswenson:
I've been starting on some ideas that will allow analysis (and sampling) to be sped up by using this. I'll start opening PRs on those in the near future.

jhprinz:
Hooray, tests pass now. WIll add some more testing and then it should be ready for review.

jhprinz:
@dwhswenson suggested to call it `PseudoAttribute` instead of `Attribute`

jhprinz:
good name, will change that.

jhprinz:
fix the strange netcdf bad chunksize error as well

jhprinz:
Should work now with a newer version of netcdf4. Means we can release all the package constraints. I also renamed to `PseudoAttribute`

jhprinz:
Let's see, if I missed something.

jhprinz:
I am a little concerned that after merging Python3 support this PR will be practically unmergeable... Will see when I have time...

jhprinz:
I guess I will have to do a rebase... Well, let's hope for the best...

jhprinz:
This will be closed and is replaced by #691 . Initial description is copied.

PR #665, opened 2017-02-16
Faster TIS analysis

dwhswenson:
Just a few lines of code to provide a significant speed-up for TIS analysis.

Caching previous results of max_lambda, so we don't check it again if we have the same trajectory as before. For the toy model, this nearly doubles the speed (with 20k trajectories).

Before:

image

After:

image

Same results; we just get them faster.

### DIFF
--- a/openpathsampling/high_level/transition.py
+++ b/openpathsampling/high_level/transition.py
@@ -257,10 +257,18 @@ class TISTransition(Transition):
         hist_data = {}
         buflen = -1
         sample_buf = []
+        prev_sample = {h: None for h in run_it}
+        prev_result = {h: None for h in run_it}
         for sample in in_ens_samples:
             for hist in run_it:
                 hist_info = self.ensemble_histogram_info[hist]
-                hist_data_sample = hist_info.f(sample, **hist_info.f_args)
+                if sample is prev_sample[hist]:
+                    hist_data_sample = prev_result[hist]
+                else:
+                    hist_data_sample = hist_info.f(sample,
+                                                   **hist_info.f_args)
+                prev_result[hist] = hist_data_sample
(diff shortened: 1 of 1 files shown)


dwhswenson:

I'll be doing other stuff in other PRs to make this even faster (and when #662 is done, that will make it possible to do other tricks to speed it up.)

The speed-up from this PR will be better with more ensembles or with lower acceptance ratios.

Interestingly, profiling the rate matrix calculation now says that, for 20k steps in the toy MSTIS example, the main cost is in identifying which steps had minus moves (and therefore need to be included in the flux calculation.) I think I can find a way to fix that. (Currently, it compares the ensembles; ensemble equality is checked based on the string representation of the ensemble; making those strings actually takes 30% of the total analysis time!)

jhprinz:
Looks good, neat trick. There is one line I think can be moved for probably unnoticeable additional improvement :)

openpathsampling/high_level/transition.py
```diff
@@ -257,10 +257,18 @@ def ensemble_statistics(self, ensemble, samples, weights=None, force=False):
         hist_data = {}
         buflen = -1
         sample_buf = []
+        prev_sample = {h: None for h in run_it}
+        prev_result = {h: None for h in run_it}
         for sample in in_ens_samples:
             for hist in run_it:
                 hist_info = self.ensemble_histogram_info[hist]
```
jhprinz: I would move this in the else sub clause. I think it is only called when a new sample is found correct?

dwhswenson: True -- but I'll add that to #666. You might as well just merge that one, which will auto-merge this as well.

PR #666, opened 2017-02-17
Faster minus flux

dwhswenson:
Told you in https://github.com/openpathsampling/openpathsampling/pull/665#issuecomment-280488217 that I'd find a way to fix this. Didn't think I'd do it tonight, though!

Since `Ensemble`s are immutable, their strings should also be immutable. So we calculate the full string once, and then cache it.

The rate matrix speed is even better. Before it was ~40 seconds, now:

image

At this point, the performance bottlenecks for the rate matrix calculation are the chaindict (I don't think these CVs are cached on disk, so that's more expensive in this case) and within the `collection.Counter` that manages the histogram (there's room to improve the way data is sent to that, but I'll save that for the rewrite of analysis that I'm working on). Loading the data (opening the file as `AnalysisStorage`) is a much more significant cost than either of these for this system (about 2-3 minutes, compared to rate calculation analysis in 20 seconds -- both for 20k steps of the MSTIS toy model).

Note that this includes #665, and therefore merging this will also merge that.

### DIFF
--- a/openpathsampling/ensemble.py
+++ b/openpathsampling/ensemble.py
@@ -209,6 +209,7 @@ class Ensemble(StorableNamedObject):
         A path volume defines a set of paths.
         """
         super(Ensemble, self).__init__()
+        self._saved_str = None  # cached first time it is requested
 
     def __eq__(self, other):
         if self is other:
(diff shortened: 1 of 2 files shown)


dwhswenson:
Passes tests, ready for review. Merging this will also merge #665.

jhprinz:
lgtm

PR #667, opened 2017-02-17
`ShootFromSnapshotsSimulation`

dwhswenson:
This refactors some of the committor simulation. It will already be useful for reactive flux, and also for some other things I'm playing with that won't go straight into OPS core.

There are a lot of simulation types that start from a user-provided list of snapshots, and then shoot from those snapshots (often doing something like randomizing the velocities before the dynamics) and run until some ensemble condition is satisfied. This refactors the reusable code from the `CommittorSimulation` into a base class. This was [already suggested by Andreas Singraber](https://gitlab.e-cam2020.eu/Classical-MD_openpathsampling/RF/issues/4), who is developing a reactive flux module for OPS.

- [x] Refactor `CommittorSimulation` into `ShootFromSnapshotsSimulation`
- [x] add `to_dict` and `from_dict`
- [x] return test coverage in `CommittorSimulation`
- [x] specialized tests of `ShootFromSnapshotsSimulation` when not subclassed

### DIFF
--- a/openpathsampling/__init__.py
+++ b/openpathsampling/__init__.py
@@ -128,7 +128,7 @@ from pathmover import (
 
 from pathsimulator import (
     PathSimulator, FullBootstrapping, Bootstrapping, PathSampling, MCStep,
-    CommittorSimulation, DirectSimulation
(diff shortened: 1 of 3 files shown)


dwhswenson:

Issue #660, opened 2017-01-27
NetCDF Chunksize Error


dwhswenson:
It appears that a new error will be introduced when we move to later versions of netCDF. Because we're currently pinned to avoid a `conda`-based issue (https://github.com/ContinuumIO/anaconda-issues/issues/1360), our tests are using:

* `libnetcdf 4.3.3.1-3`
* `netcdf4 1.2.2-np110py27_0`

On my local machine, I'm using:

* `libnetcdf 4.4.1_2` (called `netcdf` on MacPorts)
* `netcdf4 1.2.3_1` (called `py27-netcdf4` on MacPorts)

This said, I get the following error:

```python traceback
======================================================================
ERROR: testcollectivevariable.test_FunctionCV.test_storage_cv_function
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/dwhs/Dropbox/msm-tis/openpathsampling/tests/testcollectivevariable.py", line 125, in test_storage_cv_function
    storage_w.save(cv1)
  File "/Users/dwhs/Dropbox/msm-tis/openpathsampling/netcdfplus/netcdfplus.py", line 587, in save
    return store, store_idx, store.save(obj, idx)
  File "/Users/dwhs/Dropbox/msm-tis/openpathsampling/netcdfplus/stores/named.py", line 442, in save
    reference = super(UniqueNamedObjectStore, self).save(obj, idx)
  File "/Users/dwhs/Dropbox/msm-tis/openpathsampling/netcdfplus/stores/named.py", line 254, in save
    reference = super(NamedObjectStore, self).save(obj)
  File "/Users/dwhs/Dropbox/msm-tis/openpathsampling/netcdfplus/stores/object.py", line 875, in save
    self._save(obj, n_idx)
  File "/Users/dwhs/Dropbox/msm-tis/openpathsampling/storage/stores/collectivevariable.py", line 18, in _save
    self.add_diskcache(cv)
  File "/Users/dwhs/Dropbox/msm-tis/openpathsampling/storage/stores/collectivevariable.py", line 96, in add_diskcache
    chunksize=chunksize
  File "/Users/dwhs/Dropbox/msm-tis/openpathsampling/storage/stores/snapshot_wrapper.py", line 673, in add_cv
[...]

PR #669, opened 2017-02-18
Fast ensemble check

dwhswenson:
This adds a new argument to `Ensemble.__call__`, called `candidate`. If `candidate is True`, then shortcuts can be used to calculate whether the trajectory is in the ensemble. The assumption is that this is a "candidate trajectory," as described in path ensemble theory; i.e., a trajectory that could have been generated by shooting. 

For example, this will allow us to assume that all frames except first and last are not in any state for a (flexible length) TPS ensemble or a TIS ensemble. For ensembles where there is no shortcut to be had, `candidate` just has no effect.

This is the first step toward using the generalized CVs of #662 in analysis (simply need to replace the `TISEnsemble`'s use of max(orderparameter(trajectory)) by a CV that stores that result). That will be a huge improvement on analysis speed. If the max lambda is known, a TIS `candidate` trajectory can be checked by looking at the first frame, the last frame, and the max lambda. Currently, we loop over all frames, find the max lambda, and check that only the first and last are actually in the state. This will also make a big difference for sampling speed for toy models.

Note that we'll still have the occasional `sanity_check` during simulation, which will *not* use the candidate assumption. This is just to make sure that nothing goes terribly awry during the sampling.

This `candidate` approach is very similar to what we already had as `trusted` for `can_append` and `can_prepend`. Unfortunately, there appear to be a couple cases where the `__call__` with `trusted=True` is being already being used in a way that would conflict with replacing `__call__`'s `trusted` with what I'm doing with `candidate`, so it looks like `candidate` might be separate from `trusted` (Annoying, because they're almost the same, and `__call__` with `trusted=True` almost never happens. When it does, it is deep in `SequentialEnsemble`, so there might be a way to fix it, but it'll take some thought.)

- [x] add `candidate` flag to `Ensemble.__call__`
- [x] add speed-up for `TISEnsemble` using `candidate`
- [x] tests for `TISEnsemble` using `candidate` (including where candidate is too trusting)
- [x] add the use of `candidate` in replica exchange checks, path reversal checks

### DIFF
--- a/openpathsampling/ensemble.py
+++ b/openpathsampling/ensemble.py
@@ -2792,8 +2792,8 @@ class TISEnsemble(SequentialEnsemble):
         self.orderparameter = orderparameter
         # TODO: add max_orderparameter as a traj CV
         self.lambda_i = lambda_i
-        self._initial_volumes = volumes_a
-        self._final_volumes = volumes_b
+        self._initial_volumes = volume_a
+        self._final_volumes = volume_b | volume_a
 
     def __call__(self, trajectory, trusted=None, candidate=False):
         use_candidate = (candidate and self.lambda_i is not None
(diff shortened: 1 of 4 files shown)


dwhswenson:
This is ready for review. I couldn't do a performance analysis because #660 prevents me from running simulations locally, but this should speed up the sampling (especially for toy models).

This faster check only comes into play when an ensemble overrides the implementation of `__call__`. So far, that's only in the `TISEnsemble`. We don't have a separate `TPSEnsemble`, but I think I'll add one (and a `FixedLengthTPSEnsemble`) in order to use these speed-ups -- particularly useful in analysis.

PR #672, opened 2017-03-03
Fix bug in OpenMMSystemTopology

dwhswenson:
You can't get the number of atoms from `self.system` until you set `self.system`. (Apparently this initialization isn't covered in tests!)

Should be ready to merge as soon as tests pass.


dwhswenson:
Passed tests.

However, can this `Topology` even be used? It doesn't seem to work if I give it to the OpenMM `Engine`. (Can't convert this to mdtraj when setting up the `simulation`. Not sure *why* we convert to MDTraj though.)

jhprinz:
Actually, I am not sure where this is used at all? I think it was some legacy for OpenMMTools support. But that is not necessary anymore. I think we can remove this, can we?

dwhswenson:
> I think we can remove this, can we?

That may be better. I was looking for an easy way to set up an engine without having to load the PDB twice (once for OpenMM, once for OPS). It seemed like all the info that OPS needs should already be in the OpenMM `system` object, and so when I saw this `Topology`, I tried using it. Didn't do any good.

In general, I think we could do with some more topology-to-topology direct translations (the topology we're using should be easily made from or made into MDTraj or OpenMM topologies).

jhprinz:
Well, no, unfortunately, the system objects of openmm contains practically a topology but not in an accessible information. It is merely a complete description of the used force field. So, its enough to simulate but not enough to get the complete structure back.

I think this is a merging of the topology with all the ff parameters. So we really need both. Or you change to topology + forcefield + simulation options like cutoffs, etc... That would be smaller than topology + system, but much more effort to work with and less sure to remain the same between OpenMM versions.

I would remove it.

jhprinz:
Do you want to remove the class in this PR? And we just agree that we will not use it?

dwhswenson:
> Do you want to remove the class in this PR?

Done.

I do think we should clean up how we handle `Topology` objects. For example, we currently **require** a topology based on MDTraj to run OpenMM. This seems like we're conflating two codes. There are a number of points where I've been confused as to which kind of topology was expected; mostly the docs help, but it would be nice if we could find a way to make it less confusing. (I think one solution is that many objects, e.g. `MDTrajCVFunction`, could take *either* a `paths.MDTrajTopology` or a `md.Topology`)

jhprinz:
merging...

PR #673, opened 2017-03-13
Badchunksizebug

jhprinz:
Temporary fix for #660 . PR #662 also fixed this in a cleaner way.

It is still a bug in netCDF4 which prohibits setting a chunk size different from 1 after objects have been written to related dimension. See https://github.com/Unidata/netcdf4-python/issues/566


dwhswenson:
Can you add a comment to the code pointing to that issue in netcdf4-python? Just something to make it clear within the code that we're waiting on a resolution of that, and to make it easy to check if it has been resolved.

dwhswenson:
(Other than that, this looks ready to merge)

jhprinz:
Done.

PR #675, opened 2017-03-22
Fix deserialization of `openmmtools.integrator`s

dwhswenson:
When OpenMM deserializes subclasses of `CustomIntegrator`, it creates a `CustomIntegrator`, meaning that any additional methods defined in the subclass are lost. So, for example, we lose the `getTemperature()` method in `openmmtools.integrator.VVVRIntegrator`.

Fortunately, @andrrizzi came up with a nice solution for that in https://github.com/choderalab/openmmtools/pull/130. So all this does is wrap around his solution, and put it into a function that could be extended for integrators other than `openmmtools` (keeping the code in one place for readability). 

If the restored integrator isn't a subclass of `openmmtools.integrators.ThermostatedIntegrator`, or if, for some reason, the interface can't be restored, then nothing is different. If the interface can be restored, we'll have access to it.

### DIFF
--- a/openpathsampling/engines/openmm/engine.py
+++ b/openpathsampling/engines/openmm/engine.py
@@ -13,13 +13,13 @@ logger = logging.getLogger(__name__)
 
 def restore_custom_integrator_interface(integrator):
     """
-    Restore the interface to a custom integrator. 
+    Restore the interface to a custom integrator.
 
     When subclasses of CustomIntegrator are deserialized, they only show up
     as `CustomIntegrator`, and therefore lose any additional interface
     provided by the subclass.
 
-    So far, only subclasses of `openmmtools.ThermostatedIntegrator` are
+    So far, only subclasses of `openmmtools.RestorableIntegrator` are
     supported. Hopefully, we can establish some sort of more widely-used
     approach based on the tricks established in there.
     """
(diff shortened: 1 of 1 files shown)


andrrizzi:
I'm glad you found that useful. FYI, I've just merged choderalab/openmmtools#140 in which I've split the code responsible to restore the original `CustomIntegrator` subclass interface into its own `RestorableIntegrator` to make it more general than just getter/setter for temperature. The `ThermostatedIntegrator` inherits from that anyway, so this code should still work as expected.

openpathsampling/engines/openmm/engine.py
```diff
@@ -11,6 +11,35 @@
...
+    as `CustomIntegrator`, and therefore lose any additional interface
+    provided by the subclass.
+
+    So far, only subclasses of `openmmtools.ThermostatedIntegrator` are
+    supported. Hopefully, we can establish some sort of more widely-used
+    approach based on the tricks established in there.
+    """
+    # for openmmtools integrators
+    try:
+        import openmmtools
```
jhprinz: is it possible that you need to `import openmmtools.integrators` here? travis cannot find `ThermostatedIntegrator`

dwhswenson:
Thanks @andrrizzi : I'll switch this over to using the `RestorableIntegrator`! One question: will there be a new release of openmmtools soon? Tests in this branch are failing because, of course, the new code isn't in the omnia channel build we use.

@jhprinz: We can either wait for a new release of openmmtools or we can change the dependency to the current GitHub master of openmmtools. I vote for waiting for the release; this PR provides a more elegant solution to a problem I was having, but the workaround (manually enter the temperature again) isn't so hard in practice.

andrrizzi:
We are planning to release very soon, actually. Probably today or early next week.

andrrizzi:
`openmmtools 0.9` is up on Anaconda. Sorry it took a little longer than I thought.

dwhswenson:
Thanks @andrrizzi -- this passes tests now. @jhprinz -- Review this when you will.

Note that this technically means that we'll require `openmmtools` 0.9.1 or later (maybe 0.9.0? but 0.9.1 fixes a Python 2 bug).

andrrizzi:
0.9.1 fixes a python 2 bug in `openmmtools.alchemy`, so 0.9.0 should work for you.

PR #676, opened 2017-03-29
Improvements on progress output

dwhswenson:
This PR fixes a few minor annoyances with the progress output during a path sampling simulation. In particular, it includes the following:

- [x] Switch the speed reporting from "steps per second" to "seconds per step": TPS simulations of large systems have less than 0.01 steps per second, so the old version was uninformative; even our fast systems are around  0.1-0.2 seconds per step
- [x] Convert times (elapsed and remaining) from seconds to days, hours, minutes, seconds (easier to interpret once you get to tens of thousands of seconds remaining)
- [x] Move the whole progress string generation into `paths.tools`, so that it can be re-used by other objects. Just needs the time elapsed, the steps completed, and the total number of steps
- [x] Unit tests for new code

New output looks something like this:

```
Working on Monte Carlo cycle number 215
Running for 1 minute 44 seconds -  0.91 seconds per step
Estimated time remaining: 5 hours 2.52 minutes
```

The time elapsed is always the full string (`1 day 2 hours 3 minutes 4 seconds`). The time remaining always contains the first two (occupied) "levels" (day, hour, minute, second) with the remainder to 2 decimal places in the last. So (`1 day 2.05 hours`, or as above, `5 hours 2.52 minutes`). I thought this would be the most readable (and it acknowledges that the time elapsed is exact, whereas the time remaining is a rough estimate).

### DIFF
--- a/openpathsampling/ensemble.py
+++ b/openpathsampling/ensemble.py
@@ -908,15 +908,19 @@ class Ensemble(StorableNamedObject):
             the number of attemps on a trajectory to extend
         """
 
+        logger.info("Starting extend_sample_from_trajectories with level "
+                    + str(level))
         if level == 'native':
             sub_ensemble = self
         else:
             if not hasattr(self, 'extendable_sub_ensembles'):
+                logger.info("Missing ensemble.extendable_sub_ensembles")
                 return None
 
             sub_ensembles = self.extendable_sub_ensembles
 
             if level not in sub_ensembles:
+                logger.info("Missing level: " + repr(level))
                 return None
 
             sub_ensemble = sub_ensembles[level]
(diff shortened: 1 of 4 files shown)


dwhswenson:
This also added some additional logging to `Ensemble.extend_sample_from_trajectories` that I needed for some debugging I was doing.

dwhswenson:
Now that #681 is in, this passes! Ready for review.

The `pretty_print_seconds` isn't as pretty of code as it might be (more elegant approaches may exist), but it does some smart rounding (it knows to round "2  days 1 hour 59 minutes 50 seconds" to "2 days 2 hours" -- which includes changing the singular "hour" to plural "hours.") So I'm happy enough with it for now.

jhprinz:
Lookgs good to me. The formatting seems to be a lot of code, but I don't know a better solution. Will merge now.

dwhswenson:
> The formatting seems to be a lot of code

I'm pretty sure there is a better solution, but it would involve a complete rewrite. The problem was that my first version forgot about rounding up, so I kinda had to hack that in. Something to rewrite someday; lots of other things much higher on the priority list!

PR #677, opened 2017-03-30
Remove unused requirements

jhprinz:
There is some legacy requirements of packages not needed anymore.

Remove

```
pyyaml
pymbar  # might be added in the future again
docopt  # was never used
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ install:
 
 script:
   - echo travis_fold:start:conda.build.package
-  - conda install --yes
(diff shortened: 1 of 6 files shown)


jhprinz:
I think this is ready.

jhprinz:
Ready

dwhswenson:
It looks like the unpin here causes problems for downstream projects (OPSPiggybacker started crashing its nightlies after this). Should we re-pin? (but to the better choice of `hdf4 4.2.12`)

PR #678, opened 2017-03-30
[WIP] Metapackage to load recommended dependencies

jhprinz:
A metapackage that will also install all recommended dependencies. Not just the minimum.

### DIFF
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -20,7 +20,6 @@ requirements:
     - numpy
     - scipy
     - pandas
-#    - hdf4 4.2.12
     - netcdf4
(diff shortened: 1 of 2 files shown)


dwhswenson:
Closing this as no longer relevant: we have other tools to accomplish this task, and we're moving toward only having conda recipes in the necessary ([`conda-forge` feedstock](https://github.com/conda-forge/openpathsampling-feedstock)) repo. For now, our conda recipe is a placeholder for install stuff, but that should move to a `requirements.txt` that will work with either conda or pip.

At some point, we may make separate `pip_extras.txt` and `conda_extras.txt` files if we encounter cases where packages are named differently in pip and conda. For now, that's not necessary.

PR #684, opened 2017-04-20
Resampling statistics

dwhswenson:
Very basic setup for doing some resampling to get error estimates. The main idea is to create an overall framework that takes a function that generates a `pandas.DataFrame` as output (e.g., our `rate_matrix` calculation) and can return the statistics we're interested in. This is all designed for a case where we're using `pandas.DataFrame` as a relatively small table (as it is with our `rate_matrix` and most other analyses where we output a `DataFrame`.)

There are three main parts to this:

1. Selecting sets of samples to use to recalculate (currently, just implementing block resampling; others could be added later)
2. Applying the (user-defined) function to each set of samples, resulting in a set of `pandas.DataFrame`s for each result
3. Getting statistics from the set of `pandas.DataFrame`s (to help estimate uncertainty)

Tasks:

- [x] Get statistics out of `pandas.DataFrame`s
- [x] `ResamplingStatistics` object to hold relevant statistics and results from subsamples
- [x] tests for `ResamplingStatistics`
- [x] `BlockResampling` object to select sequential blocks of samples
- [x] tests for `BlockResampling`
- [x] docstrings
- [x] update MISTIS example to include error calculation

Note that this will cover stuff for getting errors from `DataFrame`s. If the output is one of our `LookupFunction`s (as with crossing probability and some of our internal histograms), then most of the statistics (mean and stddev) are already included in `numerics/lookup_function.py`.

### DIFF
--- a/examples/toy_model_mistis/toy_mistis_3_analysis.ipynb
+++ b/examples/toy_model_mistis/toy_mistis_3_analysis.ipynb
@@ -411,7 +411,7 @@
    },
    "outputs": [],
    "source": [
-    "resampling = paths.numerics.BlockResampling(storage.steps, n_blocks=1)"
+    "resampling = paths.numerics.BlockResampling(storage.steps, n_blocks=n_blocks)"
    ]
   },
   {
--- a/examples/toy_model_mistis/toy_mistis_3_analysis.ipynb
+++ b/examples/toy_model_mistis/toy_mistis_3_analysis.ipynb
@@ -23,7 +23,7 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "mistis.nc\n"
+      "../toy_mistis_1k_OPS1.nc\n"
      ]
     }
    ],
@@ -370,7 +370,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 18,
+   "execution_count": 3,
    "metadata": {
     "collapsed": true
    },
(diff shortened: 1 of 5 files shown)


dwhswenson:
This PR also changes the index/column labels of the `rate_matrix`. Currently, we use the `Volume` itself as the label. However, this has several problems:

1. It prints with the full string description of the volume, not the user-defined name. That's annoying and hard to read.
2. Since pandas 0.18.1, using a callable as index/column is problematic, because when you try to select, pandas will [apply that callable to the `DataFrame`](http://pandas.pydata.org/pandas-docs/stable/indexing.html#selection-by-callable) instead of treating it as a dictionary key. This was causing problems for the way I was calculating percentiles in this PR (and was likely to lead to user confusion).

In this PR, I switch it to using the `Volume.name`, which is the user-defined name if one is available, or a simplified string (e.g., `0.0<cv<1.0`) if the user did not set a name. This means that setting/access requires `state.name` instead of just `state`.

In summary: better behavior (and avoid problems that could really confuse the user), plus better looking output in our analysis notebooks.

dwhswenson:
This is ready for review and merge.

This updates to the MISTIS example are only to show how to use this; I still need to look more carefully at those results (especially why it is that I'm seeing rates vary by an order of magnitude). Also, note the that error here is technically incorrect for MISTIS, because it assumes no error in the (separately calculated) flux. In practice, though, the error in the flux calculated this way will be tiny compared to the error in the crossing probability.

I'll add some more tools to look at what's going on with the variance in the rate, but I wanted to get this PR merged so that @jhprinz could use it to start the error analysis of the AD MSTIS example.

jchodera:
Awesome! 

How is the number of blocks selected? Seems to default to 20 blocks unless otherwise specified---was that just an arbitrary choice?

dwhswenson:
Yeah, 20 is a pretty arbitrary default. It's about the smallest number where I feel like "standard deviation" can mean anything; no other reason. The current setup really expects the user to figure out appropriate blocks.

You mentioned something about finding a better guess for the appropriate block length (some time correlation method in pymbar, right?) -- can you point me to more on that? This PR is lower-level, easily reusable code, but better tricks to automatically select the number of MC steps per block would be useful in higher-level analysis (and with the MISTIS example, which I've been analyzing this afternoon).

jchodera:
If you have some quantity whose correlation time is indicative of the correlation time for the sampler---such as a state indicator function---you can use [`pymbar.timeseries.statisticalInefficiency`](http://pymbar.readthedocs.io/en/latest/timeseries.html#pymbar.timeseries.statisticalInefficiency) to estimate the minimum correlated block size.

dwhswenson:
@jhprinz : this is now fixed to work with Python 3. Merge at will!

jhprinz:
does this also merge the py3 changes or is it only potentially compatible...

dwhswenson:
The Py3 changes from your version of the PR are already merged. This now includes this PR plus Py3 fixes so this PR works with the testing on 3.6.

Remaining fixes in #689 are mainly to get the notebook tests working. :-) We're already basically Python 3 compatible!

jhprinz:
Cool... So, I will merge this...

PR #685, opened 2017-04-21
range -> xrange in DirectSimulation.run

dwhswenson:
This fixes a really stupid mistake in `DirectSimulation` that causes problems when you want to run a billion MD steps. (In Python 3, it would have been correct!)


dwhswenson:
This can be reviewed at any time (it's a 1-character fix; should be trivial...)

PR #686, opened 2017-05-13
New TIS Analysis

dwhswenson:
This PR provides an alternate approach to do TIS analysis. Overall, I'm suggesting a 3-step process for updating our analysis subsystem; this PR will do step 1:

1. Have this as an alternate approach to the current approach. It has the advantage that it is more flexible, but the disadvantage that it is a little harder to use than the old version. For now, both exist side-by-side.
2. In a future PR (1.0 or 1.x), keep the simplified API of the old version, but internally use the new version. That means we'll be getting rid of old code that doesn't have unit tests, and replacing it with code that does.
3. Create better default approaches, and (probably in 2.0) replace the current defaults with defaults that will be even easier (and less error-prone) for users (see, e.g., #551).

What's new in this approach:

First, none of the TIS analysis code will be located inside the "high-level" objects like `Transition` or `TransitionNetwork`. Moving the analysis outside of that makes it easier to customize the analysis approach. For example, it is now very easy to change the details (like cutoff) of the WHAM procedure, or to replace it with something else (once we have something else — MBAR, TISBAR — implemented). It also makes it easier to add very different approaches to obtain the rates from TIS (e.g., using path-type analysis), and provides a more natural way to switch between a pre-calculated flux (as we need for MISTIS) and the minus move flux.

Another big idea in this is that most of the classes are basically functions. That is, they don't cache their results. Caching is convenient for efficiency, but really obnoxious when trying to re-use parts in some interesting way. The only objects that cache results are the main `TISAnalysis` object (and its subclass, `StandardTISAnalysis`). That way the "outermost" objects in the analysis heirarchy cache the results for efficiency, and the details of how things are calculated are treated more like functions.

This is a huge PR, but nearly all of it is in new files. A few more details will follow in the comments below. Before it is complete, I'll also add a detailed overview of this analysis system to the docs; once it is there, that might be the best place to go in order to understand this PR.

- [x] New TIS analysis approach for crossing probability
- [x] Unit tests for crossing probability
- [x] New TIS analysis approach for conditional transition probability
- [x] Unit tests for conditional transition probability
- [x] New TIS analysis approach for pre-calculated flux
- [x] Unit tests for pre-calculated flux
- [x] Main `TISAnalysis` class and `StandardTISAnalysis` subclass
- [x] Unit tests for `StandardTISAnalysis`
- [x] Test new TIS against MISTIS example
- [x] New TIS analysis approach for minus move flux
- [x] Unit tests for minus move flux
- [x] Test new TIS against MSTIS example
- [x] Full test coverage
- [x] Performance improvement on minus move flux/sequential ensemble
- [x] Docstrings
- [x] Example notebooks
- [x] Detailed overview docs
- [x] Split into smaller files in `analysis.tis` subpackage


dwhswenson:
A few comments on some changes you might notice (outside the main new files):

#### Flux

The minus move flux here re-uses code from #435 instead of the older code from #254. The results are not identical, because when the minus move has two excursions from the interface, the old code only used the first in its statistics. This uses both. (And a checked [by eye] that the lengths found for "inside" were the same and that alternating length from the "outside" were the same as the code from #254.)

#### `TrajectoryTransitionAnalysis.flux_from_flux_dict`

Refactored so that this part of the code can be re-used elsewhere.

#### Changes in `network.py`

I added some convenience dictionaries for mapping between sampling transitions and analysis transitions.

dwhswenson:
This is ready for review.

It is an incredibly large PR. My suggestion for reviewing it:

1. It includes new documentation in `docs/topics/tis_analysis.rst` and the visual description in `docs/topics/tis_analysis_structure.png`. Look at those first, to get the big picture.
2. See it in use in the new notebook `example/toy_model_mstis/toy_mstis_A3_new_analysis.ipynb`.
3. Most of the actual additions changes are in `openpathsampling/analysis/tis/`, a new directory with several new files. This is the main thing to review. These things are *not* in the main `openpathsampling` namespace; you have to use them by importing `openpathsampling.analysis.tis`.
4. Other changes are relatively minor, but include some performance stuff (especially in sequential ensemble), some new conveniences in networks, some minor docs updates, and some PEP8 stuff.

openpathsampling/analysis/tis/flux.py
```diff
@@ -0,0 +1,330 @@
...
+
+    def _get_minus_steps(self, steps):
+        """
+        Selects steps that used this object's minus movers
+        """
+        return [s for s in steps
+                if s.change.canonical.mover in self.minus_movers
+                and s.change.accepted]
+
+    def trajectory_transition_flux_dict(self, minus_steps):
```
jhprinz: The flux_dict looks like a complicated object. Is there a reason not to create a class for it? I mean, is a user meant to create this object by himself for some reason? I just thought we could hide some of the complexity this way.

jhprinz:
In general, this looks fine to me. I have a few suggestions or think we might think about.

You are using some complicated looking dict constructs to hold specific data. It looks like a very flexible thing to use, but it makes it - as you mention above - also complicated to use. I think most objects are already generated in the correct structure, but I was wondering if it would be possible to hide this complexity to some degree in classes that are returned. A user that does not create these objects does not really care how the data is internally structure and might be _scared_ by an object like this. 

I guess you know what I mean. E.g. the fluxes could just be a `FluxData` or `FluxDescriptor` object that holds the data with some convenience function to access specific data and some helper constructors to make it easy to construct these. But we can do that in a later version anyway.

I like the figure that is also for understanding TIS a good image to have when talking about the decomposition of mathematical objects needed to get the rate, etc. And so I see the idea to make it flexible and replace the methods used. Say replace flux computation by given fluxes from an external simulation. 

Overall the analysis part is (and I have so solution to this) one of the more complicated parts, which I hope will not stay that way. Say in a perfect world, the definition of ensembles should imply the way to compute all properties and also which properties are accessible and not require knowledge about the whole analysis process. I mean, that would be pretty cool, but possible a multi-year project :) ...

I say, go ahead an merge as soon as you want to. It looks good the way it is.

dwhswenson:
Thanks; I'll merge, but add discussion here to clarify a few points so you can see if there's anything that need to be modified in the near-ish future. Note that I consider this still a somewhat experimental API; even if it goes into 0.9.3 as-is, we should be allowed to change it before 1.0. (In fact, I already know one thing I'll be changing, which is to get the WHAM code to return weights as well, which we will need for reweighted path ensemble -- a quick addition, if I wasn't coming up on a deadline!)

To your specific comments:

> You are using some complicated looking dict constructs to hold specific data.

Overall, I would split these into 3 categories:

1. The (very complicated) `results` dictionary in the `TISAnalysis` objects.
2. The dictionary used as input to the `DictFlux` object (or output from the `MinusMoveFlux`)
3. The dictionary based on the `TrajectoryTransitionAnalysis` for fluxes, which has keys of `'in'` and `'out'`, each mapping to a `TrajectorySegmentContainer` holding specific frames of a trajectory representing "inside the interface" and "outside the interface" segments.

I'll take these in order 2, 1, 3:

Only the second should ever be created by a user. That one is of the format `{(state, interface): flux_value}`, which is, I think, already the simplest possible way to make it. It's the flux out of a given state and through a given interface, so you need the state and interface as key, and the flux as a value. Also, the `MinusMoveFlux` outputs in the same format that the `DictFlux` takes as input, so that can simplify some things. (Don't remember if `DirectSimulation` outputs in the same format, but if not, I should make that change.)

The first category, the `results` dict, does almost exactly what you suggest, except without a new class. The `results` dictionary should almost never be used directly; instead, you use convenience methods inside the `TISAnalysis` or `StandardTISAnalysis` class to access it. But instead of using a container class, we can just save the `results` dictionary (into `storage.tags`) and load it into the class. That stores the final results as well as any intermediates that might be worth analyzing. By just using the dictionary instead of a class, we don't have to worry about how it is handled in storage, don't have to worry about the user understanding immutability, and we can keep a clean separation of the "function" and "data" parts of the analysis (the setup of the way you do the analysis is a "function," the results are the "data;" once you mix the two, we start getting the problems with extensibility that this PR is trying to avoid).

I considered having a separate results class, but the problem was that then you'd need a different results class for each analysis setup, and that would be too hard to manage (not to mention needing to create stores for them). I think that properly dealing with these intermediates requires lots of documentation. This PR already includes a lot, but there may be a need for even more.

The third category is leftover from `TrajectoryTransitionAnalysis` (#435, #448). This could be moved into a `namedtuple` instead of a `dict`, but hasn't been yet. This is only an intermediate part of the calculation, and won't usually be accessed by users (but it is saved because it could be useful for troubleshooting, if a user knows how to use it.)

> A user that does not create these objects does not really care how the data is internally structure and might be scared by an object like this.

I see what you're saying, but my hope is that these objects would normally be hidden from users, in the same way that the fact that the minus move is actually a complicated `ConditionalSequentialMover` is hidden -- always accessible if you know how to look for it, but not something that you *must* deal with. Digging into the details at that level is useful for troubleshooting, but hopefully not something you'll need to do.

The only exception is if you use the `DictFlux`, there's number 2 above. But I think that is as straightforward as it can be.

For the simplest cases, we can (and probably should) directly replace the current analysis in the `TransitionNetwork` objects with this. By using default behavior, usage there is just as easy as it currently is. What's new in this PR is the ability to set up an analysis that doesn't use the defaults. Overall, I'd recommend pushing users toward using these analysis classes instead of the network methods, because I think that will likely help them in the future, even if it technically takes one more line of code (creating a `StandardTISAnalysis` object).

> I like the figure that is also for understanding TIS a good image to have when talking about the decomposition of mathematical objects needed to get the rate, etc. And so I see the idea to make it flexible and replace the methods used.

Glad you like that figure! I'm pretty happy with the docs on the TIS analysis here; I think it's a clearer picture of the connection between code and math than is often made. And yes, the flexibility is the primary goal of this PR.

> Say in a perfect world, the definition of ensembles should imply the way to compute all properties and also which properties are accessible and not require knowledge about the whole analysis process. I mean, that would be pretty cool, but possible a multi-year project :)

Haha -- multi-year or multi-lifetime? :-) I think the biggest challenge is finding the right amount of extra troubleshooting information to provide. I mean, the rates are easy enough to calculate, but most of the additional information stored in the `results` dict are about troubleshooting and sanity-checking your results. (This is also where the hypothetical "red flags" module I've talked about for years plays a role -- I worry that inexperienced users will get nonsense results and not know how to check their data for signs that they can't trust it.) But that's definitely stuff to explore in the future.

PR #689, opened 2017-06-28
[WIP] Python 3 compatibility 

dwhswenson:
For some collaborative stuff I'm working on with the PyRETIS devs, we needed OPS to become Python 3 compatible (PyRETIS is py3-only). So I've spent the last couple days updating OPS to add Python 3 support.

- [x] Update Travis script to test py-2.7 and py-3.6
- [x] Fix nosetests to work with both 2.7 and 3.6
- [ ] Fix notebook tests to work with both 2.7 and 3.6
- [ ] Check all local notebooks
- [ ] Clean up any automated additions (especially in testing: zip, range, old_div)

Note that it might be worth spending some time double-checking performance on some of these changes. @jhprinz, I assume you'll tell me if you see anything that might be a performance killer, especially with the significant changes in making netcdfplus/storage ready for Python 3.

PR #690, opened 2017-06-28
Python3 compatibility with storage

jhprinz:
This aims to make storage and netcdfplus compatible with Python 3.

### DIFF
--- a/openpathsampling/engines/dynamics_engine.py
+++ b/openpathsampling/engines/dynamics_engine.py
@@ -22,7 +22,8 @@ logger = logging.getLogger(__name__)
 if sys.version_info > (3, ):
     property_name = lambda p: str(p.fget.__name__)
 else:
-    property_name = lambda p: str(p.fget.func_name)
+    # property_name = lambda p: str(p.fget.func_name)
+    property_name = lambda p: str(p.fget.__name__)
 
 # =============================================================================
 # SOURCE CONTROL
--- a/openpathsampling/engines/dynamics_engine.py
+++ b/openpathsampling/engines/dynamics_engine.py
@@ -19,12 +19,6 @@ from .delayedinterrupt import DelayedInterrupt
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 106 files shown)


openpathsampling/engines/dynamics_engine.py
```diff
@@ -12,13 +12,19 @@
...
+from .delayedinterrupt import DelayedInterrupt
 
 logger = logging.getLogger(__name__)
 
+if sys.version_info > (3, ):
+    property_name = lambda p: str(p.fget.__name__)
+else:
+    # property_name = lambda p: str(p.fget.func_name)
+    property_name = lambda p: str(p.fget.__name__)
+
```
dwhswenson: Ah -- since this is also available in 2.7, we should just use `fget.__name__` instead of wrapping a function here. See also line 272. I had kept the use of `fget.func_name` that had been in the original code, guess it is already aliased.

openpathsampling/engines/dynamics_engine.py
```diff
@@ -262,8 +268,8 @@ def __getattr__(self, item):
                     # alternately, trust the fixed result with
                     # return result  # miraculously fixed
                     raise AttributeError(
-                        "Unknown problem occurred in property" + 
-                        str(p.fget.func_name) + ": Second attempt returned"
+                        "Unknown problem occurred in property"
+                        + property_name(p) + ": Second attempt returned"
```
dwhswenson: See other comment: this should be replaced with the 2.7/3.x compatible `str(p.fget.__name__)` (may not require `str`? can't hurt though)

jhprinz: thanks, will do. Yes, I tried that in both and it seemed to work... No idea why I did not use that in the first place...

jhprinz:
passes tests...

dwhswenson:
woohoo! I'll merge this, then continue more complete Python 3 support in #689. (Still need to test notebooks, etc, with Py3.)

PR #691, opened 2017-07-01
[Python3] Trajectory Variables

jhprinz:
Rebase of branch in PR #662 to work with python 3. For completeness here the original description.

This adds the feature to use CVs for every object. To be precise, it replaces CVs by some general feature of NetCDF+ and now CVs are just a special case of these general `Attribute` object.

One example is an Trajectory attribute.

 ```py
max_cv = FunctionAttribute('max_cv', Trajectory, lambda t, cv: max(cv(t)), cv=cv)

# print all maxima of `cv` for all trajectories
print max_cv(storage.trajetories)
```

pretty easy now. You can of course _annotate_ all sorts of objects like this. Storage and caching works exactly like for `CV`s. Only difference is that a normal `Attribute` does not have a `reversible` property. This only makes sense for `Snapshot`s that have this quality.

The main reason to push this was that I could not analyse the long Alanine trajectory. It just took too long. One way to solve this would have been to increase the cache sizes, but this new approach makes it possible to cache the _max_lambda_ values efficiently and hence speed up the rate computation a lot.

This still would require the analysis to use this or as @dwhswenson suggested to make this attributes part of a network or whatever object is responsible for the computation of the max_lambda in the analysis.

I think to really get more speed ups, we also need a way to cache information about the _minus moves_. This should be close to the annotated trajectories.

#### Other changes

The generalisation of CVs also made some changes to the package structure.

1. `chaindict.py` is now part of `netcdfplus` which I think is a good place. it was always weird to be in ops
2. I also tried to split all the generic parts and the snapshot specific ones and the general parts are now in `netcdfplus` while the rest is in `storage`

#### Note

This will make a change in the Storage so that you can still read new and old files, but you cannot append old style files. At least I disabled that by default although it should be possible to append these files as long as no new attributes/cvs are added

#### Todo

- [x] fix notebooks
- [x] add some more testing
- [x] add docs
- [x] do dome speed simplifications

### CODE REVIEW

**Comment by dwhswenson** [comment]
Is this ready for review?

**Comment by jhprinz** [comment]
Found some speed problems in CVs and solved them. Should be faster by a factor of 2-3 or even more now.

**Comment by jhprinz** [comment]
fixed some really bad error that caused the cache to not work correctly and break some tests.

**Comment by jhprinz** [comment]
@dwhswenson This is done, but there was another question that popped up and was about reversibility in trajectories. Just saying that it makes sense to implement special storage for trajectories and its reversed like we do for snapshots. Especially since we now have trajectory CVs.
This implies changes to the storage format though.

**Comment by dwhswenson** [comment]
> This is done, but there was another question that popped up and was about reversibility in trajectories.

I see the idea, and the reason it would be more elegant than what we have now. But what we have now will work, right? Just that functions of a trajectory have to be recalculated for reversed trajectories; not optimal, but not so bad.

Keep in mind that many types of simulations won't have reversed trajectories (or will only use one of the reversed/not reversed). Only those with path reversal moves really use both.

I think that the major changes it would require in storage (breaking backward compatibility, right?) mean we should not do it. In any case, I think we should plan a better storage system for trajectories based on the idea that, instead of a list, you can define a trajectory as a continuous slice of the snapshot storage from `begin` to `end`, and therefore only need to save `begin` and `end` for each traj, and then load each segment with one call to the netCDF instead of (as I understand we do now) one call for each snapshot. But all of that is for 2.0 or later.

**Comment by dwhswenson** [comment]
(If not changing the storage of trajectories means that this is ready for review, remove the WIP tag and I'll review it!)

**Comment by jhprinz** [comment]
Go ahead...

File: openpathsampling/netcdfplus/chaindict.py [hunk 1]
```diff
@@ -57,6 +57,8 @@ class ChainDict(object):
 
     def __init__(self):
         self._post = None
+        self._iterables = (list, tuple)
```
dwhswenson: should `set` be included as an iterable?

openpathsampling/storage/storage.py
```diff
@@ -131,15 +129,26 @@ def _initialize(self):
...
-        # been initialized yet
-        self.finalize_stores()
+        self.set_caching_mode()
 
     def _restore(self):
         self.set_caching_mode()
 
+        if hasattr(self, 'cvs'):
+            logger.info('Opening an old version that handles CVs differently. '
+                        'You cannot extend this file, only savely read it.')
```
dwhswenson: "Safely" not "savely" (or just remove the word entirely)

openpathsampling/storage/stores/snapshot_wrapper.py
```diff
@@ -143,7 +142,7 @@ def load(self, idx):
                     return self.storage.fallback.stores[self.name].load(idx)
                 else:
                     raise ValueError(
-                        'str %s not found in storage or fallback' % idx)
+                        'long %s not found in storage or fallback' % idx)
```
dwhswenson: "long" of just "int" (or maybe "object ID")? Thinking that Py3 doesn't distinguish `long`

openpathsampling/netcdfplus/base.py
```diff
@@ -171,7 +171,7 @@ def base(cls):
         return cls._base
 
     def __hash__(self):
-        return hash(self.__uuid__)
+        return self.__uuid__ & 1152921504606846975
```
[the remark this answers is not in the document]
dwhswenson: (Addendum: I know that it is a mask; but can you explain what the mask is for?)

dwhswenson:
bounced assignment back to @jhprinz to reply to my review ... should be pretty easy!

openpathsampling/netcdfplus/base.py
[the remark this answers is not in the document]
jhprinz: The UUID is a 16-byte long int and I wanted to pick the last half. This picks the last 60 bits to mimick the `id(obj) / 8` from the default implementation and one for the sign. We can change that, but this is fast and should work.

openpathsampling/netcdfplus/chaindict.py
jhprinz: yes. The check for __iter__ or Collections.iterables is better but much slower. `set`, `frozenset` make sense though.

openpathsampling/storage/storage.py
jhprinz: thanks...

openpathsampling/storage/stores/snapshot_wrapper.py
jhprinz: changed

jhprinz:
@dwhswenson Have a look. I made some changes...

dwhswenson:
You have an indentation error in your last commit ... all tests fail. This appears to be due to incomplete code at pathmover.py line 633: you start defining functions and don't actually fill them out.

Also, why push the implementation into `_move`? Is this going to screw up things that users (ESDW participants) have been implementing? What's all the `promise` stuff? This does not appear to be conceptually related to this PR. Are you mixing in something from another project here?

dwhswenson:
I think you need to roll back the changes to `pathmover.py` from this. The other changes look good. I think something like `git checkout 481dd96 -- pathmover.py`, followed by commit and push, should do it.

jhprinz:
ups, I was experimenting with an async implementation of the PathMover... Seems that also got checked in... Hmmm, will fix that... Sorry.

jhprinz:
That fixed it. Was only pathmover.py.

jhprinz:
@dwhswenson Passes tests...

dwhswenson:
Excellent! Looks good now. Merging.

PR #698, opened 2017-08-15
Add `storage.tags` shortcut to access `storage.tag`

### LINKED ISSUE #643: `storage.tag` or `storage.tags`?
I think the subject says it all. Almost all of our stores are plural. However, it is `storage.tag`, not `storage.tags`.

My suggestion: make one an alias of the other, so that both work and refer to the same object.

(I bring it up because I keep typing `storage.tags['name'] = object`, getting an error, and having to correct it.)

jhprinz:
Resolves #643. Super simple, but is more consistent in the naming.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,8 +35,8 @@ addons:
     - pandoc
 
 env:
-    - CONDA_PY=2.7  CONDA_NPY=1.10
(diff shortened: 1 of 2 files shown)


dwhswenson:
Long awaited! Not a hard change to make, it looks. Should we deprecate the singular form before long?

Merging.

PR #704, opened 2017-08-22
Quick fix for problem with move_summary

dwhswenson:
This fixes a problem with `MoveScheme.move_summary` that @hejung found. I had used `is` in a case where I should have used `==` (to use UUID to check that the mover is the same). Was only a problem in some edge case (not fully identified, since it depended on the scheme) where the scheme was already in memory. For some reason, movers loaded from `storage` were re-instantiated (@jhprinz, any idea how that might have happened)? This meant that the mover in the `change` will be `==` to the one in the scheme, but not `is`.

### DIFF
--- a/openpathsampling/high_level/move_scheme.py
+++ b/openpathsampling/high_level/move_scheme.py
@@ -765,7 +765,7 @@ class MoveScheme(StorableNamedObject):
             group = my_movers[groupname]
             for mover in group:
                 key_iter = (k for k in self._mover_acceptance.keys()
-                            if k[0] is mover)
+                            if k[0] == mover)
 
                 for k in key_iter:
                     stats[groupname][0] += self._mover_acceptance[k][0]


jhprinz:
Hmm, I can only imagine that there is a problem in the recreation of the movescheme that it will recreate part of it and hence these objects will be recreated and be identical. Not sure why `==` would work in that case if UUID is used. This should be different as well. I will have a look when I have some time.

jhprinz:
also, if one of the object is a proxy then `==` will work, but not `is`

dwhswenson:
> I will have a look when I have some time.

Might require the example from @hejung. It worked fine with one move scheme, but not with another. No idea why that was.

> also, if one of the object is a proxy then `==` will work, but not `is`

These are `PathMover`s, so there shouldn't be any proxies. But the general point is important, and the reason that this should be `==` instead of `is` anyway.

In any case, I'll go ahead and merge this. It sounds like @jhprinz has seen enough to approach the (2-character) changes!

hejung:
My example can be found in https://gitlab.e-cam2020.eu:10443/hejung/sr_shooter/blob/master/examples/OneWayShooting_vs_TwoWayShooting.ipynb

if i use scheme.move_summary(steps) for the twoway shooting scheme that still lives in memory on the storage that it used to write its data to it works.
if i use scheme.move_summary(steps) for the same object that still lives but on the newly created analysis storage (just reopened the store i closed before) it does not work.
if i also load the scheme from that reopened storage it works.

PR #706, opened 2017-09-01
Updates to direct simulation

dwhswenson:
This PR involves several small updates and corrections to the `DirectSimulation` object.

- [x] Rates and fluxes now use `engine.snapshot_timestep` (if available) to get correct time/units, instead of returning numbers based on number of snapshots. Number of snapshots is a fallback when `snapshot_timestep` isn't available.
- [x] Correction to multiple state rate calculation. Previous calculation was only correct for 2-state systems.
- [x] Update to use `.name` of volumes for labels in pandas objects (this is the problem with using callables as column/index labels in recent versions of pandas)

### DIFF
--- a/openpathsampling/pathsimulator.py
+++ b/openpathsampling/pathsimulator.py
@@ -9,6 +9,8 @@ from openpathsampling.netcdfplus import StorableNamedObject, StorableObject
 import openpathsampling as paths
 import openpathsampling.tools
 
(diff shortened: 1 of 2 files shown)


dwhswenson:
Passes tests; ready for review!

jhprinz:
Looks good. Merging...

PR #708, opened 2017-09-08
Python3 notebooks, part 1

dwhswenson:
This is a PR to merge parts of #689 as they come through. (Perhaps #689 should be converted to an issue, instead of a PR).

Basic problem is this: Notebook merge conflicts are a pain to deal with. And it will take a while to get all the notebooks fixed up for Python 3 support. In the meantime, it is better to merge them in quickly as they get updated for Py3, rather than have to deal with merge conflicts in the notebooks.

This PR ensures that I can run `ipynbtest` successfully in either Python 2.7 or Python 3.6 for the following notebooks in `examples/`:

- [x] `tests/langevin_integrator_check.ipynb`
- [x] `tests/test_snapshot.ipynb`
- [x] `tests/test_snapshot_modifier.ipynb`
- [x] `tests/test_openmm_integration.ipynb`
- [x] `tests/test_netcdfplus.ipynb`
- [x] `misc/tutorial_path_histogram.ipynb`

I think `tests/test_pyemma.ipynb` should also work, but I can't get PyEmma working on my machine because apparently it used some hidden API calls to MDTraj which have since changed, and I can't seem to upgrade it with `pip` due to some compilation problem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,6 @@ addons:
 
 env:
     - CONDA_PY=2.7
-    - CONDA_PY=3.4
     - CONDA_PY=3.5
     - CONDA_PY=3.6
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ script:
   - conda info --envs
   - source activate test
   - conda install --yes future
-  - conda install --yes jupyter nose nose-timer python-coveralls msmbuilder pyemma #ipynbtest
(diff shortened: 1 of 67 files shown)


dwhswenson:
Passes tests. This is ready for review.

Note that this also adds testing for 3.5 (so we test 2.7, 3.5, 3.6). Notebooks are only tested in 2.7, but we run nosetests in all of them. (The point of this PR is to move toward having notebook testing in all, as well.)

dwhswenson:
Merging this one in. It is the first step toward #714.

PR #709, opened 2017-09-08
Release 0.9.3

dwhswenson:
We're long overdue for a release. This PR is in preparation to do one soon (probably after a few PRs that are nearly ready get merged in).

I'm also using this PR to simplify the install procedure. We should have all requirements to run tests (except the test running code, i.e., `nose` and `ipynbtest`) as part of the conda recipe.

Open issues/PRs that should be finished before this is merged:

- [x] #686 (New TIS Analysis)
- [x] #713 (New opcode in Py 3.6)
- [x] #708 (Some notebook updates for Python 3)
- [x] #711 #715 (Deployment of documentation)


dwhswenson:
This is ready for review and merge. Mostly, it bumps version numbers.

It also tries to satisfy the rule that you should be able to immediately run tests from installing based on the conda recipe. So I moved (most) things that are used in testing into requirements to be installed.

There's one exception, which is MSMBuilder. For some reason, conda installing MSMBuilder at the same time as everything was giving us an old version of numpy in py-2.7. This was leading to linking issues with the compiled libraries (see https://travis-ci.org/openpathsampling/openpathsampling/jobs/273355445 for an example). For whatever reason, this isn't a problem if MSMBuilder is installed *after* everything else.

In order to still satisfy the rule that tests should run based on the conda recipe, I made the MSMBuilder tests optional (in `testcollectivevariable.py`). In our Travis runs, we do the extra installation, so those tests will run. But if `msmbuilder` can't be imported, the tests that require it will be skipped.

Once this is merged, I'll do a release.

jhprinz:
Looks good.

PR #712, opened 2017-09-18
Use Trajectory "PseudoAttributes" in TIS

dwhswenson:
This PR will add caching of maximum values of the order parameter for TIS. This should speed up sampling and analysis.

In #691, @jhprinz added a generalization of collective variables that can take objects other than snapshots are inputs. The basic idea is that OPS CVs are wrappers for function that takes a CV as input and gives some output. The CV object is clever about caching results in memory, and, if desired, to disk. This makes analysis a lot faster, because you don't have to reload entire snapshots and recalculate the CVs.

The generalization in #691 makes it so that the same abilities are available for functions of other objects. In this case, we'll use it for functions of a trajectory -- in particular, the maximum value of the order parameter. This gives us a couple advantages in TIS:

* **Sampling speed-up:** in combination with `candidate` trajectory check from #669, this should make repex and path reversal movers O(1) instead of O(L) for trajectory length L. This will be notable for toy model RETIS simulations.
* **Analysis speed-up:** rather than checking every frame when looking for the maximum value of the order parameter, it will already be calculated and cached/saved. This should lead to much faster analysis (rate calculations, etc.)

Tasks:

- [x] Associate a `cv_max` function with each `TISEnsemble` (built with `TISTransition` object)
- [x] Ensure that the `cv_max` is built in MSTIS and MISTIS
- [x] Use `cv_max` in sampling
- [x] Check sampling performance improvements
- [x] Use `cv_max` in analysis
- [x] Check analysis performance improvements
- [x] Update tests for coverage

Note that this includes #686, so it should only be merged after that.


jhprinz:
I am really looking forward to see the speed up... 👍

dwhswenson:
### Sampling performance

The effect on sampling speed is pretty modest, even for the toy model. The reason is that this just isn't the bottleneck. When I dug into the profiling, calling the TIS ensemble from the movers only made up a small fraction of the total cost.

The biggest computational cost right now is saving each step, which takes 68% of the total simulation time. (@jhprinz, any ideas on how to improve this? Nearly all of that time is in saving snapshots. This means that most of the overhead is in storage.)

About 27% of the time is spent doing moves. Path reversals and replica exchange are now very fast, each taking less than 1% of the total simulation time (for ~20-30% of the moves each). Selecting subtrajectories for the minus move takes 7% of the total simulation, which is much longer than it should -- but I see how to fix it, and that will write a PR to do that soon. The moves that do dynamics (shooting moves and extension part of the minus moves) take a total of 14% of the simulation time, although only 2% of the total simulation time is spend in the actual dynamics. The rest of that is the OPS overhead to monitor the can-append/can-prepend methods. I'm not sure what improvements can be made in there; it will take a more careful profiling analysis to figure that out. (Keep in mind that, for a 2D system, we *expect* overhead to be significantly more expensive than dynamics! The overhead cost should be roughly constant with system size; the dynamics cost will grow significantly.)

Again, the best change we could make here would be to speed up storage, specifically of snapshots.

dwhswenson:
### Analysis performance

As a test, I ran the MSTIS example until it had ~100 shooting moves per ensemble (2100 total MC steps). I used the `%%prun` magic to analyze the total crossing probability calculation for one interface set, using the new TIS analysis framework. Note that this doesn't include the time spent loading the `AnalysisStorage`, which can be significant.

Before the analysis improvements in this PR, the generation of the histograms of the maximum value of the order parameter took about 32% of the total analysis time. Now it takes 6%. In both cases, the main cost in the TCP calculation is the WHAM procedure, although that will be independent of the total number of samples in the simulation. Now the second most expensive part is the `steps_to_weighted_trajectories` method, which is re-used by other parts of the analysis.

Overall, for this setup, the total crossing probability takes well under a second, so I don't think we need to worry about further optimizing it for a while. The main costs for the rate calculation are opening the `AnalysisStorage` and calculating the flux from the minus move.

dwhswenson:
This is just waiting on resolution of #721.

dwhswenson:
Remaining problems were fixed by #738. This is ready for review and merge!

jhprinz:
Looks good to merge. I did not completely understand the use of `_reset`, but I assume that is something for testing and not intended to be used in normal production, right?

PR #713, opened 2017-09-19
Fix different opcode width in python 3.6

jhprinz:
Turns out that Python 3.6 uses different bytecode widths and this causes the function that checks stored CVs, etc to fail for python 3.6. I have checked this with python 2.7 and 3.6. And it works now. it also cleans up the code a little and the first toy mstis notebook is fixed. Hope this helps


dwhswenson:
The changes look good; could you add a nosetest to check this? (Shouldn't that be as easy as saving and reloading a CV object to make sure that it works?)

Is the reference to the Python bytecode change is the line "The Python interpreter now uses a 16-bit wordcode instead of bytecode which made a number of opcode optimizations possible" in  https://docs.python.org/3/whatsnew/3.6.html? (Want to have the reference here for either future me, or for anyone else who stumbles across similar problems.)

jhprinz:
Not really sure. I was fiddling around and found some reference here

https://docs.python.org/3/library/dis.html#opcode-HAVE_ARGUMENT

Might try to do a test tomorrow. This is a problem in `dictify.py:ObjectJSON._find_var` and is purely disassembly stuff. And storing is tested in the notebook and there is a CV test that does that. But it requires to import a function inside the CV that is tested...

dwhswenson:
> Not really sure. I was fiddling around and found some reference here

Ah, this is also clear that it is a 3.6 change. Good -- thanks for the reference.

> Might try to do a test tomorrow.

Please do if you can. Right now only nosetests are included in the Python 3 Travis tests (as I slowly migrate notebooks over). Besides, nosetests count for coverage; notebook tests don't. It's helpful to know what might still be missing coverage, especially in these updates for Python 3.

jhprinz:
Added a simple test that will check if CVs with imports inside will be correctly pickled.

dwhswenson:
Thanks! I just tested that this test raised the expected error in a version of the code running on 3.6 without these changes. It does. Changes all look good; merge once this passes tests.

jhprinz:
Done

jhprinz:
@dwhswenson

dwhswenson:
Great! Merging this; it's the last PR we need for the 0.9.3 release. I'll try to draft release notes tomorrow and hopefully that we can do the release in the next couple days!

PR #715, opened 2017-09-24
Docs deployment

### LINKED ISSUE #711: Update (fix deployment) of online documentation
This has 2016 in the bottom. I guess it is not up-to-date.

dwhswenson:
Apparently we've had some deployment issues with our docs for a while. This PR fixes that, and also makes future problems a little easier to troubleshoot.

It used to be that the only way to even build docs (or prep for conda deployment) was if we were both *not* a PR, and on the master branch.

Now we build docs/prep for deployment if we're not a PR, regardless of branch. This still means that in the most common case (PRs), Travis won't take the time to build docs or prep deployment. However, if we want to debug the deployment procedure, we can now create a branch under the `openpathsampling` fork, and that should do all the work up to, but not including, deployment. Ideally, this will be useful for future debugging.

Fixes #711.

### DIFF
--- a/devtools/ci/after_sucess.sh
+++ b/devtools/ci/after_sucess.sh
@@ -1,4 +1,7 @@
 #!/usr/bin/env bash
+
+DEPLOY_PY="2.7"
+
 if [[ "$TRAVIS_PULL_REQUEST" != "false" ]]; then
     echo "This is a pull request. No deployment will be done."; exit 0
 fi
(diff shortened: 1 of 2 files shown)


dwhswenson:
This is ready for review and merge. Note that we have 2 different Travis runs here, since this is a branch of openpathsampling's fork, not my own. The "PR" version does not build docs or prep deployment, because it is a PR. The "push" version does build docs and prep deployment, but does not do any deployment, because the branch is not "master". It will only actually deploy when we merge to master (but I think it will actually deploy now!)

jhprinz:
Looks good.

dwhswenson:
It looks like this did not completely fix the problem. We're missing an `AWS_ACCESS_KEY_ID`, which suggests to me that there's something wrong with our setup of the `secure` section of our `.travis.yml`. See the end of https://travis-ci.org/openpathsampling/openpathsampling/jobs/279281878

This might be something where we need help from @jchodera -- at least, I don't have our AWS info, so I can't verify whether the encrypted strings are correct (or even what they represent).

Also, there are a lot of warnings that didn't show up with my version of sphinx (1.4), so I'll upgrade and see if I can fix some of those as well.

dwhswenson:
Rest of the fix is in #717. Looks like we don't need anything from @jchodera, because @jhprinz found the problem and fixed it.

PR #716, opened 2017-09-24
Documentation cleanup

dwhswenson:
We were getting a lot of warnings when building the docs. This fixes most of them (leaving only a couple because the developer documentation is still in progress and missing some files that need to be written).

Most of the warnings had to do with files that had been renamed or relocated, so it was straightforward to fix.

This changes no code -- only docs -- and should be pretty straightforward to review and merge. It's ready as soon as it passes tests.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -46,14 +46,14 @@ Ensemble
 Volume
 ******
 
-.. automodule:: volume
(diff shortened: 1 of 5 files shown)


jhprinz:
Looks good. WIll merge.

PR #717, opened 2017-09-24
Fix encryption keys

jhprinz:
Not sure, but I think, the problem is a misconfigured .travis.yml file not missing keys. Turns out that the keys are not set in the current script. This might fix it. Hard to test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,12 +35,11 @@ addons:
     - pandoc
 
 env:
+  global:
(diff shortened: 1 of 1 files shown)


jhprinz:
Try to merge it and see what happens...

dwhswenson:
merging, with fingers crossed that this will fix it (no need to worry about tests; this only affects structure of the `.travis.yml`, and we won't know if it works until it is merged to master anyway)

dwhswenson:
Excellent! This fixed it. Still have lots of warnings to fix in the docs build, but at least the docs are getting uploaded! (I really needed that, since I want the report I need to finish this week to point to the new docs on TIS analysis -- thanks for figuring out the problem and quickly fixing it!)

jhprinz:
No way, really? Wow, that was a long shot... Great...

dwhswenson:
I needed this to exist: http://openpathsampling.org/latest/topics/tis_analysis.html

Now it does! :-)

PR #719, opened 2017-09-29
Updates to AD MSTIS example

bolhuis:
This is the new correct AD PDB file.


dwhswenson:
Some of the tests in `examples/tests/test_cv.ipynb` depend on the AD initial frame, which was replaced by a correctly equilibrated version in this PR. @bolhuis, if you re-run that notebook and commit the results, I think that will fix the test failures. (Is that right, @jhprinz? It checks against what was in the notebook, so re-running should put the correct results in the notebook?)

dwhswenson:
Everything passes now; merging with the corrected version of the initial frame for the AD examples and the updated setup for the AD MSTIS example.

PR #725, opened 2017-10-27
Fix Shooting Point Analysis

### LINKED ISSUE #722: ShootingPointAnalysis items() and values() not working/result in error
The issue here is that the keys are strings but that \_\_getitem\_\_ expects them to be snapshots and tries to hash them. Since they are already hashed this results in an error, because naturally the strings do not have a coordinate attribute.
I have a fix for this behavior on  https://github.com/hejung/openpathsampling/tree/fix_SPA
I have added a conditional in \_\_getitem\_\_ checking if it is a string (->use key to take the snapshot from hash_representatives and use that as a key) or not (-> take the key as key).
EDIT: this breaks as soon as the keys are not strings anymore. The only general fix I can think of is (re)implementing .items() and .values() instead of inheriting them from Mappable but I am not sure if that is a good idea.

hejung:
closes #722
I reimplemented the TransformedDict.__iter__() method and updated the corresponding test.
Now self.values(), self.keys() and self.items() work consistently and return counter, snapshots and tuples of them respectively.

### DIFF
--- a/openpathsampling/analysis/shooting_point_analysis.py
+++ b/openpathsampling/analysis/shooting_point_analysis.py
@@ -40,7 +40,7 @@ class TransformedDict(abc.MutableMapping):
         del self.hash_representatives[hashed]
 
     def __iter__(self):
-        return iter(self.store)
+        return iter(self.hash_representatives.values())
 
     def __len__(self):
         return len(self.store)
@@ -58,8 +58,8 @@ class TransformedDict(abc.MutableMapping):
         velocities, the resulting mapping would be invalid. It is up to the
         user to avoid such invalid remappings.
         """
-        return TransformedDict(new_hash, 
-                               {self.hash_representatives[k]: self.store[k] 
+        return TransformedDict(new_hash,
+                               {self.hash_representatives[k]: self.store[k]
                                 for k in self.store})
 
 
@@ -71,7 +71,7 @@ class SnapshotByCoordinateDict(TransformedDict):
     """
     def __init__(self, *args, **kwargs):
         hash_fcn = lambda x : x.coordinates.tostring()
-        super(SnapshotByCoordinateDict, self).__init__(hash_fcn, 
+        super(SnapshotByCoordinateDict, self).__init__(hash_fcn,
                                                        *args, **kwargs)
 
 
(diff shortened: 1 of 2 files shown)


dwhswenson:
Quick side note @hejung (not specific to the code; review will come later): It looks like you don't have matching email addresses in you git config and GitHub account. This is either because your git config doesn't have an email set, or because the email you set isn't linked to GitHub. See more: 

* https://help.github.com/articles/setting-your-commit-email-address-in-git/
* https://help.github.com/articles/setting-your-commit-email-address-on-github/

Setting this helps us track contributions, and ensures that you get credit for your contributions!

dwhswenson:
@jhprinz : Our policy was to hold off on user contributions until the paper is accepted. This will be one of the first PRs to be merged once we accept external contributions!

dwhswenson:

hejung:

Now rebased on the current master.

Just as a reminder what changed:
`ShootingPointAnalysis.keys()`, `ShootingPointAnalysis.items()` and iterating over the `ShootingPointAnalysis` now hide that we in fact that we use the string representation of the snapshots coordinates instead of the snapshot as key.
I.e. we now return the snapshots associated with the hashstring whenever we are asked for a key, the strings are only accessible via the keys of the `self.hash_representative` or `self.store` dictionaries.

dwhswenson:
This technically changes tested behavior, but I'm of the opinion that the tested behavior was incorrect from a user perspective, hence a bug. (This is the difference between API breaking and not.)

Just to check that no one is relying on the old behavior (that keys in shooting point analysis results are internally string reps or the coords, not the snapshots), mentioning @sroet @arjunwadhawan -- if either of you have a good reason to be concerned about this PR, let me know. If not, I'll doubt that it has become important to anyone's workflow. I leave this open for comment for 48 hours, merging no earlier than Wednesday, 26 Feb.

This also opens the way to something else I was thinking about: since this hides the actual keys from users, they can be different for different engines. The UUID of the `statics` object in snapshots from OpenMM can be used, instead of the coordinates. I think this might make the analysis faster (might be able to only load UUIDs, not the whole snapshot).

sroet:
@dwhswenson I don't think this should raise any issues on my codes, and I agree that the current tested behavior is incorrect. So if it raises issues on my code I should be able to alter it pretty quickly.

dwhswenson:
No complaints, so merging it!

PR #726, opened 2017-11-08
Fix bug in DirectSimulation flux

dwhswenson:
Fixes a bug that @bolhuis and I noticed in the `DirectSimulation` flux when there are recrossings of the interface before re-entering the state. Includes a test for this case.


dwhswenson:
This passes tests and is ready for review/merge.

PR #728, opened 2017-11-14
Dev install script

dwhswenson:
Several projects I'm working on are based on OPS, and I like to test them against the most recent development branch of OPS. In addition, users might want to test their own OPS-based packages against modifications in their own forks of OPS.

In order to make that easier, I've made this little script that does a developer install of OPS (assuming you have `conda` installed). The premise is that, if something (requirements?) changes between conda releases, this script will be updated to make sure that someone has a working version of the most recent version of OPS. Since the repo is downloaded, it also gives them the option to (after the script) add remotes and switch to a branch in a remote.

The idea is that they could `curl` down this script and use it to install during their CI, and we'd try to keep it up to date. I'm already using a version of this script in my [`ops_additional_examples`](https://gitlab.e-cam2020.eu/dwhswenson/ops_additional_examples/) repo, where I recently added CI using `ipynbtest`.

This involves no changes to the OPS code, so review should be trivial. I haven't written a test for it, but that would be pretty easy (add a new element to the matrix installing using this; run nosetests). But it seems wasteful to have an extra Travis matrix test for this on every push; maybe it's better to count on downstream repos to notify when something when wrong? (since I know I'll get notifications from several downstream repos!)

### DIFF
--- a/devtools/conda_ops_dev_install.sh
+++ b/devtools/conda_ops_dev_install.sh
@@ -1,0 +1,13 @@
+#!/bin/sh
+
+# This assumes you've already installed conda. Then it installs OPS as a
+# developer install.
+
+conda install -y -c omnia openpathsampling
+conda update -y --all
+conda uninstall -y openpathsampling
+git clone https://github.com/openpathsampling/openpathsampling.git
+cd openpathsampling && pip install -e . && cd ..
+
+# this puts to the OPS dev install in a directory so you can change to that
+# directory and add remotes/switch branches as necessary


dwhswenson:
Passes tests. Ready for review/merge.

jhprinz:
Very nice. Will merge this.

jchodera:
Would it be easier to just push a `dev` label build each night, like OpenMM does?
You could just add the recipe here: https://github.com/omnia-md/conda-dev-recipes

OpenMM can then be installed with the `-c omnia/label/dev` argument.

dwhswenson:
@jchodera : That only serves some of the use cases I'm thinking of. For example, it won't work for someone who needs to test against a branch in someone else's fork, which has come up a few times (in ESDWs and elsewhere). This gives a starting point that would work for that case, by doing an editable install.

jhprinz:
As a sidenote. We had some discussion about a kind of dev install option exactly for this purpose on conde, but I am not sure if they already implemented it. Some options that allow you to only install dependencies.

jchodera:
Wouldn't an [`environment.yml`](https://conda.io/docs/user-guide/tasks/manage-environments.html#creating-an-environment-from-an-environment-yml-file) or a conda recipe to just install build/run deps work just as well? Why build new custom infrastructure for a solved problem?

dwhswenson:
An `environment.yml` would also work, although it would have to be maintained as we updated OPS requirements. This would add an extra place where we need to add pins when changes to upstream make that necessary (I already find it annoying that we should be pinning in both the conda recipe and the `requirements.txt`, which we don't actually use or test).

And I'm not sure that a 5-line shell script really deserves the description "a new custom infrastructure"! This was just intended to be a way to avoid writing the same 5-line shell script in every repo I have that uses OPS, and to give people like ESDW participants a quick way to set up an editable install. It's also a central script so that if something goes wrong with this process, I can fix the script once and it will fix build errors in all my projects at once. This had been a problem for me a while ago when the `openpathsampling-dev` recipe stopped working.

jchodera:
BTW, don't you want to install the *current* github development version's conda dependencies, rather than the release version's conda dependencies (which may lag behind the development version's)?

Seems like you might want to check out the repo, build and install `devtools/conda-recipe/`, uninstall it, then pip install.

dwhswenson:
> Seems like you might want to check out the repo, build and install `devtools/conda-recipe/`, uninstall it, then pip install.

You're right. That's a better way to do it. I'll try to update the script sometime soon-ish (lots of other things are keeping me busy right now).

PR #731, opened 2017-11-15
snapshot.md box_vectors

### LINKED ISSUE #710: .md attribute does not pass unitcell information
I expected the following code to do the same thing, but `.md` does not parse the unitcell information. Is this the intended behaviour?
```python
import openpathsampling.engines.openmm as peng_omm

# Use the examples pdb
ops_dir = paths.__file__.rsplit('/',2)[0]
snap = peng_omm.snapshot_from_pdb(ops_dir+"/examples/resources/AD_initial_frame.pdb")
traj = paths.Trajectory([snap])

print traj[0].md
print traj.to_mdtraj()[0]
```
returns
```
<mdtraj.Trajectory with 1 frames, 1651 atoms, 546 residues, without unitcells>
<mdtraj.Trajectory with 1 frames, 1651 atoms, 546 residues, and unitcells>
```

dwhswenson:
Fixes #710, plus a test to prove it.

In addition, I changed things so that engine developers who want to support conversion to MDTraj just need to add a `mdtraj_topology` attribute/property to their engines (and added relevant docs).

Also made it so that `snapshot_from_testsystem` allows you to select a nonperiodic snapshot. This was useful for testing.

### DIFF
--- a/docs/developers/advanced_engines.rst
+++ b/docs/developers/advanced_engines.rst
@@ -3,7 +3,6 @@
 Advanced Engines
 ================
 
-
 Custom snapshot features
 ------------------------
 
@@ -12,5 +11,55 @@ dynamics, even things that have nothing to do with molecular dynamics. If
 that's your interest (or if you're interested in methods like *ab initio* or
 quantum dynamics which might make use of a definition of state beyond
 standard MD), then this is the section for you.
+
+Standard snapshot features
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Adding features with new stored data
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Adding features that are proxies to stored data
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Adding features that are properties
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Support for additional OPS modules
+----------------------------------
+
+The basic engine setup is easy to do, and enables the core of
+OpenPathSampling's functionality. However, some additional functionality can
+be added with very little effort, for engines where this might be relevant. 
+
+Shooting Point Velocity Modification
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The original descriptions of TPS used two-way shooting with a ":math:`\Delta
+p`" velocity modification. That is, the direction of the velocity on one
+particle was changed.
+
+OPS has the ability to do this using the
+:class:`.VelocityDirectionModifier`, but this procedure becomes extremely
+complicated in condensed phase (periodic) systems with holonomic
+constraints. Therefore, OPS only allows this for engines where the snapshots
+have the ``n_degrees_of_freedom`` feature. In addition, the engine can tell
+OPS not to bother removing linear momentum (e.g., if the engine represents a
+particle rolling on a surface, like our toy engine) by setting
+``engine.ignore_linear_momentum = True``.
+
+MDTraj support
+~~~~~~~~~~~~~~
+
+Several aspects of `MDTraj <http://mdtraj.org>`_ support require an
+:class:`mdtraj.Topology` object. The developer can add these by adding a
+property/attribute called ``mdtraj_topology`` to their engine. With this,
+you automatically get:
+
+* trajectory conversion to :class:`mdtraj.Trajectory`, which enables output
+  to various file formats, and integration with other projects such as
+  nglview for trajectory visualization.
+* ??? TODO: can we make MDTraj CV support automated as well? Check for topol
+  on first run? ???
+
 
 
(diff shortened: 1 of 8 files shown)


dwhswenson:

openpathsampling/engines/trajectory.py
```diff
@@ -534,21 +534,27 @@ def to_mdtraj(self, topology=None):
...
         :class:`mdtraj.Trajectory`
             the trajectory
         """
-
-
         if topology is None:
-            topology = self.topology.mdtraj
+            # TODO: maybe add better error output?
+            # if AttributeError here, engine doesAn't support mdtraj
+            topology = self[0].engine.mdtraj_topology
```
jhprinz: I think this fails for Trajectories with no frames. If I remember correctly, we had a similar feature before and removed it. But I do not remember why...

Just check if length is zero and return an empty mdtraj.Trajectory should work.

docs/developers/advanced_engines.rst
```diff
@@ -13,4 +12,54 @@ that's your interest (or if you're interested in methods like *ab initio* or
...
+
+Several aspects of `MDTraj <http://mdtraj.org>`_ support require an
+:class:`mdtraj.Topology` object. The developer can add these by adding a
+property/attribute called ``mdtraj_topology`` to their engine. With this,
+you automatically get:
+
+* trajectory conversion to :class:`mdtraj.Trajectory`, which enables output
+  to various file formats, and integration with other projects such as
+  nglview for trajectory visualization.
+* ??? TODO: can we make MDTraj CV support automated as well? Check for topol
```
jhprinz: I need to think about this. But it definitely might break immutability, since the actual CV depends on how you first call it. It would basically mean, that you define the CV independent of the topology. That might work. So the CV always grabs the topology of trajectory when it is used the first time.
We could really do that. But it might require a different CV class to not conflict with existing ones.

jhprinz:
I had one potential error for ´Trajectory([]).to_mdtraj()´ might fail. Otherwise go ahead.

openpathsampling/engines/trajectory.py
dwhswenson: Interesting point. But I don't think that MDTraj allows empty trajectories (follow https://github.com/mdtraj/mdtraj/issues/1319 for more). For the time being, I added stuff so that we explicitly forbid conversion of a length-zero trajectory to MDTraj (raise a `ValueError` with a clear message), and a test to check it.

docs/developers/advanced_engines.rst
dwhswenson: > It would basically mean, that you define the CV independent of the topology.

Actually, I think this is okay. Compare with `FunctionCV`. If I say "give me the x-coordinate of atom 5", this CV will give different (semantic) results depending on what atom 5 is (i.e., change in topology). Same if I say "use MDTraj to give me the dihedral of atoms 11, 12, 13, 14" -- result might be nonsense for some other topology, but that's user error.

> But it might require a different CV class to not conflict with existing ones.

I was thinking that we could do it by keeping the existing `topology` argument, but making it optional. If `self.topology is None`, we set it when we first run. That maintains API compatibility.

Not sure the storage details here; we may need a special case when `topology` is `None`.

jhprinz: > I was thinking that we could do it by keeping the existing topology argument, but making it optional. If self.topology is None, we set it when we first run. That maintains API compatibility.

Yes, that should work. So we change the documentation and agree that if you create a CV without topology this means that it works with the implicit defined topology. I think just using the first one and keep it might be problematic, but with a little simple logic that should work, too. We could cache it if the same engine is used.

dwhswenson: > We could cache it if the same engine is used.

Yep, I think that's the best approach. Something like this:

```python
if not self._cached_topology:
   self._cached_engine = snapshot.engine
   self._cached_topology = snapshot.mdtraj_topology

if snapshot.engine is self._cached_engine:
   topology = self._cached_topology
else:
   topology = snapshot.mdtraj_topology
```

dwhswenson:
Tests pass... haven't heard anything on whether zero-length trajectories are possible in MDTraj. For now, it raises an error. Merge as is, or wait for word from MDTraj team?

dwhswenson:
> I had one potential error for `Trajectory([]).to_mdtraj()` might fail. Otherwise go ahead.

From https://github.com/mdtraj/mdtraj/issues/1319, MDTraj *can* have zero-length trajectories. However, it would be impossible for us to correctly create a zero-length MDTraj trajectory from a zero-length OPS trajectory. The reason is that engine, number of atoms, etc, is all stored in *snapshots*, not trajectories. The trajectory gets that information by checking its first snapshot.

That means that, without a first snapshot, we can't create the correct MDTraj topology (or even the correct MDTraj vector size.) So I propose leaving it as it stands, where we disallow conversion of zero-length trajectories to MDTraj.

We've talked in the past about my ideas for restructuring `Trajectory` in 2.0 -- making it so that the `engine` is directly associated with the `Trajectory` is part of that, and will fix this. But that's more work than I can do before 1.0, plus breaks file format.

I updated the docstring note about this, but I think the current implementation is the way to go. @jhprinz, further thoughts?

PR #732, opened 2017-11-15
Fix problem with HistogramPlotter2D limits

### LINKED ISSUE #730: Bug in HistogramPlotter2D limits
When trying to run the following code in jupyter, changing the x-limits doesn't behave as expected. 
```python
%matplotlib inline
import pickle
from openpathsampling.numerics import HistogramPlotter2D
import matplotlib.pyplot as plt

with open('test_pdh.p', 'r') as f:   
    pdh = pickle.load(f)

HistogramPlotter2D(pdh).plot(xlim=(0.0, 2.5),
                             ylim=(0.0, 2.5),
                             vmin=0, vmax=1,
                             cmap='Blues')
plt.show()
plt.close()

HistogramPlotter2D(pdh).plot(xlim=(0.5, 2.5),
                             ylim=(0.0, 2.5),
                             vmin=0, vmax=1,
                             cmap='Blues')
plt.show()
plt.close()
```
returns
working
not_working

where I expected the second one to be scaled from 0.5 to 2.5 but instead runs from 1 to 3?

[test_pdh.zip](https://github.com/openpathsampling/openpathsampling/files/1471527/test_pdh.zip) contains the pickled path density histogram. (without the cv objects as cvs, as those were not happy being pickled, so I changed them to strings). The pdh was build using ```left_bin_edges=[0,0], bin_widths=[0.025, 0.025]```

dwhswenson:
Resolves #730.

Had an incorrect test that had to be corrected -- misunderstood how things worked!


dwhswenson:
This fixes the problem @sroet was having, and also still works great on the examples in the alanine dipeptide notebook.

Passes tests, ready for review and merge!

PR #733, opened 2017-11-15
Update path tree colors

### LINKED ISSUE #727: Change path tree visualization default colors?
This follows on #682 -- As I'm preparing (what I hope will be) the final versions of figures for the papers, I think we should consider changing the path tree visualization to be more accessible: i.e., to be something other than literally the worst possible color scheme for the most common kind of colorblindness.

@bolhuis : Is there any reason we should not switch from red/green to red/blue? You're the one who started the historical red/green color scheme; do you think it would cause problems with people's understanding if we switched the default green to blue?

@jhprinz : I'm think from previous discussion that you agreed with this change, right? It would probably be good to also change whatever is currently blue to something else (I think that was repex, which has a different default line type -- maybe use green for that, and let the shape be the distinguishing feature?)

At some point, we might move to a better overall default color palette (like how matplotlib switched to the d3 colors). That can be saved for future changes, though.

dwhswenson:
Resolves #727.

Switch green <-> blue so that standard paths aren't green/red.

Longer term, we should make a better color palette.

- [x] Update CSS
- [x] Update example notebooks


dwhswenson:
Ready for review. Includes new results for the repex-shoot-repex example to be used in paper 2.

dwhswenson:
Made fixes for the two things from @jhprinz's review. Passes tests, so merging!

PR #735, opened 2017-11-19
Flux to pandas

dwhswenson:
This adds a new method, `paths.analysis.tis.flux_matrix_pd`, which converts our flux results (a dict with 2-tuple of (state, interface) as keys and with the flux as the values) into a `pandas.Series` (with `MultiIndex` for the 2-tuple).

Previously, we have been putting the flux into NxN `DataFrame`s in the analysis examples (nonsensical, but done for convenience based on other inputs when I wasn't very familiar with `pandas`). The other way of viewing it was as the raw `dict`, which is useful internally, but not very human-readable.

This now makes a decent human-readable `pandas.Series`. It also sorts the data from the input dict so that all fluxes with a given state are together. You can specify a sorting function for the indices; the default is alphabetical by state name, then alphabetical by interface name.

More features may be added to this later (select index labels in a way other than `.name`?) but for now, this works well enough, and has full testing.

Note that this `dict`-based form of the flux is used in several places: TIS, but also `TransitionTrajectoryAnalysis` and `DirectSimulation`. The function here can be used for all of them.


dwhswenson:
passes tests, ready for review

PR #740, opened 2017-12-09
Remove explicit name parameter from PseudoAttributes

jhprinz:
After short discussion with @dwhswenson this change makes sense. CVs are nameable objects like other storable objects in OPS. The explicit name is not required and the API should reflect that. It is confusing to say that you need to name it although that is not necessary.

With all the changes in storage, etc the name for a CV is not required anymore. If set, it still needs to be unique otherwise loading by name does not work or is not useful.

This PR removes the `name` from `CVSomething(name, ...)` and leaves it with `CVSomething(...)`.


jhprinz:
All nosetests passed, so the general code and tests word, but all the notebooks would need to be adjusted. I currently have little time to do that. It is wasy though tedious. Feel free to do that. The general replacement is

```
SomeCV("name", **attributes) -> SomeCV(**attributes).named("name")
```

like for all the other objects. Of course unnamed CVs are valid as well!

dwhswenson:
> all the notebooks would need to be adjusted. I currently have little time to do that. It is wasy though tedious

Let's hold off on this for a while.

I agree with the idea, but the timing now is a problem. This is API-breaking, and I have a TON of stuff in the works that this will break (like all the Python 3 notebook updates I'm working on, and updates to the examples for the paper).

Is there any way to make it so existing files can still load with this? We've stated that the file format shouldn't change, and we have enough users that I don't want to screw them over (a number of users beyond our group).

I don't mind telling users they have to change their scripts; I do mind if they can't open the files they've generated using months of computer time.

jhprinz:
Fixed the code and the notebooks. Please have a look.

dwhswenson:
Marking this for closure; it breaks the API, and therefore can't go in until 2.0, at which point netcdfplus will be replaced by SimStore. This particular behavior is already implemented in SimStore.

I will leave this open for comment for at least 7 days, closing no earlier than Wed 30 Dec 06:00 GMT (07:00 local).

PR #742, opened 2017-12-17
Fix problem in conda-build

dwhswenson:
There seems to be a problem with the stage that installs OPS into a test environment that is causing build failures. I'll see if I can fix it; if not, a version pin on conda or conda-build might do it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ install:
   - export PYTHONUNBUFFERED=true
   # Sets up a miniconda environment
   - source devtools/ci/install.sh
-  = conda install --yes conda-build
+  - conda install --yes conda-build
   - conda build --quiet devtools/conda-recipe
   - conda create --quiet --yes --name test --use-local openpathsampling-dev
   - source activate test
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,15 +18,7 @@ install:
 script:
   - python --version
   - source devtools/ci/git_hash.sh
-  #- echo "Building OPS (with conda)" && echo -en 'travis_fold:start:build\\r'
-  #- echo -en 'travis_fold:end:build\\r'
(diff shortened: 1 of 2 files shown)


dwhswenson:
Weird. This seems to be working fine in PRs, but I can't get the nightlies to pass (no matter how many times I re-run). We're not the only one seeing this error (https://github.com/conda/conda/issues/6468).

There were version bumps in conda-build, but none in conda since the error started (and the error reported above doesn't appear to involve conda-build.

There was an update in the Travis images a few days before this started happening. Still, don't understand why the PRs are passing but nightlies aren't.

dwhswenson:
The suggestion in https://github.com/conda/conda/issues/6468 is to use the `--quiet` flag for conda commands (apparently some `conda` tools are now too verbose for Travis? I think this is a Travis change.)

This PR does that. I also moved all the installation process into the `install` section of the Travis script, so the `script` section just lists the installed versions and runs the tests. Aside from being the logical setup, this has the extra advantage that steps in `install` automatically have folds in Travis, so it's easier to read Travis longs.

This is ready for review/merge. If it doesn't get reviewed in the next couple days, I'm going to go ahead and merge it, since our builds fail until then.

jchodera:
We've been struggling with this as well, and your work around seems to be the best solution for now.

dwhswenson:
Merging this in the hopes that it will fix the problem that all our builds show up as failing, even though the tests pass.

The weirdest part of this is that our PR builds are fine, but the master builds fail. I don't see how the `--quiet` flag fixes that. Browising around issues on this, I saw something about a `filter-secure` flag (or something like that) in Travis -- if this doesn't fix, I'll look into that. It seems more credible, since the difference between PR builds and master builds is access to the secure environment variables.

PR #741, opened 2017-12-10
Python 3 support in (tested) toy MSTIS notebooks

dwhswenson:
This is will add some of the notebook tests to our Python 3 Travis runs (which is the next step in #714). Specifically, this is the toy MSTIS example (only worrying about the tested notebooks for this PR).

- [x] `toy_mstis_1_setup.ipynb` runs in py-3.6 local
- [x] `toy_mstis_2_run.ipynb` runs in py-3.6 local
- [x] `toy_mstis_3_analysis.ipynb` runs in py-3.6 local
- [x] `toy_mstis_4_repex_analysis.ipynb` runs in py-3.6 local
- [x] Remove `test_storageview.ipynb` (https://github.com/openpathsampling/openpathsampling/issues/714#issuecomment-350475987)
- [x] Add new already-run .nc file for py-3.6
- [x] Run add toy MSTIS notebook tests to Travis for py-3.6
- [x] Get these notebook tests running in py-3.5

Issue #714, opened 2017-09-20
Python 3 compatibility


dwhswenson:
Converting #689 to an issue, so we can close that PR and instead do it in several smaller PRs. This covers the process of updating OPS to work with Python 3. See also #70.

Big picture tasks:

- [x] Update Travis script to test 2.7, 3.5, 3.6
- [x] Fix nosetests to work with 2.7, 3.5, 3.6
- [x] Fix notebook tests to work with 2.7, 3.5, 3.6
- [ ] Check that all untested notebooks work in 3.6 as well
- [ ] Clean up automated additions in testing (zip, range, old_div; some changes not needed)

---

Notebooks, divided by directory. An asterisk after the notebook name indicates that it is part of the standard test suite.

**`examples/tests/`**
- [x] `langevin_integrator_check.ipynb`
- [ ] `storage_mem_test.ipynb`
- [x] `test_cv.ipynb`*
- [x] `test_netcdfplus.ipynb`*
- [x] `test_openmm_integration.ipynb`*
- [x] `test_pyemma.ipynb`*
- [x] `test_snapshot.ipynb`*
- [x] `test_snapshot_modifier.ipynb`
- [x] ~~~`test_storageview.ipynb`~~~

**`examples/toy_model_mstis/`**

- [x] `toy_mstis_1_setup.ipynb`*
- [x] `toy_mstis_2_run.ipynb`*
- [x] `toy_mstis_3_analysis.ipynb`*
- [x] `toy_mstis_4_repex_analysis.ipynb`*
- [ ] `toy_mstis_5_srtis.ipynb`
- [ ] `toy_mstis_A1_split.ipynb`
- [ ] `toy_mstis_A2_split_analysis.ipynb`

**`examples/toy_model_mistis/`**
- [x] `toy_mistis_1_setup_run.ipynb`*
- [ ] `toy_mistis_2_flux.ipynb`
- [x] `toy_mistis_3_analysis.ipynb`*

**`examples/misc/`**
- [ ] `committors.ipynb`
- [ ] `custom_strategy_repex_shoot_repex.ipynb`
- [ ] `fun_with_pathmovers.ipynb`
- [ ] `move_strategies_and_schemes.ipynb`
- [ ] `tutorial_handle_nan.ipynb`
- [ ] `tutorial_movetree_visualization.ipynb`
- [x] `tutorial_path_histogram.ipynb`
- [ ] `tutorial_pathtree_visualization.ipynb`
[...]

PR #743, opened 2017-12-17
Toy MISTIS Example: Updates and Python 3 notebook tests

dwhswenson:
Update example analysis; make tests work for it in Python 3.5, 3.6. This will complete the next step of #714.

- [x] Update MISTIS example to work in Python 3.5, 3.6
- [x] Add precalculated mistis.nc file for Python 3
- [x] Update `tutorial_storage.ipynb` for Python 3
- [x] Use single `ipynbtest.sh` test script for all Python versions
- [x] Complete testing for IPython notebooks in Python 2.7, 3.5, 3.6.

Requires #741. Merge after that.


dwhswenson:
A few of the updates for the paper need me to make more significant changes to the code (we're going to try it with two-way shooting, which should be more efficient for this system.) Because of that, I'll put final results into a separate PR.

This one focuses on getting all the Py3 test notebooks running, so we can do full testing in 2.7, 3.5, and 3.6. It is ready for review and merge. Note that it includes the changes in #741.

PR #744, opened 2017-12-18
More docs cleaning

dwhswenson:
Several more little things to improve the documentation:

* Instead of having an copy of the sphinx RTD theme inside OPS, use the conda/pip-installable version (let them maintain it so we don't have to!)
* Instead of the (uninformative) "What's new" section we had before, I add a comment on the main (index) page pointing to the GitHub releases. All release note/changelog type things will be tracked in the notes to each GH release; I don't see a reason to replicate that information elsewhere
* Added the logo to the website. This also necessitated changing some of the color scheme to better match the logo. More improvements to color scheme are welcome!
* A little more documentation on adding engines


dwhswenson:
this is/has been ready for review

PR #745, opened 2017-12-19
Fix warnings that show up in nosetests

dwhswenson:
There are a few warnings that show up in our tests. This PR fixes or silences them.

* `dataframe.set_value` [is deprecated as of pandas 0.21.0](https://pandas.pydata.org/pandas-docs/stable/whatsnew.html#whatsnew-0210-deprecations), and this is raising lots of warnings in tests. This PR will replace usage of `.set_value` with the `.at` or `.iat` approaches, as recommended.
* For a long time, numpy has been throwing warnings for divide by zero for some of the WHAM. We actually *want* to generate NaNs there and then ignore them after (the speed of numpy and increased simiplicity of the code make it easier to create the NaNs then filter them). Used to be, numpy didn't warn on this; now it does. In this PR, we silence the numpy warnings for the lines of code that are expected to cause them.

- [x] Fix pandas `set_value` warning
- [x] Silence numpy division warnings
- [x] Double-check that silencing numpy warnings is safe

### DIFF
--- a/openpathsampling/analysis/channel_analysis.py
+++ b/openpathsampling/analysis/channel_analysis.py
@@ -422,8 +422,7 @@ class ChannelAnalysis(StorableNamedObject):
         switch_count = collections.Counter(switches)
         df = pd.DataFrame(index=sorted_labels, columns=sorted_labels)
         for switch in switch_count:
-            df.set_value(index=switch[0], col=switch[1],
-                         value=switch_count[switch])
+            df.at[switch[0], switch[1]] = switch_count[switch]
 
         df = df.fillna(0)
         return df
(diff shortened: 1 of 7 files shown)


dwhswenson:
This is ready for review.

dwhswenson:
> Although allowing NaNs seems hacky

Eh, the idea is that the NaN only occurs when you have `0.0/0.0`, which we would otherwise ignore anyway. So either write the code to check everything and ignore before doing the division, or ignore invalid results after doing the division. The extra division is done by numpy, and so it is a faster to just let numpy do it than to add Python if-else checks.

PR #746, opened 2017-12-22
Fix "length 5000" error

### LINKED ISSUE #632: Fix for the 5000 max length problem
We see often a 5000 step max out error in the example. Way more often than we would expect. Let's keep an open eye on this one.

dwhswenson:
I think this might fix the stochastic failure we had with the "length-5000" error. Really, we had all the machinery in place, but the extra lines (defining forbidden states in the full bootstrapping initial trajectory) got either removed from or never added to the MSTIS example. They exist in the MISTIS example.

I ran this locally 20 times without errors.

There were also some other, less frequent, stochastic errors; I'm not sure whether this will also fix those. But I think it may have fixed our most common stochastic test failure.

Note: To avoid conflicts, this builds off of the PRs that added Python 3 support to notebooks. Therefore, this should only be merged after #743 and #741.

Resolves #632.


dwhswenson:
Ready for review!

PR #748, opened 2018-01-01
fix ipynb CSS style in RTD Sphinx theme

jhprinz:
This just fixes the use of some CSS. It has always been there but was not used. This add the correct file and uses it.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -14,7 +14,7 @@ endif
 # Internal variables.
 PAPEROPT_a4     = -D latex_paper_size=a4
 PAPEROPT_letter = -D latex_paper_size=letter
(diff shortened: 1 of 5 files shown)


dwhswenson:
All of this looks good to me. I'd been holding off on reviewing this until I figured out how to fix the problem that I fixed in #749. Since that might also be a problem here (problems correctly uploading CSS files), I recommend merging this after #749. Otherwise, looks ready to merge.

Side note: We might consider switching to [nbsphinx](nbsphinx.readthedocs.io) for adding notebooks to sphinx docs. I've been using this in [my contact maps](http://contact-map.readthedocs.io/) package, and it works reasonably well. It has support for making PDFs, which is nice, and we don't have to maintain it. I'll make a separate issue with the pros and cons of that sometime soon.

dwhswenson:
The build after merging #749 passed (https://travis-ci.org/openpathsampling/openpathsampling/builds/336904078), so merging this!

PR #749, opened 2018-01-10
Docs deployment updates

dwhswenson:
We were having some problems with the OPS website. For some reason, the S3 upload was incorrectly identifying MIME types, meaning that even though our CSS was being uploaded, it wasn't actually being used as CSS on the website. (This was not an easy thing to debug!)

In this PR, I mainly changed some of the structure of how the `push-docs-to-s3.py` script works. It could do with more cleanup, but this is all to enable quick fixes if something goes wrong with the S3 upload process in the future. The website is already fixed, because I could force the upload from the `docs_deploy` branch, where I did this work.

### DIFF
--- a/devtools/ci/after_sucess.sh
+++ b/devtools/ci/after_sucess.sh
@@ -45,6 +45,8 @@ conda list -e
 
 (cd docs && python premake.py && make html && cd -)
 ls -lt docs/_build
+ls -lt docs/_build/html
+ls -lt docs/_build/html/static/
 pwd
 echo travis_fold:end:build.docs
 
--- a/devtools/ci/after_sucess.sh
+++ b/devtools/ci/after_sucess.sh
@@ -54,7 +54,10 @@ echo travis_fold:start:upload.docs
 if [[ "$TRAVIS_BRANCH" == "master" ]]; then
     python devtools/ci/push-docs-to-s3.py
 elif [[ "$TRAVIS_BRANCH" == "docs_deploy" ]]; then
-    python devtools/ci/push-docs-to-s3.py --clobber
+    # change the behavior for the docs testing branch (docs_deploy branch in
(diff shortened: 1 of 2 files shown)


dwhswenson:
oops... forgot to ask @jhprinz to review this -- anyway, this has been ready for review ever since I opened the PR.

devtools/ci/after_sucess.sh
```diff
@@ -45,13 +45,20 @@ conda list -e
...
 pwd
 echo travis_fold:end:build.docs
 
 echo travis_fold:start:upload.docs
-if [[ "$TRAVIS_BRANCH" != "master" ]]; then
+
+if [[ "$TRAVIS_BRANCH" == "master" ]]; then
+    python devtools/ci/push-docs-to-s3.py
+elif [[ "$TRAVIS_BRANCH" == "docs_deploy" ]]; then
+    # change the behavior for the docs testing branch (docs_deploy branch in
```
jhprinz: interesting idea to have a branch for docs...

jhprinz:
Merging...

devtools/ci/after_sucess.sh
dwhswenson: yeah, that brach has actually been around for a while to test deployment behavior, since PRs from my own fork don't get the AWS keys.

jhprinz: but you disabled this now, correct?

dwhswenson: oh, right now it behaves the same as a fork PR. But the trick is that to make that deploy (if testing why normal deployment fails), all you have to do is switch out which like is commented there. It specifically has the (commented out, but ready to use) `--clobber` option which will overwrite (instead of sync diffs). I figure that's probably what you want if you're in that situation ;-)

PR #750, opened 2018-01-17
Add velocities to trajectory_from_mdtraj

dwhswenson:
This makes it possible for a user to provide an `np.array` of velocities to `trajectory_from_mdtraj`, and attach those velocities to the results OPS trajectory. Note: since this is part of the interface with MDTraj (as opposed to OpenMM), this does not require explicit units. Instead the implicit units must be nm/ps.


dwhswenson:
Test pass; ready for review.

PR #751, opened 2018-01-17
More docs (simulation setup docs)

dwhswenson:
The main thing here is a new doc for the "topics" section discussing things that the user has to do before running path sampling (mostly from the standpoint of practical issues). This mainly covers:

* defining states and interfaces
* obtaining initial trajectories

It also includes a couple other cleanups of the docs, most significantly some additional discussion of how to use the `RandomVelocities` snapshot modifier (which really wasn't obvious in OpenMM's unit system without an example).

### DIFF
--- a/docs/topics/simulation_setup.rst
+++ b/docs/topics/simulation_setup.rst
@@ -1,0 +1,182 @@
+.. _simulation_setup:
+
+########################
+Setting up path sampling
+########################
+
+*Or: What OpenPathSampling does not do for you*
+
+While OpenPathSampling does a lot to simplify the process of doing a path
+sampling simulation, it doesn't do everything for you. In particular,
+setting up a path sampling simulation can still be challenging.
+
+Just as configurational Monte Carlo requires an initial configuration and
+a description of the ensemble (e.g., setting a temperature in an :math:`NVT`
+simulation), path sampling requires an initial trajectory and a description
+of the path ensemble. However, both describing the ensemble and creating an
+initial trajectory are more difficult in path sampling than the equivalent
+problems in configurational sampling.
+
+This document discusses the general idea of setting up path sampling
+simulations. In includes a few references to specifics of OPS, but is mainly
+focused on the concepts. The examples section of the website includes many
+practical uses of these ideas.
+
+-----
+
+Defining the ensemble
+=====================
+
+The nature of the ensemble depends on the specific algorithm being used.
+Here, we'll discuss ensembles for TPS and TIS. In OPS, the :ref:`transition
+network <transitions-and-networks>` is used to efficiently create *all* the
+ensembles to be used in a simulation. For TPS, there is only one ensemble to
+be sampled. For TIS, there is one ensemble per interface (plus additional
+helper ensembles, such as the minus ensemble). In multiple state systems,
+TIS can involve hundreds of ensembles.
+
+Monte Carlo methods (configurational or path sampling) draw samples from a
+statistical ensemble. For configurational sampling, the thermodynamic
+ensemble to be sampled is defined by the underlying Hamiltonian of the
+system (which includes the masses of the particles and the interaction
+energies) and the natural variables of the ensemble (e.g., for the
+canonical, or :math:`NVT`, ensemble, these are the number of particles, the
+volume, and the temperature).  Similarly, a path ensemble will be defined by
+the underlying dynamics (including the underlying thermodynamic ensemble)
+and additional restrictions (such as state definitions).
+
+Selecting the extra restrictions for path sampling is much more difficult
+that selecting the thermodynamic state to sample: whereas the temperature
+and density usually come from experimental considerations, the volume of
+phase space that defines a stable state is not an experimentally observable
+quantity. Typically, defining the state requires performing MD simulations
+in each stable state to characterize its bounds, and there is not a unique
+appropriate definition when studying a particular problem of interest.
+
+In TPS, only the states need to be defined. In TIS, one must also define the
+interfaces.
+
+Defining good states
+--------------------
+
+In TPS, the main traits defining the path ensemble are the state
+definitions. The essential point is that states should be stable: a
+trajectory started from any point in the state should be expected to take a
+long time before visiting any other state.
+
+The topic of state definitions has been included in several review articles.
+For more, see:
+
+* `Bolhuis and Dellago. "Trajectory-based rare event simulations." in
+  Reviews Computation Chemistry vol. 27 (2010)
+  <http://onlinelibrary.wiley.com/doi/10.1002/9780470890905.ch3/summary>`_
+
+In TIS, the role of the minus move in decorrelating trajectories should also
+be considered. Ideally, a single minus move trajectory should be just long
+enough to fully decorrelate (lose all memory) within the state. This puts
+restrictions on both the volume of the state as well as that of the
+innermost interface.
+
+
+Defining good interfaces
+------------------------
+
+Much of the literature on TIS uses the expression that "interfaces foliate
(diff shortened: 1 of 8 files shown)


dwhswenson:
Passes tests, ready for review. Requesting review from @bolhuis, because this is documentation dealing with some general topics related to path sampling.

docs/topics/simulation_setup.rst
```diff
@@ -0,0 +1,182 @@
+.. _simulation_setup:
+
+########################
+Setting up path sampling
+########################
+
+*Or: What OpenPathSampling does not do for you*
+
+While OpenPathSampling does a lot to simplify the process of doing a path
```
bolhuis: doing -> performing

docs/topics/simulation_setup.rst
```diff
@@ -0,0 +1,182 @@
...
+sampling simulation, it doesn't do everything for you. In particular,
+setting up a path sampling simulation can still be challenging.
+
+Just as configurational Monte Carlo requires an initial configuration and
+a description of the ensemble (e.g., setting a temperature in an :math:`NVT`
+simulation), path sampling requires an initial trajectory and a description
+of the path ensemble. However, both describing the ensemble and creating an
```
bolhuis: Sampling the paths using the TPS MC algorithm then would lead to the required equlibrated path ensemble.

docs/topics/simulation_setup.rst
```diff
@@ -0,0 +1,182 @@
...
+trajectory started from any point in the state should be expected to take a
+long time before visiting any other state.
+
+The topic of state definitions has been included in several review articles.
+For more, see:
+
+* `Bolhuis and Dellago. "Trajectory-based rare event simulations." in
+  Reviews Computation Chemistry vol. 27 (2010)
+  <http://onlinelibrary.wiley.com/doi/10.1002/9780470890905.ch3/summary>`_
+
```
bolhuis: Practical and conceptual path sampling issues
PG Bolhuis, C Dellago
The European Physical Journal Special Topics 224 (12), 2409-2427

docs/topics/simulation_setup.rst
```diff
@@ -0,0 +1,182 @@
...
+In TIS, the role of the minus move in decorrelating trajectories should also
```
bolhuis: Refer to replica exchange TIS rather than TIS.

docs/topics/simulation_setup.rst
```diff
@@ -0,0 +1,182 @@
...
+------------------------
+
+Much of the literature on TIS uses the expression that "interfaces foliate
+phase space." In the OPS picture, where interfaces are volumes instead of
+surfaces, this means that inner interfaces (closer to the state) must be
+fully contained by outer interfaces. In other words, for every volume in the
+ordered set of the state, interface 0, interface 1, etc.; any point in phase
+space that is inside interface :math:`i` is also inside interface :math:`i+1`.
+Importantly, this includes the state: the innermost interface must entirely
+encapsulate the state. Note that there is no way for OPS to check this for
```
bolhuis: Technically there is a way. If a configuration inside a volume is also inside all higher interface volumes, for all configurations, than the definition is correct. This could be coded, and a fault given when this is not obeyed.

docs/topics/simulation_setup.rst
```diff
@@ -0,0 +1,182 @@
...
+and will be added to the core code in the future.
+
+-----
+
+Getting an initial trajectory
+=============================
+
+Even with configurational sampling, initial conditions are not always
+completely trivial. A PDB entry might be missing hydrogens, and probably
+needs to be solvated (and energy-minimized). However, obtaining initial
```
bolhuis: Add: "Also for MC of e.g. a liquid it is often easier to start with a lattice, that then can be melted:

docs/topics/simulation_setup.rst
```diff
@@ -0,0 +1,182 @@
...
+conditions for path sampling of rare events is much more difficult. Path
+sampling makes it possible to generate many rare trajectories once a first
+rare trajectory is given as input. But how can you get that first rare
+trajectory?
+
+The answer is, "any way that works." Some approaches include running long
+MD, or using a ratcheting approach (such as, or similar to, adaptive
+multilevel splitting) to get natural trajectories that go further and
```
bolhuis: or forward flux sampling

docs/topics/simulation_setup.rst
```diff
@@ -0,0 +1,182 @@
...
+Note that a very bad initial trajectory could equilibrate to a nonsensical
+region of trajectory space, just like a bad configuration could equilibrate
+to a nonsensical metastable state in configurational Monte Carlo. Selecting
+a "nearly" physical path is important.
+
+One approach that has shown promise is to use metadynamics to obtain a first
+transition. After the first transition, metadynamics usually will not have
+significantly altered the potential energy surface near the barrier, so this
+trajectory is likely to equilibrate to a good initial trajectory.
+
```
bolhuis: Another option that often works is a high temperature trajectories. However, here the dynamics is often altered dramatically, and so special attention is needed. One option is to do a quick committor analysis on the high temperature trajectories, close to the expected transition. A configuration with a non zero committor can be used as a valid good initial trajectory.

docs/topics/simulation_setup.rst
```diff
@@ -0,0 +1,182 @@
...
+the next ensemble, then it switches to the new ensemble and samples that.
+The process continues until the all ensembles in the transition have
+trajectories. It is essentially a version of adaptive multilevel splitting
+that has been discretized along the progress parameter. This approach sounds
+promising, but in reality is very dependent on the quality of the order
+parameter. It is very efficient some simple models, which is why we use it
+for the examples like the :ref:`toy-mstis` notebooks. In complicated
+systems, it may fail.
+
+
```
bolhuis: As pointed out above, a committor calculation can help. The committor calculation is available in OPS

docs/topics/simulation_setup.rst
```diff
@@ -0,0 +1,182 @@
...
+Loading from a file
+-------------------
+
+You can also load a trajectory from a file. This enables you to get an
```
bolhuis: Besides creating the trajectory from scratch,

bolhuis:
I reviewed this. @dwhswenson : can you make some changes?

docs/topics/simulation_setup.rst
dwhswenson: I'm not entirely sure what you meant here, although I tried to add something. See new text. By "describe the ensemble" I'm trying to say something akin to setting N, V, and T for an NVT ensemble.

docs/topics/simulation_setup.rst
dwhswenson: Adjusted to just say that we don't do it.

The better way might actually be for the state to be defined as `state | innermost_interface`, which does formally include this, but makes the `network` objects more confusing and means that the state that is sampled may not be what the user requested.

If using our standard InterfaceSet objects, I think that successive interfaces (not including the state) should be forcibly correct in this regard.

docs/topics/simulation_setup.rst
dwhswenson: I put this in the next section instead.

dwhswenson:
@bolhuis : I made the changes. You can see the diff from the previous review at https://github.com/openpathsampling/openpathsampling/pull/751/commits/fc044a6c26136e68643941959eab4452f31e2f37.

I wasn't 100% sure on what you were suggesting with https://github.com/openpathsampling/openpathsampling/pull/751#discussion_r163573771, but I tried to add something along those lines.

I think the rest should work!

bolhuis:
checked. Looks good. Please merge.

PR #752, opened 2018-01-22
Fix bug in BackwardFirstTwoWayShooting

dwhswenson:
This fixes a bug in the `BackwardFirstTwoWayShootingMover` as well as adding several new tests for two-way shooting. The bug was uncovered through the tests, which were implemented because @bolhuis was suspicious of some behavior in two-way shooting in the MISTIS toy example.

The bug was that the wrong shooting point was being used after the backward-first modification (shooting point was from the original trajectory; failed to account for the fact that the new backward segment had affected the shooting point index!)

The new tests include the following:

1. For TPS and TIS, forward-first and backward-first 2-way shooting can abort early in certain circumstances. For example, with forward-first shooting in TPS, if the forward shot lands in state A, there is no reason to do the backward shot. The new tests include these early rejection tests for both TPS and TIS ensembles (TIS-like, actually) in forward-first and backward-first dynamics.

2. For deterministic dynamics,  shooting with no modification of the shooting point should recover the original trajectory. There is now a test that checks this with two-way shooting.

### DIFF
--- a/openpathsampling/pathmover.py
+++ b/openpathsampling/pathmover.py
@@ -2292,11 +2292,16 @@ class BackwardFirstTwoWayShootingMover(AbstractTwoWayShootingMover):
 
         bkwd_partial = self._make_backward_trajectory(trajectory, modified,
                                                       shooting_index)
+        #logger.info("Complete backward shot (length " +
+                    #str(len(bkwd_partial)) + ")")
         # TODO: come up with a test that shows why you need mid_traj here;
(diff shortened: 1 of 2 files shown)


dwhswenson:
Passes tests. This should be immediately ready for merge.

jhprinz:
I am not 100% sure if I understand what needs to be the correct point, but I assume the test will check that. Otherwise it looks good as usual :)

Issue #753, opened 2018-01-28
File structure reorganization


dwhswenson:
Currently, we have a few modules that are extremely long and should be split into several modules in a subpackage. This would help new developers find their way to the relevant classes more easily.

The goals are:

1. Do not change the API. Everything in the root `openpathsampling` namespace should stay there.
2. Create shorter files. 1000 raw lines (including comments/docs/blanks) should be a rough maximum (this is a pylint suggestion, and one that I agree with).
3. Create a clearer file structure for new devs.

Note that, at this point, I'm not proposing changing tests at all. I want all tests to pass with the only change being perhaps the way things are imported. Eventually, the tests should be also be reorganized so that there is one test module per code module, but I'm not planning to implement that now.

Tasks:

- [ ] Reorganize `ensemble.py`
- [ ] Reorganize `pathmover.py`
- [ ] Reorganize `movestrategy.py`
- [ ] Reorganize `pathsimulator.py`

Others may also be added to this list later, but these are the ones where new devs are likely to be looking for models to implement their own things or to understand how the code works.

PR #754, opened 2018-01-28
Split pathsimulator.py

dwhswenson:
As mentioned in #753, we should reorganize our file structure. I'm planning to do some work on parallelizing some path simulators soon, so I'm starting by reorganizing the `pathsimulator.py` file. Here's how I'm splitting it:

* `path_simulator.py`: `PathSimulator`, `MCStep` (will move MCStep elsewhere later)
* `shoot_from_snapshot.py`: `ShootFromSnapshotsSimulation`; `CommittorSimulation`
* `path_sampling.py`: `PathSampling`
* `bootstrap_init_conds.py`: all bootstrap stuff
* `direct_md.py`: `DirectSimulation`

Note that the approach I've used to split involves some git acrobatics in order to preserve history on lines within the split files. See https://beyermatthias.de/blog/2014/09/24/splitting-files-while-preserving-history-in-git/.


dwhswenson:
Passes tests. Ready for review/merge.

dwhswenson:
Thanks -- I have some rough plans for doing the same with `ensemble.py` and `pathmover.py`, and will do those later. Will merge this one now!

sroet:
I would have liked an `API Break` or `module rename` label on this PR. 
It broke the import of `pathsimulator` by the renaming to `pathsimulators`. This change is also not mentioned in the comments.

dwhswenson:
What were you importing from `openpathsampling.pathsimulator`? Isn't everything that is public also directly importable from the main `openpathsampling` namespace?

In order to make sure everything only shows up in only one place in the (official, public) API, I think that only the least-nested level should be used. For most of OPS, that means the root `openpathsampling` namespace.

sroet:
@dwhswenson 
>What were you importing from openpathsampling.pathsimulator? Isn't everything that is public also directly importable from the main openpathsampling namespace?

You are correct, thanks for the tip. This came up in one of my old E-CAM modules (the transition state ensemble). It is now fixed, and imports from the  `openpathsampling` namespace.

PR #757, opened 2018-02-03
Allow unicode for string options in engine option check

dwhswenson:
The `options` for engines has a bunch of type checking, and if the type given isn't the same as the default type, it raises a `ValueError`. I got this error when I tried to replace a `str`-typed default with a `unicode` object. (Note that this will only be a problem in Python 2, since `str` **is** `unicode` in Python 3.)

This PR fixes that problem, and includes a regression test for it as well.


dwhswenson:
passes tests; ready for a (very easy) review

jhprinz:
Looks good. Will merge.

Issue #756, opened 2018-02-01
Switch tests to pytest


dwhswenson:
We should switch from nosetests to pytest. Nose has been dead (no longer under active development) for a while. It seems like pytest is what "all the cool kids" are using these days.

This will take a couple PRs to implement, which is why I'm putting this in an issue. The main steps I see, in order to get this working reasonably well:

- [ ] Fix namings so that they work with pytest (current tests should also run under pytest)
- [ ] Switch to using pytest on Travis
- [ ] Remove nose-specific imports; replace with pytest-native approaches

The last step might be completed in several steps, since we can develop new tests with pytest-native approaches while still using the old nose imports within pytest.

At some point in the future, it would be really nice to also make some of the tests run in a more pytest-native way (there are additional fixtures and things like parametrization that can be done in pytest that aren't available in nose; we accomplish the same in our tests by writing more code). That overhaul may happen piece by piece after the above tasks are complete -- at least, **I** don't  have time to sit down and do a full overhaul of the test suite.

PR #758, opened 2018-02-03
Pytest compatibility

dwhswenson:
First step toward #756.

Here I'm standardizing our test module and class names. Modules (files) will be test_some_module.py, and classes will be TestSomeClass. In particular, early on I misunderstood a line in the nose documentation as meaning that the test class should be called testSomeClass (first t lowercase). Nose accepts this, but also accepts TestSomeClass. Pytest only accepts TestSomeClass (which is PEP8 anyway). So many classes need their names changed.

Other changes that were required:
* pytest also requires the function be called `setup` and not `setUp`, but that only affects a few classes.
* Module-level setup needs to be called `setup_module`, not just `setup`, in order to work for both
* test modules cannot have a superclass with `__init__`; that behavior must be put into the superclass's `setup` method
* I think files need to be `test_*.py` (I changed this; I'm not entirely sure whether it was necessary for pytest, but it runs in both).

Also includes a little cleanup of extra spaces (overall, this PR should be *really* easy to review). Once it passes Travis tests, it should be ready to review. Locally, I see the same number of tests run and skipped in both nose and pytest.


dwhswenson:
Last commit fixed a typo (well, badly-chosen input) from a test in #752, which was causing stochastic failures depending on which shooting point was chosen.

Tests now pass, and this is fully ready for review/merge! Mostly just renaming test stuff so that both pytest and nosetests can find them.

dwhswenson:
Last commit fixes problem that only occurs with recent versions of netcdf4 (apparently only in the Py-3.5 conda branch so far, but already in all versions for conda-forge). Thanks to @sroet for finding the error and how to fix it.

jhprinz:
Looks good. Feel free to merge!

PR #759, opened 2018-02-13
Make DelayedInterrupt context optional

dwhswenson:
This was interfering with task distribution using dask.distributed (thanks to Mariusz Uchronski for identifying the problem).

All this does is make the `interrupter` into an attribute of the `Engine`. Defaults to `DelayedInterrupt`; added an `EmptyContext` that can be used to remove the signal catcher with:

```python
from openpathsampling.engines.delayedinterrupt import EmptyContext
engine.interrupter = EmptyContext
```

(Note that the `interrupter` attribute is the *class* itself, not an instance of it. That's important!)

Still WIP while we're experimenting with it.

### DIFF
--- a/openpathsampling/engines/delayedinterrupt.py
+++ b/openpathsampling/engines/delayedinterrupt.py
@@ -37,5 +37,5 @@ class EmptyContext(object):
     def __enter__(self):
         pass
 
-    def __exit__(self):
+    def __exit__(self, type, value, traceback):
         pass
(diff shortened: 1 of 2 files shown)


dwhswenson:
I think this can be reviewed and merged. There's really nothing to it, and it is necessary for some of the parallelization stuff.

dwhswenson:
Giving this until Saturday for a review, then I'll merge it.

PR #764, opened 2018-04-09
Fix openmmtools.RestorableIntegrator error

dwhswenson:
Tests have been failing due to a renaming in `openmmtools` (see https://github.com/choderalab/openmmtools/pull/336#issuecomment-375892878 and following). This PR will fix the errors.

This took a while because I had trouble reproducing the error in any of my own environments. However, it should be pretty straightforward.


dwhswenson:
This seems to fix it. I will merge directly.

PR #770, opened 2018-04-17
updated OpenMMEngine docstring, adresses #769

### LINKED ISSUE #769: `OpenMMEngine` ignores `platform` option
`OpenMMEngine` ignores the documented kwarg 'platform' if passed through options.
Quote from docstring:
> options : dict
            a dictionary that provides additional settings for the OPS engine.
            Allowed are[...]
                   `platform` : str, default: `fastest`,
                                  the openmm specification for the platform to be used,
                                  also 'fastest' is allowed   which will pick the currently
                                   fastest one available

This can result in quite strange behaviour like `Exception: Error initializing FFT: 5` if you request 'CPU' because you know your GPU is busy. 

I think the reason is that `platform` does not get stored when `__init__` is called, because it has no default value in `self._default_options`.
Additionally the `.simulation` property does not call `self.initialize()` with a platform argument so even if it is stored somewhere (where?) it does not get passed when creating the simulation.

I am not 100% sure if am grasping where the problem is coming from, but I have a fix in a commit https://github.com/hejung/openpathsampling/commit/3f958ec8f0e3cbe5b092f1f9f6a846b4bc5ed427
If this is the correct approach I can make a pull request with a fix and tests.

hejung:
Changes to the docstring to reflect current implementation.

(Maintainer edit: closes #769)

### DIFF
--- a/openpathsampling/engines/openmm/engine.py
+++ b/openpathsampling/engines/openmm/engine.py
@@ -80,7 +80,11 @@ class OpenMMEngine(DynamicsEngine):
             the openmm integrator object
         openmm_properties : dict
             optional setting for creating the openmm simuation object. Typical
-            keys include GPU floating point precision
+            keys include GPU floating point precision.
+            Note that by default the engine selects the fastest currently
+            available OpenMM platform.
+            If you want to specify the platform you will have to call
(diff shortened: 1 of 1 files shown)


dwhswenson:

hejung:

PR #776, opened 2018-05-28
Fix problem in to_mdtraj

dwhswenson:
A recent update to netcdf seems to be causing errors when we use `to_mdtraj`. This PR aims to fix that.

### DIFF
--- a/openpathsampling/engines/trajectory.py
+++ b/openpathsampling/engines/trajectory.py
@@ -564,7 +564,7 @@ class Trajectory(list, StorableObject):
         box_vectors = self.box_vectors
         # box_vectors is a list with an entry for each frame of the traj
         # if they're all None, we return None, not [None, None, ..., None]
-        if not np.any(box_vectors):
+        if not np.any(np.asarray(box_vectors)):
             box_vectors = None
 
         traj.unitcell_vectors = box_vectors


sroet:
This is actually happening due to an incompatibility between previous nc files and this netcdf4 version (might be a genuine upstream issue):
The following code is runs fine when fully done in a single netcdf version:
```python
store = paths.Storage('test.nc','w')
frame = paths.engines.openmm.snapshot_from_pdb('resources/AD_initial_frame.pdb') #In the examples dir
traj = paths.Trajectory([frame,frame])
store.save(traj)
store.close()
#update netcdf from 1.3 to 1.4 here to break
store = paths.Storage('test.nc','r')
traj = store.trajectories[0]
print(traj.box_vectors)
traj.to_mdtraj()
```
and the print statement returns 
```python
[[[2.558 0.    0.   ]
  [0.    2.558 0.   ]
  [0.    0.    2.558]]

 [[2.558 0.    0.   ]
  [0.    2.558 0.   ]
  [0.    0.    2.558]]] nm
<class 'simtk.unit.quantity.Quantity'>
```
however if the storage is saved in version 1.3 and loaded in version 1.4 the code breaks and the print statement is
```python
[Quantity(value=masked_array(
   data=[[2.558000087738037, 0.0, 0.0],
         [0.0, 2.558000087738037, 0.0],
         [0.0, 0.0, 2.558000087738037]],
   mask=[[False, False, False],
         [False, False, False],
         [False, False, False]],
   fill_value=9.96921e+36,
   dtype=float32), unit=nanometer), Quantity(value=masked_array(
   data=[[2.558000087738037, 0.0, 0.0],
         [0.0, 2.558000087738037, 0.0],
         [0.0, 0.0, 2.558000087738037]],
   mask=[[False, False, False],
         [False, False, False],
         [False, False, False]],
   fill_value=9.96921e+36,
   dtype=float32), unit=nanometer)]
```

dwhswenson:
Seems like they're aware of the possibility that this could break existing code: https://github.com/Unidata/netcdf4-python/pull/787#issuecomment-376235414. At least the fix for that was easy, although it's yet another time that a minor release of netCDF broke our code.

There are still some other errors in these runs. I'll try to track them down before merging this.

dwhswenson:
This is ready for review. Because the changes are small (no new features; only updates so tests still pass after upstream changes) and because they are important (tests are failing until this is merged, especially affecting users who upgrade netCDF), I will leave this up for review for 24 hours. After that, I'll merge it myself.

jhprinz:
Looks fine from my limited quick view. Please go ahead and merge!

PR #777, opened 2018-06-10
Separate `SampleMover.move` into two functions

dwhswenson:
* `._get_samples_from_sample_set`: use the input sample set to select relevant samples
* `.move_core`: uses selected samples; returns move change

This will facilitate parallelization.


jhprinz:
@dwhswenson What is the status on this?

dwhswenson:
@jhprinz : Question -- should the method to get samples from a sample set (for a given mover) be hidden API or public API? (Use underscore or not?) I've wavered back and forth on this.

Other than deciding on that, this is ready. Note that this only does it for some moves; support for a few other moves (e.g., `MinusMover`) will be added in future PRs.

jhprinz:
I think it should be public, since I don't see a reason to hide this. It could be used standalone for testing and experiments or doing statistics, etc.

dwhswenson:
Made the change to make that public. This can be reviewed at any time.

PR #781, opened 2018-06-18
Fix pandas deprecation warning : replace df.as_matrix() with df.values

### LINKED ISSUE #775: Pandas 0.23 raises FutureWarning about as_matrix()
Since Pandas version `0.23` the tests start raising `FutureWarning: Method .as_matrix will be removed in a future version. Use .values instead.` 
It is mostly used inside the tests, but also in `openpathsampling/openpathsampling/numerics/wham.py`. This could result in this warning popping up for users when doing analysis.

jhprinz:
Thought we might do this as the recommended way to replace as_matrix() which is kept for backwards compatibility.

Resolves #775 .


dwhswenson:
LGTM.

PR #782, opened 2018-06-19
omit using rejected paths in correlation visualization

### LINKED ISSUE #773: Path tree sees rejected paths as decorrelated
Step 3 has a rejected path that is marked as decorrelated (step 1 has the same issue as #771)

cc: @dwhswenson

jhprinz:
Resolves #773 .

Apparently the visualization does not take rejected or _transparent_ marked trajectories in the path tree. This PR changes this to only consider non-transparent ones. Note, that there are different possibilities to mark paths as transparant. The default is rejected, but it can also intermediate (auxiliary) paths which happen inside of movers.

### DIFF
--- a/openpathsampling/visualize.py
+++ b/openpathsampling/visualize.py
@@ -1648,7 +1648,9 @@ class PathTreeBuilder(Builder):
         for pos_y, data in enumerate(self._plot_sample_list):
             sample = data['sample']
 
-            if pos_y > 0:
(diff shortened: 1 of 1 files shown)


dwhswenson:
Merging this, then releasing. Next release will officially correspond to the (first) submitted version!

(Will ask @Nicole-de-Groot to check that this PR did fix her problem tomorrow.)

PR #785, opened 2018-06-21
Fix push-docs-to-s3 for pip 10

dwhswenson:
As of pip 10.0, the `pip.get_installed_distributions()` no longer exists (https://github.com/pypa/pip/issues/5243), so our docs haven't been pushing. This fixes the problem.

It also fixes some problems missing docs after the pathsimulator split #754.

Note that to test this, the docs already had to get pushed (so the openpathsampling.org website already matches this PR).


jhprinz:
GO ahead!

PR #796, opened 2018-09-08
Correlation function TPS

dwhswenson:
Two reviewers on the OPS articles mentioned the ensemble where a fixed-length trajectory starts in $A$, and must visit $B$, but does not necessarily end with the final frame in $B$. See http://dx.doi.org/10.1063/1.478569.

This implements that ensemble, and also does significant cleanup of the tests for `TPSNetwork`/`FixedLengthTPSNetwork`. 

Currently calling it `NewNetwork`... obviously `New` should changed. `CFTPS`? (for correlation function TPS?) Something else?

Any naming advice, @bolhuis or @jchodera?

- [x] Code
- [x] Tests
- [x] Naming
- [x] Docstrings

### DIFF
--- a/examples/tests/test_cv.ipynb
+++ b/examples/tests/test_cv.ipynb
@@ -27,7 +27,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "np.set_printoptions(precision=6)"
+    "# force numpy print options for test comparison\n",
+    "np.set_printoptions(precision=6, formatter={'float_kind': lambda x: \"{:.6f}\".format(x)})"
    ]
   },
   {
@@ -114,7 +115,7 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "338489518249846260551333267048467791884\n"
+      "283860163354172696480456782077640048652\n"
      ]
     }
    ],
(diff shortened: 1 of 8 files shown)


jchodera:
Isn't this used in the calculation of rate constants, where a reversible work calculation is used to compute the free energy of confining one endstate to state B through a series of intermediate staged calculations?

My understanding of this (see Section II.B) is that the ensemble must start in `A` but ends in `B[i]`, where `i` denotes the intermediate number such that the final intermediate is `B[N] = B`, the initial intermediate `B[0]` is all configuration space, and the intermediates all overlap with their neighbors to a significant degree. (This is the path ensemble version of a staged alchemical free energy calculation.) This series of intermediates and endpoints would indeed be a great choice to design a `TransitionNetwork` from so that you could do this via replica exchange or expanded ensemble among path ensembles.

I have never heard of a fixed-length path ensemble where the trajectory starts in A, must pass through B, but need not end in B. Where did you get that concept from?

If indeed you're looking to name a network associated with the linked paper, I'd suggest something like `StagedTPSRateCalculation` (because it is used in TPS rate computations) or `StagedStateCorrelationFunctionCalculation` (because it actually computes a fixed-time correlation function for originating from some state).

bolhuis:
About the rate calculation: The original TPS approach was to construct and staged (umbrella-type) sampling in which B[i] was slowly changing. This would be sufficient for computing the rate, but requires long paths, to capture all possible transition (barrier crossing) path lengths. In this paper it was useful to split the calculation into an staged umbrella sampling with shorter paths,  and a flux calculation with longer paths. This later flux calculation needed to have the H_B function, that allowed leaving the state B, to avoid spurious jumps at the end of the path, and hence a wrong flux estimate. In my opinion, this approach has been made obsolete by the flexible path length approach and the development of TIS. However, if people really want this, we can easily include it in OPS, not?

I agree with @dwhswenson that the naming should be consistent. The network is part of the FixedLengthTPSNetwork family, and thus should be called something like  CaptitalHBFixedLengthTPSNetwork or VisitingBFixedLengthTPSNetwork or AnyFrameInBFixedLengthTPSNetwork or PartInBFixedLengthTPSNetwork or something similar.

dwhswenson:
> In my opinion, this approach has been made obsolete by the flexible path length approach and the development of TIS. However, if people really want this, we can easily include it in OPS, not?

I fully agree. My main reason to add this is that I think that adding the network to OPS is the easiest way to answer reviewer questions about it (and with two reviewers mentioning it, it is probably worth adding). I don't want to add the associated analysis, in part because I think TIS is a better approach anyway.

> The network is part of the FixedLengthTPSNetwork family, and thus should be called something like CaptitalHBFixedLengthTPSNetwork or VisitingBFixedLengthTPSNetwork or AnyFrameInBFixedLengthTPSNetwork or PartInBFixedLengthTPSNetwork or something similar.

These get a bit long..... `paths.PartInBFixedLengthTPSNetwork` is, I think, as long as we should go. That's already 35 characters, which is a lot on an 80-character terminal. I guess the main question is if that name would mean enough for users -- it describes the ensemble involved, but nothing about the use case. (On the other hand, I guess we don't want to encourage non-experts to use this, since TIS is probably a better choice.)

dwhswenson:
This is ready for review. Name is `PartInBFixedLengthTPS`, and I think the docstrings are sufficiently clear.

PR #797, opened 2018-09-13
Fix notebook tests

dwhswenson:
We have notebook tests failing in `test_cv.ipynb` and `test_netcdfplus.ipynb`. This seems to be because of something with the numpy string representation (because `ipynbtest` compares string representation here).

This PR will fix it (or find a way around it).


dwhswenson:
Tests pass again. Giving 24 hours for review before self-merge (because this is a really small change).

PR #768, opened 2018-04-16
Attempt to fix for old openmmtools as well as new

dwhswenson:
This, combined with pinning to `openmmtools==0.14`, may be a (temporary?) fix for #767.

Update: thanks to @andrrizzi's suggestion in https://github.com/openpathsampling/openpathsampling/issues/767#issuecomment-381629371, this PR now resolves #767 in two ways:

* If the user has `openmmtools 0.14` or older installed, we still work with that (although that will be removed in 2.0, if not before)
* We also work directly with integrators in `0.15`.

PR #798, opened 2018-09-30
Updates to Examples

dwhswenson:
This includes several updates I've made to the examples, most significantly a bunch of changes that make some examples match the papers better.

This also includes #768. (Although I don't think that problem is fully solved.)

This should be merged soon, since these changes were mentioned in our response to the reviewers.


examples/toy_model_mistis/toy_mistis_3_analysis.ipynb
```diff
@@ -965,21 +965,21 @@
...
    },
    "file_extension": ".py",
    "mimetype": "text/x-python",
    "name": "python",
    "nbconvert_exporter": "python",
-   "pygments_lexer": "ipython3",
-   "version": "3.6.1"
+   "pygments_lexer": "ipython2",
+   "version": "2.7.15"
   }
```
jhprinz: I think this change from python 3 to 2 should not matter, right? These notebooks should be compatible with both anyway...

dwhswenson: This changes to whatever you used the last time you saved the notebooks. I was in Py 2.7 when I made the updates here. But these notebooks are run in our test suite, so as long as tests pass, they must work with both Python 2 and 3!

PR #799, opened 2018-10-03
Load file with reduced box vectors

dwhswenson:
This adds tools to simplify loading input trajectories, especially loading Gromacs trajectories for the OpenMM engine.

Gromacs trajectory files, such as TRRs, round the box vectors that they store. OpenMM has a strict requirement that box vectors be [in reduced form](http://docs.openmm.org/7.0.0/userguide/theory.html#periodic-boundary-conditions). Because of the rounding in the TRR files, Gromacs input often needs to be massaged to work. OpenMM has a way of handling this, but because we load the trajectories with MDTraj, we bypass the part of OpenMM that does this. Internally, we'll just import the method that OpenMM uses and apply it to our snapshots.

The PR adds tools to simplify this, including a particular method to load TRR files for the OpenMM engine.

- [x] Add functions for reducing box vectors in snapshots and trajectories
- [x] Add convenience function for loading OpenMM-engine-ready trajectory from TRR file
- [x] Docstrings
- [x] Unit tests


dwhswenson:
This should be ready for review and merge. I think the only possible concern is that we're importing a somewhat-hidden feature of OpenMM. However, to the extent possible, I'd prefer to have *them* maintain it, rather than rewriting their code as part of OPS. If they move the functionality to another name, we can update easily enough.

Overall, this will make it much more convenient to use existing Gromacs trajectories as input to OPS:

```python
from openpathsampling.engines.openmm.tools import load_trr
traj = load_trr("file.trr", top="conf.gro")
```

This should include velocities, and also fix box vector rounding.

dwhswenson:
This should also fix some problems with notebook tests; a recent change somewhere made ipynbs in Py 3.6 stricter about the order of `from __future__` imports.

PR #800, opened 2018-10-09
Update interface-constrained shooting example

dwhswenson:
Follow-up on #791.

- [x] Remove extraneous example notebook (repeats toy MSTIS 1)
- [x] Rename notebook as another in the appendix series (A4) to be consistent with other notebook names
- [x] Move `InterfaceConstrainedSelector` to main namespace (import in main `__init__.py`)
- [x] Remove extraneous cells from example notebook
- [x] Add some more text to explain what users should see from the changes.

Notebook seems to run fine after this. Should be immediately ready for review/merge.


dwhswenson:
Ready for review/merge. I think this is the last thing before I prep a 0.9.6 release!

PR #795, opened 2018-08-20
Add docs on getting velocities from a file

dwhswenson:
This has been promised as "coming soon" in the docs for a while. It was implemented in #750, which was merged a long time ago.

This PR just adds a quick update to show how one would use this, as part of the simulation setup topic docs.

Docs only; if no review in 2 days, I'll self-merge (22 August).

PR #790, opened 2018-07-19
PEP8 cleanup after #788

dwhswenson:
Also caught a number of my own old style errors, as well as the ones introduced in #788.

This shouldn't change any functionality; it just updates code style to be more PEP8 compliant.

PR #801, opened 2018-10-10
OpenPathSampling 0.9.6

dwhswenson:
This is waiting on #800. The text below will be the release notes.

---
OpenPathSampling 0.9.6 includes several new features, as well as many improvements to the examples and documentation. Highlights include:

* Interface constrained shooting, an efficient technique in TIS
* Better ways to load trajectories for the OpenMM engine (especially from Gromacs trajectories)
* bugfix for minus interface ensemble in cases where the transition is not particularly rare

## New features
* Tools to get correctly reduced box vectors for OpenMM (#799)
* Simple function to load TRR files (#799)
* "Part-In-B" network (used in TPS rate calculation) (#796)
* Interface constrained shooting (#788, #791, #790, #800) [parts by @bolhuis]

## Bugs fixed
* maskedarray fixes (#803)
* bugfix in minus ensemble (#803)
* Attempt to fix for old openmmtools as well as new (#768)
* Fix push-docs-to-s3 for pip 10 (#785)

## Miscellaneous improvements
* Updates to Examples (#798)
* Fix notebook tests (#797)
* Updates to docs (#795, #793)
* Changes to .gitignore files (#792)


dwhswenson:
Unless there are objections, I'll cut this release tomorrow.

jhprinz:
No objections. Go ahead.

dwhswenson:
Actually, I want to hold off on this for one bugfix. A bug that was brought to our attention is a problem with netCDF returning masked arrays in Python 3 (even though I tell it not to? Not sure why it ignores that.) Anyway, I have a (somewhat frustratingly messy) fix, and will make a PR of it soon. I'd rather put the almost-completed fix into 0.9.6.

jhprinz:
I would report this as a bug to netcdf4-python. They seem to actively work on this in the last changelogs...

jhprinz:
I just had a peak into the source code of `netcdf4-python` and if I get this right, then the masking is only used if the attribute `.mask` is set. This will be set to `True` by default, which is not, what we want. You can use `set_auto` to change this, but it will only change the setting of existing variables, not for variables created next. It will not change the default behaviour. Does that help?

Why there is a difference between python3 and python2. No idea. Is it clear that the used version of netcdf4 is the same?

dwhswenson:
> I just had a peak into the source code of netcdf4-python and if I get this right, then the masking is only used if the attribute .mask is set. This will be set to True by default, which is not, what we want. You can use set_auto to change this, but it will only change the setting of existing variables, not for variables created next. It will not change the default behaviour. Does that help?

Thanks; note that #803 is the fix I'm proposing for this, at least for now.

I agree that it isn't a perfect fix, but basically I'm tired of getting burned by changes in `netcdf4-python`. This is the second time I've had to fix problems with `maskedarray`s in about 6 months. Maybe we missed that my fix 6 months ago didn't work for Py3, but still; this was a problem introduced in 1.4.0, and the behavior may have changed in 1.4.1. Minor and possibly even patch releases? It makes me think we can't trust them to keep their API stable on these things, so we have to enforce stability on our side.

Part of this is also that I don't know all the details of how things work in the current storage. That's your expertise, so if you see a more elegant fix, please, open a PR. Mine is admittedly a hack, but I think we absolutely must get some sort of fix out as part of this release.

PR #808, opened 2018-11-24
Switch to pytest on Travis

dwhswenson:
This is the second step in #756. For now, we'll still require `nose` (because our tests still use things imported from `nose`), but new tests can be written as `pytest`-only, and we'll eventually rewrite the tests that use `nose` (probably over the course of several PRs).


dwhswenson:
This also fixes a deprecation warning in `shooting.py` (`UniformSelector.pick`; numpy now uses `random.randint` instead of `random.random_integers`) Also added a test for `UniformSelector.pick`, since we didn't have one before.

Ready for review and merge. I'll merge on my own on Friday if it hasn't been checked by then.

PR #809, opened 2018-11-25
Refactor move strategies

dwhswenson:
This PR involves some refactoring of things in move strategies. It is designed to make it easier to create new move strategies, without breaking any backward compatibility. It also reduces repeated code, and improves code quality (according to CodeClimate score).

One of the main things this introduces are the `get_per_mover_ensembles` and `get_parameters`  methods in strategies. Strategies are factories for movers. What the factory needs is to combine the type of mover with the initialization parameters for each mover instance to be created. The `get_parameters` method returns a list, one entry per mover, containing the parameters to initialize each mover, which makes the `make_movers` factory function very simple.

In particular, `.get_parameters` handles things like the idea that some parameters may be given once but repeated for all movers in the strategy (e.g., the engine for shooting movers). On the other hand, parameters may be given as either a single object to be broadcast to all instance or as a per-mover list (such as with shooting point selectors, which can vary depending on the mover).

This PR also involves some simplification by using `collections` objects, in particular `defaultdict`. There's a lot of code in OPS where we should be using `collections`, but aren't!


dwhswenson:
This is ready for review. In addition to stuff listed above, it also removes some of the path mover factories (the rest should be removed soon, as well, as should our other factory objects). These have been replaced by the move scheme/strategy approach.

It also fixes a problem we've been having in tests, because conda kept trying to upgrade Python from 3.5/6 to 3.7. (Soon we should drop 3.5 and add 3.7 support.)

There's a slight drop in test coverage. This is because I made well-covered code shorter. There are no new uncovered lines, just there are fewer lines overall!

dwhswenson:
Leaving this up for review for another 2 days. If no review by Saturday night, I'll merge it myself.

sroet:
A small note: this also adds `pytest` as a requirement to run the test suite.

dwhswenson:
> A small note: this also adds `pytest` as a requirement to run the test suite.

Technically, that happened as of #808. However, this might be the first PR with tests that required `pytest`. Currently we also still require `nosetests` (for the imports). As discussed in #756, the next stage in upgrading tests is to remove the `nosetests` import requirements.

PR #811, opened 2018-12-01
Add conda-forge to channels

dwhswenson:
There is some conflict with MSMBuilder (at least, for Py3). This PR will fix that.

I don't see it in my own Py3 installations, where I use the `conda-forge` channel, so I'm hoping that adding the `conda-forge` to our channels will fix the problem. 

In any case, `conda-forge` is now the recommended channel for Omnia projects, unless there's a specific reason to use the `omnia` channel, so we might as well search it first.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,20 +22,28 @@ install:
   - echo $PACKAGES
   - conda install -y -q -c conda-forge -c omnia $PACKAGES
   - python setup.py install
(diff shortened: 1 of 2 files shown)


dwhswenson:
This fixes the problem. Since this will cause all Python 3 tests to break until merged, and since it only changes parts of how Travis installs OPS, I'll merge it on Monday if not reviewed by then.

Issue #812, opened 2018-12-28
Storage: "too many indices" IndexError with libnetcdf 4.6.2


dwhswenson:
We're currently having a bug with storage when using the newest versions of netcdf. @jhprinz, could you look at this? I've identified the bug, but since it is in your part of the code, it would be better if you could fix it.

### Which version causes the bug

This error pops up in our MSTIS example, notebook 2, when loading from storage (as well as other places in the examples). I tried several versions on netcdf4/libnetcdf (conda-forge installs), running just that notebook in Python 3.6. Here's what I found:

* netcdf4 1.4.1 / libnetcdf 4.6.1 works
* netcdf4 1.4.2 / libnetcdf 4.6.1 works
* netcdf4 1.4.2 / libnetcdf 4.6.2 fails

In the last update I made (causing the failure), there were a few other packages that change as well, in particular h5py 2.8->2.9 and hdf5 1.10.3->1.10.4.

### Minimal script to recreate the bug

Here's a simple script to raise the error (I'm using the `mstis_bootstrap.nc` file generated in the `toy_model_mstis` example):

```python
import openpathsampling as paths
storage = paths.Storage("./mstis_bootstrap.nc", mode='r')
store = storage.samples
bad_keys = store.vars['trajectory'].variable[[0,1]]
store.vars['trajectory'].getter(bad_keys)
```

This raises the error:

```pytb
IndexError                                Traceback (most recent call last)
<ipython-input-1-2db5aa8b0c6b> in <module>
      3 store = storage.samples
      4 bad_keys = store.vars['trajectory'].variable[[0,1]]
----> 5 store.vars['trajectory'].getter(bad_keys)

~/openpathsampling/openpathsampling/netcdfplus/netcdfplus.py in <lambda>(v)
    886                 None if w[0] == '-' else store.load(int(UUID(w)))
    887                 for w in v
--> 888             ] if get_numpy_iterable(v) else \
    889                 None if v[0] == '-' else store.load(int(UUID(v)))
    890

~/openpathsampling/openpathsampling/netcdfplus/netcdfplus.py in <listcomp>(.0)
    885             getter = lambda v: [
    886                 None if w[0] == '-' else store.load(int(UUID(w)))
--> 887                 for w in v
    888             ] if get_numpy_iterable(v) else \
    889                 None if v[0] == '-' else store.load(int(UUID(v)))

IndexError: too many indices for array
```

Continuing in a interactive session after that, you can see the following:

```python
[...]

PR #814, opened 2019-01-15
Pin libnetcdf to 4.6.1; update install process

dwhswenson:
This is a (temporary) fix for #812.


dwhswenson:
This fixes the problem when the builds complete, but now the msmbuilder install takes so long that Travis times out. I may need to do some more to reorganize our installation process.

PR #815, opened 2019-01-21
Fix warnings for Python 3.7

dwhswenson:
I've switched to using Python 3.7 as my primary development environment -- a big jump from 2.7! This PR fixes a few warnings I was getting when running the tests in 3.7.

1. `inspect.getargspec` is deprecated and will be removed in 3.8. Using a try/except to prefer `inspect.getfullargspec` when available (Py3k) and fall back to `inspect.getargspec` when not. (This only occurs once, in `netcdfplus/base.py`.)

2. We were getting a weird error (multiple times) at the end of pytest:

        Error in sys.excepthook:
        
        Original exception was:

    This didn't actually cause the tests to fail, but certainly looked bad. It turns out that this was happening in tests that had opened a file for writing, but not closed the file after. Closing the file fixes. Figured it out pretty quickly thanks to https://github.com/pytest-dev/execnet/issues/30#issuecomment-316962021 -- same error message, also solved by closing "files" they were writing to.

I'm still seeing a few warnings (about importing from `collections.abc` vs. just `collections`), but these are emitted by pandas, networkx, and mdtraj. It looks to me like the next releases of those packages may fix.

Also, this isn't intended to add testing for Python 3.7 yet. For one thing, we can't do full tests until msmbuilder works on 3.7 (they seem to have a problem with recent Cython).

### DIFF
--- a/openpathsampling/netcdfplus/base.py
+++ b/openpathsampling/netcdfplus/base.py
@@ -10,6 +10,11 @@ if sys.version_info > (3, ):
 
 logger = logging.getLogger(__name__)
 
+try:
+    getfullargspec = inspect.getfullargspec
+except AttributeError:
+    getfullargspec = inspect.getargspec
+
 
 class StorableObject(object):
     """Mixin that allows objects of the class to to be stored using netCDF+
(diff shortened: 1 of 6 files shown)


dwhswenson:
This is ready for review and merge. Small changes, so if not reviewed in 4 days, I'll go ahead and merge.

PR #816, opened 2019-01-28
Refactor networks

dwhswenson:
This PR makes progress toward refactoring the `TransitionNetwork` objects to make them easier to understand and easier to reuse for constructing user-defined networks. The changes here are driven both by things that will improve the CodeClimate score, as well as things that will make facilitate some projects (involving new networks) that I'm currently working on.


dwhswenson:
This is ready for review. It mainly involves

* refactoring to reuse code
* making many methods private

It also adds some comments about a number of things that don't seem to be used anymore.

Finally, I changed the input file for the `tutorial_storage` notebook. The was occasionally failing with an error about floats not being subscriptable. This was (I think) fixed by always using a pre-written file, which will be the same on every run (instead of using a file generating during testing.)

I'll give this one 72 hours for review, and will merge on Sunday if no comment.

PR #818, opened 2019-02-24
Deprecations and API cleanup prior to 1.0

dwhswenson:
This does a little preparation for releasing 1.0, and managing things after the 1.0 release. I've been playing with some metaprogramming lately to create a few tools that will help us obey [SemVer](http://semver.org) after 1.0. The primary things are:

* `deprecations`: A module in OPS to handle deprecations. Issues warnings and handles updates to docstrings.
* `define_api`: A script (not in OPS; [repo is here](https://github.com/dwhswenson/define_api)) to identify all the classes/bare functions/constants that are officially "in" the API. After defining those, it is easy to write a test that ensures they are still there. It finds all things that could be imported.

In addition, I've made a few updates to the API and deprecated a few things.

More details on each:

## Deprecations

The main things here are a couple of decorators: (1) `@deprecate`, which should go on any function/method/class that has been deprecated, and (2) `@has_deprecations`, which should go at the module-level for any object that has deprecations. In addition, there is a `Deprecation` class that provides a simple tool to summarize specific deprecations.

The `@deprecate` decorator adds a deprecation warning to the function call (`__init__` for classes). It also tags the function/class as needing to have its docstring modified. It takes a `Deprecation` instance as argument, and uses information from that in the warning message and docstring modification. The `@has_deprecations` decorator actually modifes the docstrings.

If docstrings don't need to be modified (e.g., extra branches of code that manage backward compatibility), then you can just use the `deprecation.warn()` method from the correct instance of `Deprecation`.

The instances of `Deprecation` should be kept in the file `deprecations.py`, which enables us to easily find all of them.

There's room to improve this approach, but what I've included in here should be pretty good. You can see it in use for several of our deprecations.

New code: 

- [x] Implement `deprecation` decorators
- [x] Tests for `deprecation` code
- [x] Docstrings for `deprecation` code

## API definition

The total API can be defined by two things: (1) how to import each object in the API; (2) what functions/arguments are defined for each object. Here, I'm only dealing with the first part; where each object is located.

The concept here is that each thing in the API should have exactly one canonical location. In practice, however, there are often multiple ways to get to a given object. For example, all of the following work:

```python
from openpathsampling import AllInXEnsemble
from openpathsampling.ensemble import AllInXEnsemble
from openpathsampling.tests.test_ensemble import AllInXEnsemble
# and if you really want to be silly
from openpathsampling.high_level import network
network.paths.AllInXEnsemble
```

I think it is clear that we want the canonical API location to be the simplest one: `openpathsampling.AllInXEnsemble`. This is the one that we guarantee won't change.

The `define_api` (external) script creates a list of where things are found, according to the rule that the shortest path (fewest dots) to an object is canonical. It allows us to create a file with all such paths. Then it is easy to write a test to ensure that all objects remain accessible at those paths.

New code:

- [x] Add `test_api.py` (and `api_OPS1.txt`)

## API Changes and Deprecations

At the same time, I'm doing some clean-up of the API in advance of the release. This involves cleaning up issues related to the above topics, as well as getting rid of some unused code. Here are the changes I'm suggesting.

API changes:

- [x] `openpathsampling.openmm` should be accessed as `openpathsampling.engines.openmm`
- [x] `openpathsampling.toy` should be accessed as `openpathsampling.engines.toy`
- [x] Remove `EntersXEnsemble`, `ExitsXEnsemble` from code and docs
- [x] Remove `VolumeFactory` and `EnsembleFactory` from code and docs
- [x] Remove `VoronoiVolume` from core `openpathsampling` API (leave the code in place; it isn't tested or used, but may be used later; this just means it isn't at `paths.VoronoiVolume`)

Deprecations:

- [x] Deprecate `MoveDetails` for generic `Details`


dwhswenson:
Various users (@jhprinz @sroet @hejung): Will the proposed API changes cause problems for any of you? I don't think anyone is using the factory objects, and I think all our examples used `openpathsampling.engines` to get to our engines, but best to check whether you think these will cause headaches.

dwhswenson:
The main code that will be added here has been added. Although this is still WIP until I finish the last few tasks, please feel free to start review already, @jhprinz. I expect to be unable to write code for a few days, but when I'm back, I'll quickly finish this PR. Once completed, I'll leave 24 hours for review before merging.

sroet:
> Will the proposed API changes cause problems for any of you? I don't think anyone is using the factory objects, and I think all our examples used `openpathsampling.engines` to get to our engines, but best to check whether you think these will cause headaches.

Thanks for the heads-up, I don't expect this to become a problem from the current description (did not look at the code). 
Is there an easy way for the deprecation warnings to be turned into errors (mainly for future continuous testing purposes).

dwhswenson:
> Is there an easy way for the deprecation warnings to be turned into errors (mainly for future continuous testing purposes).

Yes: If you're using pytest, see [its docs on warnings](https://docs.pytest.org/en/latest/warnings.html). In general, see the [Python docs on the warnings filter](https://docs.python.org/3/library/warnings.html#the-warnings-filter).

hejung:
Same here, thanks for the heads-up and I also do not expect trouble.

dwhswenson:
@jhprinz : Ready for review. I'll actually leave this up for 48 hours, because I'm hoping that @sroet and @hejung can take a look at the API file (new file [`openpathsampling/tests/test_data/ops1_api.txt`](https://github.com/openpathsampling/openpathsampling/blob/8a793cc58a437aee9d52c799dacf5c03235a7745/openpathsampling/tests/test_data/ops1_api.txt)) to see if there's anything obviously missing or wrong based on their uses. These will be the official, will-not-change, ways to access these objects (there's a test in place in ensure that). Other paths to these objects might work, but aren't guaranteed.

openpathsampling/tests/test_data/ops1_api.txt
```diff
@@ -0,0 +1,270 @@
...
+openpathsampling.UnionEnsemble
+openpathsampling.UnionVolume
+openpathsampling.VelocityDirectionModifier
+openpathsampling.Volume
+openpathsampling.VolumeCombination
+openpathsampling.VolumeEnsemble
+openpathsampling.VolumeInterfaceSet
+openpathsampling.WrappedEnsemble
+openpathsampling.condense_repeats
+openpathsampling.git_HEAD
```
sroet: What would this API be?

openpathsampling/tests/test_data/ops1_api.txt
```diff
@@ -0,0 +1,270 @@
...
+openpathsampling.analysis.PathHistogram
+openpathsampling.netcdfplus.StorableNamedObject
+openpathsampling.numerics.SparseHistogram
+openpathsampling.analysis.tis.ConditionalTransitionProbability
+openpathsampling.analysis.tis.DictFlux
+openpathsampling.analysis.tis.EnsembleHistogrammer
+openpathsampling.analysis.tis.FullHistogramMaxLambdas
+openpathsampling.analysis.tis.MinusMoveFlux
+openpathsampling.analysis.tis.MultiEnsembleSamplingAnalyzer
+openpathsampling.analysis.tis.PathLengthHistogrammer
```
sroet: is this only implemented for tis and not for TPS?

openpathsampling/tests/test_data/ops1_api.txt
dwhswenson: `git_HEAD` is trick that I added to get around our static values in the `version.py` file. If you have a developer installation of OPS, the actual git revision is available from `paths.git_HEAD()`. (Basically, it gives the output of `git -C $OPS_DIR rev-parse HEAD`.)

In the future, I'll add the dynamic `version.py` that I've used in more recent projects, but not before 1.0.

openpathsampling/tests/test_data/ops1_api.txt
dwhswenson: The specific `analysis.tis.PathLengthHistogrammer` should work on either TPS or TIS data, but its benefit is that it returns path length histograms per ensemble (as needed for TIS). As with other per-ensemble computations in the `analysis.tis` namespace, this first goes through all the steps and finds the (Monte Carlo) weight of each trajectory within each ensemble. Then, in a second pass, it calculates the value (length in this case) and updates the histogram with the correct weight.

When there are multiple ensembles and/or an expensive value to calculate, this approach is efficient. But for TPS, there's only one ensemble, and the length is cheap to calculate. So the path length histogram can be plotted easily (using only one pass over the data) with:

```python
plt.hist([len(s.active[0].trajectory) for s in steps])
```

PR #819, opened 2019-03-15
Gromacs engine

dwhswenson:
This is the Gromacs engine for OPS. Note that this includes #144, and should be merged after that.


dwhswenson:
This is ready for beta testing, which is why I made the PR. Since this is the first *real* usage of the external engine, I'd like to have a few experts put it through the paces before we deliver it to the general audience. This will be part of either the 1.1 release or the 1.2 release.

A few people who I hope might be willing to beta test (based on hearing that you need Gromacs): @jvreede @hejung (others feel free as well!)

hejung:
Just as a short info: I started testing gromacs, including if everything works as expected for MARTINI.
I will create a list of things that I find here and edit this comment if the list grows.

1. `Trajectory.to_mdtraj()` does not work for Gromacs trajectories, it works if we use `Trajectory.to_mdtraj(tra.topology.mdtraj)`; the reason beeing that `to_mdtraj()` checks for a (mdtraj) topology at `engine.mdtraj_topology` and gromacs engine has it at `engine.topology.mdtraj`.

dwhswenson:
Thanks @hejung -- I just added `mdtraj_topology` (the idea of a general `engine.mdtraj_topology` came after I had implemented most of the Gromacs engine).

Thanks especially for checking on MARTINI; that was something I was curious about. Are there no issues with topologies from MARTINI? This was one of my concerns (as in, that the engine can't even be built because the internal topology causes trouble -- maybe you can't export MARTINI trajectories to MDTraj, but that would be less of a concern.)

hejung:
@dwhswenson I think there is a typo in the fix from yesterday. You named the property `mdtraj_trajectory` instead of `mdtraj_topology`

dwhswenson:
> You named the property mdtraj_trajectory instead of mdtraj_topology

oops! Should be fixed now. This is why one should always include tests for everything! haha

dwhswenson:
> Using `gmx_engine.engine_command()` before calling `gmx_engine.set_filenames()` raises an AttributeError since the output filename is not set yet.

Actually, this was easy to fix: in principle, one should always set all instance variables in `__init__` anyway. Now it should return some placeholders -- let me know if you think the placeholder nature of the text should be made more clear by adding `$` before names (currently all caps).

> Setting `nstxout` and `nstvout` to zero in the `.mdp` file results in the gmx_engine never terminating, although gmx does (and possibly even generated a trajectory in `.xtc` format).

Added mention of restrictions on these values in the docstring. You're correct that we should also make this clear in other documentation (such as examples).

> `.gro` files do not contain bond information.

I thought that MDTraj figures it out when you're working with residues it is familiar with (mainly amino acids). I've modified this so that you can set the `mdtraj_topology` property. That enables you to create an MDTraj topology according to whatever approach (e.g., using a PDB) and associate it with the engine. By default, it will use the one that MDTraj builds (and that will still be at `engine.topology.mdtraj_topology`). 

For MARTINI, you'd probably need to build custom topologies. To create custom "elements", take a look in [here](https://github.com/mdtraj/mdtraj/blob/master/mdtraj/core/element.py). I'm not sure what will happen if you add an "element" whose symbol already exists (e.g., `"W"`).

For now this custom topology override will not be saved/reloaded; I need to think about how to do that (you shouldn't be able to reload it and then change it to a different topology, because then the reloaded object is different from the saved object.)

> I will look into adding MDAnalysis function CVs (which would take an MDAnalysis universe as input) as solution

This is something I've been interested in for a while. I haven't worked much with MDAnalysis, but it's definitely a package we eventually want to integrate with!

---

By the way, as you find more things, please feel free to add additional comments, instead of editing the existing. I get notifications on new comments, not on edits!

hejung:
Concerning the bond info for atomistic simulations: I checked for ala, here mdtraj sets no bonds if using the `.gro`. But since `grompp` also takes `.pdb` as inputs (`-c`) at least since gmx v4.6, I think the easiest solution is to directly pass a `.pdb` instead of `.gro` to the ops gmx engine. This then results in mdtraj loading the `.pdb` instead of the `.gro` and reading the bonding information from there. So basically just another documentation change to inform users that giving a `.pbd` instead of a `.gro` might make their life easier. :)

dwhswenson:
@hejung : I updated the docstring for the `gro` argument to explain that other filetypes are possible, and best MDTraj integration may come from PDB. Thanks again!

hejung:
A bit off-topic, but since it was last mentioned here:
I have a preliminary implementation of openpathsampling with a MDAnalysisFunctionCV (see https://github.com/hejung/openpathsampling/tree/gromacs_mdanalysis).
To get it to work I modified the `openpathsampling.engines.openmm.topology.MDTrajTopology` to an `openpathsampling.engines.topology.MDTopology`. This new topology is heavily oriented at the old, but now also stores the topology as MDAnalysis topology. To that end it has two classmethods that convert MDTraj topolgies to MDAnalysis and vice versa. (For now) it only saves the mdtraj topology to storage, but that should be sufficient since the round-trip MDAnalysis->MDTraj->MDAnalysis gives the same topology. Replacing the old with the topology results in all tests passing.
@dwhswenson Would this be a viable solution? In that case I would write tests and open a pull request from a clean branch where I rebased the changes on openpathsampling:master.

Concerning the gromacs engine: While writing the topology I realized that all other openpathsampling engines take a topology as input to `__init__`. This results in the user explicitly loading the topology as with `openpathsampling.engines.openmm.tools.topology_from_pdb()`. This in turn can minimize the assumptions made when loading topologies by requiring explicit flags (like MARTINI/atomistic; guess bonds; standartize names; etc). Getting these things right automatically seems hard to me so I would opt for making all engines taking an explicit topology. This would enable the user to e.g. load a MARTINI topology from `.tpr` with MDAnalysis and have everything work as expected (if using the new topology together with the custom MARTINI bead elements I defined even MDTrajFunctionCVs).

hejung:
@dwhswenson Some additional info on the `MDAnalysisFunctionCV` since I realized I started writing where I stopped thinking:
The branch in which I added it is based on your gromacs_engine (I was not sure if I would need it), so this diff is probably most informative: https://github.com/dwhswenson/openpathsampling/compare/gromacs_engine...hejung:gromacs_mdanalysis
The essential addition is the new openpathsampling MDTopology, otherwise I oriented myself at how the `MDTrajFunctionCV` works, i.e. using `Trajectory.to_mdanalysis()` to call the `MDAnalysisFunctionCV`.

dwhswenson:
@hejung : Sorry for the delay; I spent the last week dealing with apartment-hunting stuff. Could you start a new issue with the MDAnalysis discussion? Regarding converting between topologies, I think this is really useful code, but I wonder whether it might be even better to put it into one of the other packages (MDAnalysis, MDTraj, or maybe something like ParmEd?) so it can be used by a broader audience -- after all, it would be useful to people outside the path sampling community.

The stuff specific to the CV wrapper looks mostly good. The topology stuff has a few points that I think might be better if done in a way that's more flexible (first step: instead of `martini` being a `bool`, have it be something like `particle_set`, which can be `'atomistic'` or `'martini'` or others, and create classes to handle specialized behavior for each -- this will make it much easier to, in the future, add support for particle models). We can discuss this in more detail in the separate issue.

dwhswenson:
> Getting these things right automatically seems hard to me so I would opt for making all engines taking an explicit topology. This would enable the user to e.g. load a MARTINI topology from .tpr with MDAnalysis and have everything work as expected

I was aiming to require minimal input from the user. If added, a separate topology argument would be an optional argument for `__init__`. There's a topology that is implicit from the Gromacs input files, and for many users, that will be sufficient. If the user provides an override, how can we verify that the user-provided topology is for the same system as the one determined from the inputs? The fear here is that the user might do something that doesn't make sense -- but maybe checking number of particles is enough?

In general, I think we want to keep the simplest setup as simple as possible, and as much like the native engine (helps new OPS users). OpenMM's `Simulation` class requires a `topology`, `system`, and `integrator`, so the OPS engine requires the same. For Gromacs, `grompp` requires files for the `-c`, `-f`, and `-p` arguments, so the OPS engine requires those files. Additional things beyond that would be optional, but we need to consider potential conflicts. (The existence of a `topology` argument in the toy engine comes from a long debate between me and @jhprinz.)

Overall, I think our topology objects need to be significantly restructured. I often don't know whether something is supposed to take an OPS topology or an MDTraj topology (or maybe an OpenMM topology?) But that will probably be API-breaking, and has to wait for 2.0.

dwhswenson:
@hejung @jvreede Checking in with people who have been beta-testing this: Does it seem ready for production? If so, I'll remove the WIP status from it, and it will be ready for review.

arjunwadhawan:
Hi @dwhswenson . Where can I find an example notebook which uses Gromacs as an engine ?
I realized you had made one with alanine dipeptide but cant locate it online

dwhswenson:
@arjunwadhawan It's only in the branch for this PR -- since the Gromacs engine hasn't been merged into the master, it wouldn't make sense to have the example in master yet.

With a [developer install of OPS](http://openpathsampling.org/latest/developers/install_developers.html), check out this branch, and you'll see it in the examples. You can also find the notebooks (and related files) here: https://github.com/dwhswenson/openpathsampling/tree/gromacs_engine/examples/gromacs

arjunwadhawan:
Ah alright. Makes sense. We are going to use it and will get back if any issue 
Thanks

dwhswenson:
Recent commits address stuff discussed in #855.

dwhswenson:
The most recent commits are not backwards compatible with files generated by earlier versions of the OPS Gromacs engine. Here's the explanation and the workaround:

#### What changed and why?

Now OPS will store the contents files used to set up the simulation (e.g., `.gro`, `.mdp`, `.top`) inside the engine when saving the engine to our `.nc` storage.

OPS simulation objects should be immutable (at least, mostly). When you reuse an engine, you should always get the same behavior.

However, until these commits you could reload an Gromacs engine from storage, and change the behavior of the engine by changing the files. For example, you could change the integrator by editing the mdp file.

This is a problem because we want to ensure that you have a record of provenance in your data: you should to be able to tell exactly what integrator was used to create a specific snapshot.

In the new implementation, OPS will check for the existence of a file with the correct name. If it doesn't exist, OPS will create the file (writing the stored contents to disk). If it does exist, OPS will check whether the sha1 hash matches the hash of the stored data. If so, it uses the existing file. If not, it raises an error.

We raise an error on mismatch because we're not sure whether wanted to use the engine as is (in which case, maybe you don't want to overwrite the file?) or if you thought that you could change the behavior by changing the file (which breaks immutability and provenance checking).

#### Why will this raise an error with existing storages?

Before these commits, the file contents weren't stored. When OPS tries to reload the contents of these files from our nc storage, the engine will fail to load. For some reason, it appears to fail silently, too -- so you'll get `None` for the engine. (This is behavior that will be different in the new storage.)

#### How can I work around the error?

Write a custom `from_dict` function and monkey patch it to the engine class (before reloading!):

```python
from openpathsampling.engines import gromacs as ops_gmx
def custom_from_dict(cls, dct):
    return super(ops_gmx.Engine, cls).from_dict(dct)

ops_gmx.Engine.from_dict = classmethod(custom_from_dict)
```

Note that after monkey-patching, you will (within a given Python session) only be able to load old-format engines, but will store in the new format.

Tagging people who might need to know this: @hejung @arjunwadhawan @IdsTeepe

dwhswenson:
I'm calling this one ready for review (FINALLY!!) The last changes were to move the `ExternalMDSnapshot` out of Gromacs-specific code (because it can be used by other engines, too) and to make the `ExternalEngine` abstract, putting a subclass with the reference/test implementation in the test suite.

This is a huge PR, and includes #144, which will not be merged separately. I'll leave this up for a week for review, merging no earlier than Saturday 7 December. If anyone is actually going to review it and wants longer than that, let me know. If there are no comments expressing an interest in reviewing, I'll assume everything is okay, and will merge.

I plan to release 1.1 shortly after merging this.

Thanks to @hejung  and @IdsTeepe for doing some real-world testing of this PR. Just to make sure everyone who might care is getting the message: @jhprinz @arjunwadhawan.

PR #821, opened 2019-03-22
Refactor path histograms

dwhswenson:
This refactors a few parts of the path histogram code. Technically, this is an API break: there were methods that *should* have been private, but were part of the public API (methods related to the interpolation). This PR refactors so that there is a separate class `VoxelInterpolator`, with subclasses `NoInterpolation` and `SubdivideInterpolation`. Note that this does not affect the tests at all; all the tested behavior remains the same.

This will make it much easier to implement alternative interpolation algorithms (which may be much faster than what we have now).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,7 @@ install:
   #- conda install --quiet --yes pytest 
   - conda config --env --add pinned_packages libnetcdf=4.6.1
   - conda config --env --add pinned_packages pytest-cov=2.5.1
+  - conda config --env --add pinned_packages ipython!=7.4.0
   - conda update --all --yes --use-local --quiet
   # pin pytest-cov b/c: https://github.com/z4r/python-coveralls/issues/66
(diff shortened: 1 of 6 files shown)


dwhswenson:
This is immediately ready for review and merge. In addition to the primary purpose (cleaning up the API of path histograms), this also fixes a couple problems with tests:

* The most recent IPython has a bug that causes our notebooks to fail. See https://github.com/ipython/ipython/issues/11659. This will cause our tests to fail until fixed; I suggest using this PR as the fix.
* For some reason, I was having trouble with the notebook tests in Python 3.5 (occasionally getting opcode errors?), so I created a separate analysis test file for Python 3.5. Note that soon, we should switch to testing 3.6 and 3.7 instead of 3.5 and 3.6.

I need this improvement for an upcoming workshop, and this includes the fix for testing errors. Therefore, I'm giving this until Monday afternoon before I merge it myself, if there is no review by then.

PR #823, opened 2019-03-31
Refactor move scheme

dwhswenson:
This is the last bit of cleanup so we can lock down the API for 1.0.

It creates a new class `MoveAcceptanceAnalysis`, which will replace most of the code in `scheme.move_summary` and allows removal of related (`scheme.move_acceptance`, `scheme._move_acceptance_line`). Also, this new class is tested, unlike the old stuff.

I'm still finishing up the tests, but the code should otherwise be ready for review. I'd like to merge this on Tuesday, so @jhprinz,  please review ASAP if you plan to review.


dwhswenson:
I've confirmed that this gives the same results as the previous version with both normal TIS simulations and SRTIS simulations (which behave differently because of the null moves). This was done by running the notebooks (but not committing the results).

Since the text is the same in both cases, I'm going to replace the old methods with something that calls the new methods.

dwhswenson:
Replaced the old version with the new. This is now definitely ready for review.

PR #824, opened 2019-04-01
Switch to 2.7/3.6/3.7; deploy docs on 3.7

dwhswenson:
This updates our testing and deploy setup. Previously, we tested on Python 2.7/3.5/3.6, and deployed based on the 2.7 version. Now we'll test on 2.7/3.6/3.7 (dropping 3.5) and deploy docs on 3.7 (from now on, the most recent Python will be our deploy basis, instead of the oldest version.)

UPDATE: This also include several improvements on #840. In particular:

* Better way of getting the existing Python version number (old approach didn't work for Anaconda defaults Python)
* Switch installs to `pip install --no-deps` to avoid having setuptools go seeking extra requirements
* Add `$WORKAROUNDS` to `conda_install_reqs.sh` to add dependencies of dependencies when they're missing (currently using at as a workaround for https://github.com/choderalab/openmmtools/issues/428)


dwhswenson:
This is now just waiting on a Py 3.7 version of ipynbtest (although it would be nice if msmbuilder could get a Py 3.7 version up, too.)

dwhswenson:
This is ready for review! I'll leave it up until Friday afternoon. The new version of the `conda_ops_dev_install` script will fix some issues I had with some other projects.

PR #825, opened 2019-04-12
Minor updates to WHAM, add numerics docs

dwhswenson:
This makes it possible to access the `lnZ` from a WHAM calculation after the combined histogram has been made. This is useful for things like the reweighted path ensemble.

It also adds API documentation for the `numerics` subpackage. (Docstrings were there, but they didn't have a place in a toctree!)

### DIFF
--- a/docs/api_sections.rst
+++ b/docs/api_sections.rst
@@ -18,6 +18,7 @@ OpenPathSampling API
     network
     engines/index
     netcdfplus/index
+    numerics
     tis_analysis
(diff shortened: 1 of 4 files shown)


dwhswenson:
This is ready for review. Small changes to code, so review should be easy. I'll leave it up for 5 days (until Wednesday 17 April) unless the ESDW participant who was starting on the RPE really needs it sooner. ;-)

PR #826, opened 2019-04-19
"Visit all states" ensemble

dwhswenson:
A few times, we've used an ensemble that visits all states (for example, to create initial trajectories for the alanine dipeptide examples). The implementation of this ensemble is a nice demonstration of what you can do with OPS, but understanding how it works is definitely a bit tricky. In addition, @jhprinz wanted behavior that gave feedback about what states had been visited, and how many MD steps had been run. So when he used this in the AD MSTIS example, he added a bunch of extra code to give that info (code that is, in my opinion, hard for beginners to follow).

So what I've done is to create a special ensemble, `VisitAllStatesEnsemble`, which combines these two ideas. The ensemble itself is a wrapper around the simple ensemble we've described before (well, slightly modified, but basically the same). But it also takes options related to progress reporting, and the `can_append` method is adjusted so that if those are set, it gives output similar to what @jhprinz did in his example. 

This can be seen as an example of how an ensemble can report status while running the can-append method. (Just requires using the ensemble cache.)

This will also (eventually) make the `interval` parameter in `iter_generate` unnecessary. At least, as far as I can tell, it was only used for reasons like this (as in @jhprinz’s setup of the alanine dipeptide trajectory.)

All of that, along with tests, are in this PR. I've put the new ensemble in a separate file in order to facilitate upcoming splitting of the ensemble.py file. I updated the AD MSTIS example to use this instead of @jhprinz's old code.

Tasks:
- [x] Implement `VisitAllStatesEnsemble`
- [x] Tests for `VisitAllStatesEnsemble`
- [x] Example for `VisitAllStatesEnsemble`
- [x] Docstrings


dwhswenson:
This should be ready for review. No real rush on this one; will leave it up for a week (will merge on Friday 26 April if there's no discussion before then.)

PR #827, opened 2019-04-20
Shooting selector tests

dwhswenson:
A small PR to add missing tests for shooting point selectors (and to touch up some *very* old code).


dwhswenson:
This is ready for review now. It mainly just adds some tests; other than that, I removed a few things that were unused (`.identifier` for shooting point selectors? looks like an old storage thing; also removed `__init__` that did nothing but call `super`), and converted triple-single docstrings to triple-double (i.e., `'''` becomes `"""`, which is both our standard style and, I believe, the recommended style).

Since this should be a pretty straightforward review, I'll leave it up until Wednesday 24 April before merging it myself.

PR #829, opened 2019-04-28
Improve conda performance

dwhswenson:
Our Py 3.5 builds are failing on Travis because conda is taking too long to solve the environment. Granted, we should be moving away from 3.5, but conda performance is an issue in general. It just isn't quite bad enough to cause Travis failures for other Python versions.

To make this faster, I'll work through some of the steps described here: https://www.anaconda.com/understanding-and-improving-condas-performance/

### DIFF
--- a/devtools/ci/miniconda_install.sh
+++ b/devtools/ci/miniconda_install.sh
@@ -25,7 +25,11 @@ bash $MINICONDA -b
 
 export PATH=$HOME/miniconda${pyV}/bin:$PATH
 
-# add omnia and update
+# this puts the channel priority to (1) conda-forge; (2) omnia (3) defaults
 conda config --add channels http://conda.anaconda.org/omnia
 conda config --add channels http://conda.anaconda.org/conda-forge
+
+# this may speed up conda's package resolution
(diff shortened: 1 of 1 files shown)


dwhswenson:
It looks like just using a strict channel priority was enough for a major speed-up (I guess the problem was solving packages that exist in all of conda-forge/omnia/default). This seems to fix the problem, so this is now ready for review, and I'll merge it tomorrow if there's no opposition. (Basically, one line has changed.)

I will not, for now, use conda-metachannel. I suspect that some of the ideas from conda-metachannel will eventually be incorporated into the core conda code, so it's better to see if the conda team can speed things up without us having to maintain code to create the URL for conda-metachannel.

As an aside, I think there are two separate issues with installation performance. The first is cases like this, where we literally can't run the tests because installation takes too long. The second is that it is annoying to have to wait a long time for the installation when we're working on PRs.

I think the second problem can be ameliorated by using Travis's caching tools. The idea is that the nightly build would update the cached requirements (keeping us up-to-date), but regular builds as a part of a PR would just use the most recent cache.

PR #830, opened 2019-05-01
fix bug when getting the name of the function used to define the cv

ajsilveira:
func_name is not an attribute of FunctionCV

### DIFF
--- a/openpathsampling/netcdfplus/dictify.py
+++ b/openpathsampling/netcdfplus/dictify.py
@@ -388,13 +388,13 @@ class ObjectJSON(object):
                 err += '\n4. be passed as an external parameter ' \
                        '(not for imports!)'
                 err += '\n\n        my_cv = FunctionCV("cv_name", ' + \
-                       c.func_name + ', \n' + \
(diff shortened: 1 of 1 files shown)


dwhswenson:
Actually, this is a Python 2.7/3.x issue; the issue is with the function that the `FunctionCV` is wrapping around. See https://docs.python.org/3/whatsnew/3.0.html#operators-and-special-methods. We still support Python 2.7, so this should be set up in a way that works in both 2.7 and 3.x. You probably need something like this (at the top of the file; module-level):

```python
if sys.version_info > (3, ):
    get_callable_name = lambda c: c.__name__
else:
    get_callable_name = lambda c: c.func_name
```

(then use `get_callable_name` to get the name).

Also, I assume that you found this because something happened that raised an error message (since this is in an error message). I just wrote up some docs on making CVs that will, I hope, make it a little clearer! See #831.

ajsilveira:
Yes. I was not importing the modules. Thanks for that link!.

dwhswenson:

ajsilveira:

Issue #832, opened 2019-05-07
box vector initialization for non-periodic systems


ajsilveira:
I am trying to simulate a system in implicit solvent.
When initializing the variables to be saved in the netCDF file, a variable
 with shape (1,3,3) is created even for non-periodic systems.
The error is:
```
File "netCDF4/_netCDF4.pyx", line 4358, in netCDF4._netCDF4.Variable.__setitem__
ValueError: cannot reshape array of size 0 into shape (1,3,3)
```

The following script reproduces the issue.
The PDB is here: https://gist.github.com/ajsilveira/3c7b9171383de68e966d958a557fb215

```
import openpathsampling
import openpathsampling.engines.openmm as omm

template = omm.snapshot_from_pdb('porin-ligand.pdb', periodic=False)
storage = openpathsampling.Storage("myfile.nc", "w")
storage.snapshots.save(template)
```

dwhswenson:
Could you double-check the script you provide? I get that `periodic` is not a valid keyword for `snapshot_from_pdb`, which matches what I see in the code, and there are no recent changes to that function.

Are you perhaps actually using the `periodic` keyword for a `snapshot_from_testsystem`? As I recall, @jchodera has previously mocked a testsystem with a namedtuple (in code for the host-guest system, IIRC -- I'm hoping he can point you to it more easily that I can try to dig it up). That would provide a model that should create a minimal example.

I suspect that the underlying problem is an issue with how netcdfplus handles things when the box vector is `None`. @jhprinz, could you look into this? That's your code.

ajsilveira:
Yes. You are right. I had modified the code to include that option (periodic), because I was getting this error:
```
  File "/home/silveira/miniconda3/envs/myenv/lib/python3.6/site-packages/openpathsampling-0.9.6-py3.6.egg/openpathsampling/engines/openmm/tools.py", line 108, in snapshot_from_pdb
    box_vectors=u.Quantity(pdb.unitcell_vectors[0], u.nanometers)
TypeError: 'NoneType' object is not subscriptable
```

dwhswenson:
The `snapshot_from_pdb` code creates snapshots using a separate code base from everything else (the main code to build snapshots from MDTraj trajectories is and should be `trajectory_from_mdtraj`. I'll fix it so that it is based on that code. The entire function will probably just be to return `ops_load_trajectory(pdb_file)[0]`, so you can just use `ops_load_trajectory` instead.

[...]

PR #833, opened 2019-05-08
snapshot_from_pdb should use same code path as other functions

dwhswenson:
As discussed in #832, `snapshot_from_pdb` should use the same code path (for creating velocities; setting box vectors) as related functions. Namely, the main functionality should be in `trajectory_from_mdtraj`. This PR makes it so that it uses `ops_load_trajectory`, which in turn uses `trajectory_from_mdtraj`.

In the future, I think we should deprecate the `simple_topology` option. The topology should only be made once, which would mean that the performance concern is not a worry.


dwhswenson:
This is ready for review. It's pretty simple; will merge it on Tuesday (giving 4 days) if no comment before that.

PR #834, opened 2019-05-08
fix c.func_name in ObjectJSON.callable_to_dict for python 3

king-michael:
The property ``func_name`` was substituted by ``__name__`` in python 3.
* see: https://docs.python.org/3/whatsnew/3.0.html#operators-and-special-methods

This pull requests implements a fix to be able to use the code in python 2 and python 3.


king-michael:
Sorry it's already handled in #830

PR #835, opened 2019-05-12
Fix problems with storage for nonperiodic systems

### LINKED ISSUE #832: box vector initialization for non-periodic systems

dwhswenson:
Resolves #832. Currently, netcdfplus creates a storage space for box vectors as `n_spatial` x `n_spatial` arrays. If you create a nonperiodic system, the box vectors are `None`. When trying to save the snapshot, it raises an error, since `None` can't be reshaped into the correct size array.

Here, I modify the `StaticContainerStore` such that it will convert box vectors `None` to an array of all zeros for storage, and when loading, will convert a box vector array of all zeros to `None`.

### DIFF
--- a/openpathsampling/engines/features/shared.py
+++ b/openpathsampling/engines/features/shared.py
@@ -132,7 +132,13 @@ class StaticContainerStore(ObjectStore):
     def _save(self, configuration, idx):
         # Store configuration.
         self.vars['coordinates'][idx] = configuration.coordinates
-        self.vars['box_vectors'][idx] = configuration.box_vectors
+        box_vectors = configuration.box_vectors
+
+        if box_vectors is None:
+            n_spatial = configuration.coordinates.shape[1]
+            box_vectors = np.zeros((n_spatial, n_spatial))
+
+        self.vars['box_vectors'][idx] = box_vectors
 
     def get(self, indices):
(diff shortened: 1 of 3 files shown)


dwhswenson:
This is ready for review. @ajsilveira : I assume that this (including your addition) has fixed the problem for you? If so, then I'll leave this up until at least the end of Friday for comments before merging it myself.

ajsilveira:
@dwhswenson yes, it worked for me. Thanks!

PR #838, opened 2019-05-28
Fix bad documentation

dwhswenson:
This PR fixes two problems with the documentation, which were noticed by @bolhuis:

1. Corrects the docstring for the `GaussianBiasSelector`, which previously referred to the width parameter `alpha` as the variance. Correctly, it is inversely proportional to the variance. The Gaussian we use is $exp(-\alpha (\lambda(x) - l_0)^2)$.

2. Removes the `fun_with_pathmovers.ipynb` example notebook. This example no longer runs, and most of the features it seemed to highlight have been removed. In particular:

    * `move_change in path_mover` does not work
    * `path_mover.enum` does not exist
    * `path_mover.printtree()` does not exist

Understandably, this caused confusion!

I'll leave this up for comment for 3 days; will merge on Friday 31 May if there is no discussion.


dwhswenson:
I went ahead and fixed two more things, while I was at it:

* Python 3.5 builds were failing because of some weird issue with PyEMMA and strict channel priority (https://github.com/markovmodel/PyEMMA/issues/1419). This PR turns off strict channel priority in Python 3.5. (This was causing all builds to fail already.)
* Switch from `python-coveralls` to just `coveralls` for integration with coveralls.io. `python-coveralls` has an annoying version conflict with pytest and pytest-cov, and it doesn't seem to be maintained anymore (the conflict has been around for ages).

PR #839, opened 2019-06-15
Improve performance on trajectory_from_mdtraj

dwhswenson:
I was working on a system where I read in a really long trajectory (10^5 frames) with MDTraj, and then convert that to an OPS trajectory using `paths.engines.openmm.tools.trajectory_from_mdtraj`. It was agonizingly slow, so I looked into the code.

In our code, we obtained the box vectors from the MDTraj trajectory inside the loop over frames. However, MDTraj's `unitcell_vectors` is a `@property`. This means that, for every frame, we were recalculating *all* the unitcell vectors, and then selecting the one relevant to the frame we wanted, leading to quadratic scaling.

In this PR, I fix that by moving the box vector calculation outside of the loop over frames, so that it is only done once. Still passes all tests, and was *much* faster for my use case!

This should be immediately ready for review.

### DIFF
--- a/openpathsampling/engines/openmm/tools.py
+++ b/openpathsampling/engines/openmm/tools.py
@@ -220,6 +220,11 @@ def trajectory_from_mdtraj(mdtrajectory, simple_topology=False,
         empty_vel = u.Quantity(np.zeros(mdtrajectory.xyz[0].shape),
                                vel_unit)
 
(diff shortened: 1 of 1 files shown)


dwhswenson:
Will leave this up until Wednesday. It should be a very easy review.

PR #840, opened 2019-06-26
Update install process (fix Travis-CI problems)

### LINKED ISSUE #822: conda_ops_dev_install.sh should only download dependencies
One small thing that I noticed recently. 
The current conda_ops_dev_install.sh installs and then uninstalls OPS, which is a bit redundant. 

The file now looks like this
```
conda install -y -c omnia openpathsampling
conda update -y --all
conda uninstall -y openpathsampling
git clone https://github.com/openpathsampling/openpathsampling.git
cd openpathsampling && pip install -e . && cd ..
```
which could be converted to (using the `--only-deps` option of conda):
```
conda install -y -c omnia openpathsampling --only-deps
conda update -y --all
git clone https://github.com/openpathsampling/openpathsampling.git
cd openpathsampling && pip install -e . && cd ..
```

dwhswenson:
Tests on older Python versions are failing. This seems to be related to some change in what's available upstream (especially related to our current pins). This PR is to try to figure out what the problem is.

UPDATE: In the process, I've made major changes (and simplification) to our installation process.

* There's a new script, `devtools/conda_install_reqs.sh`, that installs all the requirements
* Pinned packages now go in `pinned`
* The conda recipe is deprecated, and will be removed at some point in the future. We're now only using it as requirements list, although pins put there will also be respected
* New version of `devtools/conda_ops_dev_install.sh`, which makes use of the new `conda_install_reqs` script. Now the correct way to do a developer install for testing is to wget/curl `conda_ops_dev_install.sh`, and run it. Resolves #822
* We no longer use strict channel priority. That had been to speed up conda's solver, but the conda team has made major speed improvements so that it is no longer necessary (and it seemed to cause problems).
* Removed old pin to IPython != 7.4. That was no longer needed after the IPython 7.5 release.
* Removed old pin to libnetcdf==4.6.1. Apparently it got fixed upstream (see https://github.com/openpathsampling/openpathsampling/issues/812#issuecomment-506378825)

There's still more cleanup to be done in terms of the install procedure (I'd like to put *all* requirements in setup.py as they should be, instead of just using the conda recipe as a placeholder for that, and I'd like to reorganize our setup.py.) But that can wait for a later PR.


devtools/conda_install_reqs.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/bash
+
+# USAGE
+# ./conda_install_reqs.sh
+#   Install the requirements. If the environment variable OPS_ENV is set,
+#   then create a new environment called $OPS_ENV. Note that you MUST set
+#   the environment variable CONDA_PY (the version of Python to use, e.g.,
```
sroet: This (`CONDA_PY`) is an extra requirement for downstream testing, could you build-in a fallback like 
```
CONDA_PY=`python --version | cut -d' ' -f2`
```
if `CONDA_PY` is not defined?

dwhswenson: Actual code is a bit more complicated, because (1) `python --version` goes to stderr, so have to redirect; (2) I only want to restrict to major/minor; don't care about patch-level. But this is done.

dwhswenson:
> I would like a fallback for the CONDA_PY variable. (Also because it is not mentioned as a requirement in the installation docs)

Done. Also, thanks for reminding me that this is mentioned in the docs; I'd completely forgotten about that. I updated the docs to discuss the new functionalities.

> * line 22-23 of conda_install_reqs.sh

Leaving this in because I would rather it work that way, but it doesn't with the current conda. For some reason, conda is willing to ignore existing pins unless they're explicit -- meaning that installing something can actually cause your pinned Python version to change (so 3.5 suddenly becomes 3.6 or something like that). However, if you list a pinned requirement on the command line, it still obeys the pin. (I think this may be a bug to file with conda, if I can figure out a minimum working example.)

dwhswenson:
@jhprinz : Leaving this up for you to review until the evening of Sunday, 30 June.

If anyone is getting install errors on other projects based on OPS (I'm having this for the ops_additional_examples and the OPSPIggybacker), I recommend copying the current `conda_ops_dev_install.sh` script, and invoking it as `source conda_ops_dev_install.sh dwhswenson travis`. That will give you a working install; once this PR is merged, you can remove the `dwhswenson travis` arguments and go back to `curl`ing down the most recent version of the script.

PR #847, opened 2019-07-23
Update docs; remove need for conda (library) in dev install

### LINKED ISSUE #846: Dev install script is not installing matplotlib for py35
Hey running the following
```
conda create -n py35 python=3.5
conda activate py35
curl -OLk https://raw.githubusercontent.com/openpathsampling/openpathsampling/master/devtools/conda_ops_dev_install.sh
bash conda_ops_dev_install.sh
cd openpathsampling/
nosetests
```

results in
```
/home/blub/temp/openpathsampling/openpathsampling/numerics/histogram.py", line 4, in <module>
    import matplotlib.pyplot as plt
ImportError: No module named 'matplotlib'
```

I expected this to be work, as the website docs still mention `python 3.5+`

edit:
After installing `matplotlib` it is still missing `openmm`, `ujson`, `NetCDF4`,  `networkx`, `openmmtools`

edit2:
As another issue, `nose` can't handle the pytest decorators 
```
TypeError: test_f() missing 1 required positional argument: 'frame'
```

while the docs still state
>  The test suite uses both pytest and, for the time being, nose.

dwhswenson:
Fixes #846, plus a few other issues (in particular, I noticed that our docs were building on all Python versions, not only the deploy version — not a problem,  but a waste of time on Travis.)

- [x] Fix docs license info
- [x] Fix mention of Python 3.5+
- [x] Fix discussion of Pytest/Nose
- [x] Fix install script to not require conda
- [x] Fix issue with docs building for all versions
- [x] Add acknowledgments to TOC so the link is in the sidebar

### DIFF
--- a/devtools/ci/after_sucess.sh
+++ b/devtools/ci/after_sucess.sh
@@ -2,8 +2,17 @@
 
 DEPLOY_PY="3.7"
 
+echo "Running after success for:"
+echo "* Python version: ${CONDA_PY}\n* Branch: ${TRAVIS_BRANCH}"
+echo "* Pull request: $TRAVIS_PULL_REQUEST"
+
 if [[ "$TRAVIS_PULL_REQUEST" != "false" ]]; then
     echo "This is a pull request. No deployment will be done."; exit 0
+fi
+
+if [[ "$CONDA_PY" != "$DEPLOY_PY" ]]; then
+    echo "Only deploy from Python ${DEPLOY_PY}. This is Python ${CONDA_PY}."
+    exit 0
 fi
 
 #if [[ "$TRAVIS_BRANCH" != "master" ]]; then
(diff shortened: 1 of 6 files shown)


dwhswenson:
@sroet : Can you verify that this fixes for you? Since the change is in a secondary script in the repo, not the `conda_ops_dev_install.sh` script itself, you can run the normal procedure, except do:

```bash
bash conda_ops_dev_install.sh openpathsampling docs_deploy
```

That will check out the `openpathsampling/docs_deploy` branch, where these changes are, and run the install based on the information in that repo (using the updated script to get requirements, which shouldn't require that conda is installed.)

Also, could you review the changes here, since most of them are to address lack of clarity in the docs that you identified in #846?

dwhswenson:
> fails for `3.5` because of `yaml` versions

Just curious: what do you mean by this?

(Also, currently having some weird flaky test fail in test_pyemma notebook; I'll try to track that down before merging. I can't imagine it has anything to do with this PR, but there also don't seem to be any version changes in requirements).

sroet:
On `python=3.5` this still fails to install matplotlib, due to
```
  File "devtools/install_recipe_requirements.py", line 42, in <module>
    required_packages = recipe_to_requirements(recipe_yaml)
  File "devtools/install_recipe_requirements.py", line 34, in recipe_to_requirements
    reqs = yaml.load(recipe_yaml, Loader=yaml.FullLoader)['requirements']
AttributeError: module 'yaml' has no attribute 'FullLoader'
```
which is probably due to the fact that `pyyaml=3.13` for `py3.5` and `5.1.1` for `py3.6`.

dwhswenson:
Ah, yeah, the script to find reqs will now require newer versions of pyyaml, and I guess they never released those for Python 3.5. The explicit `Loader` silences an annoying warning in pyyaml. Story behind that [here](https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation).

dwhswenson:
We're definitely getting flakey errors in the test_pyemma notebook. Something to do with storage, it appears. But I can't find any version differences that make the problem evident.

Anyway, @sroet says that this fixes his problems, and it's mainly docs anyway, so I'm going to ahead and merge this.

PR #848, opened 2019-07-23
Flaky test_pyemma.ipynb tests

dwhswenson:
In #847, I started seeing some flaky test behavior in the `test_pyemma.ipynb` notebook. It would often fail on Py 3.6 and/or 3.7. It looks like it might have something to do with storage (it fails after reloading CVs from storage).

I can't reproduce it locally, so this PR first an attempt to see if I can diagnose the problem. I've turned on verbose output for that notebook.

### DIFF
--- a/examples/ipynbtests.sh
+++ b/examples/ipynbtests.sh
@@ -67,7 +67,7 @@ ipynbtest.py --strict "test_netcdfplus.ipynb" || testfail=1
 date
 ipynbtest.py --strict "test_cv.ipynb" || testfail=1
 date
-ipynbtest.py --strict --verbose "test_pyemma.ipynb" || testfail=1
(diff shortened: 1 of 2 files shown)


dwhswenson:
I think there may be some condition where the variable needs to be prepared in some way before loading directly from it (which is what the example did). Even adding some other code trying to inspect before the error made it so that I couldn't reproduce the error (ran each trial several time). This was only a problem in this one example, so I don't think it is likely to be a problem for users.

I think I've fixed it by as a side effect of fixing the name of the variable in the ignored cell before where the error was occuring. That line was already causing errors (which didn't stop the test from passing, because they were ignored, but were still annoying.)

The changes are small, so I'll leave this up for a day before (will merge no earlier than Thursday morning).

PR #849, opened 2019-07-23
Add packages as a space for user contributions

dwhswenson:
In the near future, (I hope) we will have a number of user contributions in several areas, including ensembles, path movers, and CV wrappers. I made a few empty subpackages that will be a good place to keep those. Eventually, we should also migrate our existing versions of these things into those subpackages.

UPDATE: I went ahead and moved my `VisitAllStatesEnsemble` into the `ensembles` subpackage, so that gives a first example of how this will be done.

In detail, here's what I've implemented:

This PR is basically saying that we'll use something like what was suggested in #753 for future contributions. In the long run, I'd like for most of our existing code to be restructured into this format as well (as said in #753).

New ensembles, CV wrappers, and path movers will go into special subpackages for those. The idea is that everything related to adding one of these will usually fit in a single file, and this will help people browsing the source find the thing they're interested in. We already do this (to some extent) for engines and for path simulators.

By managing the correct imports in the subpackage, so that, e.g., `ensembles/__init__.py` only includes things that should be public-facing, we can safely `from .ensembles import *` in our main `__init__.py`, and the root namespace has all the public-facing stuff.

For ensembles and CVs, this is essentially saying that our huge single-file modules for each should be a package. That makes sense. For path movers, I'm suggesting something more significantly different from what we currently do. Path movers usually have a move strategy associated with them, and sometimes also have a special move scheme. I'm suggesting that users put path mover, strategy, and scheme all in a single file. This will be useful for new developers who want to create something similar to what already exists -- a complete example in one file is better than bits of code scattered across several files.

In these cases, the developer will then import the mover into `pathmovers/__init__.py`, while the scheme and strategy will be imported into `pathmovers/move_schemes.py` and `pathmovers/move_strategies.py`, respectively. Again, each of these intermediate imports serves as the gatekeeper for the public API, so we can do a `from foo import *` in the main `__init__.py`.

Previously `openpathsampling.strategies` had basically been an alias of `openpathsampling.high_level.move_strategy`. Now I've made it into a proper subpackage that contains everything from `high_level.move_strategy` as well as `pathmovers.move_strategies`.


dwhswenson:
This is ready for review. See update of the top post for a more detailed description.

Since this is significant for future development (it is more an "enhancement proposal" than a new feature), I'd like an actual review on this, instead of just waiting for a period of comment. @jhprinz, please look this over if you can. Also, it would be nice to get input from potential future contributors: @sroet and @hejung, if either of you have a chance, I'd appreciate your input.

I won't merge until there's at least one review of this.

dwhswenson:
> I am missing a subpackage for the shootingpoint selectors and modifiers?

I think these can go directly in the `pathmovers/__init__.py`. Keep in mind, at least for the 1.x cycle, the purpose of these packages is just to organize things so that they can be imported into the main `openpathsampling` namespace. People should not be doing `from openpathsampling.ensembles import VisitAllStatesEnsemble` when they can instead do `from openpathsampling import VisitAllStatesEnsemble`.

In any case, even if we move things out of the main `openpathsampling` namespace in a later version, it still makes sense for shooting point selectors/snapshot modifiers to be in the `pathmovers` package. After all, they're really only needed in the context of path movers, right? (especially shooting point selectors) In the spirit of "Flat is better than nested," I think we should keep user-facing namespaces as flat and few as possible, until it starts to feel like the namespace is too crowded to find things.

Final point: I definitely think that, unless the resulting file reaches ~1000 lines, it is best to put the shooting point selectors/snapshot modifiers specific to a new mover in a single file with the new mover. Again, this goes to the idea that a single file with all the code will be easier for a new developer to understand. This just means that the import statement in `pathmovers/__init__.py` is `from my_module import MyMover, MySelector`.

sroet:
>In any case, even if we move things out of the main openpathsampling namespace in a later version, it still makes sense for shooting point selectors/snapshot modifiers to be in the pathmovers package. After all, they're really only needed in the context of path movers, right? 

Yes, I think so.

> Final point: I definitely think that, unless the resulting file reaches ~1000 lines, it is best to put the shooting point selectors/snapshot modifiers specific to a new mover in a single file with the new mover. Again, this goes to the idea that a single file with all the code will be easier for a new developer to understand. This just means that the import statement in pathmovers/__init__.py is from my_module import MyMover, MySelector.

Sounds reasonable, I was just wondering about them, as those classes were not mentioned in the pathmovers/README.md (like `MoveScheme`)

dwhswenson:
> Sounds reasonable, I was just wondering about them, as those classes were not mentioned in the pathmovers/README.md (like `MoveScheme`)

Added a parenthetical about "helper classes" for your pathmover. Hope that clears it up!

sroet:
> Added a parenthetical about "helper classes" for your pathmover. Hope that clears it up!

It does, thank you.

dwhswenson:
>  In this case I would suggest to rename cv_wrappers to collectivevariables for consistency reasons.

Followed this suggestion. In general, we have consistency problems with CVs. Do we call it `cv` or `collectivevariable`? Depends on which class/function. This is something we should clean up in he next major release. I prefer `cv` (assuming that context will prevent confusion with "curriculum vita"). But for some reason, I don't like the plural, `cvs`. Flashbacks to the bad old days before git, I guess.

dwhswenson:
To give @jhprinz a chance to review this, I'll leave it open until the afternoon of Sunday, 28 July. However, based on the reviews from @sroet and @hejung, I'm hopeful that this will make things clearer for new contributors! (Especially once a few examples are added to each subpackage.)

PR #850, opened 2019-08-06
Add spring shooting to OPS core

sroet:
This adds the `spring shooting` algorithm as described by [Brotzakis and Bolhuis](http://dx.doi.org/10.1063/1.4965882) to the core of OPS. It has been developed previously on an E-CAM 2020 workshop.

* [development repository](https://gitlab.e-cam2020.eu/sroet/spring_shooting) 
* [E-CAM 2020 documentation](https://e-cam.readthedocs.io/en/latest/Classical-MD-Modules/modules/OpenPathSampling/ops_spring_shooting/readme.html)


sroet:
@dwhswenson This one should be ready for another round of comments

dwhswenson:
Started on another round of review on this; won't finish it today. But one question: Something weird happens at step 124 in the spring shooting path tree in the example:

image

Any idea what's up with that? I think it only happens once in the whole tree, but it is strange.

sroet:
> Any idea what's up with that? I think it only happens once in the whole tree, but it is strange.

So I looked into it and it seems that happens when we select the "last" index as a shooting point (in my new test it was `38` of a trajectory of length `39`). This results in a `0` md-step acceptance, which the tree does not like.

Now, I am not too sure what to do with this one (or index `0`); we can reject it like other "illegal" indices (negative or ones that are outside of the trajectory) or keep it accepted. It shouldn't matter for the MC acceptance, however it might alter the efficiency of the "spring" (as it does update the `last_accepted_index`).

I will keep it as the only reference for this edge case mentioned in the paper is:
> Reject the entire move if the index is outside of the current path.

And in the paper the `path` it is not to well defined, as it claims length `L` but indices `[0, L]`, so I assume it is an `off-by-one` error and the paper actually means length `L+1`, start index `1`, or final index `L-1` and should include both end-points.

Do you have another view on what is the "correct" implementation?

sroet:
After a high bandwidth discussion between me and @dwhswenson, the algorithm is slightly altered . 

It now does not allow for frames in the state to be selected anymore, and now uses these indices to ensure 0-md step shots for illegal shooting points (so it selects index `0` for a backwards shot and `len(traj)-1` for a forward shot).

sroet:
Reading through the papers that cite OPS following #967 , I noticed that there is one [that uses spring shooting with OPS](https://pubs.acs.org/doi/10.1021/acsomega.0c01434). @dwhswenson do you know if they used this code/E-CAM code or if they implemented their own? (Was just wondering, thought it was cool to see this/similar algorithm being used)

dwhswenson:
@sroet I don't know, but I would assume it was your E-CAM code, since Christoph may have known about that. That work was part of a thesis that came out in 2019, so I doubt it used this PR.

dwhswenson (changes requested):
Looks good. A few very minor changes.

I'm a little uncomfortable with the use of `SpringShootingSelector.acceptable_snapshot` as a way to transfer information between methods. However, I'm not sure how easy it would be to change that, so no need to (unless you see another approach).

sroet:
> I'm a little uncomfortable with the use of SpringShootingSelector.acceptable_snapshot as a way to transfer information between methods.

Yeah, it is not ideal. Keep in mind that this is just a backstop to correctly count the rejections, if this is not done correctly this just turns into an identity move (as the trial-trajectory should always be identical to the input trajectory if an illegal frame is selected).

The only issue this might lead to (other than a misrepresentation of the number of rejected paths) is less efficient sampling due to that (illegal) point being picked as the anchor point for the next try. None of these outcomes should actually break detailed balance. 

I can make it underscored, or name mangled (double underscored) to indicate that users should not mess with it, if that would help to make it more comfortable?

dwhswenson:
> I can make it underscored, or name mangled (double underscored) to indicate that users should not mess with it, if that would help to make it more comfortable?

Yeah, probably adding an underscore would help. How reusable is that class likely to be in other use cases? It might be possible to add methods that aren't part of the standard selector API that give a different return value. So `spring_pick` might return `index` and a bool as to whether the snapshot was acceptable.

In the regular shooting mover, the selector is a separate class because different selectors are interchangeable. That isn't true with spring shooting -- here there's a bit of separation of concerns, but it's doing more to just mimic the standard selectors for the purpose of familiarity.

Anyway, that's a bigger change that doesn't need to be done now. Just documenting the thought. For now, just add the underscore, please.

sroet:
@dwhswenson thanks for the review, all your comments should be handled now :)

dwhswenson (approved):
LGTM. In the future, we might change the example to only show the first 50 or so moves in the path trees, but will merge this now!

PR #853, opened 2019-09-17
Deprecate MSMBuilder

dwhswenson:
MSMBuilder is not being maintained. Its `conda-forge` build pins numpy to 1.12. OpenMM 7.4 requires numpy>=1.14, (although this requirement is implicit, not explicit, so we end up with test failures when we try to run OpenMM with an old version of numpy, see https://github.com/openmm/openmm/issues/2385).

I think the solution on our end is to deprecate the integration with MSMBuilder (`MSMBFeaturizerCV`). In principle, we can keep it around, but we can't test it in the same environment as the newest OpenMM.

I'm marking it as deprecated and to be removed in 2.0, with the thoughts that:

1. I don't think anyone was using this functionality in OPS.
2. It's more important to keep up to date with OpenMM.
3. If someone picks up MSMBuilder maintenance, we can always "undeprecate."

### DIFF
--- a/devtools/optional_packages.txt
+++ b/devtools/optional_packages.txt
@@ -1,1 +1,0 @@
-msmbuilder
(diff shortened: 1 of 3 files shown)


dwhswenson:
This is ready for review. I'll leave this up for comment for at least 24 hours (will not merge before evening of 18 September).

PR #865, opened 2019-10-16
Updates to docs/examples

dwhswenson:
After checking our web analytics, I looked over the pages that seem to be most frequently visited. It seems that the "introductory examples" are the most often viewed. This is good, because this is where we want people to go. But this is bad because some of the examples are really old and don't illustrate our best practices.

Problems in the toy model MSTIS example: 

* Getting from trajectories to sample sets is horribly confusing. The notebook shows multiple approaches that predate the (much better) `scheme.initial_conditions_from_trajectories` method. In this PR, I add a new "topic" doc to preserve the legacy approaches, but use the much easier (and recommended) approach in the example.
* Saved analysis is based on very little data, and includes hard-coding of colors that date back to avoiding a very old matplotlib default color cycle (that wasn't friendly for RG-colorblind). In this PR, I improve the analysis and the discussion of the analysis.

In addition, I'm adding a "topic" doc on my workflow, which is basically setup in ipynb; run simulation with .py script; analysis in ipynb. I see far too many people doing setup in straight Python scripts; life is so much easier if you save the sim objects to a file and load from there.

- [x] Topic: Setting up sample sets
- [x] Topic: "How I use OPS" -- DWHS's workflow
- [x] Simplify MSTIS example creation of initial conditions
- [x] Improve MSTIS example analysis (partially)


dwhswenson:
This now includes pretty much a complete rewrite of the toy MSTIS example, which is used in testing. In particular, I've rewritten everything about getting initial samples. I *think* the stuff I've written should work for edge cases like spontaneous transitions in the innermost ensemble (which makes it harder to get an initial minus trajectory), but it's worth keeping an eye out if we get some stochastic errors on that.

dwhswenson:
Updated the MSTIS analysis example. I think I need to re-run some of this soon to figure out performance, because the current performance sucks. Probably need to use `AnalysisStorage` here. However, the overall thing is a major improvement over the current docs, so it is time to get this PR merged in.

dwhswenson:
I didn't update the MSTIS analysis, beyond just using results with better statistics. In addition to the points discussed above, this PR makes a bunch of other improvements to docs:

* Add analysis tools to the API docs
* reorder API sections to be a little more relevant (I think this will need further work)
* reorganize the API docs for CVs
* remove unnecessary "Functions" and "API" labels from API subsection headers
* Add Transitions API docs
* Improve docs on the purpose of the replica ID
* Update install info
* Miscellaneous typo fixes and improvements to wording

Anyway, every few weeks I like to check the website analystics to see what pages people are actually reading, and I try to see if there's anything on those pages that need improvement. So more docs PRs to come, depending on what people actually use/read.

dwhswenson:
Oh, and since I otherwise would have buried the lede, a separate comment for this: This PR also adds a "Videos" section to docs (did I forget to mention that [we have a YouTube channel](https://www.youtube.com/channel/UC8zKLYGrKFGxhDDL7xAwKaA)?

This will be open for comment for 24 hours. I will merge no earlier than Thursday morning (Jan 30).

PR #875, opened 2019-11-14
Path density: Bresenham interpolation

dwhswenson:
Alternate (and much faster) interpolation for path density plots, using a variant of the [Bresenham algorithm](https://en.wikipedia.org/wiki/Bresenham%27s_line_algorithm). Not as exact as the current version, which catches all voxels the line passes through, but *much* faster.

Edit to summarize what is included:

* `BresenhamInterpolation` : based on Bresenham line-drawing algorithm
* `BresenhamLikeInterpolation`: inspired by Bresenham, but works in float space instead of int space. I thought this made a more correct-looking interpolation.
* major refactor of path histogram interpolation testing to make it much easier to test new interpolations
* `BresenhamLikeInterpolation` is now the default for `PathDensity`. I argue that this does not count as breaking the API because we guarantee a path density with interpolation, but we don't guarantee that the interpolation has to bitwise the same. Any code that worked with `SubdivideInterpolation` will work with `BresenhamLikeInterpolation`, but will return a slightly different path density. Anyone who wants the old behavior back can pass `interpolate=SubdivideInterpolation` at instantiation.


dwhswenson:
This is ready for review/merge. Note that it does change the default implementation of path density. And wow, it is a **lot** faster. Not quite as pretty -- I may still use the slow version for publications. But this renders quickly, with interpolation.

If no comments, I will merge this on or after Mon 18 November (giving at least 48 hours).

PR #876, opened 2019-11-15
Fix minimal install

dwhswenson:
Something seems to have changed between the last test run of #867 and when I merged it, and the minimal install is now giving an error. Fixing that here.


dwhswenson:
This seems to have fixed it. Merging immediately

PR #877, opened 2019-11-17
Fix coveralls parallel build issues

dwhswenson:
This should fix some occasional problems with coveralls not combining coverage correctly. I haven't noticed the problem in OPS as much as in other projects, but I expect we'd have the same issues.


dwhswenson:
I'll leave this up for 24 hours for review/comment. The 1.8% bump in coverage in this PR suggests to me that it is working.

PR #879, opened 2019-11-21
Simplify Trajectory.__getattr__

### LINKED ISSUE #878: Unable to export rejected trajectories to mdtraj.Trajectory on gromacs_engine
Let's say I have a variable `storage` with some accepted steps and some rejected steps. If I want to export all accepted steps to trr files, I might write a code block like this:
```
for step in storage.steps:
    if step.change.accepted:
        index = step.idx(storage)
        traj = step.active[0].trajectory.to_mdtraj()
        traj.to_trr(f'prefix_{index:03d}.trr')
```
And that works fine. Recently however, I have tried to export the rejected trajectories for some analysis. [As all my steps have a unique sample associated with them](https://github.com/openpathsampling/openpathsampling/issues/868#issue-510723127), I thought of writing a very similar code block:
```
for step in storage.steps:
    if not step.change.accepted:
        index = step.idx(storage)
        traj = storage.samples[index].trajectory.to_mdtraj()
        traj.to_trr(f'prefix_{index:03d}.trr')
```
However, here, something goes wrong with exporting the box vectors:
```
  File "/home/10433791/ops/openpathsampling/openpathsampling/engines/trajectory.py", line 564, in to_mdtraj
    box_vectors = self.box_vectors
  File "/home/10433791/ops/openpathsampling/openpathsampling/engines/trajectory.py", line 144, in __getattr__
    np.copyto(out[idx], getattr(s, item)._value)
AttributeError: 'numpy.ndarray' object has no attribute '_value'
```
[Now I recognize this error](https://github.com/openpathsampling/openpathsampling/issues/860#issue-500902728), and it has to do with the box vectors sometimes being a `simtk.Quantity` and otherwise being a (3,3,3) `numpy.ndarray`. 

The fix seems to be straightforward; when following the error to `openpathsampling/engines/trajectory.py`, `Trajectory.__getattr__()`, we see an explicit code block for these types in lines 135-171:
```
135:                 if type(first) is u.Quantity:
136:                     inner = first._value
137:                     if type(inner) is np.ndarray:
138:                         dtype = inner.dtype
139: 
140:                         out = np.empty(tuple([len(self)] +
141:                                              list(inner.shape)), dtype=dtype)
142:                         # print(f'inner:{inner}, item:{item}')
143:                         for idx, s in enumerate(list.__iter__(self)):
144:                             # print(f's.item:{getattr(s, item)}')
145:                             np.copyto(out[idx], getattr(s, item)._value)
146: 
147:                         return out * first.unit
148:                     else:
149:                         out = [None] * len(self)
150: 
151:                         for idx, s in enumerate(list.__iter__(self)):
152:                             out[idx] = getattr(s, item)
153: 
154:                         return out
155:                 elif type(first) is np.ndarray:
156:                     dtype = first.dtype
157: 
158:                     out = np.empty(tuple([len(self)] +
159:                                          list(first.shape)), dtype=dtype)
160: 
161:                     for idx, s in enumerate(list.__iter__(self)):
162:                         np.copyto(out[idx], getattr(s, item))
163: 
164:                     return out
165:                 else:
166:                     out = [None] * len(self)
167: 
168:                     for idx, s in enumerate(list.__iter__(self)):
169:                         out[idx] = getattr(s, item)
170: 
171:                     return out
```
And after removing the `._value` on line 144, which threw the error, my loop works fine and I'm able to export the rejected trajectories.

I'm not entirely sure why the original line didn't work, as it seems to me that `getattr(s, item)` is supposed to return a `Quantity` (line 135). However, with this fix the code on line 151 is mirrored (and it works) so it seems alright to me.

dwhswenson:
This simplifies `Trajectory.__getattr__` and may fix #878.

It adds a small restriction on snapshot "features" -- if you have a snapshot feature named `foo`, and if you create a function-type snapshot feature called `trajectory_foo`, then `trajectory_foo` must be a function that takes a trajectory and returns the values for `foo`. (I don't think this is going to cause any problems for anyone.)

I'm using those trajectory function features for the following: With the OpenMM engine, a trajectory returns a `simtk.unit.Quantity` wrapping a numpy array of the results of some snapshot feature, as opposed to making a list/numpy array containing multiple `Quantity` objects. I've now moved the code that supports this behavior into the OpenMM engine (as trajectory function features specific to it), instead of the base `Trajectory` object.

This allows us to remove more OpenMM-specific code that made `Trajectory.__getattr__` a much more complicated and hard-to-maintain function before.

### DIFF
--- a/openpathsampling/engines/features/base.py
+++ b/openpathsampling/engines/features/base.py
@@ -284,11 +284,14 @@ def attach_features(features, use_lazy_reversed=False):
                     if type(content) is str:
                         content = [content]
 
+                    def func_or_static(fnc):
+                        return callable(fnc) or (hasattr(fnc, "__func__")
+                                                 and callable(fnc.__func__))
                     if name == 'functions':
                         for c in content:
                             if hasattr(feature, c):
                                 fnc = getattr(feature, c)
-                                if callable(fnc):
+                                if func_or_static(fnc):
                                     if hasattr(cls, c):
                                         raise RuntimeWarning(
                                             'Collision: Function "%s" from feature %s already exists.' % (c, feature))
--- a/openpathsampling/engines/openmm/features/__init__.py
+++ b/openpathsampling/engines/openmm/features/__init__.py
@@ -8,3 +8,4 @@ else:
     from openpathsampling.engines.features.shared import StaticContainer, KineticContainer
     from . import masses
     from . import instantaneous_temperature
+    from . import traj_quantities
--- a/openpathsampling/engines/openmm/features/traj_quantities.py
+++ b/openpathsampling/engines/openmm/features/traj_quantities.py
@@ -13,20 +13,19 @@ def _trajectory_units(traj, feature, unit_):
         return obj
 
     vals = [getattr(snap, feature) for snap in traj]
-    print(feature, vals)
-    result = [strip_unit(val) for val in vals]
+    result = np.asarray([strip_unit(val) for val in vals])
     non_none = any([x is not None for x in result])
     if non_none:
         result = result * unit_
-    print(result)
     return result
 
 def trajectory_coordinates(traj):
-    _trajectory_units(traj, 'coordinates', unit.nanometer)
+    return _trajectory_units(traj, 'coordinates', unit.nanometer)
 
 def trajectory_box_vectors(traj):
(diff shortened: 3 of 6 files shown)


dwhswenson:
I'll merge this in 48 hours (Sunday 8 December) if no comments. I'll assume that it does fix the problem in #878 that @IdsTeepe raised.

PR #881, opened 2019-11-25
fix docstring Deprecation warning

sroet:
Since the update to the docstrings of the toy pes I got a couple of: `openpathsampling/openpathsampling/engines/toy/pes.py:198: DeprecationWarning: invalid escape sequence \s
    """` on python `3.6`

This MR fixes that issue(by making them raw strings).
It also makes sure the math is actually compiled through latex for the website, and some pep fixes.

### DIFF
--- a/openpathsampling/engines/toy/pes.py
+++ b/openpathsampling/engines/toy/pes.py
@@ -31,7 +31,8 @@ class PES(StorableObject):
         """
         v = sys.velocities
         m = sys.mass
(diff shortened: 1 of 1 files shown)


dwhswenson:
Actually, I guess it did (barely) touch the docstrings, but I still would be surprised that removing spaces so that a blank line is actually blank (i.e., just "\n") causes the warnings to pop up.

PR #882, opened 2019-12-07
Add progress meter adapter classes

dwhswenson:
One of the annoying things when working with OPS is that sometimes analysis takes a while, and there's no indicator of how long it will take. This PR starts to add optional progress indicators (using `tqdm`) to our analysis tools, by creating a mix-in class that allows users to set the `progress` attribute. Classes that use the mix-in should not define either a `progress` attribute or a `_progress` attribute.

The code here doesn't require tqdm, but uses it if it is present on the user's system (which, I think, it almost always is).

Really, these are adapter classes and we could interface with other progress meters as well, not just tqdm. Also, although I'm mainly focused on analysis, there's no reason (in principle) that this couldn't be used for simulations as well.

Eventually we'll move any long-running analysis tools toward using this, but note that you can already just wrap an iterable (like `storage.steps`) with tqdm (even without this PR) and you'll usually get the desired behavior. There are a few exceptions, where the iterable is pre-processed before the main analysis is performed.

(I actually wrote most of this code months ago, but in the process I found and fixed a bug in tqdm. I was waiting until that bugfix made it into a tqdm release, and in the meantime, completely forgot about this.)

### DIFF
--- a/openpathsampling/progress.py
+++ b/openpathsampling/progress.py
@@ -1,0 +1,95 @@
+"""
+Wrappers for progress bars. Mainly intended to use tqdm if it is available
+on the user's system, but in principle, other progress bars could be used as
+well.
+"""
+
+try:
+    from tqdm.auto import tqdm
+except ImportError:  # pragma: no cover
+    HAS_TQDM = False
+else:  # pragma: no cover
+    HAS_TQDM = True
+
+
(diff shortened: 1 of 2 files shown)


dwhswenson:
I'll leave this up for at least 48 hours for comment/review, not merging before the evening of Sun 15 Dec.

PR #883, opened 2019-12-10
Use Autorelease for version/release management

dwhswenson:
This PR will move OPS over to the [Autorelease](http://github.com/dwhswenson/autorelease) workflow that I've been using with several other projects for the last few years. This will facilitate PyPI releases starting with 1.1.

One of the main ideas of Autorelease is that branches are either "release" or "development" — release branches should only include releases. We'll call our release branch `stable`; our primary development branch is `master`. 

The automated release process in Autorelease 0.1 (which I cleaned up last week) is particularly useful. I'd been copying the relevant Travis configs into each project, but last month Travis made it possible to import configs from elsewhere. So now there's a single location for all the extra Travis stuff. The automated release process works like this:

* Travis-CI config imports from Autorelease handle PRs to `stable` differently than other branches. When you open a PR to `stable`, a test release is sent to testpypi; that test release is then redownloaded and the test suite is run, to ensure that everything that needed to be in the package *is* in the package.
* When you merge the PR to `stable`, the Travis-CI config imports in Autorelease will automatically cut a GitHub release, using the PR text as the release notes. A GitHub release creates a new tag in the repo; when Travis sees that new tag, Autorelease will again run tests (so that you have a pip-freeze type of record of the requirement version this should work with) and then deploys to PyPI.
* Autorelease includes testing tools to ensure that you don't accidentally release a development version or leave a release version number on the development branch. These tools also check for things like ensuring you don't accidentally skip a version, and if your version appears in multiple places (as ours does), that all the versions agree.

From a day-to-day perspective, the release process is as easy as updating the version number (removing the `.dev0`) and making a PR against the `stable` branch. Everything else is automatic. Once the release is out, make a branch off of `stable`, bump the version, and merge into `master` (this approach minimizes likelihood of merge conflicts on the version line, although that's an easy conflict to resolve.)

In addition, Autorelease provides a few scripts to vendor into our repo:

* smart `version.py` that keeps up-to-date versions for developer installs, while allowing a single location to store a "one true version"
* `setup.py` that is independent of the project itself. All project information is stored in `setup.cfg`. Only `setuptools` is required for installation.

I've already done Autorelease 0.1.0-based releases on two projects, and prepared a 3rd to use it the next time it releases. I think this is ready for OPS (although it may take a few commits to make OPS ready for it!)


dwhswenson:
Seems to work! Of course, the real tests of whether it works will come when we do the next release (hopefully later this week). Note that I've also switched our nightly build from the external service nightli.es to Travis's built-in cron support (which didn't exist when we first started doing nightlies). On `stable`, this will be necessary to ensure that the Autorelease configs can tell the difference between a cron job, a merge, and a PR.

I'll leave this up for at least 24 hours for comment/review, and will merge on Wednesday 11 December if no objections. Note that much of the code added here (`setup.py` and `version.py`) are vendored from Autorelease. Essentially, that means that the effort for maintaining them should be in Autorelease, not here.

PR #884, opened 2019-12-10
Add PathSampling.run_until_decorrelated

dwhswenson:
This adds a function to the `PathSampling` path simulator to tell it to run until all trajectories in the initial sample set are decorrelated (in the one-way shooting sense of "not having any configurations in common with the original trajectories.")

This definition of decorrelation is an absolute bare minimum requirement for equilibration, and so you could imagine equilibration strategies such as "run twice as long as it takes to get a decorrelated trajectory" or "run as long as it takes to get a decorrelated trajectory, plus N additional steps." This function will facilitate that.

### DIFF
--- a/openpathsampling/pathsimulators/path_sampling.py
+++ b/openpathsampling/pathsimulators/path_sampling.py
@@ -1,5 +1,6 @@
 import time
 import logging
+import os
 
 import openpathsampling as paths
 from .path_simulator import PathSimulator, MCStep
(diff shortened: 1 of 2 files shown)


dwhswenson:
This was intended for 1.2 ... now that 1.1 is out, we can start merging in things for 1.2! This is a very simple PR, so I'll leave it up for review/comment at least another 24 hours, not merging before the evening of Sat 14 Dec.

PR #894, opened 2020-01-26
`run_until_decorrelated` should account for time reversibility

dwhswenson:
PR #884 added a `run_until_decorrelated` function to the `PathSampling` path simulator. However, that function didn't use the decorrelated trajectory check that accounts for time reversibility (i.e., it allowed a path reversal to count as decorrelating a trajectory). Since this is intended for equilibration, that doesn't make sense: it should require that the frame be completely replaced (since initial frames may come from nonphysical trajectories).

This fixes that by default, providing `run_until_decorrelated` with a `time_reversal` argument like that in `Trajectory.is_decorrelated`. In this case, that argument default to `True`.

This also updates so that the output for the final step makes it clear that all trajectories have decorrelated, and switches the testing to use a TIS network with path reversals possible.

### DIFF
--- a/openpathsampling/pathsimulators/path_sampling.py
+++ b/openpathsampling/pathsimulators/path_sampling.py
@@ -187,7 +187,7 @@ class PathSampling(PathSimulator):
         n_steps_to_run = n_steps - self.step
         self.run(n_steps_to_run)
 
-    def run_until_decorrelated(self):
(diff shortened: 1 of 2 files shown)


dwhswenson:
I'll merge this in tomorrow (Tues 28 Jan) if there are no comments.

PR #895, opened 2020-01-29
Progress indicators for TIS analysis

dwhswenson:
#882 introduced adapters for progress indicators (namely, tqdm). This PR integrates that into the "new" TIS analysis. (The "new" analysis, which is about 2 years old, is described in [this document](http://openpathsampling.org/latest/topics/tis_analysis.html) and [this example notebook](https://github.com/openpathsampling/openpathsampling/blob/master/examples/toy_model_mstis/toy_mstis_A3_new_analysis.ipynb). It isn't the default yet because it isn't compatible with files created by very old version of OPS -- I think pre 0.9.3 -- which we use for our tests.)

If you have tqdm installed, you'll automatically get progress bars. You can also customize the behavior, including easily silencing the progress bars.

The goal of this PR is to get these fully working for the `StandardTISAnalysis`, along with conveniences for changing some behavior.


dwhswenson:
This is ready for review/merge. I will leave it up for at least 24 hours, merging no earlier than Sunday night, 9 February.

PR #896, opened 2020-01-30
Fix pandas API change: set_value -> at

dwhswenson:
Pandas 1.0 has been released! And long-deprecated the `.set_value` method for setting dataframe values has been removed. This fixes the places where we were using that (in TIS rate matrix calculations).


dwhswenson:
Merging this immediately, since it is a small change and CI fails until this gets merged.

PR #901, opened 2020-02-15
Add GromacsEngine to docs

dwhswenson:
We've had the Gromacs engine working in OPS since version 1.1, and the docstrings were in place and there was an example in the `examples/` directory. However, none of this stuff had made it to the website. This PR puts it there.

While I'm at it, I'm going to try to clean up a few complaints that I see coming from sphinx.


dwhswenson:
This is ready for review. Only involves docs. Will leave up for 24 hours, merging no earlier than Thurs 20 Feb.

PR #903, opened 2020-02-23
storage.filename is now an absolute path and a read-only property

### LINKED ISSUE #774: Some Storage properties break when changing working directory
The following example code produces an error at the second print statement:
```python
import os

storage = paths.Storage('test.nc', 'w')
os.chdir('..')
print(storage.filename, storage.filepath())
print(storage.file_size)
```
Error: 
```
test.nc test.nc

FileNotFoundError                         Traceback (most recent call last)
<ipython-input-37-628e081a85fb> in <module>()
      5 os.chdir('..')
      6 print(storage.filename, storage.filepath())
----> 7 print(storage.file_size)

~/Documents/Master/OPS/openpathsampling/openpathsampling/netcdfplus/netcdfplus.py in file_size(self)
    355     @property
    356     def file_size(self):
--> 357         return os.path.getsize(self.filename)
    358 
    359     @property

~/conda-envs/NNOPS/lib/python3.6/genericpath.py in getsize(filename)
     48 def getsize(filename):
     49     """Return the size of a file, reported by os.stat()."""
---> 50     return os.stat(filename).st_size
     51 
     52 

FileNotFoundError: [Errno 2] No such file or directory: 'test.nc'
```

I think the reason is that `storage.filename` returns a relative path, as does the netCDF function `storage.file_path()`.
Could it be a solution to use `self.filename = os.path.abspath(filename)` in https://github.com/openpathsampling/openpathsampling/blob/692ee2c88beb5d38b69fac6c6ac97ed8610aff08/openpathsampling/netcdfplus/netcdfplus.py#L237 ?
Or maybe add an `self.abspath = os.path.abspath(filename)` ?

hejung:
All in the title.
Closes #774

### DIFF
--- a/openpathsampling/netcdfplus/netcdfplus.py
+++ b/openpathsampling/netcdfplus/netcdfplus.py
@@ -246,7 +246,7 @@ class NetCDFPlus(netCDF4.Dataset):
                 "Open existing netCDF file '%s' for reading - "
                 "reading from existing file", filename)
 
-        self.filename = filename
+        self._filename = os.path.abspath(filename)
         self.fallback = fallback
 
         # this can be set to false to re-store objects present in the fallback
@@ -367,6 +367,10 @@ class NetCDFPlus(netCDF4.Dataset):
 
     def _create_simplifier(self):
         self.simplifier = UUIDObjectJSON(self)
+
+    @property
+    def filename(self):
+        return self._filename
 
     @property
     def file_size(self):


dwhswenson:
LGTM. Since `os.path.abspath` also normalizes (i.e., among other things, turns Posix-style paths into Windows-style paths on Windows), I think this also brings us much of the way to Windows support.

IIRC, at one time OpenMM wasn't playing nicely on Windows, so we never tried to support Windows. But I think that Windows support for us is literally a matter of wrapping filenames (mostly in tests) in platform-independent ways.

PR #905, opened 2020-03-04
DirectSimulation should use engine.start/.stop

dwhswenson:
From an email report from a user:

> I tried to follow this example: 
>
> http://openpathsampling.org/latest/examples/mistis.html#toy-model-mistis
>
> to set up openpathsampling.DirectSimulation for my flux_pairs, but got this error: 
> 
> ```python
> AttributeError: 'GromacsEngine' has no attribute 'n_frames_since_start', nor does its options dictionary
> ```

Two problems, both solved in this PR:

1. `ExternalEngine.n_frames_since_start` is only set in its `.start()` method. It should be set (to `None`) in `__init__`.
2. `DirectSimulation` is not using the `.start()` and `.stop()` methods. This didn't matter for OpenMM/toy engine, but does for external engines like Gromacs.

NOTE: I would not recommend using `DirectSimulation` on external engines. `DirectSimulation` is mainly useful to get fluxes from toy engine runs or other simulations where you don't want to save trajectory frames. It's better to run a trajectory normally and use `TrajectoryTransitionAnalysis`.

### DIFF
--- a/openpathsampling/engines/external_engine.py
+++ b/openpathsampling/engines/external_engine.py
@@ -78,6 +78,7 @@ class ExternalEngine(DynamicsEngine):
         self.first_frame_in_file = first_frame_in_file
         self._traj_num = -1
         self._current_snapshot = template
+        self.n_frames_since_start = None
 
     @property
     def current_snapshot(self):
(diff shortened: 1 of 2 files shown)


dwhswenson:
I will leave this open for another 24 hours for comment; merging no earlier than Friday afternoon, 5 March.

Issue #907, opened 2020-03-09
With Gromacs Openpathsampling unregular errors in snapshot.py


jasperdenduijf:
So with Gromacs I ran an Openpathsampling simulation and created 30 trajectories. They are all stored neatly in an netCDF file called `file15.nc`. I wanted to analyse them, so I have a very simple snippet of code:

```
flexible = paths.Storage("file15.nc", mode = "r") #open storage
d_WC = flexible.cvs["d_WC"] #796  for certain cv
plt.plot(d_WC(flexible.trajectories[**20**])) #plt is matplotlib
```
I get an error when I check certain trajectories like 20, but not 6 or 29. It has nothing to do with rejected trajectories, I plotted a svg tree without a problem.
The error goes as follows:

```
IndexError                                Traceback (most recent call last)
~/miniconda3/lib/python3.6/site-packages/openpathsampling/engines/external_snapshots/snapshot.py in load_details(self)
     49                 self.file_name,
---> 50                 self.file_position
     51             )

~/miniconda3/lib/python3.6/site-packages/openpathsampling/engines/gromacs/engine.py in read_frame_data(self, filename, frame_num)
    199         try:
--> 200             f.seek(offset=frame_num)
    201             data = f._read(n_frames=1, atom_indices=None, get_velocities=True)

mdtraj/formats/xtc/trr.pyx in trr.TRRTrajectoryFile.seek (mdtraj/formats/xtc/trr.c:11389)()

IndexError: index 152 is out of bounds for axis 0 with size 66

During handling of the above exception, another exception occurred:

NameError                                 Traceback (most recent call last)
<ipython-input-4-d20e77c8f2e0> in <module>
----> 1 plt.plot(d_WC(flexible.trajectories[20]), color='g')
      2 plt.plot(d_WC(flexible.trajectories[0]))

~/miniconda3/lib/python3.6/site-packages/openpathsampling/netcdfplus/chaindict.py in __getitem__(self, items)
    227 
    228     def __getitem__(self, items):
--> 229         return self._post[items]
    230 
    231     __call__ = __getitem__

~/miniconda3/lib/python3.6/site-packages/openpathsampling/netcdfplus/chaindict.py in __getitem__(self, items)
    240 
    241     def __getitem__(self, items):
--> 242         return np.array(self._post[items])
    243 
    244     __call__ = __getitem__

[...]

PR #908, opened 2020-03-09
Fix missing names in ExternalMDSnapshot

dwhswenson:
`ExternalMDSnapshot` was moved to a different file, and didn't get all its module-level names redefined. Bug uncovered by @jasperdenduijf in #907.

~~Right now, this still requires tests~~ (tests added!). We're missing coverage on the lines that caused that part of the issue in #907, so we should have tests covering those lines.

### DIFF
--- a/openpathsampling/tests/test_external_snapshots.py
+++ b/openpathsampling/tests/test_external_snapshots.py
@@ -73,7 +73,6 @@ class TestExternalMDSnapshot(object):
 
     def test_load_details_indexerror(self):
         engine = ErrorMockEngine(self.engine.sequences)
-        engine.accessed
         snap = ExternalMDSnapshot(file_name='foo',
                                   file_position=0,
                                   engine=engine)
--- a/openpathsampling/tests/test_external_snapshots.py
+++ b/openpathsampling/tests/test_external_snapshots.py
@@ -1,0 +1,134 @@
+import pytest
+
+import openpathsampling as paths
+import numpy as np
+
+from openpathsampling.engines.external_snapshots.snapshot import (
+    ExternalMDSnapshot
+)
+
+class MockEngine(object):
+    def __init__(self, sequences, sleep_ms=0):
+        self.sequences = sequences
+        self.sleep_ms = sleep_ms
+
+    def read_frame_data(self, filename, position):
+        return self.sequences[filename][position]
+
+
+class ErrorMockEngine(MockEngine):
+    """Mock engine used to create the IndexError in load_details"""
+    def __init__(self, sequences, sleep_ms=0):
+        super(ErrorMockEngine, self).__init__(sequences, sleep_ms)
+        self._sequences = self.sequences
+        self.sequences = {k: [] for k in self.sequences}
+        self.accessed = False
+
+    def read_frame_data(self, filename, position):
+        if self.accessed:
+            self.sequences = self._sequences
+
+        self.accessed = True
+        return super(ErrorMockEngine, self).read_frame_data(filename,
+                                                            position)
+
+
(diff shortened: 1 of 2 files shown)


sefalkner:
Tested this on my Gromacs TPS setup that also had these occasional Errors (#907) after some 10-30 mc cycles, now runs smooth for > 300 cycles.

In general I thought about contributing to the docs for the gromacs engine. For a new user it is unclear that setting n_frames_max to the gromacs nsteps option is crucial as there is little documentation about it.

dwhswenson:
> Tested this on my Gromacs TPS setup that also had these occasional Errors (#907) after some 10-30 mc cycles, now runs smooth for > 300 cycles.

Thanks for the confirmation that this fixes -- I'll make finishing this PR a priority, but things are a little bit busy for the next week. (Out-of-town guests.)

> In general I thought about contributing to the docs for the gromacs engine. For a new user it is unclear that setting n_frames_max to the gromacs nsteps option is crucial as there is little documentation about it.

Please do! Improvements to the docs are always welcome. External engines (such as Gromacs) are newer, and neither the code nor the docs have had as much stress testing, so the help is appreciated.

That said, you might wait until, say, Monday. I've been working on a branch to improve a few things in the Gromacs engine docs -- docstrings and, especially, the example notebooks -- so it might be easier to build on that than to write (possibly conflicting) changes. I'll try to make a PR of that very soon.

jasperdenduijf:
This might be an obvious question, but once I've installed openpathsampling, how do I pull these commits and add them to my current PATH?

dwhswenson:
@jasperdenduijf : I think most of what you want is described at http://openpathsampling.org/latest/developers/install_developers.html (after getting the dev install script with the curl command at http://openpathsampling.org/latest/install.html#developer-install-with-conda). The quickest way to do a bugfix test install is under the heading "Quick bugfix/developer installation" in the first link. That creates a separate conda environment for the bugfix version of OPS. For this specific branch, you'd use remote `dwhswenson` and branch `fix_ext_snapshot_missing` (in place of `experimental_feature` in the docs).

Assuming the tests that I just added pass, then I expect this will be merged to master tomorrow evening or Saturday morning. It will be included in the 1.3 release, which I hope to release early next week.

dwhswenson:
Tests pass; I'll leave this open for comment for at least 24 hours before merging. I will merge no earlier than Friday 20 March, 14:00 GMT (15:00 my local).

This PR involves adding some missing imports and then adding tests to improve our test coverage, so I can't imagine anything here is controversial.

openpathsampling/tests/test_external_snapshots.py
```diff
@@ -0,0 +1,134 @@
...
+            snap.load_details()
+            expected_xyz = np.array([[0.1 * i, 0.0]])
+            expected_vel = np.array([[1.0, 0.0]])
+            np.testing.assert_array_equal(snap._xyz, expected_xyz)
+            np.testing.assert_array_equal(snap._velocities, expected_vel)
+            np.testing.assert_array_equal(snap._box_vectors, self.box)
+
+    def test_load_details_indexerror(self):
+        engine = ErrorMockEngine(self.engine.sequences)
+        engine.accessed
```
sroet: why is this line here? 
just to test that the attribute is there? 
or is it leftover debug code?

dwhswenson: Thanks  for catching this: It was there because `ErrorMockEngine` had an `__init_` method instead of an `__init__` method, and it took me an embarrassingly long time to track down the missing underscore. This was part of that process. Removed!

sroet: ah, classic :laughing:

PR #909, opened 2020-03-13
Gromacs engine/example updates

dwhswenson:
Main improvements:

* Major rewrite of the Gromacs example notebooks to provide a lot more explanation, and to use newer, more convenient approaches within OPS
* The grompp and mdrun command strings are now class variables, making them more easily accessible from outside. The commands to create the fully formatted strings have not changed (that's the API).
* The default values for filenames (before file number is set) are more informative, and include the directories where things end up.
* Gromacs tests are better at cleaning up after themselves
* Added logging conf file to facilitate engine debugging


dwhswenson:
I'll leave this up for comment/review for at least 48 hours, merging no earlier than evening of Sun 15 March (European time).

PR #911, opened 2020-03-31
"Hooks" for PathSampling

hejung:
This builds on #755 and should only be merged after that.

It adds hook support to the `PathSampling`, the `PathSamplingOutputHook` and the `LiveVisualizerHook` including tests.

Please let me know if htere is anything to improve!
I also have some open questions/TODOs, in particular:
1. if a user sets any attribute that controls the behaviour of the hook on the `PathSimulator` object it is only read once at the start of the simulation and only if the hooks value for that attribute is `None` (see e.g. `output_stream`, `save_frequency` and `storage`). This is potentially confusing, as I would expect to be able to set these attributes on the `PathSimulator` to influence the behaviour of the currently attached hooks. I am mot really sure how to adress this. Potential solutions include removing these attributes from the `PathSimulator` and only set them on hook creation. Another option is to make them properties to either (re)set the respective attributes on the hooks (or at least warn the user that it could not work as intended) if he/she sets the attributes on the `Pathsimulator`

2. Previously the LiveVisualizer was called before the MCstep but visualizing the previous step. I moved the LiveVisualization to after step. This enables visualization of the step that just finished and avoids that the beginning of the next step depends on the previous one. But it also means that visualization only works if the serial OutputHook has `allow_refresh=False`, otherwise the plot will get refreshed away.

3. The `step_info` passed to the `after_step` hooks from the `PathSampling` is specifically tailored for the `PathSamplingOutputHook`, i.e. it only takes the variables the OutputHook needs (current step, total steps, total elapsed time). @dwhswenson any ideas what else we could want/need in there in the future?

### DIFF
--- a/openpathsampling/beta/hooks.py
+++ b/openpathsampling/beta/hooks.py
@@ -106,7 +106,10 @@ class StorageHook(PathSimulatorHook):
     def after_step(self, sim, step_number, step_info, state, results,
                    hook_state):
         if self.storage is not None:
-            self.storage.save(results)
+            try:
+                self.storage.stash(results)
+            except AttributeError:
+                self.storage.save(results)
             if step_number % self.frequency == 0:
                 self.storage.sync_all()
 
--- a/openpathsampling/beta/hooks.py
+++ b/openpathsampling/beta/hooks.py
@@ -41,19 +41,64 @@ class PathSimulatorHook(StorableNamedObject):
 
 
 class StorageHook(PathSimulatorHook):
-    """Standard hook for storage.
+    """
+    Standard hook for storage.
+
+    NOTE: Arguments passed to init take precedence over the corresponding
+          parameters of the PathSimulator this hook is attached to. They can
+          only be accessed through this hook, e.g. as hook.live_visualizer.
+
+    Parameters
+    ----------
+    storage : :class:`.Storage
+              where to save to; default ``None`` uses the simulation's
+              ``storage``
+    frequency : int
+                save frequency measured in steps; default ``None`` uses the
+                simulation's value for ``save_frequency``
+
     """
     implemented_for = ['before_simulation', 'after_step',
                        'after_simulation']
     def __init__(self, storage=None, frequency=None):
         self.storage = storage
         self.frequency = frequency
+        self._simulation = None
+
+    @property
+    def frequency(self):
+        if self._frequency is not None:
+            return self._frequency
+        elif self._simulation is not None:
+            return self._simulation.save_frequency
+        else:
+            raise SimulationNotFoundError("'frequency' has not "
+                                          + "been set and no hosting "
+                                          + "simulation known to get a value."
+                                          )
+
+    @frequency.setter
+    def frequency(self, val):
+        self._frequency = val
+
+    @property
+    def storage(self):
+        if self._storage is not None:
+            return self._storage
+        elif self._simulation is not None:
+            return self._simulation.storage
+        else:
+            raise SimulationNotFoundError("'storage' has not "
+                                          + "been set and no hosting "
+                                          + "simulation known to get a value."
+                                          )
+
+    @storage.setter
+    def storage(self, val):
+        self._storage = val
 
     def before_simulation(self, sim, **kwargs):
-        if self.storage is None:
-            self.storage = sim.storage
-        if self.frequency is None:
-            self.frequency = sim.save_frequency
+        self._simulation = sim
 
     def after_step(self, sim, step_number, step_info, state, results,
                    hook_state):
@@ -76,6 +121,10 @@ class ShootFromSnapshotsOutputHook(PathSimulatorHook):
 
     Updates every time a new snapshot is shot from.
 
+    NOTE: Arguments passed to init take precedence over the corresponding
+          parameters of the PathSimulator this hook is attached to. They can
+          only be accessed through this hook, e.g. as hook.live_visualizer.
+
     Parameters
     ----------
     output_stream : stream
@@ -89,12 +138,42 @@ class ShootFromSnapshotsOutputHook(PathSimulatorHook):
     def __init__(self, output_stream=None, allow_refresh=None):
         self.output_stream = output_stream
         self.allow_refresh = allow_refresh
+        self._simulation = None
+
+    @property
+    def output_stream(self):
+        if self._output_stream is not None:
+            return self._output_stream
+        elif self._simulation is not None:
+            return self._simulation.output_stream
+        else:
+            raise SimulationNotFoundError("'output_stream' has not "
+                                          + "been set and no hosting "
+                                          + "simulation known to get a value."
+                                          )
+
+    @output_stream.setter
+    def output_stream(self, val):
+        self._output_stream = val
+
+    @property
+    def allow_refresh(self):
+        if self._allow_refresh is not None:
+            return self._allow_refresh
+        elif self._simulation is not None:
+            return self._simulation.allow_refresh
+        else:
+            raise SimulationNotFoundError("'allow_refresh' has not "
+                                          + "been set and no hosting "
+                                          + "simulation known to get a value."
+                                          )
+
+    @allow_refresh.setter
+    def allow_refresh(self, val):
+        self._allow_refresh = val
 
     def before_simulation(self, sim, **kwargs):
-        if self.output_stream is None:
-            self.output_stream = sim.output_stream
-        if self.allow_refresh is None:
-            self.allow_refresh  = sim.allow_refresh
+        self._simulation = sim
 
     def before_step(self, sim, step_number, step_info, state):
         snap_num, n_snapshots, step, n_per_snapshot = step_info
--- a/openpathsampling/beta/hooks.py
+++ b/openpathsampling/beta/hooks.py
@@ -17,6 +17,24 @@ class SimulationNotFoundError(RuntimeError):
     pass
 
 
+def _self_or_sim_property_or_err(obj, prop_name, sim_prop_name=None):
+    # helper function to get values either from the hook or the parent sim
+    if sim_prop_name is None:
+        # make it possible to have different names for the sim and
+        # self attributes (but default to same name)
+        sim_prop_name = prop_name
+    if getattr(obj, "_" + prop_name) is not None:
+        return getattr(obj, "_" + prop_name)
+    elif obj._simulation is not None:
+        return getattr(obj._simulation, sim_prop_name)
+    else:
+        raise SimulationNotFoundError("'" + prop_name + "' has not "
+                                      + "been set and no hosting "
+                                      + "simulation known to get "
+                                      + "simulation." + sim_prop_name + " ."
+                                      )
+
+
 class PathSimulatorHook(StorableNamedObject):
     """Superclass for PathSimulator hooks.
 
@@ -50,7 +68,7 @@ class StorageHook(PathSimulatorHook):
 
     Parameters
     ----------
-    storage : :class:`.Storage
+    storage : :class:`.Storage`
               where to save to; default ``None`` uses the simulation's
               ``storage``
     frequency : int
@@ -68,15 +86,7 @@ class StorageHook(PathSimulatorHook):
 
     @property
     def frequency(self):
-        if self._frequency is not None:
-            return self._frequency
-        elif self._simulation is not None:
-            return self._simulation.save_frequency
-        else:
-            raise SimulationNotFoundError("'frequency' has not "
-                                          + "been set and no hosting "
-                                          + "simulation known to get a value."
-                                          )
+        return _self_or_sim_property_or_err(self, "frequency", "save_frequency")
 
     @frequency.setter
     def frequency(self, val):
@@ -84,15 +94,7 @@ class StorageHook(PathSimulatorHook):
 
     @property
     def storage(self):
-        if self._storage is not None:
-            return self._storage
-        elif self._simulation is not None:
-            return self._simulation.storage
-        else:
-            raise SimulationNotFoundError("'storage' has not "
-                                          + "been set and no hosting "
-                                          + "simulation known to get a value."
-                                          )
+        return _self_or_sim_property_or_err(self, "storage")
 
     @storage.setter
     def storage(self, val):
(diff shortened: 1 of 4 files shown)


dwhswenson:
Great to see this! Sorry that I'm not going to be very responsive in the next week or two. Somehow I'm significantly busier on lockdown than at normal times.

The current test failure is a known issue with the most recent conda-forge release of svgwrite. It should be fixed very soon -- in principle it is fixed; needs me to reply to someone though (one of the many things on my list).

Quick comments on your questions/todos:

> 1. [RE: `output_stream`, `save_frequency`] I would expect to be able to set these attributes on the `PathSimulator` to influence the behaviour of the currently attached hooks.

This is a good point. I'm being quick here, so I'm not even checking the code, but what should probably happen is to treat the `PathSimulator` attributes as a default fallback: (1) By default, hooks don't set this; instead they get it from the PathSimulator (every hook knows its simulator, right?). This means if the user specifically changes it in `simulator` and never set it in `hooks`, it behaves as expected. (2) If a user sets it in the hook and in the simulator, I think the hook takes precedence.

What should not happen is, hook is added, implicitly takes the simulator's value. Then user changes simulator's value, and it has no effect. This would be a bug.

> 2. I moved the LiveVisualization to after step.

This seems fine; but is there also a way to ensure that the 0th step is visualized? (Maybe a `before_step` hook and also has some internal state so it only gets run once?) I'm not sure the current behavior, but it would make sense to show the initial conditions if you're running a simulation that takes a while to do the first step.

The caveat on refreshing output is completely fine. There's no way around that, so as long as the docstrings are clear on the topic, all good.

> 3. The `step_info` passed to the `after_step` hooks from the `PathSampling` is specifically tailored for the `PathSamplingOutputHook`, i.e. it only takes the variables the OutputHook needs (current step, total steps, total elapsed time)

Good question. As I recall, the idea was for `step_info` to carry all metadata that might possibly be of interest to a hook. For committors, I believe there were 4 things: snapshot number, total number of snapshots, step number (inner loop), and total number of steps per inner loop.

Is the time info actually used by anything other than the output hook? You can probably move the time info into the hook -- get initial time at `before_simulation`, then update at each stage.

hejung:
No problem. 

Concerning the points:
1.  Atm it is exactly as you said, if it is set in the hook the hook takes precedence otherwise the hook takes the defaults from the `PathSimulator`.  The only situation in which I see potential trouble is the following:
   First user adds implicit default hooks (without value). Then user runs a round of sampling (i.e.   `before_simulation` is executed and the hooks get their implicit defaults from the `PathSimulator`). If the user now changes the attributes on the `PathSimulator` and runs another round of sampling the hooks  will not update, because they already got their (default) values when `before_simulation` was ran the first time.
  In addition to making these attributes properties or removing them from the `PathSimulator` (and  moving them to the hooks altogether), I think another solution would be to add a state-flag to the hook: If hook is initialized with values it should always keep them. If it was initalized without it should always take the values from the associated `PathSimulator` when `before_simulation` is run. What do you think?

2. Previously the first step was also not visualized. Currently the `StepVisualizer2D` needs/uses the MC change to visualize. It was just that the visualization for step n was ran at the start of the next iteration, i.e. directly before step (n+1). But I will look into how/if it is possible to visualize the first step too.

3. No, at the moment the timing info is not used anywhere else. I could move the timing to the hook, but the `PathSamplingOutputHook` still needs the total number of snapshots and the current number to be able to calculate the progress string, this would then be similar to the committors, where we also have total iterations and loop counter.
 Something else where the time info could be useful is something I have been thinking about: Maybe it would be useful to write a "gracious kill" hook to make ops work nicer with the time limits on queuing systems. This would be hook that knows the maximum walltime and simply checks every step if we expect to finish the next step given the time the previous steps took on average. If not it would sync the storage and exit the simulation loop.

dwhswenson:
First off, the error you have now should be fixed as soon as you merge in master. It's a problem with the miniconda install (they moves the web page we use for checksums).

Point-by-point:

> Atm it is exactly as you said, if it is set in the hook the hook takes precedence otherwise the hook takes the defaults from the `PathSimulator`.

Let me sketch out in code what I meant, which I think will address the issue. Let's pretend that the simulator has a property called `frequency`. The hook *should* do something like:

```python
class MyHook(PathSimulatorHook):
    def __init__(self, frequency=None):
        self._simulation = None
        self._frequency = frequency

    def before_simulation(self, sim):
        self._simulation = sim

    @property
    def frequency(self):
        if self._freq_val is not None:
            return self._frequency
        elif self._simulation is not None:
            return self._simulation.frequency
        else:
            raise RuntimeError("`frequency` has not been set")  # maybe return None?
```

That isn't how I initially implemented it, but I think it (mostly) makes more sense. It solves the problem you're thinking of, right? If you ever set it in the hook, then the hook's value is used (even if you set it in the simulator later). If you haven't set it in the hook, the simulation's value is used.

The only question is what to return when `frequency` isn't given on initialization and if the simulation is still not known (hasn't started yet). I'm leaning toward an error (probably best to actually subclass `RuntimeError` with a specialized error case), but returning `None` is also a reasonable option.

>  I will look into how/if it is possible to visualize the first step too.

If the first step can be visualized, great, but no big deal if it can't easily be done.

> I could move the timing to the hook, but the `PathSamplingOutputHook` still needs the total number of snapshots and the current number to be able to calculate the progress string, this would then be similar to the committors, where we also have total iterations and loop counter.

I'd say move timing into the hook. For any other hook that needs it, I'm not worried about the overhead of calling `time.time()` once per MC step.

> Maybe it would be useful to write a "gracious kill" hook to make ops work nicer with the time limits on queuing systems.

I think this sounds like a very good idea. In the long run, I want to have mid-trajectory checkpointing, because there are cases where a single trajectory might be as long as your allowed wall time. However, this would be a very useful improvement in the meantime.

If you implement this, please do a separate PR for it. It deserves a separate mention in the release notes.

hejung:
Yes, that solves the problem I had in mind. I went ahead and implemented it for all hooks.

I moved the timing to the PathSamplingOutputHook, which has the added benefit that the test now also test the actual timing (and not only that we are able to print the correct string given the numbers).

hejung:
@dwhswenson I would say this is ready for review. I cleaned up a bit and added the `run_until_n_accepted` method you suggested.

hejung:
@dwhswenson Not morning anymore, but the branch is rebased on the current master and updated. But it looks like I will have to write some more test tomorrow... :)

openpathsampling/beta/hooks.py
```diff
@@ -78,12 +143,42 @@ class ShootFromSnapshotsOutputHook(PathSimulatorHook):
...
+        self._simulation = None
+
+    @property
+    def output_stream(self):
+        if self._output_stream is not None:
+            return self._output_stream
+        elif self._simulation is not None:
+            return self._simulation.output_stream
+        else:
+            raise SimulationNotFoundError("'output_stream' has not "
```
dwhswenson, line 155: (Starting to glance over it, but a thought before you spend much time on remaining tests.) You might to create something like

```python
def _self_or_sim_property(obj, attr_name):
    if getattr(obj, "_" + attr_name) is not None:
        return getattr(obj, "_" + attr_name)
    ...  # etc.
```

then use it as

```python
    @property
    def output_stream(self):
        return _self_or_sim_property(self, 'output_stream')
```

This would allow you to test the underlying logic once, and not need to worry about creating multiple tests for each property. More complete integration test are also welcome, but this definitely seems like a motif that is being (and will continue to be) repeated.

dwhswenson: (You could add a `return_none=False` to that private method to make it return `None` instead of erroring when that it appropriate)

dwhswenson (changes requested):
Mostly looks good, and thanks for the contribution!

A few mostly minor changes. In particular:

* probably best to move `sanity_check` to its own hook
* I don't think `LiveVisualizerHook` should be a default hook
* please patch `time.time` in your unit tests -- this should make a big difference on performance and reproducibility

BTW, also thanks for the TODO & other comments with questions. It makes very easy to find points that should be considered.

openpathsampling/pathsimulators/path_sampling.py
```diff
@@ -103,15 +105,19 @@ def from_dict(cls, dct):
 
         return obj
 
+    def attach_default_hooks(self):
+        self.attach_hook(hooks.StorageHook())
+        self.attach_hook(hooks.PathSamplingOutputHook())
+        # TODO: or should we make self.live_visualizer a property?
+        # and only attach the visualizerhook on 'request', i.e. by setting it?
+        self.attach_hook(hooks.LiveVisualizerHook())
```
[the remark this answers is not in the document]
hejung: This TODO stems from the goal to have no change in functionality through the hooks as compared to before. The current implementation of the `PathSampling` has set `live_visualizer` to `None` and checks every step if there is one (and visualizes if so). Therefore I made it a default hook that checks at every step if the hosting simulation has its `live_visualizer` set to not change any functionality (e.g. that setting `live_visualizer` during a simulation will immediately start the visualization). I think the same could be achieved by making `live_visualizer` a property and attach the corresponding hook only when this is set. 
From what you wrote, I guess the property is the way to go? This should make deprecation easy and we can keep the current functionality for now without making visualization a default hook.

openpathsampling/beta/hooks.py
```diff
@@ -34,37 +41,95 @@ def after_simulation(self, sim, hook_state):
...
         if self.storage is not None:
             self.storage.save(results)
             if step_number % self.frequency == 0:
+                if sim.sample_set is not None:
+                    # some PathSimulators never set their sample_set
+                    # but PathSimulator.__init__ sets it to None
+                    sim.sample_set.sanity_check()
                 self.storage.sync_all()
 
     def after_simulation(self, sim, hook_state):
```
hejung, line 115: This is something I am unsure about in both versions. In master we check for a storage not to be `None` that we then do not sync (which is strange, except if the two storages are always the same) and my version feels quite cumbersome (but I wasnt sure if we can expect the two storages to be the same).

Another possibility I thought about is the following:
```python
if self.storage is not None:
    self.storage.sync_all()
```
Given the fact that we set the `simulation.storage` as `self.storage` this should also work (assuming that the hook will not be transplanted to another simulation after calling `before_simulation`).

dwhswenson: The methods names on storage are really confusing, and don't do what you would expect from their names. (This is somewhat fixed in the new storage.)

In SimStore (new storage), the following is true, and I think the behavior is the same in netcdfplus:

* `save` is a function called `stash` in SimStore. It basically puts data into a staging area, but does not actually save it to disk
* `sync_all` dumps everything from the staging area to disk (and will probably be renamed in OPS 2.0 to something like `save_stashed`.) This is the part that actually takes significant time.

So you *always* want to put the data in the staging area (don't lose anything!) but you might only infrequently want to actually dump that to disk (large fraction of cost for toy models is saving to disk, so batch it together).

Actually, looking a little more carefully at this, you'll probably want to add something based on the try/except clause that has been added to `save_current_step`:

https://github.com/openpathsampling/openpathsampling/blob/b6da5ddf14981c6624c2607f3b00ce0ec120c85d/openpathsampling/pathsimulators/path_sampling.py#L115-L120

(where `storage.steps.save` can be replaced by `storage.save` to save any general object.)

openpathsampling/pathsimulators/path_sampling.py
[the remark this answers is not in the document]
dwhswenson: > From what you wrote, I guess the property is the way to go?

Definitely. Then in 2.0 the *only* way to add a live visualizer is via the hook -- functionality is available, but there's not some ghost attribute hanging around when not needed.

hejung:
@dwhswenson All done (I think).

- I moved the `sample_set.sanity_check()` to its own hook (which is a default hook for `PathSampling`) + wrote tests for that hook
- `LiveVisualizerHook` is not default anymore, instead it is attached only upon setting the `PathSampling.live_visualizer` property to something different then `None`. I now also throw an error from the hook if the `live_visualizer` is not set, which should be ok since the hook is not attached by default anymore. + wrote tests for that hook
- I fixed/improved the tests for the `PathSamplingOutputHook` according to your suggestions. Thanks for pointing me to the `patch` function, that solved the knot I had in my brain :)
- After sleeping a night on the `StorageHook.after_simulation()` method I think what we want is this:
```python
if self.storage is not None:
    self.storage.sync_all()
```
This should work in all cases, namely if the storage set only in the hook and also if the storage is taken from the simulation (because then it will be at `hook.storage` if the hook does not have a storage).

hejung:
So I just realized that I did not address the `save` try/except clause. I would/will do it together with the `StorageHook.after_step()` if what I did there is not what we want @dwhswenson : Let me know what you think is best there.

dwhswenson:
Looks good, as soon as the try/except clause for SimStore vs netcdfplus saving is added!

> * After sleeping a night on the `StorageHook.after_simulation()` method I think what we want is this:
> 
> ```python
> if self.storage is not None:
>     self.storage.sync_all()
> ```

I agree. I'm not sure why there was a thing to avoid doing a second `sync_all`. I'm not entirely sure what all happens in netcdfplus, but in SimStore, that just tries to save all objects in an empty list (so there's no real computational cost, and no side effect.) I would think that a second `sync_all` would also be idempotent in netcdf/netcdfplus.

hejung:
> Looks good, as soon as the try/except clause for SimStore vs netcdfplus saving is added!

Done.
 
> I agree. I'm not sure why there was a thing to avoid doing a second `sync_all`. I'm not entirely sure what all happens in netcdfplus, but in SimStore, that just tries to save all objects in an empty list (so there's no real computational cost, and no side effect.) I would think that a second `sync_all` would also be idempotent in netcdf/netcdfplus.

That was me beeing confused from what we previously had 
```python
if self.storage is not None:
    sim.storage.sync_all()
```
which did not make sense to me (checking `self.storage` but using `sim.storage`)...and the tests call `assert_called_once` which I did not want to touch :)

dwhswenson (approved):

dwhswenson:
> which did not make sense to me (checking `self.storage` but using `sim.storage`)...and the tests call `assert_called_once` which I did not want to touch :)

Ah, I understand. The `self` vs `sim` was probably just a slip of the fingers (or a missed change in a refactor).`assert_called_once` was probably because there didn't seem to be a way for it to be called more than once, but the important thing was that it should have been called at least once!

Anyway, all looks good now! Merging -- thanks again!

PR #787, opened 2018-06-23
S-shooting

singraber:
Added S-shooting method to OPS, includes example and tests.

Developed as part of E-CAM ESDW in Leiden in August 2017 and finished on a sunny day in June 2018 in Vienna.

Link to the gitlab repository:
https://gitlab.e-cam2020.eu/singraber/S-Shooting

PR #913, opened 2020-04-30
Fix: OPS hangs when external engine errors/exits unexpectedly

oliverdutton:
Problem: I was using GROMACS and OPS couldn't tell when GROMACS had exited due to errors.

Reason: External engine get's checked if it's still running using subprocess.Popen.process.is_running() call. This doesn't actually communicate with the process to ask if it's still running, it returns the value corresponding to last time it was checked. In this case, when the process was set up, so case of not running never gets called.

Solution: Altered to use subprocess.Popen.process.poll() - return None is process still running, and the return code if it isn't. 

I've also altered the stop signal to copy with the process already having exited of it's own accord


oliverdutton:
Ah, that explains it. I wasn't entirely sure what .is_running was doing. Zombies explain it. 

I don't have lots of experience with unittests so I'll leave it to you if that's okay.

dwhswenson:
Sorry for the delay on this; I meant to come back and add tests myself. Life got busy. Then it looks like a few accidental commits got added to this branch, and I was going to try to figure out how to deal with them. Life got busier. Regarding those extra commits:

* 855f0fb keeping (though note that this functionality will be deprecated and replaced in 2.0; it's too high-level of a question for the low-level code)
* 2914fac keeping: @hejung, I think that you are also working on something similar? heads-up that this is going into master
* c2d47f0 will revert; in favor of the implementation included in #924

dwhswenson:
> will revert; in favor of the implementation included in #924

Oops, I meant #787. Right author; wrong PR.

oliverdutton:
Wonderful.

hejung:
> Sorry for the delay on this; I meant to come back and add tests myself. Life got busy. Then it looks like a few accidental commits got added to this branch, and I was going to try to figure out how to deal with them. Life got busier. Regarding those extra commits:
> 
>     * [855f0fb](https://github.com/openpathsampling/openpathsampling/commit/855f0fba950f840af00061143c695705f11ea3ea) keeping (though note that this functionality will be deprecated and replaced in 2.0; it's too high-level of a question for the low-level code)
> 
>     * [2914fac](https://github.com/openpathsampling/openpathsampling/commit/2914fac642048fcefc521888307442096cf5dade) keeping: @hejung, I think that you are also working on something similar? heads-up that this is going into master

You mean the mdp parsing I guess? This should be easy to rebase/merge, I didn't touch the write_frame function :) 
However looking at the commit I was wondering if the snapshot timestep is a property that all external engines have, i.e. is there anything that hinders us to put it in the ExternalEngine properties dict?
And just as a matter of curiosity: what is the corner case we catch with the nested asarray calls? list of iterators? (numpy-doc claims it should be able to convert lists of lists, lists of tuples and lists of arrays.)

> 
>     * [c2d47f0](https://github.com/openpathsampling/openpathsampling/commit/c2d47f055712bc339356ffe9701e4de74375f116) will revert; in favor of the implementation included in #924

dwhswenson:
> You mean the mdp parsing I guess? This should be easy to rebase/merge, I didn't touch the write_frame function :)

Yep; glad it doesn't cause any headaches!

> However looking at the commit I was wondering if the snapshot timestep is a property that all external engines have, i.e. is there anything that hinders us to put it in the ExternalEngine properties dict?

Good point; completely reasonable to move that into the `ExternalEngine`. (In fact, I think it should be a property that *all* engines have, period. Defaults to `1`, but this is used in, e.g., flux calculations. For OpenMM, it carries units, so your rates etc. are reported with correct units as well.)

> And just as a matter of curiosity: what is the corner case we catch with the nested asarray calls? list of iterators? (numpy-doc claims it should be able to convert lists of lists, lists of tuples and lists of arrays.)

@oliverdutton : Do you recall the reason for this? I had assumed it might be something weird with `simtk.unit` stuff, but that isn't used for Gromacs.

PR #914, opened 2020-05-04
Gracious kill hook

hejung:
This builds on #911 and #755. It should only be merged after them. 

This adds a `GraciousKillHook` that writes the current simulation state to storage and ends the simulation loop by raising a `GraciousKillError` when a given maximum walltime is reached.
You can also pass a custom callable as `final_call` at hook construction, in case you want to do anything else than closing the storage when the simulation is terminated.

Example usage (assuming your steps are not superfast the code below should end the simulation after something sligthly smaller than 3 min 34 s):
 ```python
 kill_hook = GraciousKillHook("3 minutes 34 seconds")
# sampler is a `PathSimulator`
sampler.attach_hook(kill_hook)
try:
    sampler.run(2000)
except GraciousKillError:
    print("Simulation ended due to maximum walltime reached")
```

### DIFF
--- a/openpathsampling/beta/hooks.py
+++ b/openpathsampling/beta/hooks.py
@@ -413,7 +413,7 @@ class GraciousKillHook(PathSimulatorHook):
         self.final_call = final_call
 
     def _get_timedelta(self, time_str):
-        # https://stackoverflow.com/questions/35545091/convert-human-readable-time-difference-not-timestamp-to-something-usable-for-s/35545140
+        # https://stackoverflow.com/questions/35545140
         timespaces = {"days": 0}
         for timeunit in "year month week day hour minute second".split():
             content = re.findall(r"([0-9]*?)\s*?" + timeunit, time_str)
--- a/openpathsampling/beta/hooks.py
+++ b/openpathsampling/beta/hooks.py
@@ -5,15 +5,29 @@ These hooks group several methods together for use as part of a
 :class:`.PathSimulator` ``run`` method. They allow for additional
 calculations or output at several points in the simulation.
 """
+import re
 import time
+import logging
 import openpathsampling as paths
+from datetime import timedelta
 from openpathsampling.netcdfplus import StorableNamedObject
+
+
+logger = logging.getLogger(__name__)
 
 
 class SimulationNotFoundError(RuntimeError):
     """
     Raised when a hook tries to access its parent simulation before knowing it.
     """
+    pass
+
+
+class GraciousKillError(RuntimeError):
+    """
+    Raised when a simulation reaches the maximum walltime.
+    """
+    # makes it easy to catch **only** the max walltime but fail on anything else
     pass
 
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by dwhswenson** [comment]
With #911 in, checking on this one: it looks like the code is largely completed. It will need some conflict resolution (which is hopefully easy to accomplish -- probably some are cases of "added two functions in the same location of the old code").

Main question: Is this close enough (and do you, @hejung, have time enough to finish it up) that it is worth holding off on the 1.5 release until this is included? Thematically, it would be nice to include it with the other work on hooks in 1.5, but if it will take more than a week to get done, it might be better to include it in 1.6.

**Comment by hejung** [comment]
> With #911 in, checking on this one: it looks like the code is largely completed. It will need some conflict resolution (which is hopefully easy to accomplish -- probably some are cases of "added two functions in the same location of the old code").

I cherry-picked the commit that introduces the `GraciousKillHook` onto the current master and changed the tests to use a patched `time.time` (as for the `PathSamplingOutputHook`).
 
> Main question: Is this close enough (and do you, @hejung, have time enough to finish it up) that it is worth holding off on the 1.5 release until this is included? Thematically, it would be nice to include it with the other work on hooks in 1.5, but if it will take more than a week to get done, it might be better to include it in 1.6.

I would say it is done.
Only thing I noticed that we might want to consider, is that the `GraciousKillHook` is the only hook that uses `logging`. I did that to see in the ops main simulation log if the hook ended the simulation.
However I realized now that the reason the other hooks do not use logging might be that logging and dask do not play nicely together (similar to `logging` and `multiprocessing`), but I have no experience with logging from dask. @dwhswenson Should I rather convert the calls to logging to print statements? (Then they would/should go to std-out and would normally be captured by the queuing system in my intended use-case)

File: openpathsampling/beta/hooks.py [hunk 1]
```diff
@@ -346,3 +360,92 @@ def after_simulation(self, sim, hook_state):
...
+        self.max_walltime = max_walltime
+        self._timedelta = self._get_timedelta(max_walltime).total_seconds()
+        logger.info("Parsed time string '{:s}' as a ".format(self.max_walltime)
+                    + "timedelta of {:d} seconds.".format(int(self._timedelta))
+                    )
+        self.fuzziness = fuzziness
+        self.final_call = final_call
+
+    def _get_timedelta(self, time_str):
+        # https://stackoverflow.com/questions/35545091/convert-human-readable-time-difference-not-timestamp-to-something-usable-for-s/35545140
```
dwhswenson, line 416:
```suggestion
        # https://stackoverflow.com/questions/35545091/
```

Equivalent link for the question, keeping to 80 chars. You might actually want to change the first part to `https://stackoverflow.com/a/35545140`, which is the answer that it looks like the following is based on.  (Was orig link supposed to be https://stackoverflow.com/questions/35545091/convert-human-readable-time-difference-not-timestamp-to-something-usable-for-s/#35545140 ?)

openpathsampling/tests/pathsimulators/test_hooks.py
```diff
@@ -415,3 +415,62 @@ def test_after_simulation(self):
...
+                                          )
+        self.nokill_hook = GraciousKillHook("10 hours 2 seconds",
+                                            final_call=self.nocall_final_call
+                                            )
+
+    def test_kill(self):
+        s_num = 1  # step after which the simulation is killed
+        with patch("openpathsampling.beta.hooks.time.time",
+                   new=MagicMock(side_effect=[0, 10]),
+                   ) as MockTime:
```
dwhswenson, line 443: `as` statement doesn't appear to be needed here (or in `test_nokill`)

dwhswenson (changes requested):
Looks good. Very tiny change to keep the line lengths down -- you may want the link listed in my comment instead of the one in the single-click suggestion -- and some unused `as` clauses in tests.

As to `logging` vs. writing to an output stream:

1. The general advantage of logging is that it allows the user to control where/whether/how the output is reported. We limit user options on that in progress reporting because it might mess up our tools for refreshing the output (although users can still silence/redirect output by changing the `output_stream`, giving at least some choice). This use case seems reasonable for `logging` in serial. 

2. I think that, in order for Dask to report task logging, you need to set up the logging inside the task, and do so in a way that gets the information back to you (i.e.,  use a unique file name). (I remember investigating this in detail at one point, but don't remember exactly what I could/couldn't do -- in any case, logs from remote processes are at best a headache to process). However, `GraciousKill` probably won't be running inside a remote Dask process, since each Dask process may be running on a different job from your queuing system (so it's hard for `GraciousKill` to track time remaining).

In other words, this probably isn't an issue: the `GraciousKillHook` will be very useful for some workflows/simulation types (e.g., running TPS or other simulations where we can't do multiple shooting moves as once), but it probably won't be used in combination with Dask-based workflows.

hejung:
Done! (I also removed the unnecessary `as` from the `PathSamplingOutputHook`...I seem to always forget that you can use `with` without `as`....)

dwhswenson (approved):

dwhswenson:
LGTM. Merging! (And I will start the 1.5.0 release process soon.)

PR #921, opened 2020-06-23
support ujson v3 (part 2)

hejung:
This is the same that @sroet did in #917. I only added the `ujson_kwargs` to another call to `ujson.dumps`.


dwhswenson:
Looks good to me. Also looks like you need to associate an email GitHub knows about with the git config on the computer where you did this ;-)

hejung:
Good catch!...I have too many e-mails and GitHub did not know yet about this one :)

PR #923, opened 2020-07-20
fix numpy 1.19 DeprecationWarning: tostring() 

sroet:
Since `numpy=1.19` `array.tostring()` is deprecated, and should be replaced with `.tobytes()` (present in `numpy >=1.9`), this PR does that for OPS.

The leftover `>40000` warnings in the test suite come from `NetCDF4` which has been fixed in [this PR](https://github.com/Unidata/netcdf4-python/pull/1024) (I tested this locally and it passes without these DerpecationWarnings).

Also cleaned some `pep8` complaints on the files I touched.

### DIFF
--- a/openpathsampling/analysis/shooting_point_analysis.py
+++ b/openpathsampling/analysis/shooting_point_analysis.py
@@ -1,8 +1,6 @@
-import openpathsampling as paths
 import collections
 import pandas as pd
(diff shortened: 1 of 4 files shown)


dwhswenson (approved):
Sorry for the long delay on review! Good to merge.

PR #925, opened 2020-09-15
Fix toy_mstis_3 for change in pandas

dwhswenson:
Apparently pandas no longer allowed tuples as indices? At least, that fixed it locally.


sroet:
LGTM (and should be fine if the code output is the same)

PR #927, opened 2020-09-16
Clean up devtools

dwhswenson:
We've gone through a lot of changes and experimental approaches to handle CI over the years, and often we've left behind cruft from a previous approach (this was been especially true when either I or @jhprinz want to try a new approach, and don't know which files from the other person's work might break something else).

This PR is aimed at simplifying that, and getting rid of extra files that aren't used and only serve to add confusion to new developers. It will also replace some really old scripts with better approaches.

In particular, this completely removes the old conda recipe. We haven't used conda-build to build the project in ages (and never really needed to, anyway). That, and some of the other stuff in devtools, existed because we copied our skeleton from MDTraj.

This creates a setuptools "extras" target called `testing`. I'll add a comment below for details on the name choice: I'm open to other option (`test`? `tests`?), but I want to do whatever is most common in the broader community.

Anyway, this will make no difference to users, but may help developers contribute more. And there won't be all this extra stuff to annoy me.

- [x] Remove unneeded documents
- [x] Move as much as possible to setup.cfg (single source for most pinning)
- [x] Simpler `conda_install_reqs.sh` script
- [x] Update the devtools/README.md to be about OPS, not 5-year-old version of MDTraj
- [x] Remove `pinned`
- [x] Move `autorelease_check.py` to devtools directory

This certainly isn't final word on cleaning up our devtools directory, but this is the once-every-five-years cleanup that is sorely needed.


dwhswenson:
Regarding the name for the `testing` extras: I'm open to whatever is most commonly used in the community. For packages installed in a given environment, the script below will count the most common names of extras (assuming you have `importlib_metadata` installed. In my main development environment, I had both `test` and `testing` at the same number, and between the two, I prefer `testing`. But if there's a broader community consensus, I'll go with that.

```python
import sys
import collections
import importlib_metadata

extras = []
for mod in sys.modules:
    try:
        pkg_extras = importlib_metadata.metadata(mod).get_all('Provides-Extra')
    except importlib_metadata.PackageNotFoundError:
        pass
    else:
        if pkg_extras is not None:
            extras.extend(pkg_extras)
        
print(collections.Counter(extras))
```

hejung:
I do not have any preference on `testing` vs `test`, but the pypi sample project uses `test` (https://github.com/pypa/sampleproject/blob/52966defd6a61e97295b0bb82cd3474ac3e11c7a/setup.py#L158 ). Not sure if this is community consensus but I have so far used that for my skeletons.

dwhswenson:
Thanks @hejung : that seems like a good reason to switch to `test`. Honestly, I was mainly surprised that out of my entire dev conda env (which is huge, since it includes everything I need to develop any of my projects) there were only 8 packages testing-related extras, split evenly between `test` and `testing`.

dwhswenson:
This is now ready. A few things this does:

* Fewer locations for requirements -- previously a change to requirements, including pinning one of our reqs, meant changing several files. There are still two locations: this is because our setuptools requirements list includes MDTraj, although OPS can technically run without MDTraj (see #867), and we do test against the true minimal requirements, which are in `devtools/minimal.txt`. The standard installation is in `setup.cfg`. There is no longer a conda recipe, because we weren't using it (except as another place to list version and requirements).
* New script to install requirement from `setup.cfg` instead of the conda recipe.
* Remove old, unused files such as `setup.yaml` (an early attempt by @jhprinz at getting a single source for version/requirement info), `.project`/`.pydevproject` (something from some IDE that @jhprinz was using)
* Move testing requirements from a file and into a setuptools extra.

PR #928, opened 2020-09-17
SimStore: New storage subsystem (part 1)

dwhswenson:
For a while, I've been trying to rework our storage. This is because of several problems with the netcdfplus package:

1. It doesn't seem to play nicely with [dask](http://dask.pydata.org/), which I intend to use for parallelization
2. It is heavily tied to netCDF, and kept breaking because netCDF changed their internals frequently.
3. It doesn't have very good performance on loading data.
4. @jhprinz is no longer actively developing it, so it is becoming hard to maintain.

This PR gives the first work toward SimStore, a new storage subsystem. For now, I'm developing this in the `experimental` subpackage, because the API may change. This will replace netcdfplus in OPS 2.0. The last release before OPS 2.0 will include tools to facilitate transition from netcdfplus to SimStore, although I don't guarantee that all data will be transferable.

Some advantages of the new approach:

* Faster and more efficient storage, and easier debugging. Instead of recursive algorithms, I use iterative algorithms that group everything to be written to/loaded from a given table into a minimal number of calls to disk. I also distinguish between storage of data objects (of which there are many, but which are of predictable format) and storage of simulation objects (which are few in number, but of a wider variety). I use faster algorithms to store data objects, and more generic (but slower) algorithms to store simulation objects.
* More flexible with regards to backends. For now, the new storage uses SQL as a backend, but changing that is as easy as creating a new backend model (and I intend to do that by adding a pytables-based hdf5 backend in the future.) The backend is completely independent from the rest of the storage subsystem. In netcdfplus, storage inherits from netcdf4. That makes changing the backend much more difficult.
* Easier to add new types for JSON-based storage. The new storage uses a registration-based system for adding JSON types, whereas netcdfplus would require actually editing the core netcdfplus code. Now, a user can add another type (e.g., simtk.unit.Quantity) without having to edit the core code.

In addition, as I go forward, I'm planning to implement it in a way that is designed to work with dask, and I'm also planning to focus on analysis performance.

This is a major undertaking, so in addition to this huge PR, there will be a number of PRs following to complete the storage.

- [x] Ability to store/reload arbitrary simulation objects
- [x] Ability to store/reload all data objects from a toy simulation
- [x] Custom tables for `ensembles`, `volumes`, etc.
- [x] SQL backend support
- [x] General framework for extensible JSON support; support for numpy arrays
- [x] Support for automatic registration of new snapshot types
- [x] Lazy proxy objects

Coming up in future PRs:

* New base class for CVs (which will give better analysis performance and will enable parallelization)
* Support for OpenMM (static/kinetic containers; simtk.unit.Quantity)
* Generic "tagged" objects
* Complete unit tests

This will be the first package released in the `experimental` subpackage. The policy for the `experimental` subpackage is that any unit tests must run and pass, but they aren't counted toward overall coverage requirements. In addition, things in `experimental` and `beta` are not tested on Python 2. Unit tests have to be complete (nearly 100% coverage) before we move this into the `beta` subpackage. 

As I've been developing this, I've used a number of notebooks as additional tests (mainly to ensure that I don't break things that had been working.) Those notebooks are now online in their own repo: https://github.com/dwhswenson/ops-storage-notebooks

I'm planning on putting better examples in that repo soon so that users have the necessary information to start playing with this.


openpathsampling/experimental/storage/sql_backend.py
```diff
@@ -0,0 +1,449 @@
...
+        self.number_to_table.update({n_tables: table_name})
+
+    def _load_from_table(self, table_name, idx_list):
+        # this is not public API (assumes idx_list, which is reserved by not
+        # guaranteed)
+        table = self.metadata.tables[table_name]
+        or_stmt = sql.or_(*(table.c.idx == idx for idx in idx_list))
+        sel = table.select(or_stmt)
+        with self.engine.connect() as conn:
+            results = list(conn.execute(sel))
```
sroet, line 222: So, just commenting the fix that I had to get around the hard capped SQLite limit of 10000 expressions in the expression tree ([that can only be changed at compile time](https://www.sqlite.org/limits.html)):
1. `from itertools import zip_longest` should be added to the imports

2. 
```suggestion
        # Loop over max 1000 items at once
        # taken from https://stackoverflow.com/a/434411
        results = []
        max_items = 1000
        iter_list = [iter(idx_list)] * max_items
        full_list = zip_longest(*iter_list)
        with self.engine.connect() as conn:
            for sub_list in full_list:
                or_stmt = sql.or_(*(table.c.idx == idx for idx in sub_list
                                               if idx is not None))
                sel = table.select(or_stmt)
                sub_results = list(conn.execute(sel))
                results.extend(sub_results)
```

sroet (commented):
Just leaving the improvements that we discussed so far while stress testing.

I didn't check, but did you fix the issue where UUIDs were not into an index?

openpathsampling/experimental/storage/sql_backend.py
dwhswenson: I took the basic idea, but details of the implementation were to have a reusable tool in the `tools.py`. Changes in 1613b34.

dwhswenson:
@sroet : Thanks for the reminders!

> I didn't check, but did you fix the issue where UUIDs were not into an index?

I had not, but I think the changes in 1613b34 should fix (specifically, `sql_backen.py:127`). Still need to give it a stress test to be sure.

dwhswenson:
@sroet : If you have a chance, could you take a quick glance at the notebooks in [dwhswenson/ops-storage-notebooks/examples](https://github.com/dwhswenson/ops-storage-notebooks/tree/master/examples) and let me know if they match your usage? (If there's anything I may have left out?)

If there are any problems with them, easiest to raise an issue on that repo (rather than mixing usage discussion into this PR)

sroet:
So there a couple problems I mentioned in an email long ago (04 Jan 2020) and I don't know if you want to solve this via dwhswenson/ops-storage-notebooks/examples  or in the actual code:

1. Negative slicing does not work (`storage.steps[-1]`), this needs to be `storage.steps[len(storage.steps)-1]`
2. When restarting the simulation object does not get initialized with the right storage (it is set to `None`) and save_fequency (set to 1), so it won't safe after restarting and is really slow for toy systems if storage is set.
```python
backend = storage.sql_backend.SQLStorageBackend('test.sql', mode='a')
s = storage.ops_storage.OPSStorage.from_backend(backend=backend)
calc = s.pathsimulators[0]
print(calc.storage) # This will print "None"
print(calc.save_frequency) # This will print 1 no matter how you stored it
calc.storage = s # This line is needed to actually store the rest of the simulation when restarting
calc.save_frequency = 50 # This is needed for speed in toy models
``` 

I can move this to the other repo if you want to solve these issues on the user side.

dwhswenson:
> So there a couple problems I mentioned in an email long ago (04 Jan 2020) and I don't know if you want to solve this via dwhswenson/ops-storage-notebooks/examples or in the actual code:

Thanks; and these are definitely things to discuss here. (Although I think you're mixing US vs rest-of-world date order -- still long ago, but I'm pretty sure that was 1 Apr, not 4 Jan.)

> 1\. Negative slicing does not work (`storage.steps[-1]`), this needs to be `storage.steps[len(storage.steps)-1]`

Yes, this needs to be fixed. Will do in this PR.


> 2\. When restarting the simulation object does not get initialized with the right storage (it is set to `None`) and save_fequency (set to 1), so it won't safe after restarting and is really slow for toy systems if storage is set.

These are actually two separate issues:

* *`Storage` isn't storable.* I think I fixed that in a commit in another branch that builds on this (at least, that's what the commit message *says* it does: de565ab7e90e5921c3537a4c01bee4152d5d3ac7). I'll see if that can be cherry-picked into here without introducing problems. Note that this isn't trivial -- in particular, I don't think we really want to treat the `mode` attribute seriously; at least, not for `mode='w'`, which would erase the existing file. Kind of a `rm -rf ~/`. My guess is that the user expectation would be for the storage to be `mode='a'`. (Note: for SQL-based storage, there's actually no `'a'`/`'r'` distinction, because you don't actually open a file pointer. The `'w'` distinction is logic that *I* wrote to wipe the existing data; it doesn't come from the filesystem.)

* *Mutable properties (e.g., `save_frequency`) are not stored.* I'm pretty sure this is the same with netcdfplus. We consider these things mutable, and they get default values when the object is recreated. That's true of anything that isn't returned by `to_dict` -- obviously, we can't store it if `to_dict` doesn't mention it! Long-term plan is to allow mutable options to be stored as well. The API I would use for this is pretty straightforward, and also backward compatible: Instead of returning one dict, `to_dict` can (optionally) return 2. The first one is the immutable attributes; the second is the mutable attributes. Similarly, `from_dict` has an optional second argument. Getting it to work with the storage subsystem is another issue, and will be beyond the scope of this PR.

    My general leaning would be to always save all copies of mutable attributes, and when you reload you get the last one that was saved for the given object. This is about reproducibility. There are also a number of simulation cases where this makes sense, too (I think @hejung's S-range work could be more easily implemented by allowing a single volume-like object to have a mutable property that is saved at each step. I'm also working around mutability in my implementation of AMS in #806). The argument against this approach would be that frequent mutation and saving would bulk up file size. But I think it will be rare to save and resave objects outside the context of a simulation, where reproducibility matters.

dwhswenson:
* Negative slicing: Should be done now, although I haven't properly tested it.
* Making `Storage` storable: It's a little tricky to cherry-pick that into this branch, because it seems to use a few aspects that were developed for the storable function branch, which will be the follow-up to this PR (that's where I do the new CVs). I've cherry-picked it into `storable_functions`, so it'll be a part of that PR.

sroet:
> . (Although I think you're mixing US vs rest-of-world date order -- still long ago, but I'm pretty sure that was 1 Apr, not 4 Jan.)
Ah yeah, somehow outlook is using US dates...

> Negative slicing: Should be done now, although I haven't properly tested it.

I can confirm it works for the example I gave

> Making Storage storable: It's a little tricky to cherry-pick that into this branch, because it seems to use a few aspects that were developed for the storable function branch, which will be the follow-up to this PR (that's where I do the new CVs). I've cherry-picked it into storable_functions, so it'll be a part of that PR.

That is fine by me :+1: (assuming that PR will go into the same release as this PR)

> But I think it will be rare to save and resave objects outside the context of a simulation, where reproducibility matters.

I totally agree with that. 

I think that covered all the issues I found while stress testing this for running the test system 100_000 steps, except that that there is some optimization I do when loading a lot of steps, but I will make that an issue in the examples repo. (It is essentially resetting the uuid cache every 1000 steps)

dwhswenson:
I'm about to do a bit of renaming here. Essentially, I want the `experimental` subpackage to somewhat mimic what the `openpathsampling` package will look like in 2.0, so that all you need to do to go from experimental to 2.0 is remove the `experimental` from the imports, even for things not in the root `openpathsampling` namespace. 

Of the user-facing things, I believe this would just mean the following changes:

* `experimental.storage.sql_backend.SQLStorageBackend` => `experimental.simstore.SQLStorageBackend`
* `experimental.storage.ops_storage.OPSStorage` => `experimental.storage.Storage`  (will be in root namespace in OPS 2.0, but `experimental` only holds subpackages)
* `experimental.storage.monkey_patches.monkey_patch_all` available at `experimental.storage.monkey_patch_all` (etc for other monkey patches; also, the existing imports will still work, but a simpler one will be available.)

@sroet : Mainly, I want to warn you this change is coming. If there's a reason it would be a huge hassle, I can hold off for a bit. After that change, I think I'll mark this ready for review.

sroet:
> Mainly, I want to warn you this change is coming. If there's a reason it would be a huge hassle, I can hold off for a bit. After that change, I think I'll mark this ready for review.

Don't worry about it.  I think should be proficient enough with `git` to lock to a specific commit hash if it would be a hassle ;)

openpathsampling/experimental/storage/storage.py
```diff
@@ -0,0 +1,531 @@
...
+    def __iter__(self):
+        # TODO: ensure that this gives us things in idx order
+        backend_iter = self.storage.backend.table_iterator(self.table)
+        enum_iter = enumerate(tools.grouper(backend_iter,
+                                            self.iter_block_size))
+        for block_num, block in enum_iter:
+            row_uuids = [row.uuid for row in block]
+            loaded = self.storage.load(row_uuids)
+            if block_num % self.clear_cache_block_freq == 0:
+                self.storage.cache.clear()
```
sroet, line 432: This works for the new analysis, but the old one errors out ([details on how this was figured out](https://github.com/dwhswenson/ops-storage-notebooks/issues/1)) :

If you want to support the old analysis for a bit longer, this can be changed to:
```suggestion
            row_uuids = [row.uuid for row in block]
            loaded = self.storage.load(row_uuids)
            if block_num % self.clear_cache_block_freq == 0 and block_num != 0:
                self.storage.cache.clear()
```

or

```suggestion
            row_uuids = [row.uuid for row in block]
            if block_num % self.clear_cache_block_freq == 0:
                self.storage.cache.clear()
            loaded = self.storage.load(row_uuids)
```

dwhswenson: Thanks!

dwhswenson:
I've completed the renaming/shifting imports as proposed in https://github.com/openpathsampling/openpathsampling/pull/928#issuecomment-698280723.

As far as I'm concerned, this is now ready for review. It's not expected to be the complete implementation of the new storage, but a lot of first steps. It's already a large enough PR that it'll be hard for anyone to thoroughly review it, so I want this merged in so that subsequent steps can come in smaller doses.

That said, I'll leave this up for at least 10 days, just in case someone is brave (foolish) enough to want to give it a full review, and to provide sufficient time for other feedback. I will not merge this before Monday, 5 October, 2020.

sroet:
From me this is a +1,

I have not looked at the code, but:
- I have used this code (specifically https://github.com/dwhswenson/openpathsampling/commit/f80eb2687ebf73085cb57eb7e2802375c66b3364 ) to run multiple `100_000` step toy system calculation with custom objects.
- I have used  https://github.com/openpathsampling/openpathsampling/pull/928/commits/c3fbbe0ed9b24af86a3e9db4923066f1432ebc11 to analyze these simulations with reasonable speeds (for me it loads at about `22 steps/s`, `1.25 hours for 100_000 steps while I never got the old (netcdf) storage to even open.)
- To get to the above point I had to manually create an `sql` index which was the reason for [this comment](https://github.com/openpathsampling/openpathsampling/pull/928#pullrequestreview-490799966), which was not to painful if required (takes about 15 min)

The two things that I have not checked and am a bit worried about:
- I have not run a simulation with any code after  https://github.com/dwhswenson/openpathsampling/commit/f80eb2687ebf73085cb57eb7e2802375c66b3364 , so I don't now if the "UUID is not an SQLIndex" is actually fixed.

- I don't know if there is performance degradation when **running** long simulations because the storage cache is not cleared while running these simulations, but I do expect there to be (It used to scale as O(`N`), but might now be O(`log(N)`), with N being the number of `objects` in the `cache`).

Know issues that will be left for future PRs (@dwhswenson  please correct me here if I am promising things that you will not do :wink:  ) :
- `PathSimulator` objects don't get initialized with the storage they are loaded from
- Mutable attributes are not saved (whole different issue)
- While indexing now works for both positive and negative indices (for example `storage.steps[-1]`), it can't be sliced yet (for example `storage.steps[:5000]`)

dwhswenson:
> Know issues that will be left for future PRs

These are correct. The first one, storable storage, is already addressed in #929. Proper slicing of tables will probably come after #929, in a separate PR (keeping things smaller and easier to review). Storing mutable data is a larger problem, and may not happen until after 2.0 (but can be done in a way that is backward-compatible, so that's fine.)

Also, since this is in the `experimental` subpackage (where [SemVer](https://semver.org/) does apply), let me take a moment to state a few ways to future-proof for anyone using this. I'll focus on creating/using the `Storage` object, since that's the main thing users will care about (i.e., only devs will care about adding new `JSONCodec`s or about how storage and backend communicate).

* The examples currently show an approach of creating the backend directly with (`SQLStorageBackend(filename, mode)`) and then creating the storage with `Storage.from_backend`. This will probably remain the same. If you want to be even more future-proofed, starting with #929 I think the approach of create [SQLAlchemy engine](https://docs.sqlalchemy.org/en/13/core/engines.html) => `SQLBackend.from_engine` => `Storage.from_backend` will be guaranteed. (In the future, there will by a single-step `Storage(filename, mode)` initialization.) 
* Saving objects, either singly or as a list, with `storage.save()` is guaranteed
* Accessing true tables by single element (e.g., `storage.steps[0]`) or by iteration (e.g., `for step in storage.steps`) is guaranteed
* Accessing pseudo-tables (views on the simulation objects table) by number (e.g., `storage.cvs[0]`), by name (e.g., `storage.cvs['cv_x']`) or by iteration (e.g., `for cv in storage.cvs`) is guaranteed. Note that not all pseudo-tables have been added yet; if you find one missing, let me know.

I think these are the main ways people tend to interact with the storage. If there are other ways, please ask about them!

sroet:
alright I ran another simulation with `100_000` on commit https://github.com/openpathsampling/openpathsampling/pull/928/commits/fbe2b05d6d60753260510ca7d0fa58bd76807294

> I have not run a simulation with any code after dwhswenson@f80eb26 , so I don't now if the "UUID is not an SQLIndex" is actually fixed.

This has not been fixed and can lead to detrimental analysis performance as mentioned in [this comment](https://github.com/dwhswenson/ops-storage-notebooks/issues/1#issuecomment-701651243)

Non-indexed sql loads with `29.17s/it` while the indexed one: `17.69it/s` (notice the swap of `it` and `s` on those scales)

> I don't know if there is performance degradation when running long simulations because the storage cache is not cleared while running these simulations, but I do expect there to be (It used to scale as O(N), but might now be O(log(N)), with N being the number of objects in the cache).

The performance is still degrading somewhat, from on average `0.19 s/step` at  the start to on average `1.02 seconds per step` at `100_000`.

sroet:
Ok I completed a new run with `a2706cc`:
`Completed at 100_000 with and average speed of 0.45 s/step`
The final average speed of this version (`0.45 s/step`) is already more than `2x` the average speed of before  `1.02 s/step`.

```
Working on Monte Carlo cycle number 10000
Running for 29 minutes 36 seconds -  0.18 seconds per step
--
Working on Monte Carlo cycle number 20000
Running for 1 hour 9 minutes 40 seconds -  0.21 seconds per step
--
Working on Monte Carlo cycle number 30000
Running for 2 hours 3 minutes 14 seconds -  0.25 seconds per step
--
Working on Monte Carlo cycle number 40000
Running for 3 hours 3 minutes 21 seconds -  0.28 seconds per step
--
Working on Monte Carlo cycle number 50000
Running for 4 hours 16 minutes 47 seconds -  0.31 seconds per step
--
Working on Monte Carlo cycle number 60000
Running for 5 hours 37 minutes 18 seconds -  0.34 seconds per step
--
Working on Monte Carlo cycle number 70000
Running for 7 hours 5 minutes 29 seconds -  0.36 seconds per step
--
Working on Monte Carlo cycle number 80000
Running for 8 hours 39 minutes 29 seconds -  0.39 seconds per step
--
Working on Monte Carlo cycle number 90000
Running for 10 hours 41 minutes 45 seconds -  0.43 seconds per step
--
Working on Monte Carlo cycle number 100000
Running for 12 hours 33 minutes 34 seconds -  0.45 seconds per step
```

It is actually still growing linearly (on average), but at such a slow rate I think it is acceptable.

The if we look at the average time of the last `10_000`  the speeds are (with slightly different rounding than above):

| steps (x10000) | average | average last 10_000 | diff last 10_000 |
| --- | --- | --- | ---|
| 1   | 0.18 | 0.18 | - |
| 2   | 0.21 | 0.24 | 0.06 |
| 3   | 0.25 | 0.33 | 0.09 |
| 4   | 0.28 | 0.37 | 0.04 |
| 5   | 0.31 | 0.43 | 0.06 |
| 6   | 0.34 | 0.49 | 0.06 |
| 7   | 0.37 | 0.55 | 0.06 |
| 8   | 0.40 | 0.61 | 0.06 |
| 9   | 0.43 | 0.67 | 0.06 |
| 10 | 0.46 | 0.73 | 0.06 |

This seems to indicate that the average time to add a step grows by `0.06 s` for every `10_000` steps added. Given that this should still be sub-second for up to `130_000` steps , I am totally fine by this

- Analysis runs as fast as you would expect with an indexed sql (`36.38 it/s`)

- Appending seems to be as fast as before with an indexed sql (and in line with the `last 10_000` speeds mentioned above)
```
Working on Monte Carlo cycle number 103100
Running for 38 minutes 24 seconds -  0.74 seconds per step
Estimated time remaining: 23 minutes 33 seconds
```

This solves all the issues for me.

dwhswenson:
Thanks @sroet : I added an index on the UUIDs table, so this confirms my approach for that works and that the index persists.

It may be worth indexing other tables as well, especially snapshots and (in #929) saved CV results.

IIRC your trajectories are also rather long, right? Order of 1000 snapshots? Your analysis speed is probably limited by needing to load all snapshots (from accepted trajectories, at least) -- once I finish #929, that will no longer be necessary (if you run the simulation with the stuff in #929).

With this performance issue dealt with, I'll go ahead and merge this in, and future performance improvements can be in other PRs.

sroet:
> IIRC your trajectories are also rather long, right? Order of 1000 snapshots? Your analysis speed is probably limited by needing to load all snapshots (from accepted trajectories, at least) -- once I finish #929, that will no longer be necessary (if you run the simulation with the stuff in #929).

Normally between `100-1000` frames per trajectory.

PR #929, opened 2020-09-25
SimStore part 2: Storable functions (and more)

dwhswenson:
This is the second PR in the process of updating the OPS storage subsystem, following on #928. See #928 for overall motivation.

Major improvements in this PR:

- [x] `StorableFunction`: a new base class for CVs that is designed to handle parallel simulation
- [x] Storable storages: you can now restart a simulation without manually recreating the `storage` object it saved to
- [x] Support for Gromacs snapshots (NB: some Gromacs improvements here will be cherry-picked into a separate PR)
- [x] `snapshots` table that combines all the individual (engine-specific) snapshot tables.

This includes #928, and therefore should only be merged after that.


dwhswenson:
This is ready for review and comment. Since this is another big PR, I will leave it open for at least a week, merging no earlier than Sat 31 Oct 13:00 local (12:00 GMT).

An example showing how to use this is in dwhswenson/ops-storage-notebooks#5, in the `01_simple_usage.ipynb` notebook.

The good news is, these are the two really big SimStore PRs. From here, I can add smaller chunks that might actually be reviewable.

PR #933, opened 2020-09-27
Internalized snapshots; random filenames for external snapshots

dwhswenson:
This PR includes two significant improvements for external engines, and well as some minor improvements specifically for the Gromacs engine. Some of these commits were cherry-picked from other branches, including #929.

### General improvements to external engines

* **Internalized snapshots**: Every external engine should also be able to save full snapshot information to internal storage. We do this with the `internalize` feature of `ExternalMDSnapshot`s. This is designed to make it easier to start simulations with using external engines -- this way, a user doesn't need to worry about matching the path to the initial trajectory on a when copying simulation setup to a remote host.
* **Random filenames**: Previously, external engines created filenames as a simple increasing sequence. Now users have the option of setting random filenames. For now, the default behavior remains the same as before, but it is recommended that users start using `{'filename_setter': RandomStringFilenames()}` in their engine options. Random filenames solve two problems: (1) there's a tendancy for users to think that these TRR files contain the entire accepted trajectory (as opposed to the trial frames that were created), and then they try to analyze with Gromacs tools and get complete nonsense; (2) this will be essential for parallelization.

### Specific improvements for Gromacs

* The names of `topol.tpr` and `mdout.mdp` as generated by `grommp`, which were previously locked to defaults, can now be changed by the user. These files are automatically deleted at the end of a trajectory.
* `GroFileEngine` has moved from being defined in a nested scope to being a private class of the module.
* `snapshot_from_gro` can be accessed as `openpathsampling.engines.gromacs.snapshot_from_gro`


dwhswenson:
Ready for review. Will leave this up for 48 hours. Although this involves some changes to the underlying functionality, it should still be backward compatible.

openpathsampling/engines/external_engine.py
```diff
@@ -48,6 +53,87 @@ def _debug_snapshot_loading(snapshot):
...
+    def to_dict(self):
+        dct = super(_InternalizedEngineProxy, self).to_dict()
+        dct['engine'] = self.engine
+        return dct
+
+    @classmethod
+    def from_dict(cls, dct):
+        # ignore everything except the engine here
+        dct = {'engine': dct['engine']}
+        return super(_InternalizedEngineProxy, cls).from_dict(dct)
```
sroet, line 127: Wouldn't this be easier by just saving the engine? (The you don't have to drop the info again when loading)
```suggestion
    def to_dict(self):
        # We only need the engine to load again
        dct = {'engine': self.engine}
        return dct
```

sroet (approved):
The code looks good to me
I had one possible code reduction comment, but feel free to ignore that one.

openpathsampling/engines/external_engine.py
dwhswenson: Yeah, I've been trying to figure out if there was some reason I did it this way, but I don't see one.

PR #935, opened 2020-10-13
Improve CV storage error message about missing templates

dwhswenson:
In netcdfplus, storing CVs (`PseudoAttribute`s) requires a template object to be already stored. Originally we only had CVs that mapped snapshots to computed values, but eventually @jhprinz added the ability to use other objects (like trajectories) as inputs. However, the error message didn't get updated, and it is very confusing to get an error message saying to save a snapshot when you already have a snapshot saved (but need to save a trajectory).

This whole issue will be irrelevant with SimStore, but might as well fix the confusing error message.


sroet (approved):
LGTM, seems a lot clearer

PR #936, opened 2020-10-17
Fix collections.abc DeprecationWarning in sample.py

sroet:
Going from 5570cd31 to ace9d7e7 and running the tests surfaced another 
```
/home/sroet/github_files/openpathsampling/openpathsampling/sample.py:10: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3,and in 3.9 it will stop working
    from collections import Counter, Mapping
```

This solves that one (`Counter` still has to be imported from base `collections`)

### DIFF
--- a/openpathsampling/sample.py
+++ b/openpathsampling/sample.py
@@ -7,7 +7,8 @@ from openpathsampling.netcdfplus import DelayedLoader
 
 from openpathsampling.tools import refresh_output
 
-from collections import Counter, Mapping
+from collections import Counter
+from collections.abc import Mapping
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


dwhswenson:
I think you're gonna need a try: except: for Python 2.

sroet:
Yeah, figured that one out. Decided to explicitly check for python3 so it is easier to remove if/when OPS drops python `2.7`

dwhswenson:
Sure... I usually add a comment with the string `Py2` or something similar, so feel free to also do that if you come across others of this. Explicit version check is also fine here.

dwhswenson (approved):
Code looks good. Will merge on green.

PR #938, opened 2020-10-22
OpenMM instantaneous temp should work if positions not set

dwhswenson:
Currently, for an OpenMM snapshot, `snapshot.instantaneous_temperature` will raise `Exception: Particle positions have not been set` if no snapshot has been loaded into the engine (e.g., via setting `engine.current_snapshot`). Analysis will frequently not set that, and so it should be possible to calculate the instantaneous temperature without having done that.

This updates the tests for this and other OpenMM engine-based features to ensure that they work if the engine hasn't set its snapshot yet.

### DIFF
--- a/openpathsampling/engines/openmm/features/instantaneous_temperature.py
+++ b/openpathsampling/engines/openmm/features/instantaneous_temperature.py
@@ -18,7 +18,7 @@ def n_degrees_of_freedom(snapshot):
                                if system.getParticleMass(i) > 0*u.dalton])
     dofs_constaints = system.getNumConstraints()
     dofs_motion_removers = 0
-    if any(type(system.getForce(i)) == mm.CMMotionRemover 
+    if any(type(system.getForce(i)) == mm.CMMotionRemover
            for i in range(system.getNumForces())):
         dofs_motion_removers += 3
     dofs = dofs_particles - dofs_constaints - dofs_motion_removers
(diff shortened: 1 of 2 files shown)


dwhswenson:
This is ready for review and comment. I will leave it open for at least 24 hours, merging no earlier than Fri 23 Oct 10:00 local (08:00 GMT).

openpathsampling/engines/openmm/features/instantaneous_temperature.py
```diff
@@ -34,16 +34,24 @@ def instantaneous_temperature(snapshot):
...
-    old_snap = snapshot.engine.current_snapshot
-    snapshot.engine.current_snapshot = snapshot
-    state = snapshot.engine.simulation.context.getState(getEnergy=True)
+
+    # if no engine, error here; don't get caught in try/except below
+    engine = snapshot.engine
+    try:
+        old_snap = engine.current_snapshot
+    except Exception:  # openmm doesn't use a custom exception class yet
+        # Exception: Particle positions have not been set
```
sroet, line 43: This is a **very** broad catch (as you mention in the comments), I might have added
```suggestion
    except Exception as e:  # openmm doesn't use a custom exception class yet
        # Exception: Particle positions have not been set
        if "Particle positions" not in str(e):
            raise e
```
feel free to ignore though

sroet (approved):
I had one (maybe) improvement to make the broad `except Exception` slightly more secure. 

Feel free to ignore, LGTM otherwise

openpathsampling/engines/openmm/features/instantaneous_temperature.py
dwhswenson: Thanks. Not taking this because basically, whatever problem there is in loading current snapshot, that should be ignored here (could be missing velocities, etc.) It means the currently internal snapshot represetnation is not valid snapshot, so overriding it with the snapshot for which we want the temperature should not be a problem. The note is mainly for informational purposes (and I do wish that OpenMM returned some sort of `GetStateError`, instead of super-general `Exception`, but that's an issue for them.)

PR #940, opened 2020-10-31
SimStore: Tags

dwhswenson:
This PR adds a `tags` table to SimStore storages. As with `netcdfplus`, the `tags` table can be used to store arbitrary data. Any JSON-serializable object, or any object with a SimStore UUID, can be saved as a named tag.

This PR also includes some additional cleanup after #929 and miscellaneous support so that SimStore can be used in the OPS CLI (see also https://github.com/openpathsampling/openpathsampling-cli/pull/29).


dwhswenson:
This is ready for review and comment. I will leave it open for at least 48 hours, merging no earlier than Mon 02 Nov 15:00 GMT (16:00 local).

PR #941, opened 2020-11-07
Fix bug in SimStore when appending to non-existent files

dwhswenson:
Pretty much what the title says. Before, there was an error if you opened a new file as append, because it tried to read the non-existent schema from the file.

Also cleans up some deprecations from importing ABCs from `collections`, not `collections.abc` (SimStore is Python 3+.)

### DIFF
--- a/openpathsampling/experimental/simstore/test_sql_backend.py
+++ b/openpathsampling/experimental/simstore/test_sql_backend.py
@@ -94,9 +94,6 @@ class TestSQLStorageBackend(object):
         assert self._col_names_set('tables') == {'name', 'idx', 'module',
                                                  'class_name'}
         assert self._col_names_set('schema') == {'table', 'schema'}
-        for f in ['test1.sql', 'test2.sql']:
-            if os.path.isfile(f):
-                os.remove(f)
 
     def test_register_schema(self):
         new_schema = {
@@ -258,4 +255,3 @@ class TestSQLStorageBackend(object):
             expected = tuple(samp_dct[k]
                              for k in ['replica', 'ensemble', 'trajectory'])
             assert row[2:] == expected
-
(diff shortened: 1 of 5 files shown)


openpathsampling/experimental/simstore/test_sql_backend.py
```diff
@@ -79,13 +81,22 @@ def _col_names_set(self, table):
...
+        }[db]
+        table_names = database.engine.table_names()
         assert set(table_names) == self.default_table_names
         assert self._col_names_set('uuid') == {'uuid', 'table', 'row'}
         assert self._col_names_set('tables') == {'name', 'idx', 'module',
                                                  'class_name'}
         assert self._col_names_set('schema') == {'table', 'schema'}
+        for f in ['test1.sql', 'test2.sql']:
+            if os.path.isfile(f):
+                os.remove(f)
```
sroet, line 99: Shouldn't this be caught by your teardown?
```suggestion
```

openpathsampling/experimental/simstore/tools.py
```diff
@@ -5,7 +5,7 @@
 import logging
 logger = logging.getLogger(__name__)
 
-class SimpleNamespace(collections.MutableMapping):
+class SimpleNamespace(collections.abc.MutableMapping):
```
sroet, line 8: Any reason why this file does not do a similar thing as `memory_backend.py` (or the other way around)?
(Define `abc` and use that)

sroet (changes requested):
1 piece of excess code in the test and 1 question about coherence of `abc` between the different files inside `/simstore`

openpathsampling/experimental/simstore/tools.py
dwhswenson: No particular reason. I don't really have a consistent style when it comes to imports. I went ahead and switched this to just use `abc`. Really, I should have been using the [separate `abc` package for Py27](https://docs.python.org/2/library/abc.html), not `collections.MutableMapping`. I think the previous existence of that package is why I prefer `from collections import abc` (plus it is much shorter). On the other hand, I usually just do `import os` when using things in `os.path`, so there's not a general rule to apply.

sroet (approved):
LGTM

PR #871, opened 2019-11-01
Python 3.8 testing

dwhswenson:
This adds Python 3.8 to our testing matrix. I'm not sure how long it will take conda-forge to make it available, since I'm making this PR literally minutes after conda-forge/python merged the new version.

Resolves #864.

Issue #864, opened 2019-10-15
Python 3.8 testing plans


dwhswenson:
Python 3.8 has been released (although at this moment, not even conda-forge is serving it up). We plan to officially support Python 3.8 as soon as we can run our test suite on Python 3.8. We will maintain Python compatibility based on [NEP29](https://numpy.org/neps/nep-0029-deprecation_policy.html), meaning that we'll continue testing against Python 3.6 until at least 23 June 2020, after which point Python 3.6 support is no longer guaranteed. (Python 2.7 support will be dropped whenever OPS 2.0 is released in the first half of 2020.)

Here's a checklist of OPS testing requirements and whether we can install them into a Python 3.8 environment (empty at the moment of writing because I can't even create a non-pre-release Python 3.8 environment):

- [x] python
- [x] numpy
- [x] netcdf4
- [x] ujson
- [x] future
- [x] matplotlib
- [x] pyemma
- [x] openmmtools
- [x] networkx
- [x] scipy
- [x] setuptools
- [x] svgwrite
- [x] mdtraj
- [x] pandas
- [x] openmm
- [x] jupyter
- [x] nose
- [x] pytest
- [x] pytest-cov
- [x] python-coveralls* (switching from coveralls to python-coveralls)
- [ ] ipynbtest

sroet:
> We plan to officially support Python 3.8 as soon as we can run our test suite on Python 3.8.

Test run on python 3.8 is feasible at this moment (with skips for optional dependencies like `openmm` and `openmmtools`)

dwhswenson:
We actually distribute on 3.8 because conda-forge handles that automatically. I don't feel like managing different dependency installation procedures for different versions of Python, so I'm letting this sit.

dwhswenson:
At this point the only thing blocking Python 3.8 (and 3.9) is ipynbtest. I think we'll move the tested notebooked over to nbval instead -- I prefer the output from ipynbtest, but since it isn't maintained, better to use the maintained code.

PR #944, opened 2020-11-17
Fix SyntaxWarning (resolves #943).

### LINKED ISSUE #943: SyntaxWarning in Python 3.8
Hi! I'm looking into using OpenPathSampling. I imported the package while preparing to go through the tutorial and got a SyntaxWarning. This is non-fatal but it's an easy fix and I'll make a PR. This is related to #864 and #871.

```
>>> import openpathsampling
/.../openpathsampling/openpathsampling/engines/features/numpydoctools.py:207: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  if section is not 'HEAD':
```

Cause: https://docs.python.org/3/whatsnew/3.8.html#changes-in-python-behavior

bdice:
This PR resolves #943 by fixing a SyntaxWarning that affects Python >=3.8. Also affects #864, #871.

### DIFF
--- a/openpathsampling/engines/features/numpydoctools.py
+++ b/openpathsampling/engines/features/numpydoctools.py
@@ -204,7 +204,7 @@ class NumpyDocTools(object):
             while doc and doc[0] == '':
                 doc.pop(0)
 
-            if section is not 'HEAD':
+            if section != 'HEAD':
                 cap = ' '.join([s[0].upper() + s[1:].lower() for s in section.split(' ')])
                 doc = [
                     cap,


dwhswenson (approved):
Thanks for the fix! Since you mention in #943 that you're starting out with OPS, a quick heads-up -- we're planning to do a little online tutorial (YouTube live stream) on some new aspects on OPS sometime in early December (I need to finish a few things to release 1.4 first). Topics will include the new storage and/or the command line interface, both of which will be especially useful if you're getting started on a new project. Check the issues here (I'll pin an issue when date/time are set) or watch the OPS Twitter (<span>@</span>pathsampling) for details!

bdice:
@dwhswenson Wonderful, thanks! 😄

PR #946, opened 2020-11-17
fix for filname counter of gromacs engine (fixes #942)

### LINKED ISSUE #942: gromacs engine broken
The gromacs engine in the version on the current master is broken. Tests pass, however running the example notebooks does not work. The generated trial trajectories are continously overwritten. The reason is that using the default FileNameSetter (the one which uses numbers) is reset to the current number every time the `engine.set_filenames()` method is called. See [this line](https://github.com/openpathsampling/openpathsampling/blob/fc27a9d34fba38adea194e0f24bbf76111f58ede/openpathsampling/engines/gromacs/engine.py#L306).
Using the `RandomStringFilenames` instead of the standard `FileNameSetter` it is possible to work around the issue and run the example notebooks.

I can (and will) submit a PR with a fix, however I need your input first as I am not sure what the desired behaviour is. If I am not mistaken there are 3 possible solutions:
1. change the reset to use given number +1 (results in [this test](https://github.com/openpathsampling/openpathsampling/blob/fc27a9d34fba38adea194e0f24bbf76111f58ede/openpathsampling/tests/test_gromacs_engine.py#L221) failing)
2.  change the FileNameSetter to return the incremented value (instead of incrementing and returning the pre-increment value); essentially the same as above
3. Remove the reset completely (this way engines could start with a high trajectory number and setting the filename would have no effect for future names, but tests pass)

I tend to solution 1, changing the test slightly to pass (set the filname to -1 instead of 0).

As a side note: I think it would be good to include the `-nt 1` flag in the mdrun_args for the tests, i.e. forcing it to run with one thread (which is also the number of domains the system will be split into). This prevents gromacs crashing (and the test in turn never finishing) on machines with many cores, since gromacs is naturaly not able to find a domain decomposition that splits the small test system into enough parts. I will add this to the PR fixing this issue in case you agree.

hejung:
Closes #942. As proposed there I slightly modified the test to reset the counter via its own `reset()` method instead of via the engines `set_filenames()` method. 
Engine works and test pass with and without the counter reset in `set_filenames()`, so it is now a matter of preference if we want to have that reset in `set_filenames()`. I left it in for now, was not sure what your preference is here.

### DIFF
--- a/openpathsampling/engines/gromacs/engine.py
+++ b/openpathsampling/engines/gromacs/engine.py
@@ -303,7 +303,7 @@ class GromacsEngine(ExternalEngine):
             num_str = '{:07d}'.format(number + 1)
             self.output_file = self.trajectory_filename(number + 1)
             init_filename = "initial_frame.trr"
-            self.filename_setter.reset(number)
+            self.filename_setter.reset(number + 1)
         else:
             num_str = number
             self.output_file = self.trajectory_filename(num_str)
--- a/openpathsampling/tests/test_gromacs_engine.py
+++ b/openpathsampling/tests/test_gromacs_engine.py
@@ -90,7 +90,7 @@ class TestGromacsEngine(object):
         self.engine = Engine(gro="conf.gro",
                              mdp="md.mdp",
                              top="topol.top",
-                             options={},
+                             options={'mdrun_args': '-nt 1'},
                              base_dir=self.test_dir,
                              prefix="project")
 
@@ -227,7 +227,7 @@ class TestGromacsEngine(object):
 
         traj_0 = self.engine.trajectory_filename(0)
         snap = self.engine.read_frame_from_file(traj_0, 0)
-        self.engine.set_filenames(0)
+        self.engine.filename_setter.reset(0)
 
         ens = paths.LengthEnsemble(5)
         traj = self.engine.generate(snap, running=[ens.can_append])


hejung:
> I think it makes sense to keep it. I think this difference is this: with `reset`, you can restart your engine with `engine.set_filenames(last_num)`; without `reset`, that does not work (but you can always restart with `engine.filename_setter.reset(last_num + 1)`). Is that correct? If so, I'd say that keeping the `engine.set_filenames()` approach is probably needed to maintain the API.

Yes, if by restart you mean creating a new gromacs engine object. When loading the old engine from storage the count is preserved, i.e. no reset is necessary to continue with a loaded gromacs engine.

> 
> [Also, I forgot to mention that I was +1 on the `-nt 1` for tests. I feel like you mentioned this at some point long ago, and I agreed, but then never put it in. Thanks for that!]
Time is relative, I think I first brought it up 2 days ago :)

dwhswenson:
> When loading the old engine from storage the count is preserved, i.e. no reset is necessary to continue with a loaded gromacs engine.

It does? Across sessions? I honestly don't remember writing code that would do that, and don't see anything at a quick glance.

> Time is relative, I think I first brought it up 2 days ago :)

Haha... maybe it was someone else using the Gromacs engine who mentioned it? Can't find it in issues here, so must have been in person. Maybe @gyorgy-hantal when we were both visiting the UvA.

Apparently I'm getting old enough that my memory is failing in general!

gyorgy-hantal:
> Haha... maybe it was someone else using the Gromacs engine who mentioned it? Can't find it in issues here, so must have been in person. Maybe @gyorgy-hantal when we were both visiting the UvA.
> 
Nice, I wasn't aware this was fixed - for sure not in the version I use (released before Feb 2020) ... maybe time to update... :-D

PR #947, opened 2020-11-28
SimStore: Type identification

dwhswenson:
SimStore storable functions (CVs) no longer need to have result type provided. Instead, is it identified when the results are first saved. This changes some of the SimStore API, which is only relevant to anyone already using SimStore-based CVs (which are not only experimental, but not even in a release yet.)


dwhswenson:
This is ready for review and comment. I will leave it open for at least 48 hours, merging no earlier than Mon 30 Nov 13:00 GMT (14:00 local).

Issue #558, opened 2016-08-12
Reject trajectories that generate `NaN`


jchodera:
I'm only as far as repeatedly running into `nan` Exceptions with the bootstrapping scheme of https://github.com/choderalab/ligand-binding-pathsampling/pull/3, but it would be useful to have trajectories with `nan`s automatically rejected or else OpenPathSampling will be almost guaranteed to crash after running for long enough---especially at high temperature.
```
DONE! Completed Bootstrapping cycle step 80 in ensemble 1/5.
Working on Bootstrapping cycle step 81 in ensemble 1/5 .
Working on Bootstrapping cycle step 82 in ensemble 1/5 .
Working on Bootstrapping cycle step 83 in ensemble 1/5 .
Working on Bootstrapping cycle step 84 in ensemble 1/5 .
Working on Bootstrapping cycle step 85 in ensemble 1/5 .
Working on Bootstrapping cycle step 86 in ensemble 1/5 .
Working on Bootstrapping cycle step 87 in ensemble 1/5 .
Working on Bootstrapping cycle step 88 in ensemble 1/5 .
Working on Bootstrapping cycle step 89 in ensemble 1/5 .
Working on Bootstrapping cycle step 90 in ensemble 1/5 .
Working on Bootstrapping cycle step 91 in ensemble 1/5 .
Working on Bootstrapping cycle step 92 in ensemble 1/5 .
Working on Bootstrapping cycle step 93 in ensemble 1/5 .
Traceback (most recent call last):
  File "setup-ops.py", line 168, in <module>
    initial_sample_set = bootstrap.run()
  File "/Users/choderaj/miniconda/lib/python2.7/site-packages/openpathsampling/pathsimulator.py", line 456, in run
    bootstrap.run(n_steps_per_round)
  File "/Users/choderaj/miniconda/lib/python2.7/site-packages/openpathsampling/pathsimulator.py", line 262, in run
    movepath = bootstrapmove.move(self.sample_set)
  File "/Users/choderaj/miniconda/lib/python2.7/site-packages/openpathsampling/pathmover.py", line 2385, in move
    subchange = self.mover.move(sample_set)
  File "/Users/choderaj/miniconda/lib/python2.7/site-packages/openpathsampling/pathmover.py", line 1922, in move
    mover.move(sample_set),
  File "/Users/choderaj/miniconda/lib/python2.7/site-packages/openpathsampling/pathmover.py", line 2245, in move
    movepath = mover.move(subglobal)
[...]

PR #954, opened 2020-12-24
fix omt is None error in test suite

sroet:
If `OpenMM` is installed, but `OpenMMTools` isn't,  the test `TestFeatures.test_copy_with_replacement_openmm`
fails with:
```python
    def test_copy_with_replacement_openmm(self):
        if not paths.integration_tools.HAS_OPENMM:
            raise SkipTest
        # test an openmm snapshot
>       sys = omt.testsystems.AlanineDipeptideVacuum()
E       AttributeError: 'NoneType' object has no attribute 'testsystems'
```

This is due to the code-block at the start of that file:
```
try:                                                                            
    import openmmtools as omt                                                   
except ImportError:                                                             
    omt = None   
```

This alters the test such that it is also skipped if `omt is None`

Also cleans up some `pep8` complains in that file (mostly unused imports)

### DIFF
--- a/openpathsampling/tests/test_features.py
+++ b/openpathsampling/tests/test_features.py
@@ -1,12 +1,10 @@
 from __future__ import absolute_import
 from builtins import hex
 from builtins import object
-from nose.tools import (assert_equal, assert_not_equal,
-                        assert_almost_equal, raises, assert_in)
+from nose.tools import raises
 
-from nose.plugins.skip import Skip, SkipTest
-from .test_helpers import (true_func, assert_equal_array_array,
-                           make_1d_traj, assert_items_equal, u)
+from nose.plugins.skip import SkipTest
+from .test_helpers import u
 
 import logging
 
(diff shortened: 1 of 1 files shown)


dwhswenson:
Good catch; will merge after #955 so that the version number on the main branch is correct.

Our testing didn't catch this because we only test with no integrations or "all" integrations ("all" in quotes because some integrations, e.g., Gromacs, are only tested indirectly). This is because it is just too much of a pain in CI to test each integration separately. But it's possible that, as here, there's a problem when only part of a set of related integrations is installed.

However, as we switch to GitHub Actions, we might try adding scheduled actions (maybe weekly?) to check various individual integrations. It could give us better coverage of user environments without adding burden to the CI for the normal development cycle.

Final note: We usually use OpenMMTools for its integrators. Beware of the built-in OpenMM integrators. They're not reversible, and that's not only theoretically wrong for path sampling, but it actually leads to real problems. See the extensive discussion in #558.

PR #957, opened 2020-12-27
Fix so SimStore tests pass without simtk.unit

dwhswenson:
Without `simtk.unit` installed, SimStore wasn't working. We only test SimStore with full integrations (similar problem to #954).

This also adds an `extras` for SimStore, so (in future releases) `pip install openpathsampling[simstore]` will install the extra requirements for SimStore as well.

### DIFF
--- a/openpathsampling/experimental/simstore/test_custom_json.py
+++ b/openpathsampling/experimental/simstore/test_custom_json.py
@@ -4,7 +4,6 @@ import pytest
 
 import numpy as np
 from numpy import testing as npt
-from simtk import unit
 
 from . import test_utils
 
--- a/openpathsampling/experimental/storage/ops_storage.py
+++ b/openpathsampling/experimental/storage/ops_storage.py
@@ -40,7 +40,11 @@ from ..simstore import SQLStorageBackend  # TODO: generalize
 from . import snapshots
 from .snapshots_table import SnapshotsTable
 
-from .simtk_unit import simtk_quantity_codec, SimtkQuantityHandler
+try:
+    from .simtk_unit import simtk_quantity_codec, SimtkQuantityHandler
+except ImportError:
+    simtk_quantity_codec = None
+    SimtkQuantityHandler = None
 
 import logging
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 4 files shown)


dwhswenson:
This is ready for review and comment. I will leave it open for at least 24 hours, merging no earlier than Mon 28 Dec 14:00 GMT (15:00 local).

NB: This particular issue is blocking the next release of the OPS CLI (which will include SimStore support!) and so I'll probably go ahead and release a 1.4.1 shortly after this is in.

sroet:
LGTM, 1 quick question:
I don't see where `pip install openpathsampling[simstore]` would install `simtk`? (I only see `sqlalchemy` and `dill` added to the `cfg`)

Is `simtk` only a testing requirement for `SimStore` or did you forget to add it to the `cfg`? (just making sure it was intentional, I am not too familiar with the internals of simstore anymore)

dwhswenson:
> Is `simtk` only a testing requirement for `SimStore` or did you forget to add it to the `cfg`? (just making sure it was intentional, I am not too familiar with the internals of simstore anymore)

Sorry, the PR message wasn't clear. Two separate issues here:

1. Using SimStore should *not* require that `simtk.unit` is present in the environment (you should be able to use and test the new storage without OpenMM installed). But you couldn't use `experimental.Storage` without importing `simtk.unit` -- that's the main thing this PR fixes.

2. As a separate issue, currently I say "if you want to use SimStore, you need to install a couple extra packages..." --  adding the `simstore` extra in setup.cfg makes that process a little more elegant (at least, in some use cases).

sroet:
>Sorry, the PR message wasn't clear. Two separate issues here:

>    Using SimStore should not require that simtk.unit is present in the environment (you should be able to use and test the new storage without OpenMM installed). But you couldn't use experimental.Storage without importing simtk.unit -- that's the main thing this PR fixes.

>    As a separate issue, currently I say "if you want to use SimStore, you need to install a couple extra packages..." -- adding the simstore extra in setup.cfg makes that process a little more elegant (at least, in some use cases).

Ah, thanks for the clarification! All looks good to me then :)

PR #958, opened 2020-12-27
Remove nose from test_helpers

dwhswenson:
Small step in the direction of #756.

I have other projects that are built on OPS (e.g., the OPS CLI). Sometimes it can be convenient to use some of the tools in `openpathsampling.tests.test_helpers`, such as `make_1d_traj`. However, `test_helpers` required importing `nose`, and so every project that used `make_1d_traj` had `nose` as a requirement just so I could get that one function, even though `nose` support in OPS is deprecated. 🙄

This fixes that situation.

(Note also that many of the functions in `test_helpers` are probably better replaced by tools in `numpy.testing`. I recommend using those rather than our own hacky implementations.)

### DIFF
--- a/openpathsampling/tests/test_helpers.py
+++ b/openpathsampling/tests/test_helpers.py
@@ -23,8 +23,6 @@ try:
 except ImportError:
     md = None
 
-from nose.tools import assert_equal, assert_in, assert_true
-#from nose.tools import assert_items_equal
 from pkg_resources import resource_filename
 
 import openpathsampling as paths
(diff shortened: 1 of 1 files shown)


dwhswenson:
This is ready for review and comment. I will leave it open for at least 24 hours, merging no earlier than Mon 28 Dec 14:00 GMT (15:00 local).

openpathsampling/tests/test_helpers.py
```diff
@@ -306,18 +304,18 @@ def _wrapper(*args, **kwargs):
...
+    assert len(truth.columns) == len(beauty.columns)
+    assert set(truth.index) == set(beauty.index)
+    assert set(truth.columns) == set(beauty.columns)
     for idx in truth.index:
         for col in truth.columns:
             truth_val = truth.loc[idx, col]
             beauty_val = beauty.loc[idx, col]
             if np.isnan(truth_val):
-                assert_true(np.isnan(beauty_val))
+                assert np.isnan(beauty_val)
```
sroet, line 316: Huh, I never knew `assert_true` accepted `truthy` values, I thoughed it checked for `is True`.

But looking at the code for `assert_true`
```python
   def assertTrue(self, expr, msg=None):                                       
        """Check that the expression is true."""                                
        if not expr:                                                            
            msg = self._formatMessage(msg, "%s is not true" % safe_repr(expr))  
            raise self.failureException(msg)                                    
```

 this is indeed the correct replacement.

sroet (approved):
LGTM, was surprised by one line but the new implementation is correct

PR #962, opened 2021-01-17
Bump version to 1.4.2.dev0

dwhswenson:
Apperently I forgot to do this after the 1.4.1 release.


dwhswenson:
After significant trial and error to debug, it looks like `jupyter-client` has changed in a way that isn't backward-compatible with ipynbtest. Since ipynbtest is not maintained, the pin here is a short term solution, and the longer-term solution is to switch to nbval for notebook tests. Will merge this now.

PR #963, opened 2021-01-21
Fix broken link in Readme.md of toy mstis examples

sroet:
When looking through the repository, I noticed that the link `404`ed


dwhswenson (approved):
Weird that I managed to not fill in that URL! Thanks for the fix. Currently re-running tests (the coverage drop should have been fixed by #964); will merge after tests run.

sroet:
Currently re-running tests (the coverage drop should have been fixed by #964); will merge after tests run.

I don't think that actually runs on current master (as it is still my (unupdated) branch), will merge master to trigger another run

PR #964, opened 2021-01-22
Make more deterministic test

sroet:
In the doc update #963, I saw from codecov that the following line was not reliably hit in `shooting.py`:
Screenshot from 2021-01-22 09-18-41
 
This adds a test to force that code path in the first commit: https://github.com/openpathsampling/openpathsampling/commit/1842ed9fd04bda947cba7c01f7d7016aaf9a5269

The second commit (https://github.com/openpathsampling/openpathsampling/commit/31e11bf38db860cdcec4f74561dc90ec4b614e60) drops `nose` in the `test_shooting.py` and uses `pytest` instead as an incremental update towards #756 

Edit: (https://github.com/openpathsampling/openpathsampling/pull/964/commits/dc6e428b22e106707e6fdc528fdc30e62e5ac83d) removes the duplicate of the test added in the first commit (I had copied it to a different spot around in the first commit, but forgot to remove the original one :facepalm: )

### DIFF
--- a/openpathsampling/tests/test_shooting.py
+++ b/openpathsampling/tests/test_shooting.py
@@ -1,11 +1,8 @@
 from builtins import object
 import collections
-from nose.tools import assert_equal, assert_almost_equal, raises
-from nose.plugins.skip import Skip, SkipTest
+from nose.tools import assert_equal, raises
 from openpathsampling.tests.test_helpers import (
-    assert_equal_array_array, assert_not_equal_array_array, make_1d_traj,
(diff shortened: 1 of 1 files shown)


dwhswenson (approved):
LGTM. I think I mentioned recently that there may be a couple tests that are flaky with regards to coverage -- that is, some small fraction of lines are sometimes covered, sometimes not. This is one are where I remember that being the case. On Coveralls we allowed some small drop; we don't (yet) on CodeCov. Thanks for the fix, and the update to pytest-style!

Issue #922, opened 2020-07-15
Error when loading a saved committor simulation storage file 


mayank-agr:
I am getting errors when I load any saved .nc file after committor simulation. It works perfectly after a TPS simulation but never works after a committor simulation. Can anyone please help me with this error?

I appreciate it
Thanks,
Mayank

The error I am getting is "ValueError: Value is too big!"  and the full error is as follows:

```pytb
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-4-9c6cf6fe1e43> in <module>
----> 1 output_storage = paths.AnalysisStorage("final_committor_113_500each.nc")

~/data/magrawa3/softwares/miniconda3/envs/env_py3.6/lib/python3.6/site-packages/openpathsampling/storage/storage.py in __init__(self, filename, caching_mode)
    499 
    500         # Let's go caching
--> 501         AnalysisStorage.cache_for_analysis(self)
    502 
    503     @staticmethod

~/data/magrawa3/softwares/miniconda3/envs/env_py3.6/lib/python3.6/site-packages/openpathsampling/storage/storage.py in cache_for_analysis(storage)
    532             store = getattr(storage, store_name)
    533             with AnalysisStorage.CacheTimer('Cache all objects', store):
--> 534                 store.cache_all()
    535 
    536     class CacheTimer(object):

~/data/magrawa3/softwares/miniconda3/envs/env_py3.6/lib/python3.6/site-packages/openpathsampling/netcdfplus/stores/variable.py in cache_all(self, part)
     99             data = zip(*[
    100                 self.vars[var][part]
--> 101                 for var in self.var_names
    102             ])
    103 

~/data/magrawa3/softwares/miniconda3/envs/env_py3.6/lib/python3.6/site-packages/openpathsampling/netcdfplus/stores/variable.py in <listcomp>(.0)
     99             data = zip(*[
    100                 self.vars[var][part]
--> 101                 for var in self.var_names
    102             ])
    103 

~/data/magrawa3/softwares/miniconda3/envs/env_py3.6/lib/python3.6/site-packages/openpathsampling/netcdfplus/netcdfplus.py in __getitem__(self, key)
    123             # print(self.variable[key])
    124             # print(type(self.variable[key]))
--> 125             return self.getter(self.variable[key])
    126 
    127         def __getattr__(self, item):

[...]

PR #965, opened 2021-01-22
Update ujson pin

dwhswenson:
With the [release of ujson 4.0.2](https://github.com/ultrajson/ultrajson/releases/tag/4.0.2), we should be able to relax the pin of ujson without regressing on #922. Since that release is very new, I'm not going to assume that it is in all distribution channels yet, so the pin is kind of complicated (still allowing ujson <2 while also allowing >=4.0.2).

By the next release, I hope we'll be able to just pin `ujson >=4.0.2`.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -36,7 +36,7 @@ install_requires =
     svgwrite
     networkx
     matplotlib
-    ujson<2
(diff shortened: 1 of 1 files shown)


dwhswenson:
Forgotten I'd done this; somewhat surprised that it passed. Will go ahead and merge.

PR #972, opened 2021-02-04
Fix problem with Histogram.bin_range not getting set

dwhswenson:
This fixes an issue with `Histogram.bin_range` not getting set correctly. The bug showed up as a histogram not having the same parameters as its `empty_copy()` (which was preventing the combination of multiple histograms into one with `Histogram.sum_histograms`).


sroet:
Test failure is genuine, looking at the docstring/test the following diff is needed (I can't propose suggestions to unaltered code):
```diff
--- a/openpathsampling/numerics/histogram.py
+++ b/openpathsampling/numerics/histogram.py
@@ -226,7 +226,7 @@ class Histogram(SparseHistogram):
         used, then `bin_range` is required. If `n_bins` is used, then
         `bin_range` is optional. `n_bins` overrides `bin_width`.
 
-        If no options are given, the default is to use 40 bins and the
+        If no options are given, the default is to use 20 bins and the
         range generated by np.histogram.
         """
         # this is to compare whether another histogram had the same setup,
@@ -250,6 +250,7 @@ class Histogram(SparseHistogram):
         else:
             if n_bins is not None:
                 self.n_bins = n_bins
+                self.bin_width = None # Override if given
             else:
                 self.n_bins = 20 # default
             self.bins = self.n_bins
```
This overrides `self.bin_width` if `n_bins` is given (as stated in the docstring and leads to the test failure), this also updates the docstring to the right default number (as is tested against)

dwhswenson:
This is ready for review and comment. I will leave it open for at least 24 hours, merging no earlier than Tue 23 Feb 20:00 GMT (21:00 local).

sroet (approved):
This looks good to me

PR #974, opened 2021-02-14
fix numpy 1.20 DeprecationWarnings

sroet:
With the release of `numpy` `1.20` came a couple `DeprecationWarning`s, mainly about `np.float` or `np.bool` just being an alias for the python internal. The one Warning that is left (in my test setup) comes from `openmmtools` and is handled in https://github.com/choderalab/openmmtools/pull/497

### DIFF
--- a/openpathsampling/netcdfplus/netcdfplus.py
+++ b/openpathsampling/netcdfplus/netcdfplus.py
@@ -877,7 +877,7 @@ class NetCDFPlus(netCDF4.Dataset):
             setter = lambda v: np.array(v)
 
         elif var_type == 'bool':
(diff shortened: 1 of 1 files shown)


dwhswenson (approved):

dwhswenson:
Huh... [Release notes](https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated) on this say (emphasis mine):

> For a ***long time***, `np.int` has been an alias of the builtin `int`.

I'm pretty sure I distinctly remember my confusion when:

```python
>>> np.True is True
False
```

I must be getting old.

LGTM.

PR #975, opened 2021-02-17
Cache AllOutX volume

dwhswenson:
There is no reason to create a new volume every time we use `AllOutXEnsemble.can_append` etc.

PR #976, opened 2021-02-19
also cache the ~volume for PartOutX

sroet:
While diving further down with #892 I saw a similar solution as #975 for `PartOutX`. This also caches that volume

### DIFF
--- a/openpathsampling/ensemble.py
+++ b/openpathsampling/ensemble.py
@@ -2284,14 +2284,16 @@ class PartOutXEnsemble(PartInXEnsemble):
     """
     Ensemble of trajectories with at least one frame outside the volume
     """
-
-    def _str(self):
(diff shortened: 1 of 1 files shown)


sroet:
@dwhswenson the coverage drop seems unrelated? Anyway, this is ready for a review

dwhswenson (approved):
LGTM. We should probably add a small allowance for flaky coverage. Re-running fixed.

BTW, tonight I'm playing with managing GitHub stuff from the phone, so if my tone differs, that's why.

sroet:
> BTW, tonight I'm playing with managing GitHub stuff from the phone, so if my tone differs, that's why.

Cool, good luck. I will be opening an interesting PR later tonight (should reduce the number of Trajectory objects that are generated by 75% (for my test system))

PR #977, opened 2021-02-19
kill off getslice

sroet:
As `list.__getslice__()` has been deprecated since python `2.0` and [killed since `3.0`](https://docs.python.org/3/whatsnew/3.0.html?highlight=list%20__getslice__#operators-and-special-methods) this also removes the `Trajectory` override of it.

### DIFF
--- a/.autorelease/test-testpypi
+++ b/.autorelease/test-testpypi
@@ -1,0 +1,4 @@
+#!/bin/bash
+
+python -m pip install pytest nose
+py.test --pyargs openpathsampling.tests
--- a/.github/workflows/autorelease-default-env.sh
+++ b/.github/workflows/autorelease-default-env.sh
@@ -1,4 +1,4 @@
-INSTALL_AUTORELEASE="python -m pip install autorelease==0.2.3"
+INSTALL_AUTORELEASE="python -m pip install autorelease==0.3.0"
 if [ -f autorelease-env.sh ]; then
     source autorelease-env.sh
 fi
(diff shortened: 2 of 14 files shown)


sroet:
@dwhswenson Do we want to keep python 2.7 support alive here (I don't have a python 2 environment anymore). If so, feel free to close this PR

dwhswenson:
Our official commitment is to support Py27 until
OPS 2.0. This is to allow PhD students who started in 2.7 to finish with working code. (Hoping that's mainly @arjunwadhawan, who has now submitted his thesis!) Could you close and target this against `dev-2.0` instead?

sroet:
@dwhswenson Github allows to change the base branch :smile: , so this should not be correct

sroet:
do you want me to properly rebase onto dev-2.0, or don't you mind pulling the master/dev-2.0 diffs as well?

dwhswenson:
Doesn't matter. I've been doing the occasional merge of master into dev-2.0. If I've fallen a bit behind, no worries if your PR catches me up. As long as contributors get the appropriate credit (which (1) they would with this; (2) not an issue in this specific case) then I'm happy.

dwhswenson (approved):
LGTM. There's a drop in coverage largely due to removing Py2.7 from the testing matrix (some code paths no longer used). I'll follow this up with a PR removing some of those code paths.

Related: note that part of the [roadmap to 2.0](https://github.com/openpathsampling/openpathsampling/wiki/Roadmap) includes modifying trajectory so that it no longer inherits directly from `list` (instead wrapping a list and inheriting from `collections.abc.Sequence`).

PR #978, opened 2021-02-19
Prevent excessive trajectory slicing

sroet:
Some more UUID optimization coming from #892.

So while running the test uuid system from #892, I noticed **a lot** of `Trajectory` objects getting generated. Looking at the code, **every time you slice a Trajectory, you generate a new Trajectory**.

To test how many and were I added the following code to trajectory.py (and where) 
```diff
+++ b/openpathsampling/engines/trajectory.py
@@ -23,6 +23,7 @@ class Trajectory(list, StorableObject):
     """
 
     engine = None
+    kill_count = 1000000
 
     def __init__(self, trajectory=None):
         """
@@ -45,6 +46,10 @@ class Trajectory(list, StorableObject):
             else:
                 self.extend(trajectory)
 
+        Trajectory.kill_count -= 1
+        if Trajectory.kill_count < 0:
+            raise RuntimeError("KILLED")
+
```

and ran the following script:
```python
sset = testscheme.initial_conditions_from_trajectories([eng_trj, eng_trj2,
                                                        eng_trj3],
                                                        strategies=[
                                                            'split',
                                                            'extend-complex',
                                                            'extend-minimal'])
# Error out early
sset.sanity_check()
testscheme.assert_initial_conditions(sset)

#Equilibrate
equilibration = paths.PathSampling(
    storage=None,
    sample_set=sset,
    move_scheme=testscheme
)
initial = 100000
paths.Trajectory.kill_count=initial
equilibration.run(10)
print("trajs_generated = "+str(initial-paths.Trajectory.kill_count))
```
(important part here is that I only look at the number of trajectories created after initialization)

Here is a table of number of trajectories created from running this 6 times on `master` and 6 times on this branch
|  | main| this PR |
| --- | --- | --- | 
||77857|20728|
||64553|22852|
||73042|14508|
||76004|18205|
||75349|27016|
||82440|33964|
|*mean*| 74874 | 22879 |

This PR saves **70%** of all Trajectory objects generated (during sampling). `23_000` still an insane amount in my opinion for just `10` mc steps, but this is at least a decent step into (hopefully) not running into #892 as often.

In the long term; sliced trajectories should probably return read_only views (without an uuid)  and not new Trajectories

### DIFF
--- a/openpathsampling/ensemble.py
+++ b/openpathsampling/ensemble.py
@@ -43,6 +43,34 @@ def join_ensembles(ensemble_list):
     return ensemble
 
 
+def _get_list_traj(trajectory):
+    """Return a list of (proxy) snapshots from either a list or Trajectory
+
+    Parameters
+    ----------
+    trajectory : :class:`.Trajectory` or :class:`.list`
+       trajectory or list to convert into a list of snapshots
+
+    Returns
+    -------
+    :class:`.list`
+       list of (proxy) snapshots
+
+    Note
+    ----
+    Due to possible UUID space restrictions, we don't want to slice
+    Trajectories if we can help it (as this will generate a new Trajectory
+    object with its own UUID). This is a convenience function that will either
+    turn a Trajectory into a list of (proxy) snapshots or just list to list
+    mapping, which you can slice without generating a new UUID
+
+    For a more in-depth discussion, please see:
+    https://github.com/openpathsampling/openpathsampling/pull/978
+    """
+    itraj = getattr(trajectory, 'iter_proxies', trajectory.__iter__)
+    return list(itraj())
+
+
 # note: the cache is not storable, because that would just be silly!
 class EnsembleCache(object):
     """Object used by ensembles to enable fast algorithms for basic functions.
@@ -1601,9 +1629,8 @@ class SequentialEnsemble(Ensemble):
 
         subtraj_first = 0
         subtraj_i = 0
-        # Don't load proxies if this gets a trajectory
-        itraj = getattr(trajectory, 'iter_proxies', trajectory.__iter__)
-        ltraj = list(itraj())
+        # Make a list before slicing
+        ltraj = _get_list_traj(trajectory)
         while subtraj_i < len(self.ensembles):
             subtraj_final = transitions[subtraj_i]
             subtraj = ltraj[slice(subtraj_first, subtraj_final)]
@@ -1632,10 +1659,8 @@ class SequentialEnsemble(Ensemble):
             subtraj_final = max(last_checked, subtraj_first)
         traj_final = len(traj)
         ens = self.ensembles[ens_num]
-        # Don't load proxies if this gets a trajectory
-        itraj = getattr(traj, 'iter_proxies', traj.__iter__)
-        ltraj = list(itraj())
-
+        # Make a list before slicing
+        ltraj = _get_list_traj(traj)
         subtraj = ltraj[slice(subtraj_first, subtraj_final + 1)]
 
         # if we're in the ensemble or could eventually be in the ensemble,
@@ -1666,9 +1691,9 @@ class SequentialEnsemble(Ensemble):
             subtraj_first = min(last_checked, subtraj_final - 1)
         traj_first = 0
         ens = self.ensembles[ens_num]
-        itraj = getattr(traj, 'iter_proxies', traj.__iter__)
-        ltraj = list(itraj())
 
+        # Make a list before slicing
+        ltraj = _get_list_traj(traj)
         subtraj = ltraj[slice(subtraj_first, subtraj_final)]
         logger.debug("*Traj slice " + str(subtraj_first) + " " +
                      str(subtraj_final) + " / " + str(len(traj)))
@@ -1721,9 +1746,8 @@ class SequentialEnsemble(Ensemble):
 
         traj_final = len(trajectory)
         final_ens = len(self.ensembles) - 1
-        # Prevent proxy loading and trajectroy slicing
-        itraj = getattr(trajectory, 'iter_proxies', trajectory.__iter__)
-        ltraj = list(itraj())
+        # Make a list before slicing
+        ltraj = _get_list_traj(trajectory)
         # print traj_final, final_ens
         # logging startup
         if cache.debug_enabled:  # pragma: no cover
@@ -1892,9 +1916,8 @@ class SequentialEnsemble(Ensemble):
         subtraj_final = len(trajectory)
         ens_final = len(self.ensembles) - 1
         ens_num = ens_final
-        # Prevent proxy loading and trajectroy slicing
-        itraj = getattr(trajectory, 'iter_proxies', trajectory.__iter__)
-        ltraj = list(itraj())
+        # Make list before slicing
+        ltraj = _get_list_traj(trajectory)
 
         if self._use_cache:
             _ = cache.check(trajectory)
@@ -2237,8 +2260,8 @@ class AllInXEnsemble(VolumeEnsemble):
             logger.debug("Untrusted VolumeEnsemble " + repr(self))
             # logger.debug("Trajectory " + repr(trajectory))
             # This can sometimes get a list instead of a Trajectory
-            itraj = getattr(trajectory, 'iter_proxies', trajectory.__iter__)
-            for frame in itraj():
+            # Make sure this is a proxy list
+            for frame in _get_list_traj(trajectory):
                 if not self._volume(frame):
                     return False
             return True
@@ -2297,8 +2320,7 @@ class PartInXEnsemble(VolumeEnsemble):
         trajectory : :class:`openpathsampling.trajectory.Trajectory`
             The trajectory to be checked
         """
-        itraj = getattr(trajectory, 'iter_proxies', trajectory.__iter__)
-        for frame in list(itraj()):
+        for frame in _get_list_traj(trajectory):
             if self._volume(frame):
                 return True
         return False
@@ -2325,8 +2347,7 @@ class PartOutXEnsemble(PartInXEnsemble):
 
     def __call__(self, trajectory, trusted=None, candidate=False):
         # Don't load proxies if this is a Trajectory
-        itraj = getattr(trajectory, 'iter_proxies', trajectory.__iter__)
-        for frame in itraj():
+        for frame in _get_list_traj(trajectory):
             if self._volume(frame):
                 return True
         return False
--- a/openpathsampling/ensemble.py
+++ b/openpathsampling/ensemble.py
@@ -1601,7 +1601,9 @@ class SequentialEnsemble(Ensemble):
 
         subtraj_first = 0
         subtraj_i = 0
-        ltraj = list(trajectory)
+        # Don't load proxies if this gets a trajectory
+        itraj = getattr(trajectory, 'iter_proxies', trajectory.__iter__)
+        ltraj = list(itraj())
         while subtraj_i < len(self.ensembles):
             subtraj_final = transitions[subtraj_i]
             subtraj = ltraj[slice(subtraj_first, subtraj_final)]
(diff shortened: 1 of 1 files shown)


openpathsampling/ensemble.py
```diff
@@ -1636,12 +1641,15 @@ def _find_subtraj_final(self, traj, subtraj_first, ens_num,
         # logger.debug("Can-app " + str(ens.can_append(subtraj, trusted=True)))
         # logger.debug("Call    " + str(ens(subtraj, trusted=True)))
         # TODO: the weird while condition is handling the OVERSHOOTING
```
sroet, line 1643: I don't know if this TODO has now to be moved to the if_statemt

sroet:
@dwhswenson this is ready for a review. As you probably know this code better than I do, are there any other places where a similar strategy might be beneficial? (looking at it quickly, I can probably mimic this pretty easy for `_find_subtraj_first`, but that function has not popped up yet when I try to kill on Trajectory generation)

openpathsampling/ensemble.py
```diff
@@ -1624,6 +1625,10 @@ def _find_subtraj_final(self, traj, subtraj_first, ens_num,
         traj_final = len(traj)
         ens = self.ensembles[ens_num]
         subtraj = traj[slice(subtraj_first, subtraj_final + 1)]
+
+        # Every trajectory slice generates a new object which is exsessive
+        # So we pre-slice here and ininitialse an iterator to be used later
+        exttraj = traj[slice(subtraj_final + 1, traj_final+1)].__iter__()
```
dwhswenson, line 1631:
```suggestion
        exttraj = iter(traj[slice(subtraj_final + 1, traj_final+1)])
```

openpathsampling/ensemble.py
dwhswenson: The TODO is about the fact that the while loop checks both the `can_append` and the `__call__` of the ensemble. This is the issue that some fundamental ensembles (`LengthEnsemble`) know whether you can append to the trajectory without running an additional frame, while others (`AllInXEnsemble`) only know you've left the ensemble after the next frame is generated.

dwhswenson (changes requested):
Please do also modify `_find_subtraj_first`. That should help get rid of some more UUID creation.

What you have here works, but I think there's a much easier way (as opposed to the `exttraj` iterator, etc.) Wouldn't this do the job:

```python
# subtraj = traj[slice(start, stop)]
# replace with
subtraj = list(traj)[slice(start, stop)]
```

I think all we need is an iterable of snapshots, so this would create that with very little change to the code.

sroet:
> What you have here works, but I think there's a much easier way (as opposed to the exttraj iterator, etc.) Wouldn't this do the job:
```python
# subtraj = traj[slice(start, stop)]
# replace with
subtraj = list(traj)[slice(start, stop)]
```
> I think all we need is an iterable of snapshots, so this would create that with very little change to the code.

I did do that in https://github.com/openpathsampling/openpathsampling/pull/978/commits/1c156433c6f7c2a1d5084b770aef970d913ea258 , this would possibly break other custom ensembles, but the workaround is generally easy and the limited change/robustness in this code is worth it in my opinion. 

As ensembles are not fully covered under coverage, please do have a **really** good look at this PR. Here is a new set of numbers from this PR:

|  | master| 1c15643 | b89bce9 |
| --- | --- | --- | --- |
||71186|5850| 62 |
||75610|3257| 114 |
||76504|7526| 86 |
||67516|6842| 99 |
||71031|4789| 47 |
||67648|9378| 33 |
|*mean*| 71186 | 6274 | 74

Currently this PR prevents **99.9%** of all created trajectories in 10 mc steps And these number seem reasonable for 44 ensembles and 10 mc steps

sroet:
@dwhswenson please have another look at this

openpathsampling/ensemble.py
```diff
@@ -165,13 +166,15 @@ def check(self, trajectory=None, reset=None):
             if self.debug_enabled:
                 logger.debug("Resetting cache " + str(self))
             if self.direction > 0:
-                self.start_frame = trajectory.get_as_proxy(0)
-                self.prev_last_frame = trajectory.get_as_proxy(-1)
+                # TODO: this can be hit with trajectory is None?
```
sroet, line 169: If `trajectory is None` the `else statement on line `159` is hit, which means that if `self.direction` is set this will error out (this was already the case in the old code), the same is true for line `174`

dwhswenson (approved):

openpathsampling/ensemble.py
```diff
@@ -2212,7 +2226,8 @@ def __call__(self, trajectory, trusted=None, candidate=False):
         else:
             logger.debug("Untrusted VolumeEnsemble " + repr(self))
             # logger.debug("Trajectory " + repr(trajectory))
-            for frame in trajectory.as_proxies():
+            # This can sometimes get a list instead of a Trajectory
+            for frame in list(trajectory):
```
dwhswenson, line 2230: I think you'll want to do something here to allow this to load as proxies instead. The current implementation will require loading the full snapshot object every time (with all its data) -- could be a significant slowdown in custom analysis. Maybe add something like:

```python
iter_traj = getattr(trajectory, 'iter_proxies', trajectory.__iter__)
for frame in iter_traj():
    ...
```

openpathsampling/ensemble.py
```diff
@@ -2271,7 +2286,7 @@ def __call__(self, trajectory, trusted=None, candidate=False):
         trajectory : :class:`openpathsampling.trajectory.Trajectory`
             The trajectory to be checked
         """
-        for frame in trajectory.as_proxies():
+        for frame in list(trajectory):
```
dwhswenson, line 2289: as above; allow use of proxies

openpathsampling/ensemble.py
```diff
@@ -2297,13 +2312,12 @@ def __invert__(self):
         return AllInXEnsemble(self.volume, self.trusted)
 
     def __call__(self, trajectory, trusted=None, candidate=False):
-        for frame in trajectory.as_proxies():
+        for frame in list(trajectory):
```
dwhswenson, line 2315: as above; allow use of proxies

dwhswenson (changes requested):
A general change that you will want to do everywhere, instead of `list(trajectory)`. Discussed in more detail in a comment, but now that I think about it, it applies to the whole changeset:

You should always try to use one of the `proxy` methods from `Trajectory`.

* `get_as_proxy(idx)`: get a single snapshot, allowing proxy snapshots
* `as_proxies()`: return a list of all snapshots, which may include proxy snapshots
* `iter_proxies()`: iterator over snaphots, allowing proxies

When you call `list(trajectory)`, it will call `Trajectory.__iter__()`, which will fully reload the snapshot into memory. This will be slow and isn't needed for a lot of analysis. (I had forgotten about this when I suggested converting to list before; only remembered when I saw changes dropping the `as_proxies`.)

> this might possibly break custom ensembles

I'm not sure where this risk would be. I suppose there might be something weird if you subclass `SequentialEnsemble` in some strange way. In general, we discourage subclassing `SequentialEnsemble` (unless you're changing the frame assignment algorithm). Instead, custom ensembles should be subclasses of `WrappedEnsemble`, wrapping the appropriate `SequentialEnsemble`. This is what we do for `TISEnsemble`, `MinusInterfaceEnsemble`, and `VisitAllStatesEnsemble`, as a few examples.

The only thing I was worried about with this change was handling proxy snapshots, which you are dealing with.

> As ensembles are not fully covered under coverage, please do have a **really** good look at this PR.

Although `ensemble.py` doesn't have full coverage, the sections modified here (`EnsembleCache` and `SequentialEnsemble`) are pretty close, mainly missing coverage of logging stuff in debug guards and of a few error cases. Additionally, they're tested very thoroughly, involving what is probably thousands of individual asserts. `SequentialEnsemble` was some very complex code to write, and I didn't trust it until I had written every test I could think of at the time.

openpathsampling/ensemble.py
sroet: But this would always error out if it was ever hit with `Trajectory=None`? 
It will now error out with `NameError: name 'get_frame' is not defined` but in the old code it should have error out with `AttributeError: 'NoneType' has no attribute 'get_as_proxy'`

dwhswenson: Yeah, I think it's fine. I think that more correctly, I shouldn't have allowed `trajectory` to be `None`. I don't know what motivated that -- maybe in early iterations of the caching, I didn't want it to raise errors if it wasn't called with a trajectory.

sroet:
@dwhswenson implemented the `proxy_loading` as you advised, please have another look.

I updated the table in https://github.com/openpathsampling/openpathsampling/pull/978#issuecomment-782896532 with the new data. (still sub `10` Trajectories per md step)

openpathsampling/ensemble.py
```diff
@@ -2271,7 +2297,8 @@ def __call__(self, trajectory, trusted=None, candidate=False):
         trajectory : :class:`openpathsampling.trajectory.Trajectory`
             The trajectory to be checked
         """
-        for frame in trajectory.as_proxies():
+        itraj = getattr(trajectory, 'iter_proxies', trajectory.__iter__)
+        for frame in list(itraj()):
```
dwhswenson, line 2301: Why `list` here?

dwhswenson (approved):
Looks pretty good. A small optional change in a comment, as well as the following thought:

It might be worth just having a function called `_get_list_traj` which does this `getattr`/`list` pair of operations. I suspect CodeClimate will complain about repeated code here, and since the reason for the need is obscure enough, it might be worth it to have a single location with a clearer docstring explaining what's going on. That's optional though.

In case you want to make those changes, I will leave this open for at least 24 hours, merging no earlier than Wed 24 Feb 16:00 GMT (17:00 local). If no comment/update by then, I will merge sometime after that.

openpathsampling/ensemble.py
dwhswenson: (Note: if you do a separate `_get_list_traj` function, feel to use it here and make the list -- I'm not worried about that extra overhead. I just wondered why it was explicitly made a list on this particular line, since it doesn't seem to need to be.)

sroet: > Why list here?

No reason, just me trying to code in the early morning :laughing: . You are absolutely right.

About the 
> - I'm not worried about that extra overhead.

It would not actually be extra overhead compared to `master` as that is exactly what `as_proxies()` does (it returns a list wrapped iterator). And for the ease of implementation I will actually implement it that way.

sroet:
> It might be worth just having a function called _get_list_traj which does this getattr/list pair of operations. I suspect CodeClimate will complain about repeated code here, and since the reason for the need is obscure enough, it might be worth it to have a single location with a clearer docstring explaining what's going on. That's optional though.

Yeah, looking back at it that is the cleaner and more maintainable solution, will refactor that now.

openpathsampling/ensemble.py
dwhswenson: > It would not actually be extra overhead compared to `master`

Sure -- what I really meant was "Don't feel like you need to create two functions; one that does the iterator and one that does the list. Unless a trajectory has a million frames (which it shouldn't) you might as well just do the list. And don't think that my question about using a list here means you need to use an iterator everywhere possible."

sroet:
@dwhswenson refactored the code, it still results in numbers in the same order as 1cff417 . Please have a look if you think the docstring is descriptive enough.

By the way, thank you for the (multiple) in-depth reviews.

dwhswenson:
> Please have a look if you think the docstring is descriptive enough.

LGTM. If anything, it's probably more detailed than needed!

> By the way, thank you for the (multiple) in-depth reviews.

No problem, and thank *you* for making the improvements!

Merging this one now.

dwhswenson:
> Please have a look if you think the docstring is descriptive enough.

LGTM. If anything, it's probably more detailed than needed!

> By the way, thank you for the (multiple) in-depth reviews.

No problem, and thank *you* for making the improvements!

Merging this one now.

PR #979, opened 2021-02-20
Move n_dofs to engine; add engine.has_constraints()

dwhswenson:
This moves the calculation of the number of degrees of freedom to the engine. This is expected to be a constant for an engine -- at least, the engines we have so far don't allow number of particles to change -- so it makes sense to cache it (for OpenMM), instead of calculating it over and over with the `snapshot.n_degrees_of_freedom` feature.

This also adds an `engine.has_constraints()` method. This gives a cleaner way to abort if an engine with constraints tries to use one of the `DirectionModifier` snapshot modifiers.

### DIFF
--- a/openpathsampling/tests/test_snapshot_modifier.py
+++ b/openpathsampling/tests/test_snapshot_modifier.py
@@ -317,6 +317,26 @@ class TestGeneralizedDirectionModifier(object):
         )
         self.openmm_modifier._verify_snapshot(constr_snap)
 
+    def test_verify_engine_constraints(self):
+        ad_vacuum_constr = omt.testsystems.AlanineDipeptideVacuum()
+        constrained_engine = omm_engine.Engine(
+            topology=self.test_snap.topology,
+            system=ad_vacuum_constr.system,
+            integrator=omt.integrators.VVVRIntegrator()
+        )
+        modifier = GeneralizedDirectionModifier(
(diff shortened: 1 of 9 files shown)


dwhswenson:
This is ready for review and comment. I will leave it open for at least 24 hours, merging no earlier than Thu 25 Mar 08:00 GMT (09:00 local).

openpathsampling/tests/test_snapshot_modifier.py
```diff
@@ -317,6 +317,26 @@ def test_verify_snapshot_constraints(self):
...
+            1.2 * u.nanometer / u.picosecond,
+            engine=constrained_engine
+        )
+        # this is a hack because ndofs not defined in TestsystemEngine
+        self.openmm_engine.current_snapshot = self.test_snap
+        snap = self.openmm_engine.current_snapshot
+        # when it checks based on the engine, it should be fine
+        self.openmm_modifier._verify_snapshot(snap)
+        # when modifier overrides snap.engine, it errors
+        with pytest.raises(RuntimeError):
```
sroet, line 337: I prefer a slightly more explicit error check
```suggestion
        with pytest.raises(RuntimeError, match="constraints"):
```

sroet (approved):
One small comment about checking if the expected error is raised. Please don't hang on that, LGTM otherwise

PR #980, opened 2021-02-21
Add codecov.yml

dwhswenson:
Adding [config file](https://docs.codecov.io/docs/codecovyml-reference) for CodeCov. In particular, I'm hoping that I can get CodeCov to wait until the integration tests are done before adding its comment/status indicator. It's a little annoying to get the email (which, unlike the comment, doesn't update) saying CodeCov hates the PR after only the minimal test suite has come in. Also, I keep seeing the red X on a PR and checking to see where the tests went wrong, when it's really that the tests are still running and CodeCov has incomplete info.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -22,6 +22,8 @@ jobs:
     runs-on: ubuntu-latest
     name: "Tests"
     strategy:
+      # NOTE: when changing the test matrix, update the after_n_builds in
+      # codecov.yml to match the number of builds in the matrix
       matrix:
         CONDA_PY:
           #- 3.9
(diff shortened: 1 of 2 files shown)


sroet:
I have no real preference here, but please add a comment to `.github/workflows/tests.yml` to update this `after_n_builds` number whenever the  test-matrix size changes

dwhswenson:
> I have no real preference here, but please add a comment to `.github/workflows/tests.yml` to update this `after_n_builds` number whenever the test-matrix size changes

Done. This may solve the problem (I'm typing this when the minimal tests have been finished but the integration tests are still running; there's no complaining from codecov now.) In case there's any further need for discussion (I assume not), I will leave this open for at least 6 hours, merging no earlier than Mon 22 Feb 19:00 GMT (20:00 local).

PR #981, opened 2021-02-21
Remove some Py2-specific code

dwhswenson:
#977 dropped Python 2 from the OPS 2.0 test matrix. In the process, some code lost coverage. Here I remove the Py2-specific code for most of that (except netcdfplus, which I'm not worrying about since it will also be dropped in 2.0).

There's one line where I don't immediately understand how it lost coverage:

https://github.com/openpathsampling/openpathsampling/blob/6ff2cbbf615928c031896e1d719217d2087fa940/openpathsampling/engines/dynamics_engine.py#L256

This is in an instance test against `basestring`, but `basestring` is `str` for Py3+ (we set it at the top of the file). Any idea why that's not getting covered? (particularly @sroet)

Note that this PR is not intended to drop all Py2-related code -- there's more to do on that. This is just catching the things that lost coverage when Py2 left the test matrix.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -27,7 +27,7 @@ jobs:
           #- 3.9
           #- 3.8
           - 3.7
-          - 2.7
+          #- 2.7
         MINIMAL: [""]
         include:
           - CONDA_PY: 3.7
--- a/openpathsampling/analysis/shooting_point_analysis.py
+++ b/openpathsampling/analysis/shooting_point_analysis.py
@@ -3,12 +3,7 @@ import pandas as pd
 import numpy as np
 
 from openpathsampling.progress import SimpleProgress
-
-try:
-    from collections import abc
-except ImportError:
-    import collections as abc
-
+from collections import abc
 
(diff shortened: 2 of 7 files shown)


sroet:
Are you sure that one is not flaky? 

But looking at the code, shouldn't that always (assuming `str`) be cought by line 225 into 251?
```python
225                     if type(my_options[variable]) is type(default_value):

251                        else:
                                okay_options[variable] = my_options[variable]
```

What input type  do you expect to be a subclass of `str` without being a `str`?

dwhswenson:
> What input type do you expect to be a subclass of `str` without being a `str`?

Yeah, that's it. Py 2.7 had `str` and `unicode` that inherited from `basestring`.

I removed that line and all references to `basestring` from that file.

dwhswenson:
This is ready for review and comment. I will leave it open for at least 24 hours, merging no earlier than Tue 23 Feb 14:00 GMT (15:00 local).

sroet (approved):
This looks good to me. For any other reviewer: the changes in `ensemble.py` are identical to the #976

PR #983, opened 2021-02-24
Fix MISTIS minus ensemble not forbidding other states

dwhswenson:
The minus interface as set up by the `MISTISNetwork` didn't forbid the trajectory from entering other states during the excursion from the state. This would occasionally manifest as a failure of the sanity check when saving, stating that the trajectory for some replica did not satisfy the ensemble (because the trajectory in the TIS ensemble after a replica exchange had frames in the other state). 

Typically, this meant that the minus trajectory had to cross to the other state and then *cross back*, so it was extremely rare, but I found it when working on a simple model that isn't actually rare enough to really need TIS to get the rate. @gyorgy-hantal: I think you encountered this problem once. This is the explanation.

This fixes it -- everything was in place to forbid other states (and used in MSTIS); just needed to add it to MISTIS. Also adds a test to prevent regression.

### DIFF
--- a/openpathsampling/tests/test_network.py
+++ b/openpathsampling/tests/test_network.py
@@ -382,7 +382,6 @@ class TestMISTISNetwork(TestMultipleStateTIS):
         assert_equal(ensBA(self.traj['AB']), False)
         assert_equal(ensBA(self.traj['AC']), False)
 
-
     def test_storage(self):
         import os
         fname = data_filename("mistis_storage_test.nc")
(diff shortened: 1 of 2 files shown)


dwhswenson:
Side note: looks like #980 did fix my issues with CodeCov. Now we get the comment after the last of the pytest runs has come in, but while the notebook tests are still running. Perfect!

dwhswenson:
This is ready for review and comment. I will leave it open for at least 24 hours, merging no earlier than Thu 25 Feb 15:00 GMT (16:00 local).

openpathsampling/tests/test_network.py
```diff
@@ -365,6 +382,7 @@ def test_trajectories_strict(self):
         assert_equal(ensBA(self.traj['AB']), False)
         assert_equal(ensBA(self.traj['AC']), False)
 
+
```
sroet, line 385: excess newline?
```suggestion
```

sroet (approved):
This LGTM, thank you for adding the understandable regression test. 1 extra newline snuck in, but don't block this PR based on that

PR #988, opened 2021-03-01
Fix problem with repeated ensemble names in MISTIS

dwhswenson:
In MISTIS, there's a distinction between the "input transitions," "sampling transitions" and "(physical) transitions", as described here:

https://github.com/openpathsampling/openpathsampling/blob/d32c198fc014d26d31ff8c62a8f9fa65ff3049c9/openpathsampling/high_level/network.py#L742-L750

However, OPS was auto-naming input transitions (and associated ensembles) with the same names as sampling transitions, potentially causing confusion when trying to load from storage. This PR fixes that by adding a `name_suffix` to the transitions, which defaults to an empty string. For input transitions, we use `name_suffix=" (input)"`.

### DIFF
--- a/openpathsampling/high_level/network.py
+++ b/openpathsampling/high_level/network.py
@@ -786,7 +786,8 @@ class MISTISNetwork(TISNetwork):
             self.input_transitions = {
                 (stateA, stateB):
                 paths.TISTransition(stateA, stateB, interface, interface.cv,
-                                    name=stateA.name+"->"+stateB.name)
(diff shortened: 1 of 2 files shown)


dwhswenson:
This is ready for review and comment. I will leave it open for at least 24 hours, merging no earlier than Tue 02 Mar 18:00 GMT (19:00 local).

sroet (approved):
LGTM, can't wait to convert all these string summations manipulations to `f-strings` for OPS 2.0

PR #989, opened 2021-03-02
Use a deque to keep track of more recent time estimates

sroet:
Today, I got annoyed by the fact that OPS's `Estimated time remaining` becomes worse, the longer you run (see for example, the table in [this comment](https://github.com/openpathsampling/openpathsampling/pull/928#issuecomment-709249488) )

This fixes that (slightly) by using a `deque` with a default length of `10_000` MC steps. So that it just uses the average of the last `10_000` steps to estimate the average step time and the time remaining. 

The changes in this PR are:
- use `deque` to get a slightly better estimate of the time-left (47f76d4 and babd276)
- add a `time_per_step` override for `progress_string` to override the `time_per_step` calculation and adds a test for this behavior (47f76d4)
- removes `nose` from `test_tools.py`, partial progress towards #756 (6a8485c)

### DIFF
--- a/openpathsampling/pathsimulators/path_sampling.py
+++ b/openpathsampling/pathsimulators/path_sampling.py
@@ -254,6 +254,10 @@ class PathSampling(PathSimulator):
             self.step += 1
             logger.info("Beginning MC cycle " + str(self.step))
             refresh = self.allow_refresh
+            now = time.time()
+            if nn > 0:
+                deq.append(now-time_start)
+
             if self.step % self.status_update_frequency == 0:
                 # do we visualize this step?
                 if self.live_visualizer is not None and mcstep is not None:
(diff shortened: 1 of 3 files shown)


dwhswenson:
The idea is fine, but can you redo this PR to build off of #911? I'm pretty sure there will be significant conflicts with that. (I really need to finish reviewing hooks-related stuff; I'm insisting on that stuff being in the 1.5 release, and that's the main thing remaining for 1.5. If you can give an eye to #755 and #911 with a mental focus on "will this work if the hooks use Dask?" that would be a big help -- the mental energy of needing to ensure that, and being the only person to check it, has been a big part of why I've been slow on that.) Plus, this kind of approach is exactly what hooks should improve: let you easily customize the behavior of something like this, with code that is more isolated from the rest of the logic.

Also, probably the single biggest (and potentially easiest) approach to solve the underlying problem would be to add a SQL index on the UUID for more tables. I believe we now do that for the UUIDs table; it would probably make sense to do that for all data object tables (or even all tables; the extra cost for simulation objects isn't much). I suspect that most of the slow-down you see now is because snapshot tables don't have a UUID index, meaning that the cost of adding to them grown linearly with the number of snapshots stored.

sroet:
closing this for now, will rebuild of of #911 in the future

PR #990, opened 2021-03-11
SimStore: Periodic CVs

dwhswenson:
This PR adds information about periodicity to SimStore CVs. It reserves the kwargs `period_min` and `period_max`.

This isn't currently used anywhere in the core of OPS, but might be used in the future to combine `PeriodicCVDefinedVolume` and `CVDefinedVolume` (probably a good option for OPS 2.0).

It will be used in the CLI for simulation setup.

### DIFF
--- a/openpathsampling/experimental/simstore/test_storable_function.py
+++ b/openpathsampling/experimental/simstore/test_storable_function.py
@@ -230,7 +230,7 @@ class TestStorableFunction(object):
         expected = {2: False, 0: True, 1: 'error'}[n_nones]
         check_period = StorableFunction._check_period
         if expected == 'error':
-            with pytest.raises(RuntimeError, match='period'):
+            with pytest.raises(ValueError, match='period'):
                 check_period(period_min, period_max)
         else:
             assert check_period(period_min, period_max) == expected
--- a/openpathsampling/experimental/simstore/test_storable_function.py
+++ b/openpathsampling/experimental/simstore/test_storable_function.py
@@ -223,6 +223,12 @@ class TestStorableFunction(object):
 
         self.func = StorableFunction(get_expected)
 
+    def test_check_periodic(self):
+        pytest.skip()
+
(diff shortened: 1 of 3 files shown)


dwhswenson:
This is ready for review and comment. I will leave it open for at least 24 hours, merging no earlier than Sun 14 Mar 19:00 GMT (20:00 local).

openpathsampling/experimental/simstore/test_storable_function.py
```diff
@@ -223,6 +223,24 @@ def get_expected(uuid):
...
 
+    @pytest.mark.parametrize('min_max', [(None, None), (None, 10), (0, 10)])
+    def test_check_periodic(self, min_max):
+        period_min, period_max = min_max
+        n_nones = Counter(min_max)[None]
+        expected = {2: False, 0: True, 1: 'error'}[n_nones]
+        check_period = StorableFunction._check_period
+        if expected == 'error':
+            with pytest.raises(RuntimeError):
+                check_period(period_min, period_max)
```
sroet, line 234: maybe make sure that the expected error is raised?
Something like:
```suggestion
            with pytest.raises(RuntimeError, match='period'):
                check_period(period_min, period_max)
```

sroet (approved):
LGTM in general. I would like a slightly more specific test (just to make sure the error that you expect is raised). But please don't hang this PR on that.

openpathsampling/experimental/simstore/test_storable_function.py
dwhswenson: Agree, although I'm going to change this to a `ValueError`. Types `None` or numeric are allowed for `period_min` and `period_max`, but if one is `None` and the other is numeric, the values are in conflict.

sroet: > Agree, although I'm going to change this to a ValueError.

+1, I do agree that is a better error type

PR #984, opened 2021-02-25
Updated notebooks/experimental storage

oliverdutton:
Not sure how you want to manage it, but I updated the gromacs tps example to use the experimental SimStore. It works. As currently CVs are under 'storage_functions' I make a dict then select for choosing named cv's. Other than that it's been dropped in.

```python
cv_dict = {fn.name:fn for fn in list(flexible.storable_functions)} # Extract the cv's from storage where they're under storable_functions
phi = cv_dict['phi']
psi = cv_dict['psi']
```

Obviously don't accept this pull request. I'm doing it to highlight the commit, which may be useful to you.

PR #991, opened 2021-03-12
Add threshold to codecov.yml

dwhswenson:
Some recent PRs that should have no effect on coverage (#990, #984) have been getting complaints from CodeCov because a single line in netcdfplus has become uncovered (not sure if this is flakiness or a change in something upstream).

This PR adds a threshold so CodeCov will pass if flakiness uncovers up to 0.4% of the currently covered code. That corresponds to about 5 Python statements.

This only affects the CodeCov "project" check: the "patch" check will keep its default behavior, which is that coverage for lines changed in the PR must be at least as much as the current total project coverage.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -3,6 +3,7 @@ coverage:
     project:
       default:
         target: auto
+        threshold: 0.04%
 
 codecov:
   notify:


dwhswenson:
This is ready for review and comment. I will leave it open for at least 24 hours, merging no earlier than Sat 13 Mar 14:00 GMT (15:00 local).

codecov.yml
```diff
@@ -3,6 +3,7 @@ coverage:
     project:
       default:
         target: auto
+        threshold: 0.04%
```
sroet, line 6: You probably also want to add the 100% patch target
```suggestion
    project:
      default:
        target: auto
        threshold: 0.04%
    patch:
      default:
        target: 100%
        threshold: 0%
    
```

sroet (changes requested):
LGTM, one minor change to actually set the patch threshold to 100%

codecov.yml
sroet: Another option is changing the PR description from:
> If I understand correctly, this only affects the CodeCov "project" check: the "patch" check will still require that all lines added as part of the PR be covered.

to 
> The "patch" check will still require that the total coverage does not drop due to the added lines

dwhswenson: updated the description

sroet (approved):
With the updated description and the confirmation from #992, this LGTM!

PR #992, opened 2021-03-12
[DO NOT MERGE] Test codecov patch

sroet:
This is just a PR to make sure #991 does not alter the `patch` codecov requirements


openpathsampling/test.py
```diff
@@ -0,0 +1 @@
+print("codecov should scream about this")
```
dwhswenson, line 1:
```suggestion
def fake_func():
    print("codecov should scream about this")
```

You need to import the file somewhere -- either in the main package or in a `test_test.py` test module. Coverage doesn't know about files unless they're imported (for example, none of the OPS visualization stuff is even included in unit tests, because it has to be imported separately.)

Also, check the .coveragerc to make sure that we don't exclude anything with "test" in it from coverage.

dwhswenson:
Py27 fail is something flaky with PyEMMA in recent days. CodeCov non-fail is because coverage.py doesn't see the file you added.

sroet:
> Also, check the .coveragerc to make sure that we don't exclude anything with "test" in it from coverage.

Just the openpathsampling/tests/ directory, I did test if the coverage showed up in the coverage report if I ran `py.test --cov=openpathsampling`. However, codecov is more clever than that xD

dwhswenson:
Behaves as expected! 😄

dwhswenson:
Oh, although I should clarify: rule for patch is that it must increase total coverage, not have 100% coverage.

sroet:
>Oh, although I should clarify: rule for patch is that it must increase total coverage, not have 100% coverage.

Yeah, I was under the impression that you expected 100% coverage on any altered line. (unless, of course, explicitly untested with `pragma: no cover`

sroet:
With that, this PR can be closed :smile:

PR #993, opened 2021-03-14
Add Python 3.8, 3.9 to testing matrix

### LINKED ISSUE #864: Python 3.8 testing plans

dwhswenson:
OpenMM and OpenMMTools are now available on conda-forge (and therefore available for Python > 3.7). Resolves #864.

Because we release via conda-forge, we already release on 3.8 and 3.9, but they haven't been in our test matrix on CI.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -26,8 +26,8 @@ jobs:
       # codecov.yml to match the number of builds in the matrix
       matrix:
         CONDA_PY:
-          #- 3.9
-          #- 3.8
+          - 3.9
+          - 3.8
           - 3.7
           - 2.7
         MINIMAL: [""]
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -26,9 +26,9 @@ jobs:
       # * Update the after_n_builds in codecov.yml to match the number of
       #   builds in the matrix
       # * You may need to create another example run for the notebook tests
-      #   (see examples/ipynbtests.sh). This is because Python internals
-      #   (usually bytecode, which is saved in CVs) can differ between minor
-      #   Python versions.
+      #   (see examples/ipynbtests.sh; examples/prep_example_data.py). This
+      #   is because Python internals (usually bytecode, which is saved in
+      #   CVs) can differ between minor Python versions.
       matrix:
         CONDA_PY:
           - 3.9
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -22,8 +22,13 @@ jobs:
     runs-on: ubuntu-latest
     name: "Tests"
     strategy:
-      # NOTE: when changing the test matrix, update the after_n_builds in
-      # codecov.yml to match the number of builds in the matrix
+      # NOTE: When changing the test matrix:
+      # * Update the after_n_builds in codecov.yml to match the number of
+      #   builds in the matrix
+      # * You may need to create another example run for the notebook tests
+      #   (see examples/ipynbtests.sh). This is because Python internals
+      #   (usually bytecode, which is saved in CVs) can differ between minor
+      #   Python versions.
       matrix:
         CONDA_PY:
           - 3.9
@@ -33,7 +38,7 @@ jobs:
         MINIMAL: [""]
         include:
           - CONDA_PY: 3.7
-            MINIMAL: true
+            MINIMAL: "minimal"
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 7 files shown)


dwhswenson:
I suspect that the current failure is due to something changing in Python internals that breaks OPS storage between different versions. Most likely, a change in bytecode. OPS stores the bytecode of CVs, and bytecode can change between Python minor versions. This means that files should only be opened with the same version of Python (or need to be opened in safe mode). We had to make a change based on this when Python 3.6 came out.

sroet:
from looking at `examples/ipynbtests.sh`
it might be as easy as making another couple lines in:
```bash
case $PYTHON_VERSION in
    "2.7")
        mstis=$dropbox_base_url/1x4ny0c93gvu54n/toy_mstis_1k_OPS1.nc
        mistis=$dropbox_base_url/qaeczkugwxkrdfy/toy_mistis_1k_OPS1.nc
        ;;
    "3.6")
        mstis=$dropbox_base_url/1ulzssv5p4lr61f/toy_mstis_1k_OPS1_py36.nc
        mistis=$dropbox_base_url/76981cbgxm639m3/toy_mistis_1k_OPS1_py36.nc
        ;;
    "3.7")
        mstis=$dropbox_base_url/1ulzssv5p4lr61f/toy_mstis_1k_OPS1_py36.nc
        mistis=$dropbox_base_url/76981cbgxm639m3/toy_mistis_1k_OPS1_py36.nc
        ;;
    *)
        echo "Unsupported Python version: $PYTHON_VERSION"
esac
```
(As 3.6 and 3.7 already are identical?)

sroet:
(if that is the case, please also make a note in ` .github/workflows/tests.yml` to remind future developers to add these whenever the testing matrix is changed)

dwhswenson:
@sroet : Yes, that's pretty much it. Basically, I think all I need to do is re-run the example in each version of Python, and then save the file with 1000 MC steps (which should be enough that TIS analysis has enough overlap), stick that in my Dropbox, and update with the appropriate links.

Adding a comment to the CI yml is a good idea. It isn't that you necessarily need to do this every time (the file from 3.6 also works with 3.7), but it is a problem that can occur and should be straightforward to fix when it comes up.

Probably won't get to that today, though.

dwhswenson:
Summary of changes here:

* Added 3.8 and 3.9 to the testing matrix (including adding new example data files based on 3.8)
* Added script `examples/prep_example_data.py` to create data for analysis. It turns out that our old setups used a version of the examples that started with trajectories that met the requirements for the outermost ensembles. This ensured overlap in the TIS analysis, but this approach has long since been removed from the actual examples. The actual examples need more than 1000 steps to get overlap for TIS analysis. This new script sets things up using the old approach (and even figures out the Python version and labels the output file correctly).
* Changed `tutorial_storage.ipynb` to load CV by name, not number. The datafiles we have now seem to go back to before we had "CVs" that take trajectories as input. `storage.cv[0]` in current files is a trajectory CV giving the maximum value of an order parameter over the trajectory. This does not give a list for output, which caused issues in the notebook.
* Minor change to matrix so that minimal build shows in GitHub as "(3.7, minimal)" instead of "(3.7, true)".
* Removed omnia from channels
* Updated trove classifiers to match our tested versions

dwhswenson:
This is ready for review and comment. I will leave it open for at least 24 hours, merging no earlier than Wed 07 Apr 20:00 GMT (22:00 local).

sroet (approved):
LGTM, thanks for adding the data-generation script!

PR #995, opened 2021-03-15
SimStore storage.cvs should only be CVs

dwhswenson:
Previously, `storage.cvs` contained all the storable functions. It should only include CVs (which are specifically storable functions that take a snapshot as input). This PR fixes that issue.

### DIFF
--- a/openpathsampling/experimental/storage/ops_storage.py
+++ b/openpathsampling/experimental/storage/ops_storage.py
@@ -39,6 +39,7 @@ from ..simstore import SQLStorageBackend  # TODO: generalize
 
 from . import snapshots
 from .snapshots_table import SnapshotsTable
+from .collective_variables import CollectiveVariable
 
 try:
     from .simtk_unit import simtk_quantity_codec, SimtkQuantityHandler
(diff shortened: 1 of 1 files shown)


dwhswenson:
This is ready for review and comment. I will leave it open for at least 24 hours, merging no earlier than Tue 16 Mar 20:00 GMT (21:00 local).

sroet:
This LGTM. As you have labeled this as `bugfix`, what was the bug you encountered, anything you want to make sure doesn't regress? Or is it more a quality-of-life (and performance?) improvement for users to have an easier way loading cvs

dwhswenson:
It's not really something that I see as likely to have a regression; it's just that I chose the wrong class to use for the 'pseudotable'. The main usability improvement shows up in things like the CLI. For example, the `contents` command lists all the named objects in a file, organized by type of object (i.e., `[cv.name for cv in storage.cvs if cv.is_named]`). The idea is that the user can see what names are valid as command line options. Under CVs, SimStore storages were showing the generalized CVs that are automatically generated for TIS by the `InterfaceSet` (which enable disk-caching of maximum value of order parameter). But using those functions (which take trajectories as input) in place of a regular CV (snapshot input) would obviously cause errors.

Plus, under SimStore, there's a very clear distinction between `StorableFunction` (a generic SimStore object) and `CollectiveVariable` (an OPS-specific subclass of `StorableFunction`). If the storage attribute is called `cvs`, it should only return `CollectiveVariable`s.

PR #997, opened 2021-03-24
Potential fix for flaky PyEMMA tests in 2.7

dwhswenson:
I've been having problems with PyEMMA tests being flaky in Py2.7 ever since #987. It seems like there's something that sometimes shows up in stderr, which ipynbtest didn't check but apparently nbval does. The error I get is:

```
Notebook cell execution failed
Cell 8: Cell outputs differ

Input:
cv(storage.trajectories[0]);

Traceback:
Unexpected output fields from running code: set([u'stderr'])
```

This PR adds ignores to that cell and another near it that also may cause problems.

### DIFF
--- a/examples/tests/test_pyemma.ipynb
+++ b/examples/tests/test_pyemma.ipynb
@@ -147,6 +147,7 @@
    },
    "outputs": [],
    "source": [
+    "# NBVAL_IGNORE_OUTPUT\n",
     "cv = paths.collectivevariable.PyEMMAFeaturizerCV(\n",
     "    'pyemma', \n",
     "    pyemma_generator, \n",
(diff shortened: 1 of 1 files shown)


dwhswenson:
This is ready for review and comment. I will leave it open for at least 24 hours, merging no earlier than Thu 25 Mar 09:00 GMT (10:00 local).

sroet:
Do you mind adding a label/milestone indicating that this is fix for just python `2.7`  and should probably be cleaned/reverted as soon as python `2.7` is dropped? LGTM otherwise

sroet (approved):

dwhswenson:
> Do you mind adding a label/milestone indicating that this is fix for just python `2.7` and should probably be cleaned/reverted as soon as python `2.7` is dropped?

I agree with the principle, but no need in this case. This notebook is about old-style CVs and netcdfplus, both of which will be replaced in 2.0, when we also drop 2.7. So the whole notebook will be removed then. (I'll probably use some of the current notebook as a guide to write tests for SimStore-based `PyEMMAFeaturizerCV`s.)

PR #998, opened 2021-04-06
Add global random number generator; support in shooting pt selectors and pathmovers

sroet:
This week my CI failed with:
```python
______________________ TestGaussianBiasSelector.test_pick ______________________

self = <openpathsampling.tests.test_shooting.TestGaussianBiasSelector object at 0x7fac57d4fc10>

    def test_pick(self):
        picks = [self.sel.pick(self.mytraj) for _ in range(100)]
        pick_counter = collections.Counter(picks)
        assert set(pick_counter.keys()) == set(range(len(self.mytraj)))
        # final test: 99.5 should happen more than 32.4
>       assert pick_counter[2] > pick_counter[0]
E       assert 13 > 18

openpathsampling/tests/test_shooting.py:70: AssertionError
```

This PR attaches a random number generator to the selectors that can be seeded during tests to prevent these flaky tests.

The current implementation requires `numpy>=1.17` (that is when numpy swapped from `RandomState` to [`Generator`](https://numpy.org/doc/stable/reference/random/index.html)) @dwhswenson please let me know if that is too strict of a requirement and if I should write some code to support the Legacy `RandomState`

### DIFF
--- a/openpathsampling/pathmover.py
+++ b/openpathsampling/pathmover.py
@@ -4,7 +4,6 @@ Created on 19.07.2014
 @author: Jan-Hendrik Prinz
 @author: David W. H. Swenson
 """
-from __future__ import absolute_import
 import abc
 import logging
 import numpy as np
--- a/openpathsampling/pathmover.py
+++ b/openpathsampling/pathmover.py
@@ -4,7 +4,7 @@ Created on 19.07.2014
 @author: Jan-Hendrik Prinz
 @author: David W. H. Swenson
 """
-
+from __future__ import absolute_import
 import abc
 import logging
 import numpy as np
--- a/openpathsampling/pathmover.py
+++ b/openpathsampling/pathmover.py
@@ -7,12 +7,12 @@ Created on 19.07.2014
 
 import abc
 import logging
-import random
-
 import numpy as np
+
 import openpathsampling as paths
 from openpathsampling.netcdfplus import StorableNamedObject, StorableObject
 from openpathsampling.pathmover_inout import InOutSet, InOut
+from openpathsampling.random import default_rng, random
 from .ops_logging import initialization_logging
 from .treelogic import TreeMixin
 
@@ -132,7 +132,7 @@ class PathMover(with_metaclass(abc.ABCMeta, TreeMixin, StorableNamedObject)):
 
     def __init__(self):
         StorableNamedObject.__init__(self)
-
+        self._rng = default_rng()
         self._in_ensembles = None
         self._out_ensembles = None
         self._len = None
@@ -430,6 +430,8 @@ class PathMover(with_metaclass(abc.ABCMeta, TreeMixin, StorableNamedObject)):
                 "," + str(sample.trajectory) +
                 "," + repr(sample.ensemble) +
                 ")")
+        # TODO: This can't go through numpy.random as Samples unwrap to
+        # Snapshots when cast to arrays
         selected = random.choice(legal)
         logger.debug(
             "selected sample: (" + str(selected.replica) +
@@ -546,7 +548,7 @@ class SampleMover(PathMover):
             else:
                 probability *= sample.bias
 
-        rand = random.random()
+        rand = self._rng.random()
 
         if rand > probability:
             # rejected
@@ -1256,7 +1258,9 @@ class RandomSubtrajectorySelectMover(SubtrajectorySelectMover):
     """
 
     def _choose(self, trajectory_list):
-        return random.choice(trajectory_list), {}
+        # Needed to prevent ragged nested sequence DeprWarning in np 1.20
+        idx = self._rng.choice(len(trajectory_list))
+        return trajectory_list[idx], {}
 
 
 class FirstSubtrajectorySelectMover(SubtrajectorySelectMover):
@@ -1508,7 +1512,6 @@ class SelectionMover(PathMover):
         super(SelectionMover, self).__init__()
 
         self.movers = movers
-
         initialization_logging(init_log, self,
                                entries=['movers'])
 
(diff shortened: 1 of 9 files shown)


dwhswenson:
> The current implementation requires `numpy>=1.17` (that is when numpy swapped from `RandomState` to [`Generator`](https://numpy.org/doc/stable/reference/random/index.html)) @dwhswenson please let me know if that is too strict of a requirement and if I should write some code to support the Legacy `RandomState`

This is fine in principle: my view is that we should follow [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) with the exception of our until-2.0 support for Python 2.7 (and that the NEP 29 rule for NumPy should also apply to other dependencies.) NumPy 1.16 support could therefore be dropped as of 13 Jan 2021; see drop table in that link. However, it looks like NumPy 1.17 was not released for Python 2.7, so we need an approach compatible with Python 2.7.

That said, fixing the test is probably better than pinning the seed. That test takes a trajectory with 5 possible shooting points, and considers two of them: frame A has about 1/3 the selection probability of frame B. The test assumes that after 100 picks, frame B will have been picked more frequently than frame A. However, apparently the odds of that going the other way is enough that we've noticed multiple failures. Maybe the better approach is to increase the number of picks, or enhance the gap in probability. This test is supposed to be an integration test; if the random number generator changed in some way that it was no longer using the probabilities we provide, we would *want* the test to be flaky -- that would be a sign of a problem.

BTW, I'm +1 on moving random number generators to instance variables, although I think I'd tend to prefer defaulting to a global RNG (to simplify seeding for a whole simulation in the future), rather than a separate RNG instance for each OPS object instance. Maybe create the RNG instance as `paths.default_rng`, and then it can be set as the default in the instances.

dwhswenson:
I updated the title to better reflect what I understand this will contain; better for release notes. Please correct if the current title doesn't represent the intent of the PR.

openpathsampling/pathmover.py
```diff
@@ -132,7 +132,7 @@ class PathMover(with_metaclass(abc.ABCMeta, TreeMixin, StorableNamedObject)):
 
     def __init__(self):
         StorableNamedObject.__init__(self)
-
+        self._rng = default_rng()
```
sroet, line 135: This has to be underscored for the safe/load cycle to work

openpathsampling/random.py
```diff
@@ -0,0 +1,24 @@
+import numpy as np
+import random
```
sroet, line 2: added the random import here to (hopefully) make the import less confusing for `python2.7`

openpathsampling/random.py
```diff
@@ -0,0 +1,25 @@
+from __future__ import absolute_import
```
sroet, line 1: This is needed as `python 2.7` does not understand that I don't want to import this file whenever I say `import random` in this file...
```python
Python 2.7.15 | packaged by conda-forge | (default, Mar  5 2020, 14:56:06) 
[GCC 7.3.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from openpathsampling.random import random
>>> random.random
<module 'openpathsampling.random' from 'openpathsampling/random.py'>
>>> random.random.random
<module 'openpathsampling.random' from 'openpathsampling/random.py'>
>>> random.random.random.random
<module 'openpathsampling.random' from 'openpathsampling/random.py'>
```

sroet:
@dwhswenson Finally figured out what was wrong with python `2.7` (I don't even understand how that import could ever be resolved in the way it did...). 

While hunting this down I also added the `rng` to `PathMover`s. 
This should now be ready for a review!

openpathsampling/sample.py
```diff
@@ -1,4 +1,4 @@
-import random
+from openpathsampling.random import random
```
dwhswenson, line 1: I take it this is the necessary way of getting to stdlib random? (Long term -- being before 2.0 -- everything in OPS should be using `paths.default_rng`)

openpathsampling/tests/test_pathmover.py
```diff
@@ -11,6 +11,7 @@
 from nose.plugins.skip import SkipTest
 from nose.tools import (assert_equal, assert_not_equal, raises, assert_true,
                         assert_in, assert_not_in)
+import numpy as np
```
dwhswenson, line 14: I don't see this import being used -- is this leftover from earlier tests?

openpathsampling/random.py
```diff
@@ -0,0 +1,25 @@
...
+            return self.random_sample(*args, **kwargs)
+
+        def integers(self, *args, **kwargs):
+            return self.randint(*args, **kwargs)
+    rng = RandomState()
+else:
+    rng = np.random.default_rng()
+
+
+def default_rng():
```
dwhswenson, line 24: I don't think this should be a function. Is there a specific reason it must be? Here are my reasons:

1. A function implies to users that this is creating something.
2. The behavior is specifically different from what `np.random.default_rng()` does.

If you do your `TestRandom.test_identities` test with `np.random.default_rng()`, it will fail the `is` test.

Instead, either name `rng` as `default_rng` or `DEFAULT_RNG` (to really emphasize that it is constant). You can still import it into the main `openpathsampling` namespace.

dwhswenson: nvm. I think this has to be a function so that the underlying `rng` object can change, but newly created instances get the new `rng`. Nonetheless, please rename `rng` to something like `DEFAULT_RNG`.

If I understand correctly, in that case, setting the seed before a simulation starts would be something like `paths.random.DEFAULT_RNG = np.random.default_rng(seed)` before creating any objects. Is that right?

dwhswenson (changes requested):
Most changes very minor. One point I'm a little confused about in the implications of making `default_rng` a function, instead of just importing an instance. That's worth some additional discussion.

openpathsampling/random.py
sroet: > Nonetheless, please rename rng to something like DEFAULT_RNG

Will do

> If I understand correctly, in that case, setting the seed before a simulation starts would be something like paths.random.DEFAULT_RNG = np.random.default_rng(seed) before creating any objects. Is that right?

Yes, exactly (except that it has to go through `paths.random.RandomState(seed)` for `np<1.17`) .

openpathsampling/random.py
```diff
@@ -0,0 +1,25 @@
...
+import numpy as np
+import random
+
+
+# Set one rng for all of OPS
+if np.version.version < '1.17':
+    # Legacy support, remove when numpy 1.16 support is dropped
+    class RandomState(np.random.RandomState):
+        def __init__(self, *args, **kwargs):
+            super(RandomState, self).__init__(*args, **kwargs)
```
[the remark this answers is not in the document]
sroet: I wasn't sure if python `2.7` does reasonable things without the init being here, so I just added it to be sure.

openpathsampling/sample.py
sroet: yes, just to make name-space clashes less likely

openpathsampling/tests/test_pathmover.py
sroet: Seems like it

openpathsampling/random.py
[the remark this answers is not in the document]
dwhswenson: Try it without this `__init__` -- it should work correctly, since all you're doing is using the inherited version.

dwhswenson (approved):

dwhswenson:
LGTM!

PR #1003, opened 2021-04-07
Prevent netcdfplus from trying to load netcdfplus CVs as SimStore CVs

dwhswenson:
As of #995, it appears that netcdfplus tries to load CVs as SimStore CVs. This is apparently because netcdfplus maps a class name (not fully qualified; just the class name) to a class. And once SimStore CVs were found in the chain of subclasses, netcdfplus replaced the old CVs with the new ones. This would lead to errors when loading old CVs (which do not have the same storage dict as new CVs).

The solution is to exclude classes in the `openpathsampling.experimental.storage` subpackage from being listed as classes that netcdfplus can load.

### DIFF
--- a/openpathsampling/netcdfplus/base.py
+++ b/openpathsampling/netcdfplus/base.py
@@ -242,8 +242,10 @@ class StorableObject(object):
             The name points to the class
         """
         subclasses = StorableObject.descendants()
-
-        return {subclass.__name__: subclass for subclass in subclasses}
+        return {subclass.__name__: subclass for subclass in subclasses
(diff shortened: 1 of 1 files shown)


dwhswenson:
I have no idea why GHA isn't running this... going to close/reopen to see if that triggers a build.

sroet (approved):
LGTM

dwhswenson:
This is ready for review and comment. I will leave it open for at least 24 hours, merging no earlier than Thu 08 Apr 16:00 GMT (18:00 local).

(Forgot to add that once tests passed!)

PR #1004, opened 2021-04-08
SimStore: Make storable function modes customizable

dwhswenson:
The definition of modes for storable functions in SimStore (i.e., `'analysis'`, `'production'`, `'no-caching'`) was in a dictionary that was defined in the `__call__` method. This moves that dictionary to a private instance variable. This will allow (advanced) users to update that dictionary with custom modes.

This also removes an unnecessary skip that was shadowing the fact that some tests needed to be updated for changes in #985. The previously skipped tests now run and pass.

### DIFF
--- a/openpathsampling/experimental/simstore/storable_functions.py
+++ b/openpathsampling/experimental/simstore/storable_functions.py
@@ -325,7 +325,7 @@ class StorableFunction(StorableNamedObject):
 
     @mode.setter
     def mode(self, value):
-        allowed_values = list(self._modes.keys())
+        allowed_values = list(self._modes)
         if value not in allowed_values:
             raise ValueError("Unknown mode: '%s'. Allowed options: %s" %
                              (value, allowed_values))
@@ -562,4 +562,3 @@ class StorageFunctionHandler(object):
         missing = uuids - set(uuid_map.keys())
         missing_map = {uuid: uuid_items[uuid] for uuid in missing}
         return uuid_map, missing_map
-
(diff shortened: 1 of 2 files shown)


dwhswenson:
This is ready for review and comment. I will leave it open for at least 24 hours, merging no earlier than Fri 09 Apr 12:00 GMT (14:00 local).

openpathsampling/experimental/simstore/storable_functions.py
```diff
@@ -317,7 +325,7 @@ def mode(self):
 
     @mode.setter
     def mode(self, value):
-        allowed_values = ['no-caching', 'analysis', 'production']
+        allowed_values = list(self._modes.keys())
         if value not in allowed_values:
```
sroet, line 329: This is unnecessary (`in` always checks against `keys()` for `dict`s) :
```python
>>> a = {"a": 1, "b": 2, "c":3}
>>> "a" in a
True
>>> 1 in a
False
```
```suggestion
        if value not in self._modes:
```

sroet (changes requested):
1 unnecessary line of code, LGTM otherwise

openpathsampling/experimental/simstore/storable_functions.py
dwhswenson: It's necessary because the error message reports `allowed_values`. I thought it was more readable to set it here and re-use it than to have the error message require `list(self._modes)` (although it could be `list(self._modes)` instead of `list(self._modes.keys())`

sroet: +1, Missed the error (collapsed in the review)
```suggestion
        allowed_values = list(self._modes)
        if value not in allowed_values:
```

PR #1005, opened 2021-04-08
[WIP] Functions to convert netcdfplus CVs to SimStore CVs

dwhswenson:
This adds a constructor to SimStore-based CVs to create the appropriate SimStore CV from a netcdfplus CV. This is a necessary step toward writing a script that converts netcdfplus files to SimStore files. That will be useful now for users who want to experiment with SimStore, and will be necessary in a pre-2.0 release so that users can upgrade.


dwhswenson:
@sroet, @hejung, @gyorgy-hantal (and anyone else who has some old CVs from netcdfplus and is comfortable trying an experimental branch): Could you try this branch out and see if the `from_netcdfplus_cv` methods are working correctly? They work on the things I have tested, but those are based on the way that I create CVs; just checking to see if there are use cases I'm missing.

The following script is a rough template of I hope you can try, especially for any CV that might have an unusual setup:

```python
from openpathsampling.experimental.storage.collective_variables import MDTrajFunctionCV

storage = paths.Storage("old_file.nc", mode='r')
old = storage.cvs['phi']  # was originally defined as a paths.MDTrajFunctionCV

# create the new CV using the classmethod `from_netcdfplus_cv`
new = MDTrajFunctionCV.from_netcdfplus_cv(old)

snap = storage.snapshots[-1]  # take an arbitrary snapshot from storage
assert old(snap) == new(snap)
```

So far, the CV types in the following table have been implemented in the new storage. (Only `FunctionCV` changed its name, because SimStore naming changed a little to avoid some domain-specific things in netcdfplus.) If you have a different CV type, let me know and I can prioritize setting up those, as well -- I assume these are the most-used CV types.

| `openpathsampling.` | `openpathsampling.experimental.storage.collective_variables.`|
|---------|-----------------------------------------------|
| `FunctionCV` | `CollectiveVariable` |
| `CoordinateFunctionCV` | `CoordinateFunctionCV` |
| `MDTrajFunctionCV` | `MDTrajFunctionCV` |

PR #1008, opened 2021-04-12
make rng private for selectors

### LINKED ISSUE #1007: Problem with selector RNGs and serialization
@sroet : Merging #998 seems to have caused problems for SimStore. Somehow, even for `UniformSelector`, the `selector.to_dict()` includes the RNG, which cannot (and should not) be serialized. I just got around to investigating the "Object of type Generator is not JSON serializable" error I'm getting in the ops-storage-notebooks repo, and it points to the `UniformSelector`.

From my ipdb session:

```
ipdb> p o 
<openpathsampling.shooting.UniformSelector object at 0x7f8b52bae150>
ipdb> p vars(o)
{'rng': Generator(PCG64) at 0x7F8B4D0EBC80, '__uuid__': 205825572803023860022845415596767639408, '_name': '', '_name_fixed': True, 'pad_start': 1, 'pad_end': 1}
ipdb> p o.to_dict()
{'pad_end': 1, 'rng': Generator(PCG64) at 0x7F8B4D0EBC80, 'pad_start': 1}
```

It looks like you made it private in the movers because of issues with storage, but perhaps not in the selectors?

Options I'm fine with:

1. Property `.rng` giving access to `._rng`, which is not stored.
2. All access to the RNG via private `._rng` (if you're messing with this, you're probably doing it intentionally).
3. Excluding `.rng` from storage via some mechanism that works -- netcdfplus has some poorly documented tricks like class variables called `excluded` (or something like that) to ignore some attributes. This is my least favorite choice, because I might not support those in the future in favor of only overriding with direct overrides of `to_dict`. ("There should be one-- and preferably only one --obvious way to do it. / Although that way may not be obvious at first unless you're Dutch.")

(EDIT: Also open to other options that solve the problem, of course, but these are three that I immediately think would be reasonable.)

sroet:
closes #1007 (partly)

The notebook from that issue now works. However, It is now uncertain is the state of the generators that is loaded. 
This depends whether OPS is re-imported between loading or not.

```python
# before simulation:
{'bit_generator': 'PCG64',
 'state': {'state': 336467696170254096309584224979479325155,
  'inc': 238122374084886836437703854582102603205},
 'has_uint32': 0,
 'uinteger': 0}
<numpy.random._pcg64.PCG64 at 0x7f0b819d9d10>
 
# after running:
 {'bit_generator': 'PCG64',
 'state': {'state': 195230481313007235580189205770362297672,
  'inc': 238122374084886836437703854582102603205},
 'has_uint32': 1,
 'uinteger': 2214207804}
 <numpy.random._pcg64.PCG64 at 0x7f0b819d9d10>
 
# loaded in the same kernel session
{'bit_generator': 'PCG64',
 'state': {'state': 195230481313007235580189205770362297672,
  'inc': 238122374084886836437703854582102603205},
 'has_uint32': 1,
 'uinteger': 2214207804}
<numpy.random._pcg64.PCG64 at 0x7f0b819d9d10>
 
# loaded in a restarted kernel session
 {'bit_generator': 'PCG64',
 'state': {'state': 316665816032858718566197330052104521012,
  'inc': 90421417105856406385200004491830534511},
 'has_uint32': 0,
 'uinteger': 0}
 <numpy.random._pcg64.PCG64 at 0x7f0ab8347d10>
```
This is a broader issue to be solved, as OPS does not have a way of storaging mutable objects (such as a rng state)

### DIFF
--- a/openpathsampling/pathmovers/spring_shooting.py
+++ b/openpathsampling/pathmovers/spring_shooting.py
@@ -244,7 +244,7 @@ class SpringShootingSelector(paths.ShootingPointSelector):
         # Select the next index from the probability list and shift to range
         # from -delta_max to delta_max, 0 being the shooting index of the last
         # accepted trajectory
-        rand = self.rng.random() * self._total_bias
+        rand = self._rng.random() * self._total_bias
         dframe = 0
         prob = prob_list[0]
         while prob <= rand and dframe < len(prob_list):
--- a/openpathsampling/shooting.py
+++ b/openpathsampling/shooting.py
@@ -13,7 +13,7 @@ init_log = logging.getLogger('openpathsampling.initialization')
 class ShootingPointSelector(StorableNamedObject):
     def __init__(self):
         # Assign rng, so it can be set to something else
-        self.rng = default_rng()
+        self._rng = default_rng()
         super(ShootingPointSelector, self).__init__()
 
     def f(self, snapshot, trajectory):
@@ -77,7 +77,7 @@ class ShootingPointSelector(StorableNamedObject):
         prob_list = self._biases(trajectory)
         sum_bias = sum(prob_list)
 
-        rand = self.rng.random() * sum_bias
+        rand = self._rng.random() * sum_bias
         idx = 0
         prob = prob_list[0]
         while prob <= rand and idx < len(prob_list):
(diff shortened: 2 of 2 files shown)


sroet:
@dwhswenson this is ready for a review

dwhswenson (approved):
Looks good; thanks!

Regarding loading RNG state: It's not clear to me that a user would expect a reloaded shooting point selector to have the same state. Indeed, the way I recommend using OPS is to save all your objects to a setup.db file, and then run simulations by loading those objects (preserve UUIDs across runs). If you did that, the selector reloaded its RNG state, then "independent" path sampling runs would be identical. So I think the expectation is to have a per-session random seed unless the user explicitly requests otherwise.

PR #1010, opened 2021-04-15
Fix docstring MDTopology to MDTrajTopology

sroet:
A student asked my help initializing the `OpenMMEngine` and I could not find the `openpathsampling.engines.openmm.MDTopology`. Looking at `openpathsampling/engines/openmm/topology.py` this is actually called: `MDTrajTopology`

This PR all the docstrings that mention `MDTopology`. 
it also fixes some of the trivial `pep8` complaints in openmm/engine.py

### DIFF
--- a/openpathsampling/collectivevariables/plumed_wrapper.py
+++ b/openpathsampling/collectivevariables/plumed_wrapper.py
@@ -27,7 +27,7 @@ class PLUMEDCV(paths.CoordinateFunctionCV):
     https://doi.org/10.1016/j.cpc.2013.09.018
 
     Examples
-
+    --------
     >>> # To create a `CollectiveVariable` which calculates the dihedral psi
     >>> # formed by the atoms [7,9,15,17] in Ala dipeptide:
     >>> from openpathsampling import PLUMEDCV, PLUMEDInterface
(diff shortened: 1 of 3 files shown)


sroet:
The `-0.19%` coverage in `openmm/engine.py` is the removal of the unneeded import of `copy` (while the number of not-coverd lines stays the same, so a lower percentage is covered)

openpathsampling/collectivevariables/plumed_wrapper.py
```diff
@@ -27,7 +27,7 @@ class PLUMEDCV(paths.CoordinateFunctionCV):
     https://doi.org/10.1016/j.cpc.2013.09.018
 
     Examples
-    --------
+
```
dwhswenson, line 30:
```suggestion
    --------
```

The original was correct. Examples is a valid numpydoc section, as shown in the [numpydoc example](https://numpydoc.readthedocs.io/en/latest/example.html).

I personally don't use it very often because most examples are either so trivial or so complicated that I don't find them informative.

dwhswenson (changes requested):
Thanks for the cleanup. Looks good except one tiny request to undo one change.

General thought on this: I think that `MDTrajTopology` should be moved out of OpenMM and into `engines.topology` (with canonical import as `paths.engines.MDTrajTopology`). I also sometimes have to go hunting for it, and the point is that it isn't an OpenMM-specific tool. It's used (implicitly) in the Gromacs engine, and will be relevant for most other MD engines anyway. (Obviously such a change would come with continued support -- with deprecation -- at the current locations where it can be imported.)

openpathsampling/collectivevariables/plumed_wrapper.py
sroet: good catch, that is an accidental delete.

dwhswenson (approved):

PR #1012, opened 2021-04-15
Move MDTrajTopology

### LINKED ISSUE #1011: MDTrajTopology should be moved from engines.OpenMM to engines.topology
In #1010 @dwhswenson wrote:
> General thought on this: I think that MDTrajTopology should be moved out of OpenMM and into engines.topology (with canonical import as paths.engines.MDTrajTopology). I also sometimes have to go hunting for it, and the point is that it isn't an OpenMM-specific tool. It's used (implicitly) in the Gromacs engine, and will be relevant for most other MD engines anyway. (Obviously such a change would come with continued support -- with deprecation -- at the current locations where it can be imported.)

This is an issue to keep track of that

sroet:
closes #1011 

This PR moves `MDTrajTopology` from `openpathsampling.engines.openmm.topology` to `openpathsampling.engines.topology

This was a slightly bigger hassle than intended (because we want to warn on import). 

I ended up mimicking [`importing ABCs from collections instead of collections.abc is deprecated`](https://github.com/python/cpython/blob/3.7/Lib/collections/__init__.py#L49) 

This also makes `mdtraj` a hard requirement


openpathsampling/engines/openmm/topology.py
```diff
@@ -1,95 +1,14 @@
-import mdtraj as md
-import numpy as np
-import pandas as pd
```
sroet, line 3: @dwhswenson I assumed I could get away with not supporting `openpathsampling.engines.openmm.topology.md` etc anymore. (technically an API break, and can be added to the `__getattr__` solution if you still think we should keep this API in place)

devtools/minimal.txt
```diff
@@ -8,3 +8,4 @@ matplotlib
 svgwrite
 ujson
 psutil
+mdtraj
```
sroet, line 11: due to the move `mdtraj` became a hard requirement

setup.cfg
```diff
@@ -39,8 +39,6 @@ install_requires =
     matplotlib
     ujson!=2.*,!=3.*,!=4.0.0,!=4.0.1
     mdtraj
-# mdtraj is not technically required, but we co-package it because it is
-# required for many integrations with other packages
```
sroet, line 43: it is now a hard requirement

openpathsampling/engines/openmm/topology.py
dwhswenson: Not an API break: our API promise is for the first level (breadth-first search) where it can be imported. Since MDTraj can be imported from outside of OPS, it's not in our scope.

The entire importable API for 1.0 is listed [in this file](https://github.com/openpathsampling/openpathsampling/blob/master/openpathsampling/tests/test_data/apis/ops1x0_api.txt), and checking that we haven't violated it is literally the first test in each CI run (mainly because `test_api` comes first alphabetically). This is the official list of the "documented API" -- and because it is in unit tests, we're guaranteed to not to break it.

If you're curious, the script I used to generate that file is in my [define_api](https://github.com/dwhswenson/define_api) repo. It may have also involved some manual editing of the output file, but most of it should have been generated by `./define_api.py --runtype=in_api --api-file=ops_api_dirs.txt openpathsampling`.

However, it looks like we didn't guarantee `MDTrajTopology` in that. This is probably because it was not imported into a package, as it should have been (and I overlooked it when doing the `--runtype=not_in_api` run). Moving it now will put it on the API list when I regenerate that for 1.5 (which will be after I generate the versions for the tags at 1.1, 1.2, 1.3, and 1.4).

devtools/minimal.txt
dwhswenson: Can't you do a `try/except ImportError` and then use `paths.integration_tools.error_if_no_mdtraj('MDTrajTopology')` at the start of the `from_dict`? (Maybe also `__init__`, although `__init__` doesn't actually use MDTraj.

dwhswenson:
Note (as much to self as to others): This can be merged despite the CodeCov fails. The changes are moving code that was not fully covered. Therefore the patch code isn't covered.

I suspect that the drop in project coverage (7 lines) may be due to making MDTraj a hard requirement. (It shouldn't be, see https://github.com/openpathsampling/openpathsampling/pull/1012#discussion_r614284757 for suggested fix.)

sroet:
@dwhswenson, reverted the hard mdtraj dependency. Please have another look (even though codecov is still not happy about it).

openpathsampling/deprecations.py
```diff
@@ -154,6 +154,16 @@ def version_tuple_to_string(version_tuple):
     deprecated_in=(1, 1, 0)
 )
 
+OPENMM_MDTRAJTOPOLOGY = Deprecation(
+    problem=("openpathsampling.engines.openmm.topology.MDTrajTopology "
+             "has been moved."),
+    remedy=("Import MDTrajTopology from openpathsampling.engines.topology "
+            "instead"),
```
dwhswenson, line 161: Could you add `MDTrajTopology` to the list of things imported into `openpathsampling.engines`? (i.e., edit `openpathsampling/engines/__init__.py`). Other topologies are in that namespace. That should be the canonical one. Then this diff should read:

```suggestion
    remedy=("Import MDTrajTopology from openpathsampling.engines instead."),
```

dwhswenson (changes requested):
Looks basically good. Only changes:

1. `deprecated_in` should be *next* release (1.5.0)
2. Please import `MDTrajTopology` into the `openpathsampling.engines` namespace (via the `__init__.py`), just as `Topology` is. On this point, I do care that the deprecation message tells people to load from `openpathsampling.engines`, since that will be the canonical API location. However, I don't care as much about the docstrings. Besides, I think the more correct way to write the docstrings would be `` :class:`.MDTrajTopology` ``.

dwhswenson (approved):

dwhswenson:
Looks good now, thanks! Merging.

PR #1015, opened 2021-04-20
rename random => rng

dwhswenson:
This renames the `openpathsampling.random` module (see #998) to `openpathsampling.rng`. The fact that the name `random` shadows the std library's `random` when the main OPS source directory is my cwd was wreaking havoc with scripts I use in my usual workflow.

I'm open to other possible names for the module -- just that `random` is causing me too many headaches.


sroet:
I am sorry to hear this is wreaking havoc for you... No preference from me on the new name. 

If you do this, do you mind also removing:
`from __future__ import absolute_import` and `import random` from `openpathsampling/rng.py`? (These are not needed anymore if `stdlib random` is available again via regular imports)

dwhswenson:
> I am sorry to hear this is wreaking havoc for you.

No worries! It was kind of funny to see how many things loaded by my zsh profile/used in a script in my regular git workflow/part of a VIM plugin imported `random`.

> If you do this, do you mind also removing: `from __future__ import absolute_import` and `import random` from `openpathsampling/rng.py`?

Done.

To give anyone else an opportunity to comment on the specific name I'm changing this to, I will leave it open for at least another 12 hours, merging no earlier than Wed 21 Apr 10:00 GMT (12:00 local).

PR #1016, opened 2021-04-28
MAX_SQL_ITEMS as class variable

dwhswenson:
This converts the `max_query_size` that @sroet had introduced as an instance variable in `SQLStorageBackend` to a class variable, and adds its use in a corner case that @gyorgy-hantal came across.

It makes more sense to have this as a class variable because it is determined by the way the SQL backend (usually SQLite) was compiled on the local machine, and therefore will be constant throughout a session (for however many files the user opens). Of course, it can still be overridden on an instance-by-instance basis.

Currently WIP because I'm hoping that I can get a regression test, but I may mark it as ready to merge without such a test if it is too annoying to create. @gyorgy-hantal has confirmed that this fixed his issue.

### DIFF
--- a/openpathsampling/experimental/simstore/sql_backend.py
+++ b/openpathsampling/experimental/simstore/sql_backend.py
@@ -288,7 +288,7 @@ class SQLStorageBackend(StorableNamedObject):
         table = self.metadata.tables[table_name]
         results = []
         with self.engine.connect() as conn:
-            for block in grouper(idx_list, 1000):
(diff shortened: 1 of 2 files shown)


sroet (approved):
LGTM

dwhswenson:
I couldn't reproduce the error locally (I think because of the way my SQLite is compiled) but this fixed the error on the user end, and I *did* add tests that should, in principle, check for regression. So I'm going to go ahead and merge this.

PR #1017, opened 2021-05-03
Pin OpenMM != 7.5.1

dwhswenson:
A change in OpenMM 7.5.1 is causing our tests to fail, and could cause problems for users if they trigger a NaN during dynamics with OpenMM. See https://github.com/openmm/openmm/issues/3098.


dwhswenson:
I'm going to merge this immediately, since it is a tiny change to get tests passing again. No need for review.

PR #1019, opened 2021-05-07
Test against OpenMM RCs

dwhswenson:
Every time I see an OpenMM release candidate announced, I think, "I should really try that out and run the OPS test suite against it." But I don't. And this time we got hit by a bug (https://github.com/openmm/openmm/issues/3098) that could have been fixed in prerelease had I run those tests.

So the other day I created a GitHub Action that checks for recent prereleases of a conda package. This PR is to implement use of it here. I think I'll need to merge this PR before I can actually check that everything works, because the workflow that gets triggered to run the tests won't exist until it has been merged to master. So this may require a follow-up PR to complete everything, although the basic idea seems to be working fine in https://github.com/dwhswenson/conda-rc-check/.

Overall approach I plan to take: Every other day we check to see if an RC has been released within the last 3 days. If so, we trigger a workflow that will test against that RC. This is a simplified version of our test suite, only testing against the current default Python version from miniconda, and using the OpenMM RC. (The 2 day/3 day difference is in case GitHub's scheduler doesn't run at exactly the same time, so this ensures some overlap in the time windows.)

For now I'm using the `main` branch of my Action, but I'll release a 1.0 of that once I've tried it out here.

### DIFF
--- a/.github/workflows/check-openmm-rc.yml
+++ b/.github/workflows/check-openmm-rc.yml
@@ -1,0 +1,27 @@
+# Workflow to check for a recent OpenMM release candidate.
+#
+# If a recent RC is found, trigger a second workflow to test against it.
+# See also: test-openmm-rc.yml
+name: "Check for OpenMM RC"
+on:
+  schedule:
+    - cron: "0 9 * * */2"
+
+jobs:
+  check-rc:
+    runs-on: ubuntu-latest
+    name: "Check for OpenMM RC"
+    steps:
+      - uses: actions/checkout@v2
+      - uses: dwhswenson/conda-rc-check@main
+        id: checkrc
+        with:
+          channel: conda-forge
+          package: openmm
+          ndays: 30  # TODO: decrease to 7
+          labels: openmm_rc
+      - uses: benc-uk/workflow-dispatch@v1
+        with:
+          workflow: Test OpenMM Release Candidate
+          token: ${{ secrets.DISPATCH_TOKEN }}
+        if: ${{ steps.checkrc.outputs.hasrc == 'True' }}
--- a/.github/workflows/check-openmm-rc.yml
+++ b/.github/workflows/check-openmm-rc.yml
@@ -18,7 +18,7 @@ jobs:
         with:
           channel: conda-forge
           package: openmm
-          ndays: 30  # TODO: decrease to 7
+          ndays: 30  # TODO: decrease to 3
           labels: openmm_rc
       - uses: benc-uk/workflow-dispatch@v1
         with:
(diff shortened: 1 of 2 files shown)


dwhswenson:
> Just so that if this starts failing it is easy to (temporary) activate this for running on PRs (that might be opened in order to fix it)

If I understand what you're asking for, I think the way to do this is basically to undo 508061c. I left the lines commented to remind me how to do it in the future. So you just tell the `test-openmm-rc.yml` workflow to trigger on PR: 4b404d0 has a red X by it because it **correctly** tests against the most recent RC. The check called "Test OpenMM Release Candidate / Tests (pull_request)" fails because the issue fixed by pinning in #1017 is present in that RC.

I'll run a quick check now to see if I can also trigger the `check-openmm-rc.yml` workflow via PR. I'm not sure if `workflow-dispatch` requires the triggered workflow to be merged to the default branch or not.

FWIW, the actual code that searches anaconda.org for a prerelease is kind of an ugly hack. There must be some URL where I can get info on all the releases in JSON or XML format (as there is for PyPI), but I couldn't find it. So I ended up scraping the web pages. Writing Actions is really easy, though.

dwhswenson:
> I'll run a quick check now to see if I can also trigger the `check-openmm-rc.yml` workflow via PR. I'm not sure if `workflow-dispatch` requires the triggered workflow to be merged to the default branch or not.

It doesn't (of course) because the secret token can't be accessed from a PR coming from a fork. (The `workflow-dispatch` action I'm using requires a GitHub personal access token to authenticate with the GitHub API and trigger the workflow.)

sroet:
> FWIW, the actual code that searches anaconda.org for a prerelease is kind of an ugly hack. There must be some URL where I can get info on all the releases in JSON or XML format (as there is for PyPI), but I couldn't find it. So I ended up scraping the web pages. Writing Actions is really easy, though.

The jsons can be found at 
https://conda.anaconda.org/conda-forge/{subdir}/repodata_from_packages.json.bz2"

(Taken from the repository that is used to apply pathes on conda-forge): https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/blob/master/recipe/show_diff.py (look for the `download_subdir`)

dwhswenson:
> The jsons can be found at
> [https://conda.anaconda.org/conda-forge/{subdir}/repodata_from_packages.json.bz2](https://conda.anaconda.org/conda-forge/%7Bsubdir%7D/repodata_from_packages.json.bz2)

I believe this is for ALL the packages in that architecture. This is way more than I want to download for this. I was hoping for something similar to what we get at https://pypi.org/pypi/openpathsampling/json -- a nice JSON dict specific to the package, with key "releases" mapping to all the releases.

dwhswenson:
I'll go ahead and merge this now. I'm expecting to see an error before the end of the weekend, because I've artificially increased the time for an RC to be considered "new", and the most recent OpenMM RC raises errors. Expect a follow-up PR on this in a few days, where I'll reduce the time searched for RCs.

PR #1018, opened 2021-05-06
ignore more flaky pyemma py 2.7

sroet:
This failed for 2.7 with:
```
__________________ examples/tests/test_pyemma.ipynb::Cell 10 ___________________
Notebook cell execution failed
Cell 10: Cell outputs differ

Input:
py_cv = storage.cvs['pyemma']

Traceback:
Unexpected output fields from running code: set([u'stderr'])
```

Fixed in a similar fashion as #1009 

[edit]: [link to the failing test](https://github.com/sroet/openpathsampling/runs/2515098923)

PR #1022, opened 2021-05-12
ignore cell 15 pyemma test

sroet:
Similar to #1018 , I know the drill by now.

Today py2.7 failed with:
```
__________________ examples/tests/test_pyemma.ipynb::Cell 15 ___________________
Notebook cell execution failed
Cell 15: Cell outputs differ

Input:
py_cv_idx = storage.idx(py_cv)
print(py_cv_idx)
py_emma_feat = storage.vars['attributes_json'][py_cv_idx]

Traceback:
Unexpected output fields from running code: set([u'stderr'])
```
This also ignores that cell

### DIFF
--- a/examples/tests/test_pyemma.ipynb
+++ b/examples/tests/test_pyemma.ipynb
@@ -329,6 +329,7 @@
     }
    ],
    "source": [
+    "# NBVAL_IGNORE_OUTPUT\n",
     "py_cv_idx = storage.idx(py_cv)\n",
     "print(py_cv_idx)\n",
     "py_emma_feat = storage.vars['attributes_json'][py_cv_idx]"


sroet:
@dwhswenson Do you have any list of functionality that you want to have tested via this notebook? (So I might exchange this notebook with proper unit-tests if I run into this when I have some more time)

dwhswenson:
> Do you have any list of functionality that you want to have tested via this notebook? (So I might exchange this notebook with proper unit-tests if I run into this when I have some more time)

I would be very much in favor of converting this to proper unit tests. Also something I just haven't had time to do (it's less of a headache to just re-run when I wake up to a failing test.)

It looks to me like the notebook basically tests the following:

1. PyEMMA CVs give correct results (which might be better tested again MDTraj results on the same thing -- I think that, under the hood, PyEMMA is using MDTraj, but this can be used to ensure that it works correctly.
2. PyEMMA CV results are properly stored and can be reloaded from storage
3. PyEMMA CVs can be applied to iterables, returning an iterable (might also be good to check that snapshots return scalars)
4. PyEMMA CVs (the CV object itself) can be stored and reloaded

I suspect that actually putting that stuff into the real unit tests will make a decent bump in test coverage, because I suspect that a lot of the stuff tested here is what never gets covered anywhere else. (Of course, it's also stuff that will be removed in 2.0.)

dwhswenson (approved):

PR #1023, opened 2021-05-13
convert pyemma to unittests

sroet:
Got another error like #1022 this morning, but this time for cell 21.

Decided to make unittests instead

@dwhswenson , this should test every point from [this comment](https://github.com/openpathsampling/openpathsampling/pull/1022#issuecomment-839749850)


sroet:
@dwhswenson I am not completely certain that I covered point:
> PyEMMA CV results are properly stored and can be reloaded from storage

But I tried to mimic what was done is the notebook, let me know if that was insufficient

openpathsampling/tests/test_pyemma.py
```diff
@@ -0,0 +1,92 @@
...
+
+        def pyemma_generator(f):
+            f.add_inverse_distances(f.pairs(f.select_Backbone()))
+
+        self.cv = paths.collectivevariable.PyEMMAFeaturizerCV(
+            'pyemma',
+            pyemma_generator,
+            topology=self.ops_trajectory[0].topology
+            ).with_diskcache()
+        self.storage = paths.Storage('delete.nc', 'w')
```
dwhswenson, line 27: Use the [pytest `tmpdir` fixture](https://docs.pytest.org/en/6.2.x/tmpdir.html#the-tmpdir-fixture) here and you can save yourself the teardown hassle

openpathsampling/tests/test_pyemma.py
```diff
@@ -0,0 +1,92 @@
...
+import os
+import pytest
+from itertools import combinations
+from .test_helpers import data_filename
+from openpathsampling.engines.openmm.tools import trajectory_from_mdtraj
+pyemma = pytest.importorskip("pyemma")
+mdtraj = pytest.importorskip("mdtraj")
+
+
+class TestPyEMMA(object):
```
dwhswenson, line 12:
```suggestion
class TestPyEMMAFeaturizerCV(object):
```

Could you also rename the file similarly? At least to `test_pyemma_cv.py`; `test_pyemmafeaturizercv`  might be a bit hard to read.

examples/ipynbtests.sh
```diff
@@ -64,7 +64,7 @@ py.test --nbval --current-env \
     test_snapshot.ipynb \
     test_netcdfplus.ipynb \
     test_cv.ipynb \
-    test_pyemma.ipynb \
+    # test_pyemma.ipynb \ got covered by unittests instead
```
dwhswenson, line 67: you'll need to fully remove this line; the continuation `\` now has a comment after it, so the `||` underneath is no longer part of the same line in bash.

dwhswenson (changes requested):
Thanks! In principle, looks good; a couple small changes. I think that's as thorough of a unit test for storing CV results results as one could hope unless @jhprinz were to write it.

openpathsampling/tests/test_pyemma.py
sroet: That one does not seem to catch on `setup` functions (`setup(self, tmpdir)` triggers a `missing 1 required positional argument` error). 

So then I would have to boilerplate the storage creation for every test. I think this setup/teardown is more clear

dwhswenson (approved):
LGTM! Thanks

dwhswenson:
Huh, I'm surprised this didn't have a bigger change on coverage. It made an expected difference on `collectivevariable.py`, but I thought it might do more in storage.

PR #1024, opened 2021-05-13
Shooting point selector with arbitrary bias

dwhswenson:
For a very long time, OPS has had the ability to use a Gaussian bias in shooting point selection. That is, given a CV `f`, the  probability of selecting a shooting point `x` within a given trajectory is proportional to `np.exp(-a*(f(x)-x0)**2)`, where `a` and `x0` are parameters set by the user.

There is no reason for this to be restricted to Gaussians. This PR introduces a `BiasedSelector`, which allows one to use *any* function (wrapped in an OPS CV) as the bias function. The only reason this didn't exist before is because when we were writing shooting point selectors, I don't think we understood the flexibility that OPS CVs would have. 

Now, for example, the `GaussianBiasSelector` could be implemented as:

```python
def gaussian(snapshot, cv, a, x0):
    return np.exp(-a * (cv(snapshot)-x0)**2)

cv_gaussian = paths.FunctionCV('shooting_bias', gaussian, cv=my_cv, a=2.0, x0=0.25)
selector = paths.BiasedSelector(cv_gaussian)
```

But similarly, one could use any other function.

### DIFF
--- a/docs/shooting.rst
+++ b/docs/shooting.rst
@@ -11,3 +11,7 @@ ShootingPointSelectors
     ShootingPointSelector
     UniformSelector
     GaussianBiasSelector
+    BiasedSelector
+    InterfaceConstrainedSelector
+    FirstFrameSelector
+    FinalFrameSelector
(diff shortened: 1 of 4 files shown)


sroet (approved):
LGTM (assuming codecov is happy)

dwhswenson:
This is ready for review and comment. I will leave it open for at least 24 hours, merging no earlier than Fri 14 May 18:00 GMT (20:00 local).

PR #1027, opened 2021-06-03
Ensure that SimStore always saves movescheme

dwhswenson:
Based on an email report from @gyorgy-hantal. It seems that SimStore was not saving the move scheme in the rare case that all steps fail due to maximum length/NaN. This seems to be because of a difference in SimStore allowing you to stash several steps for saving, whereas netcdfplus immediately saves after each step.

This PR also introduces a (not well-tested) `unpatch` method to undo the SimStore monkey patches. This will mainly be useful in testing.


dwhswenson:
This is ready for review and comment. I will leave it open for at least 24 hours, merging no earlier than Fri 04 Jun 19:00 GMT (15:00 my local).

PR #1029, opened 2021-06-16
Make histogramming more robust

sroet:
Issue found by a master student:
If `cv` singletons are not unpacked the `SparseHistogram` might return broken output bins, which then fail in following code that assumes it is 1D (Like in [this line](https://github.com/openpathsampling/openpathsampling/blob/aaff2b8e41cf42a13bbb5ce55a498ff66d03e8c6/openpathsampling/analysis/path_histogram.py#L210)).

This PR forces a reshape to prevent mangled output

master:
```python
>>> from openpathsampling.numerics.histogram import SparseHistogram
>>> histo = SparseHistogram(bin_widths=(0.5, 0.3), left_bin_edges=(0.0, -0.1)) 
>>> data = ([0.0],[1.0])
>>> histo.map_to_bins(data)
(array([0., 0.]), array([2., 3.]))
>>> max(histo.map_to_bins(data))
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

This PR:
```python
>>> from openpathsampling.numerics.histogram import SparseHistogram
>>> histo = SparseHistogram(bin_widths=(0.5, 0.3), left_bin_edges=(0.0, -0.1)) 
>>> data = ([0.0],[1.0])
>>> histo.map_to_bins(data)
(0.0, 3.0)
>>> max(histo.map_to_bins(data))
3.0
```

It also fixes some pep8 issues in histogram.py

### DIFF
--- a/openpathsampling/numerics/histogram.py
+++ b/openpathsampling/numerics/histogram.py
@@ -87,7 +87,7 @@ class SparseHistogram(object):
             the bin that the data represents
         """
         # Reshape data to prevent accidental wrong output
-        #data = np.asarray(data).reshape(self.left_bin_edges.shape)
+        data = np.asarray(data).reshape(self.left_bin_edges.shape)
         return tuple(np.floor((data - self.left_bin_edges) / self.bin_widths))
 
     def add_data_to_histogram(self, data, weights=None):
--- a/openpathsampling/numerics/histogram.py
+++ b/openpathsampling/numerics/histogram.py
@@ -1,12 +1,12 @@
 import numpy as np
 import pandas as pd
-import scipy
 import matplotlib.pyplot as plt
 import math
 from .lookup_function import LookupFunction, VoxelLookupFunction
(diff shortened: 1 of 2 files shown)


sroet:
Codecov complaint is about a line that only has whitespace changes

dwhswenson:
This is fine, but users creating CVs that return len-1 arrays seems to be a recurring issue that causes problems in multiple aspects of analysis. I'll accept this PR, but in general I'd be more interested in finding ways to discourage these CVs in the first place. I'll open a PR that tries to address this by raising a warning if using the `CVDefinedVolume` with a len-1 array.

The fundamental problem is that Python (and NumPy) allow ordering of len-1 iterables, which often lets the problem slip through sampling without causing problem. However, later analyses fail when they require that these be scalars (e.g., combining them into arrays).

dwhswenson (approved):

PR #1030, opened 2021-06-17
Warn if CVDefinedVolume CVs return iterables

dwhswenson:
A recurring issue has been that users create CVs that return length-1 arrays/lists, and then use these CVs to define volumes/interfaces. Because Python/NumPy allow order comparisons on length-1 iterables as if they were scalars, sampling goes fine. However, some aspects of analysis expect the CVs to output scalars, and then users get cryptic error messages at unexpected points.

This PR emits a warning if the CV used by a `CVDefinedVolume` returns an iterable.

See also #1029. 

- [x] Initial implementation for len-1 arrays
- [x] Warn on any iterable: the CV for a `CVDefinedVolume` should not, in general, be any kind of iterable (although things that aren't orderable will fail later anyway)
- [x] Add tests

### DIFF
--- a/openpathsampling/tests/test_volume.py
+++ b/openpathsampling/tests/test_volume.py
@@ -7,9 +7,18 @@ from builtins import object
 from nose.tools import (assert_equal, assert_not_equal, assert_is, raises,
                         assert_true, assert_false)
 from nose.plugins.skip import Skip, SkipTest
-from .test_helpers import CallIdentity, raises_with_message_like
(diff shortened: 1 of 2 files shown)


dwhswenson:
This is ready for review and comment. I will leave it open for at least 24 hours, merging no earlier than Sat 19 Jun 19:00 GMT (15:00 my local).

In general, I'm not a fan of this sort of type checking. However, we've seen problems with this coming from multiple users, so I think it's worth giving a reasonably early warning that a user might notice and fix before doing a lot of sampling.

PR #1036, opened 2021-07-19
Fix DataFrame.sort_index future warning

sroet:
The test suite gave the following future warning:
```python
openpathsampling/tests/test_lookup_function.py::TestVoxelLookupFunction::test_df_2d
  /home/sroet/github_files/openpathsampling/openpathsampling/numerics/lookup_function.py:268: FutureWarning: In a future version of pandas all arguments of DataFrame.sort_index will be keyword-only
    df = df.sort_index(0).sort_index(1)
```
066f87a implements the fix for that (make the arguments keyword based instead of location based).
90c75cf solves the `pep8` complaints my coding environment was noticing

### DIFF
--- a/openpathsampling/numerics/lookup_function.py
+++ b/openpathsampling/numerics/lookup_function.py
@@ -2,6 +2,7 @@ import pandas as pd
 import numpy as np
 import collections
 
+
 class LookupFunction(object):
     """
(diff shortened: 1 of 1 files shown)


dwhswenson (approved):
LGTM. Will merge!

PR #1037, opened 2021-07-20
Fix OpenMM engine docs on n_frames_max

### LINKED ISSUE #817: openMM engine does not accept: n_frames_max=None
The docs say int or None is an accepted value but setting None I get a ValueError. Setting to integer values works.
I think the problem is the combination of the default value beeing an int for the openMM engine and the basic dynamics engine checking that the types of variables match the default type in `openpathsampling.engines.dynamic_engine._check_options()`.

hejung:
Resolves #817
I looked through the issues I opened and realized that this one is a real low hanging fruit :)


dwhswenson (approved):
Definitely an easy fix! 😄  LGTM.

Issue #1040, opened 2021-07-28
Moving in dihedral space


samarjeet:
I am new to OPS but I love its modular design! Kudos to the architects! I would like to use OPS to move between two states by moving in the space of dihedrals i.e. each MC step changes the dihedral angles of the conformation. I am wondering if I need to create a new PathMover subclass for this. I will really appreciate some guidance on this! TIA

dwhswenson:
Thanks for the interest in OPS! If I understand correctly, you want to do path sampling where your "trajectories" will come from MC (based on moves in dihedral space) instead of MD. This should be perfectly feasible, but the thing you'll want to change will be the `DynamicsEngine`.

If you can use OpenMM to calculate energies, I think the easiest path would be:

1. Use [OpenMMTools](https://github.com/choderalab/openmmtools) and create a custom [MCMC Move](https://openmmtools.readthedocs.io/en/0.18.1/mcmc.html) for your dihedral move (see the `MCDisplacementMove` and `MCRotationMove` for analogous examples -- I haven't done this before, but it looks like you just need to write a `_propose_positions` method, plus helpers like `__getstate__` which allow the MC move to be stored to disk)
2. Wrap that MC move in an OpenMM integrator (you may want to check with the OpenMMTools team for details on that -- I can imagine some problems here).
3. Once you have that integrator, it can be used directly in the existing OPS OpenMM engine -- just replace the VVVR integrator in our examples with your MC move integrator.

If you can't use OpenMM (or prefer to use an existing tool for the MC), or if it isn't possible to wrap the MC move in an OpenMM integrator, then what you'd want would be to create a subclass of `DynamicsEngine`. Docs are a bit incomplete (as always), but the [section on writing custom engines](http://openpathsampling.org/latest/developers/engines.html) has a decent start.

dwhswenson:
I've thought about this a little more, and I'm pretty sure using the OpenMM MCMC would require a new `DynamicsEngine`. Good news is, writing a new engine is pretty easy. So I did the core of it already. See #1041.

samarjeet:
That awesome! I am still working on proposing a step in dihedral space. I also have plan to propose steps in normal mode space. The new `DynamicsEngine` should be useful for this as well. Thanks!

PR #1041, opened 2021-07-28
[WIP] OpenMMToolsMCEngine

dwhswenson:
This provides an engine that wraps the tools in [`openmmtools.mcmc`](https://openmmtools.readthedocs.io/en/0.18.1/mcmc.html). This will allow users to perform path sampling simulations of Monte Carlo "trajectories."

- [x] Basic dynamics
- [x] Serialization
- [x] Support for MDTraj
- [x] Allow same input options as `MCSampler` (don't required `WeightedMove`)
- [x] Docstrings
- [ ] Example
- [x] Tests

### DIFF
--- a/examples/misc/mc_engine.ipynb
+++ b/examples/misc/mc_engine.ipynb
@@ -1,0 +1,353 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "1ede3493",
+   "metadata": {},
+   "source": [
+    "# The OpenMMTools Monte Carlo Engine\n",
+    "\n",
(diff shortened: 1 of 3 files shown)


dwhswenson:
At this point, this seems to work. However, sampling single-atom Cartesian displacements, even for alanine dipeptide in vacuum, is ***SLOW***. I think I need to improve the premise of the example before it is useful pedagogically without the context of this PR and #1040.

Performance is also not great. I haven't profiled it yet, but I wonder if we're going to and from the GPU with each MC trial. It looks to me like the OpenMMTools MCMC stuff was designed with hybrid Monte Carlo in mind, where one proposal creation would involve a significant chunk of work on the GPU.

Right now, I can't get an equilibrated trajectory because (1) trajectories are very long because single atom displacements require many frames for a transition; and (2) each frame seems to be quite slow compared to MD. (And of course (3) I need my computer for other things!)

PR #1042, opened 2021-08-02
New `iter_generate` for OPS 2.0

dwhswenson:
One of the things I want to do in OPS 2.0 is simplify code in API-breaking ways, when the API isn't something that is widely used. One case of this is `DynamicsEngine.iter_generate`. The existing `iter_generate`has a huge complexity rating (McCabe cyclomatic complexity of 46; CodeClimate "cognitive complexity" of 92). This comes because it tries to implement a bunch of user-facing options for error handling into one catch-all function, instead of letting the client code handle the errors. This is all the stuff about how to handle NaNs or max-length error (whether to restart/retry/etc.)

In path sampling, many of those options would be just wrong. A trajectory that generates a NaN is a rejected trajectory. Retrying will bias the sampling. In addition, when I tried to write tests for it, it became clear that some of that functionality isn't even implemented correctly.

So this PR rewrites `iter_generate` without the extra error handling functionality. That error handling should be done in the calling code. We raise errors that should make it clear to calling code what the problem is; special treatments in case of certain error types can be done by calling code (this is never something that should be done on a per-engine level, but something that should be done on a situational basis).

There are two places where engine writers can change the error handling:

* `engine.is_valid_snapshot()`: Check whether the created snapshot is valid. Generally will just return a bool. If you want to use the default error behavior (raise an `EngineError`), this is all you need to change.
* `engine.snapshot_error_handler()`: If you want to modify the error in some way, or if there are some errors that you can recover from, do that here. If you want an error to be raised, then return the error. If you can recover from the error and create a valid snapshot, then return the valid snapshot.

For most users, this makes no difference. It means that under the hood we're using a different function. The API break is the drop of engine-level support for error handling, and a small change in the call signature of `iter_generate` (removing the `intervals` parameter). I don't think anyone will miss these things.

- [x] Code for new `iter_generate`
- [x] Use new `iter_generate` for all tests
- [ ] Add support for correcting snapshot
- [ ] Unit tests for new `iter_generate`; full coverage
- [ ] Update docs on writing engines
- [ ] Remove options for NaN handling from engines
- [ ] Remove docstrings about NaN handling from engines

### DIFF
--- a/openpathsampling/__init__.py
+++ b/openpathsampling/__init__.py
@@ -29,6 +29,9 @@ except ImportError:  # pragma: no cover
         version.full_version += ".dev-" + version.git_version[:7]
     isrelease = str(ops_setup.preferences['released'])
 
+
+from . import utils
+
 from .analysis.path_histogram import PathDensityHistogram
 
 from .analysis.replica_network import (
--- a/openpathsampling/engines/dynamics_engine.py
+++ b/openpathsampling/engines/dynamics_engine.py
@@ -471,14 +471,16 @@ class DynamicsEngine(StorableNamedObject):
                                     trusted=False)
 
         if stop:
-            return trajectory
+            yield trajectory
+            return
 
         self.start()
         snapshot = None
         while not stop:
 
-            # first we check for maximum length
-            if len(trajectory) >= max_length:
+            # first we check for maximum length; forcible setting
+            # max_length to None will allow infinite trajectories
+            if max_length is not None and len(trajectory) >= max_length:
                 # should only hit on ==, but >= as safety
                 raise EngineMaxLengthError(
                     "Hit maximum trajectory length: %d frames" % max_length,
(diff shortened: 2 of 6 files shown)


dwhswenson:
There's a weird thing in the unit tests for the minus mover. I changed some behavior there, but I honestly don't understand why the previous behavior was occurring.

Essentially, the minus move consists of the replica exchange stage followed by an extension stage. There's a test for the case that the minus move fails because the extension hits max length, which uses this code. For some reason, the old test checked that the extension trajectory was of a length longer than the engine's maximum. I don't know why that would be. The test itself also modified the engine's error handling (setting `'on_max_length'` to `'stop'`, not `'fail'`), so I expect that was part of it. If anyone else has an idea as to why I did it that way, please let me know.

Relevant code is in tests for both minus mover and single replica minus mover:

https://github.com/openpathsampling/openpathsampling/blob/c67edf7c42f440361298a542c54f2e1fd1bf8631/openpathsampling/tests/test_pathmover.py#L1355-L1358

https://github.com/openpathsampling/openpathsampling/blob/c67edf7c42f440361298a542c54f2e1fd1bf8631/openpathsampling/tests/test_pathmover.py#L1517-L1522

dwhswenson:
Not sure why we're getting @codecov updates on a branch where the CI doesn't report coverage... I wouldn't trust the data.

PR #1043, opened 2021-08-02
Update CI for OPS 2.0

dwhswenson:
The main development branch CI is in `.github/workflows/tests.yml`. The CI for the 2.0 development branch is in `.github/workflows/test-2.0.yml`. This PR ports improvements in `tests.yml` into `test-2.0.yml` (and, of course, targets the `dev-2.0` branch).

Useful improvements:

* Adds Python 3.8 and 3.9 to the test suite. This wasn't possible (no OpenMM available) before.
* Adds notebook tests. I'm not sure why this wasn't done before, but with core changes on the way, such as #1042, I'll feel a lot better having the more thorough test suite.

PR #1044, opened 2021-08-03
Fix syntax warning

sroet:
When I looked at the tests in #1043, I noticed a:
```
  /home/runner/work/openpathsampling/openpathsampling/openpathsampling/tests/test_external_engine.py:48: SyntaxWarning: "is" with a literal. Did you mean "=="?
    if line is '':
```

This solves that warning, drops `nose` from that particular test file (part of #756), and solve some style complaints

### DIFF
--- a/openpathsampling/tests/test_external_engine.py
+++ b/openpathsampling/tests/test_external_engine.py
@@ -45,7 +45,7 @@ class ExampleExternalEngine(peng.ExternalEngine):
         # nothing exists, linecache returns '', so we return None.
         # Otherwise, try to make a snapshot and return "partial" if we fail
         line = linecache.getline(filename, first_line)
-        if line is '':
+        if line == '':
             snap = None
         else:
             try:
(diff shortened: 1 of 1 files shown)


openpathsampling/tests/test_external_engine.py
```diff
@@ -123,33 +122,33 @@ def test_start_stop(self):
...
-            assert_equal(eng.proc.is_running(), False)
+            assert not eng.proc.is_running()
         except AttributeError:
-            pass # if eng.proc doesn't exist, then it isn't running
+            pass  # if eng.proc doesn't exist, then it isn't running
 
         # start it; check that it is running
         eng.start(self.template)
-        assert_equal(eng.proc.is_running(), True)
+        assert eng.proc.is_running()
```
sroet, line 131: I arbitrarily decided to test against `truthy` here, not `is True`. Let me know if you want to make this `is True` (and `is False` further down

dwhswenson: I think truthy is the right way to go. I had a bad habit of automatically using `assert_equal` for everything -- just muscle memory on every assert.

dwhswenson (approved):
LGTM. Thanks!

PR #1045, opened 2021-08-03
Raise FileNotFoundError if sql backend tries to read a file that does not exist

sroet:
I tried reading a file that did not exist in that directory, this lead to 2 things happening:
1) the file was created (this made it so that it took way too long before I figured out what was going on)
2) the code failed with: 
```
~/github_files/openpathsampling/openpathsampling/experimental/simstore/sql_backend.py in __init__(self, filename, mode, sql_dialect, **kwargs)
    148             )
    149             engine = sql.create_engine(self.connection_uri, **self.kwargs)
--> 150             self._initialize_from_engine(engine)
    151             self.mode = mode  # in case we changed when checking existence
    152 

~/github_files/openpathsampling/openpathsampling/experimental/simstore/sql_backend.py in _initialize_from_engine(self, engine)
    154         self.engine = engine
    155         self._metadata = sql.MetaData(bind=self.engine)
--> 156         self._initialize_with_mode(self.mode)
    157 
    158     @property

~/github_files/openpathsampling/openpathsampling/experimental/simstore/sql_backend.py in _initialize_with_mode(self, mode)
    208         elif mode == "r" or mode == "a":
    209             self.metadata.reflect(self.engine)
--> 210             self.schema = self.database_schema()
    211             self.table_to_number, self.number_to_table = \
    212                     self.internal_tables_from_db()

~/github_files/openpathsampling/openpathsampling/experimental/simstore/sql_backend.py in database_schema(self)
    592             schema dictionary
    593         """
--> 594         schema_table = self.metadata.tables['schema']
    595         sel = schema_table.select()
    596         with self.engine.connect() as conn:

KeyError: 'schema'
```
This PR ( in 7d90a63) adds logic to raise a FileNotFound error with a test

Implementing the code and running that test led to some Warning/Errors being raised, so I fixed those in the following commits:

293c867 resolves an error my coding enviroment was throwing about `self` not being defined in a `classmethod`
3d8f488 resolves a couple `DeprecationWarning: invalid escape sequence \(` (or `\.` )inside regex strings and resolves some pep8 complaints in touched files
b8d6947 resolves `SADeprecationWarning: The Engine.table_names() method is deprecated and will be removed in a future release.  Please refer to Inspector.get_table_names(). (deprecated since: 1.4)` in the proposed way ([documentation link](https://docs.sqlalchemy.org/en/14/core/reflection.html#fine-grained-reflection-with-inspector)) and removes a leftover `pytest.skip()` at the end of a completed test

### DIFF
--- a/openpathsampling/experimental/simstore/test_sql_backend.py
+++ b/openpathsampling/experimental/simstore/test_sql_backend.py
@@ -273,3 +273,7 @@ class TestSQLStorageBackend(object):
             expected = tuple(samp_dct[k]
                              for k in ['replica', 'ensemble', 'trajectory'])
             assert row[2:] == expected
+
+    def test_non_existing_file(self):
+        with pytest.raises(FileNotFoundError, match="foo.sql"):
+            SQLStorageBackend("foo.sql")
--- a/openpathsampling/experimental/simstore/test_sql_backend.py
+++ b/openpathsampling/experimental/simstore/test_sql_backend.py
@@ -1,6 +1,6 @@
 from .sql_backend import *
 import pytest
-
+sqla = pytest.importorskip("sqlalchemy")
 
 class TestSQLStorageBackend(object):
     def setup(self):
@@ -88,7 +88,7 @@ class TestSQLStorageBackend(object):
             'write': SQLStorageBackend('test1.sql', mode='w'),
             'append': SQLStorageBackend('test2.sql', mode='a'),
         }[db]
-        table_names = database.engine.table_names()
+        table_names = sqla.inspect(database.engine).get_table_names()
         assert set(table_names) == self.default_table_names
         assert self._col_names_set('uuid') == {'uuid', 'table', 'row'}
         assert self._col_names_set('tables') == {'name', 'idx', 'module',
@@ -100,7 +100,7 @@ class TestSQLStorageBackend(object):
             'snapshot1': [('filename', 'str'), ('index', 'int')]
         }
         self.database.register_schema(new_schema, self.table_to_class)
-        table_names = self.database.engine.table_names()
+        table_names = sqla.inspect(self.database.engine).get_table_names()
         assert set(table_names) == self.default_table_names | {'snapshot1'}
         assert self._col_names_set('snapshot1') == {'idx', 'uuid',
                                                     'filename', 'index'}
@@ -177,7 +177,6 @@ class TestSQLStorageBackend(object):
         uuid_table_rows = self.database.load_uuids_table(uuids)
         loaded_table = self.database.load_table_data(uuid_table_rows)
         assert len(loaded_table) == 3
-        pytest.skip()
 
     def test_load_save_large_numbers(self):
         # mainly a smoke test to ensure that we reload everything correctly
--- a/openpathsampling/experimental/simstore/test_sql_backend.py
+++ b/openpathsampling/experimental/simstore/test_sql_backend.py
@@ -1,6 +1,5 @@
 from .sql_backend import *
 import pytest
-sqla = pytest.importorskip("sqlalchemy")
 
 class TestSQLStorageBackend(object):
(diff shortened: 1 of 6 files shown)


openpathsampling/experimental/simstore/test_sql_backend.py
```diff
@@ -1,6 +1,6 @@
 from .sql_backend import *
 import pytest
-
+sqla = pytest.importorskip("sqlalchemy")
```
dwhswenson, line 3:
```suggestion
```

(or `import sqlalchemy as sql`)

1. `importorskip` shouldn't be necessary. If running SimStore tests, SQLAlchemy is a hard requirement. Ignoring these tests (unless you're running from the `experimental` directory) is handled by our `pytest.ini`.
2. You already have the import (as `sql`) from `from .sql_backend import *`. Actually, it looks like the more common short name for SQLAlchemy is `sa`, but lets stay consistent within the project.

(Some stats on "as" names [here](https://import-as.github.io/import/sqlalchemy.html), though I have no idea where the numbers come from. It seems to cover all short names with at most four characters. (No `paths`, but it does have [`toys`](https://import-as.github.io/import/openpathsampling.engines.toy.html) for the toy engine!)

openpathsampling/experimental/simstore/test_sql_backend.py
```diff
@@ -88,7 +88,7 @@ def test_setup(self, db):
             'write': SQLStorageBackend('test1.sql', mode='w'),
             'append': SQLStorageBackend('test2.sql', mode='a'),
         }[db]
-        table_names = database.engine.table_names()
+        table_names = sqla.inspect(database.engine).get_table_names()
```
dwhswenson, line 91:
```suggestion
        table_names = sql.inspect(database.engine).get_table_names()
```

openpathsampling/experimental/simstore/test_sql_backend.py
```diff
@@ -100,7 +100,7 @@ def test_register_schema(self):
             'snapshot1': [('filename', 'str'), ('index', 'int')]
         }
         self.database.register_schema(new_schema, self.table_to_class)
-        table_names = self.database.engine.table_names()
+        table_names = sqla.inspect(self.database.engine).get_table_names()
```
dwhswenson, line 103:
```suggestion
        table_names = sql.inspect(self.database.engine).get_table_names()
```

dwhswenson (changes requested):
Good catch! The fact that we have to do all this file handling ourselves frustrating, but it comes from the fact that most SQL databases don't really use a file model.

Tiny changes, mainly to use the SQLAlchemy as `sql` as in `sqlbackend.py`.

SQLAlchemy 1.4 made a bunch of changes. It's essentially a preview of their 2.0 while still maintaining some backward compatibility. There may be a number of other things we will have to change at some point, but they do have a [clear migration guide](https://docs.sqlalchemy.org/en/14/changelog/migration_20.html). I think it's okay to require 1.4 even though it was just released in March, since I don't think we need to have backward compatibility for things in the experimental package.

sroet:
@dwhswenson added the suggestions, please have another look.

> Good catch! The fact that we have to do all this file handling ourselves frustrating, but it comes from the fact that most SQL databases don't really use a file model.

Yeah, I understand how from an sql perspective that they would just create a file if it does not exists. In this case we should deal with it to prevent users getting confused. I was mainly thrown for a loop when I did `ls` after getting the error and the file was actually there (it was just 0 bytes).

dwhswenson (approved):
LGTM! Merging.

Correcting my previous statement:

> since I don't think we need to have backward compatibility for things in the experimental package.

Should read: "I don't think we need to support old versions of requirements for things in the experimental package." I *do* think that at this point we need to support backward compatibility of files as much as possible. (Files made with today's version should open in tomorrow's version.)

PR #1046, opened 2021-08-06
Prevent errors in ShootingPointAnalysis with invalid (e.g., max length) trajectories

sroet:
A student tried to use ShootingPointAnalysis on a comittor simulation. However, due to the nature of his (biological) system, some of the steps were rejected due to reaching maximum length.

He thus ran into the (leftover) AssertionError.

This PR cleans up that error handling and adds an option to skip rejected steps:
e46fe67 adds the functionality and tests
2a529c4 solves a pep8 complaint from my coding environment
37e5428 drops nose from the test file and cleans up some unnecessary compatibility code (this test file should be inspected again when dropping python 2.7)

For 2.0 I would propose to break API and change the default of `skip_rejected` to `True`. (I think that rejected steps should not be included in analysis on default).

### DIFF
--- a/openpathsampling/analysis/shooting_point_analysis.py
+++ b/openpathsampling/analysis/shooting_point_analysis.py
@@ -163,7 +163,6 @@ class ShootingPointAnalysis(SimpleProgress, SnapshotByCoordinateDict):
                        "endpoints in any of the states.")
                 raise NoFramesInStateError(err)
             if total_count > len(test_points):
-                # TODO: is this actually wrong / an error?
                 err = ("The " + str(len(test_points)) +
                        " end points of the trail trajectory from step " +
                        str(step.mccycle) +
--- a/openpathsampling/analysis/shooting_point_analysis.py
+++ b/openpathsampling/analysis/shooting_point_analysis.py
@@ -265,8 +265,10 @@ class ShootingPointAnalysis(SimpleProgress, SnapshotByCoordinateDict):
         except TypeError:
             ndim = 1
         if ndim > 2 or ndim < 1:
-            raise RuntimeError("Histogram key dimension {0} > 2 or {0} < 1 "
-                               + "(key: {1})".format(ndim, key))
+            err = ("Histogram key dimension {0} > 2 or {0} < 1 "
+                   "(key: {1})").format(ndim, key)
+
+            raise RuntimeError(err)
         return ndim
 
     def committor_histogram(self, new_hash, state, bins=10):
--- a/openpathsampling/analysis/shooting_point_analysis.py
+++ b/openpathsampling/analysis/shooting_point_analysis.py
@@ -76,6 +76,14 @@ class SnapshotByCoordinateDict(TransformedDict):
                                                        *args, **kwargs)
 
 
+class NoFramesInStateError(ValueError):
+    pass
+
+
+class MoreStatesThanFramesError(ValueError):
+    pass
+
+
 class ShootingPointAnalysis(SimpleProgress, SnapshotByCoordinateDict):
     """
     Container and methods for shooting point analysis.
@@ -91,15 +99,13 @@ class ShootingPointAnalysis(SimpleProgress, SnapshotByCoordinateDict):
     states : list of :class:`.Volume`
         volumes to consider as states for the analysis. For pandas output,
         these volumes must be named.
-    skip_rejected: bool, default False
-         boolean flag to skip rejected moves
+    error_if_no_state: bool, default True
+         boolean flag to error on steps that don't end in one of the states
     """
-    def __init__(self, steps, states, skip_rejected=False):
+    def __init__(self, steps, states, error_if_no_state=True):
         super(ShootingPointAnalysis, self).__init__()
         self.states = states
-        self.skip_rejected = skip_rejected
-        if steps and self.skip_rejected:
-            steps = [step for step in steps if step.change.canonical.accepted]
+        self.error_if_no_state = error_if_no_state
         if steps:
             self.analyze(steps)
 
@@ -112,7 +118,16 @@ class ShootingPointAnalysis(SimpleProgress, SnapshotByCoordinateDict):
             MC steps to analyze
         """
         for step in self.progress(steps):
-            self.analyze_single_step(step)
+            try:
+                self.analyze_single_step(step)
+            except NoFramesInStateError as err:
+                if self.error_if_no_state:
+                    addition = ("\nTo disable this error set "
+                                "'error_if_no_state=False'")
+                    raise type(err)(str(err) + addition)
+
+                else:
+                    pass
 
     def analyze_single_step(self, step):
         """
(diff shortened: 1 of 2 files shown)


openpathsampling/analysis/shooting_point_analysis.py
```diff
@@ -138,8 +143,25 @@ def analyze_single_step(self, step):
...
+                raise ValueError(err)
+            if total_count > len(test_points):
+                # TODO: is this actually wrong / an error?
+                err = ("The " + str(len(test_points)) +
+                       " end points of the trail trajectory from step " +
+                       str(step.mccycle) +
+                       " found " + str(total_count) + " stable states."
+                       "\n Are you sure your states don't overlap?"
+                       )
+                raise ValueError(err)
```
sroet, line 163: @dwhswenson I implemented this to be consistent with the (idea of) original assertion (check if you find more states than tested points). However, I am not convinced this actually wrong/ something we should raise on.

Is there any analysis that breaks if you snaps do not uniquely map to a single state?

openpathsampling/tests/test_shooting_point_analysis.py
```diff
@@ -197,81 +220,79 @@ def test_non_shooting_steps(self):
...
-            assert_equal(committor_A[snap],
-                         old_div(float(self.analyzer[snap][self.left]), 20.0))
-            assert_almost_equal(committor_A[snap] + committor_B[snap], 1.0)
-            assert_equal(committor_B[snap],
-                         old_div(float(self.analyzer[snap][self.right]), 20.0))
+            assert (committor_A[snap] ==
+                    float(self.analyzer[snap][self.left]) / 20.0)
+            assert pytest.approx(committor_A[snap] + committor_B[snap]) == 1.0
+            assert (committor_B[snap] ==
+                    float(self.analyzer[snap][self.right]) / 20.0)
```
sroet, line 246: I removed `old_div` here as I am pretty sure in python `2.7` `float/float` is identical to the implementation in python `3`

openpathsampling/analysis/shooting_point_analysis.py
```diff
@@ -91,11 +91,16 @@ class ShootingPointAnalysis(SimpleProgress, SnapshotByCoordinateDict):
...
     """
-    def __init__(self, steps, states):
+    def __init__(self, steps, states, skip_rejected=False):
         super(ShootingPointAnalysis, self).__init__()
         self.states = states
-        if steps is not None:
+        self.skip_rejected = skip_rejected
+        if steps and self.skip_rejected:
+            steps = [step for step in steps if step.change.canonical.accepted]
+        if steps:
```
sroet, line 103: changed this from `is not None` to test against `truthy`, which seems like a robuster approach to catch empty lists etc early

openpathsampling/tests/test_shooting_point_analysis.py
dwhswenson: Tests were automatically converted from Python 2 to 3. I think the tool to do that automatically replaced `/` with `old_div` in all cases. So yeah, if you see strange use of `old_div` in tests, that's why.

(BTW, on mobile now. Will do a more complete review (much) later.)

dwhswenson (changes requested):
I like the idea, but I think `skip_rejected` is the wrong level of granularity.

If I understand your thinking here, the argument is that all rejected committor steps are missing frames in the stable states, so there's a 1:1 mapping between missing-stable-state and rejected.

This is true for the committor, but the SP analysis can also be used with output from path sampling, where rejected paths can have frames in the stable states (and could also have the real problem you're addressing here: rejections due to max length).

Maybe something like `error_if_no_state`, default True? I think a custom error thrown by `analyze_single_step` and caught by `analyze` should allow you to skip those steps.

sroet:
> This is true for the committor, but the SP analysis can also be used with output from path sampling, where rejected paths can have frames in the stable states (and could also have the real problem you're addressing here: rejections due to max length).

Feel free to correct me here, but I thought most analysis that is done on path-sampling is done on accepted steps (with their MC weight). However ShootingPointAnalysis does not do that. If this (analyzing all trials independent of acceptance) is indeed the intention, do you think catching the RejectedMaxLength/RejectedNan movechanges would be a better solution? (With something like a `skip_engine_rejection` flag)?

dwhswenson:
This is more about analysis that could be built on SP analysis than on anything we have in OPS now (although we do have an example somewhere of calculating committor histograms on MSTIS data). There are several methods that at least start from the observation that, if the system is sufficiently diffuse, each shooting move is basically a committor shot.

IIRC, those ideas were behind some of the reweighted path ensemble analysis, especially when it came to calculating projections of the committor. I think it was also part of the idea behind VIE-TPS, although that shouldn't need to use this code.

For now, I think checking endpoints is best. The problem with checking for "engine rejections" is that it wouldn't work for an early-rejection shooting move (where you pre-calculate the length at which the shooting move would be rejected). We haven't implemented that yet because I think it's dangerous in systems that you don't know well, but it's a perfectly valid shooting move. But if someone used that, then they cannot use the idea that each shooting move is a committor trial, and this analysis should fail or at least raise a lot of warnings. 

I think the actual best test would be if `EngineMover`s reported `forward_partial` and/or `backward_partial` in their `details`. Then you could easily check whether the last-generated frame was in a state. But that's beyond the scope of this PR.

sroet:
@dwhswenson I moved it to `error_if_no_state` and made sure that the intended behavior (raise on both accepted and rejected steps) is tested. I would still like your feedback on what the intention is with the error we raise when the code finds more states than endpoints

dwhswenson (changes requested):
Thanks! This looks pretty good. Minor changes still to do -- overall picture looks great, just a couple suggestions around error hierarchy and a request to warn rather than completely silence the error.

openpathsampling/analysis/shooting_point_analysis.py
```diff
@@ -107,7 +118,16 @@ def analyze(self, steps):
...
+            try:
+                self.analyze_single_step(step)
+            except NoFramesInStateError as err:
+                if self.error_if_no_state:
+                    addition = ("\nTo disable this error set "
+                                "'error_if_no_state=False'")
+                    raise type(err)(str(err) + addition)
+
+                else:
+                    pass
```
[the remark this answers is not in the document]
sroet: So, the code now warns once per `mccycle` as normally warnings are filtered per message and the MC cycle is the number that is updated. Would that be sufficient for now?

sroet:
@dwhswenson implemented the suggestion and added a test that makes sure more than 1 warning is emitted.

openpathsampling/analysis/shooting_point_analysis.py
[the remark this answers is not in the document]
dwhswenson: Yep, once per `mccycle` is what I meant. (Since this is based on trial trajectories from shooting moves, that's essentially the same as once per trajectory, but you're right that the loop is over steps/`mcccycle`.)

dwhswenson (approved):
Thanks! LGTM; merging.

PR #1047, opened 2021-08-11
Change ShootingPointAnalysisError to an Exception

sroet:
This changes the Error type of the ShootingPointAnalysis from an `AssertionError` (to ensure backwards compatability) to an more reasonable ~`ValueError`~ `Exception` for OPS 2.0. 

(I could not figure out how to do an easy DeprecationWarning on this; My current guess would be to write a metaclass that would update the MRO/`__bases__` of this error to a subclass of tuple that would warn if the index that contained the `AssertionError` was accessed, something like [this SO answer](https://stackoverflow.com/a/20832588/15100302)) 

Edit: this also merges the current master into OPS 2.0

Edit 2: after 2d95011 this now becomes an `Exception` instead of a `ValueError`

### DIFF
--- a/openpathsampling/analysis/shooting_point_analysis.py
+++ b/openpathsampling/analysis/shooting_point_analysis.py
@@ -72,8 +72,7 @@ class SnapshotByCoordinateDict(TransformedDict):
                                                        *args, **kwargs)
 
 
-class ShootingPointAnalysisError(AssertionError):
-    # TODO this should inherit from a different Error type in OPS 2.0
+class ShootingPointAnalysisError(ValueError):
     pass
 
 
--- a/openpathsampling/analysis/shooting_point_analysis.py
+++ b/openpathsampling/analysis/shooting_point_analysis.py
@@ -72,7 +72,7 @@ class SnapshotByCoordinateDict(TransformedDict):
                                                        *args, **kwargs)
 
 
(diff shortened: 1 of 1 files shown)


dwhswenson (commented):
I thought about this before merging #1046 -- if we make these into `ValueError`s, then we should use multiple inheritance in in the 1.x series so it also inherits from `ValueError` there.

I agree that `ValueError` is the best choice among the other concrete [built-in-exceptions](https://docs.python.org/3/library/exceptions.html). But what is the case for inheriting from `ValueError` over just `Exception`?

The case for inheriting from `Exception` is to avoid shadowing other errors (e.g., if someone tries to catch a `ValueError` coming from checking whether a snapshot is in a state, they would also catch these errors).

As for deprecations, I think MRO hacks are serious overkill here -- I doubt anyone is currently catching these assertion errors, and we're clearly documenting the change through these PRs.

sroet:
> I agree that ValueError is the best choice among the other concrete built-in-exceptions. But what is the case for inheriting from ValueError over just Exception?

There isn't, I just didn't remember that `Exception` was one of the options :laughing: , fixed it in 2d95011

dwhswenson (approved):
LGTM (pretty straightforward to review!) Will merge after green.

PR #1050, opened 2021-08-20
add access via slice

sroet:
Today, I got annoyed today that I could not do
`steps = storage.steps[:100]` in some test code using simstore.

This adds handling of slices in the `__getitem__` logic. 
It behaves as a `list` would do (so an empty list on an empty slice, a list with a single item for slices like `[4:5]`, a single item with `int` and a `TypeError` if it is not an `int` or `slice`)

It should also leverage batched loading, if I understood the code correctly

One assumption that I added is that the `name` of objects can only be strings (for `_name_to_uuid`, I don't know if this is currently an assumption/requirement, but all tests seem to pass on this.

This code is `python > .3.6` (f strings) and the tests are `python > 3.7` (leverages dict order)

### DIFF
--- a/openpathsampling/experimental/simstore/storage.py
+++ b/openpathsampling/experimental/simstore/storage.py
@@ -587,14 +587,28 @@ class StorageTable(abc.Sequence):
 
     def __getitem__(self, item):
         len_self = len(self)
-        if not (-len_self <= item < len_self):
-            raise IndexError("table index out of range")
+        if item is slice:
+            unpack_return = False
+            items = (i for i in range(*item.indices(len_self)))
+        elif type(item) is int:
+            unpack_return = True
+            if not (-len_self <= item < len_self):
+                raise IndexError("table index out of range")
 
-        if item < 0:
-            item += len(self)
+            if item < 0:
+                item += len(self)
+            items = [item]
+        else:
+            raise TypeError("Only acces via slice or int allowed, "
+                            f"got type {type(item)}.")
 
-        row = self.storage.backend.table_get_item(self.table, item)
-        return self.storage.load([row.uuid])[0]
+        rows = [self.storage.backend.table_get_item(self.table, item).uuid
+                for item in items]
+        if unpack_return:
+            return self.storage.load(rows)[0]
+        else:
+            return self.storage.load(rows)
+
 
     def __len__(self):
         return self.storage.backend.table_len(self.table)
@@ -647,7 +661,9 @@ class PseudoTable(abc.MutableSequence):
     def __getitem__(self, item):
         try:
             ret_val = self._sequence[item]
-        except TypeError:
+        except TypeError as err:
+            if type(item) != str:
+                raise err
             uuid = self._name_to_uuid[item]
             ret_val = self.get_by_uuid(uuid)
         return ret_val
--- a/openpathsampling/experimental/simstore/storage.py
+++ b/openpathsampling/experimental/simstore/storage.py
@@ -69,7 +69,7 @@ class GeneralStorage(StorableNamedObject):
                                                         {})
 
         # self._pseudo_tables = {table_name: dict()
-                               # for table_name in self.simulation_classes}
+        #                        for table_name in self.simulation_classes}
         self._simulation_objects = {}
         self._pseudo_tables = {table_name: PseudoTable()
                                for table_name in self.simulation_classes}
@@ -159,8 +159,8 @@ class GeneralStorage(StorableNamedObject):
         # check validity
         self.class_info.register_info(class_info_list, schema)
         # for info in class_info_list:
-            # info.set_defaults(schema)
-            # self.class_info.add_class_info(info)
+        #     info.set_defaults(schema)
+        #     self.class_info.add_class_info(info)
 
         schema_types = [type_str for attr_list in schema.values()
                         for _, type_str in attr_list]
@@ -213,7 +213,6 @@ class GeneralStorage(StorableNamedObject):
 
         return uuid_dict
 
-
     def _uuids_by_table(self, input_uuids, cache, get_table_name):
         # find all UUIDs we need to save with this object
         logger.debug("Listing all objects to save")
@@ -243,7 +242,6 @@ class GeneralStorage(StorableNamedObject):
         loaded = self.load(lazies, allow_lazy=False)
         uuid_mapping.update({get_uuid(obj): obj for obj in loaded})
         return uuid_mapping
-
 
     def save(self, obj_list, use_cache=True):
         if type(obj_list) is not list:
@@ -384,11 +382,11 @@ class GeneralStorage(StorableNamedObject):
         logger.debug("Getting internal structure of %d non-cached objects",
                      len(uuid_list))
         to_load, lazy_uuids, dependencies, uuid_to_table = \
-                get_all_uuids_loading(uuid_list=uuid_list,
-                                      backend=self.backend,
-                                      schema=self.schema,
-                                      existing_uuids=self.cache,
-                                      allow_lazy=allow_lazy)
+            get_all_uuids_loading(uuid_list=uuid_list,
+                                  backend=self.backend,
+                                  schema=self.schema,
+                                  existing_uuids=self.cache,
+                                  allow_lazy=allow_lazy)
         logger.debug("Loading %d objects; creating %d lazy proxies",
                      len(to_load), len(lazy_uuids))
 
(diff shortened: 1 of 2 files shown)


openpathsampling/experimental/simstore/storage.py
```diff
@@ -587,20 +585,33 @@ def __iter__(self):
...
+            return self.storage.load(rows)[0]
+        else:
+            return self.storage.load(rows)
 
     def __len__(self):
         return self.storage.backend.table_len(self.table)
 
     def cache_all(self):
-        old_blocksize = self.iter_block_size
+        # old_blocksize = self.iter_block_size
```
sroet, line 614: @dwhswenson this variable was unused, but I didn't know if you had any future plans with this. If not, this line can be removed

openpathsampling/experimental/simstore/storage.py
```diff
@@ -587,20 +585,33 @@ def __iter__(self):
...
+        elif type(item) is int:
+            unpack_return = True
+            if not (-len_self <= item < len_self):
+                raise IndexError("table index out of range")
+
+            if item < 0:
+                item += len(self)
+            items = [item]
+        else:
+            raise TypeError("Only acces via slice or int allowed, "
```
dwhswenson, line 600:
```suggestion
            raise TypeError("Only access via slice or int allowed, "
```

openpathsampling/experimental/simstore/storage.py
```diff
@@ -587,20 +585,33 @@ def __iter__(self):
...
+            return self.storage.load(rows)
 
     def __len__(self):
         return self.storage.backend.table_len(self.table)
 
     def cache_all(self):
-        old_blocksize = self.iter_block_size
+        # old_blocksize = self.iter_block_size
         self.iter_block_size = len(self)
         _ = list(iter(self))
```
dwhswenson, line 616:
```suggestion
        old_blocksize = self.iter_block_size
        self.iter_block_size = len(self)
        _ = list(iter(self))
        self.iter_block_size = old_blocksize
```

dwhswenson, line 616:
```suggestion
        old_blocksize = self.iter_block_size
        self.iter_block_size = len(self)
        _ = list(iter(self))
        self.iter_block_size = old_blocksize
```

openpathsampling/experimental/simstore/storage.py
dwhswenson: It looks to me like I forgot to reset the block size. See suggestion for this method (separate comment in review because of GitHub internals). I don't think this has any significant effect in practice, but it does make sense for `cache_all` to *temporarily* change the block size for iteration into "load everything at once." As the code stands, it permanently makes that change.

I don't think it matters in practice, because once everything is cached, the value of `iter_block_size` shouldn't need to be used again. But the cleaner way to structure the code is to not have non-obvious side effects. Thanks for catching this!

openpathsampling/experimental/simstore/storage.py
```diff
@@ -384,11 +382,11 @@ def load(self, input_uuids, allow_lazy=True, force=False):
...
-                get_all_uuids_loading(uuid_list=uuid_list,
-                                      backend=self.backend,
-                                      schema=self.schema,
-                                      existing_uuids=self.cache,
-                                      allow_lazy=allow_lazy)
+            get_all_uuids_loading(uuid_list=uuid_list,
+                                  backend=self.backend,
+                                  schema=self.schema,
+                                  existing_uuids=self.cache,
+                                  allow_lazy=allow_lazy)
```
dwhswenson, line 389: Not urgent, but we have an IDE mismatch thing here. Your IDE says that backslash continuation gets one level of indentation; mine says two levels. In general, I like the 2-level indentation because I feel like backslash continuation is different from other indentation (like parentheses), mainly because backslashes should be avoided if possible. Can you check if that's a setting that you can modify?

(On the other hand, your IDE does a kind of block-comment thing that mine does not -- that approach is better, and I should try to figure out if I can start doing that. I hope that's just the difference between be typing `,c ` and `,cs` -- that was the difference for C/C++ code, at least).

openpathsampling/experimental/simstore/storage.py
```diff
@@ -587,20 +585,33 @@ def __iter__(self):
 
     def __getitem__(self, item):
         len_self = len(self)
-        if not (-len_self <= item < len_self):
-            raise IndexError("table index out of range")
-
-        if item < 0:
-            item += len(self)
+        if item is slice:
```
dwhswenson, line 588: I'm utterly confused by this line. It seems to work (tests pass!), but I thought it would need to be:

```python
        if item isinstance(slice):
```

Any link to docs on why this works? I thought this would test against the builtin `slice`, and the item passed to `__getitem__` would be an instance of `slice`, not the class itself.

dwhswenson (changes requested):
Thanks -- I've also noticed this as a minor annoyance (usually I'm too busy with whatever I'm working on and just use `list(storage.steps)[start:end]`, because hacks like that are obvious to me -- that kind of hack is often NOT obvious to our users, and we shouldn't expect them to know it -- so this is a very useful improvement.)

The only required change is to fix a typo in an error message. Beyond that, I hope you'll explain why one part of the code works (because I don't get it!), and other things are fixes for a mistake I made and some general "let's agree on some code style that isn't officially defined in PEP-8" stuff.

openpathsampling/experimental/simstore/storage.py
sroet: It doesn't work :facepalm:, I guess this is why we have code review 

The test is actually not hitting this code... but instead does slicing on `PseudoTable._sequence`, which is a `list` by default, so of course it behaves as a list :P Let me ponder on how to actually test this and make sure we catch the real error xD ("tuple" also shows up if you try to slice a list with a tuple, as a good error message should do)

openpathsampling/experimental/simstore/storage.py
sroet: I have no strong preference here, so will revert this change

sroet:
> Thanks -- I've also noticed this as a minor annoyance (usually I'm too busy with whatever I'm working on and just use list(storage.steps)[start:end], because hacks like that are obvious to me -- that kind of hack is often NOT obvious to our users, and we shouldn't expect them to know it -- so this is a very useful improvement.)

Yeah, that would be a solution if I actually could load all the steps (100_000) on my local machine in a reasonable time, I can't :laughing: 

My hack around was to write something like
```python
stop = 100:
steps = []
for i, e in storage.steps:
    steps.append(e)
    if i > stop:
        break
```
Which was not great xD

With d1c8609, I actually test the correct code path. I also tested that this works (and hits the code) on the "Simstore and CLI" minitutorial.

sroet:
@dwhswenson please have another look, thanks for catching the wrong code!

sroet:
also, I would like to know if:
> One assumption that I added is that the name of objects can only be strings (for _name_to_uuid), I don't know if this is currently an assumption/requirement, but all tests seem to pass on this.

Is a reasonable assumption to enforce

dwhswenson:
> One assumption that I added is that the `name` of objects can only be strings (for `_name_to_uuid`), I don't know if this is currently an assumption/requirement, but all tests seem to pass on this.

I think this ***should*** be the case, although it looks like SimStore isn't raising an error if you try to save an object with a non-string name. The `StorableNamedObject.named` method allows non-string names. As long as the `name` attribute is something that SimStore can turn into a JSON string, it will be saved in SimStore. (netcdfplus does catch this error on saving, because it doesn't treat the `name` attribute as part of the JSON that is generated.)

PR coming to make this explicitly required.

openpathsampling/experimental/simstore/storage.py
```diff
@@ -587,14 +585,27 @@ def __iter__(self):
...
+                            f"got type {type(item).__name__}.")
 
-        row = self.storage.backend.table_get_item(self.table, item)
-        return self.storage.load([row.uuid])[0]
+        rows = [self.storage.backend.table_get_item(self.table, item).uuid
+                for item in items]
+        if unpack_return:
+            return self.storage.load(rows)[0]
+        else:
+            return self.storage.load(rows)
```
dwhswenson, line 608:
```suggestion
        uuids = [self.storage.backend.table_get_item(self.table, item).uuid
                 for item in items]
        if unpack_return:
            return self.storage.load(uuids)[0]
        else:
            return self.storage.load(uuids)
```

Rename `rows` => `uuids` for readability. Usually `row` refers to the entire entry in the database (all columns).

dwhswenson (changes requested):
One more tiny change for readability (sorry to miss this one before). Otherwise, looks good.

I was debating whether to do `isinstance(item, slice)` or `type(item) is slice`: I agree that there's no distinction, since `slice` is final (can't be subclassed), but I'd also probably look for this code by grepping "isinstance". Anyway, either one is fine.

PR #1051, opened 2021-08-21
Fixes for OpenMM 7.6 RC

dwhswenson:
Good news: the [GitHub action I wrote](https://github.com/marketplace/actions/check-conda-for-release-candidates) to automatically test against OpenMM release candidates works! The tests ran against the newest RC.

Bad news: we explode with errors on the 7.6 RC. This is because the OpenMM import has moved out of the `simtk` namespace (so it is now correct to do `import openmm`; `import openmm.unit`). The old imports still work for the `simtk.openmm`, `simtk.openmm.app`, and `simtk.unit` namespaces. However, we were also using some OpenMM internals (specifically, `simtk.openmm.app.internal.unitcell.reducePeriodicBoxVectors`.) Importing from these does not work.

This PR fixes that, and makes the non-`simtk` import the preferred choice.

### DIFF
--- a/openpathsampling/engines/openmm/__init__.py
+++ b/openpathsampling/engines/openmm/__init__.py
@@ -1,10 +1,9 @@
 def missing_openmm(*args, **kwargs):
     raise RuntimeError("Install OpenMM to use this feature")
 
-try:
-    import simtk.openmm
-    import simtk.openmm.app
-except ImportError:
+from openpathsampling.integration_tools import HAS_OPENMM
+
+if not HAS_OPENMM:
     HAS_OPENMM = False
     Engine = missing_openmm
     empty_snapshot_from_openmm_topology = missing_openmm
--- a/openpathsampling/engines/openmm/__init__.py
+++ b/openpathsampling/engines/openmm/__init__.py
@@ -4,7 +4,6 @@ def missing_openmm(*args, **kwargs):
 from openpathsampling.integration_tools import HAS_OPENMM
 
 if not HAS_OPENMM:
-    HAS_OPENMM = False
     Engine = missing_openmm
     empty_snapshot_from_openmm_topology = missing_openmm
     snapshot_from_pdb = missing_openmm
--- a/openpathsampling/tests/test_integration_tools.py
+++ b/openpathsampling/tests/test_integration_tools.py
@@ -12,5 +12,5 @@ def test_chain_import(modules):
     assert mod is os.path
 
 def test_chain_import_error():
-    with pytest.raises(ImportError):
+    with pytest.raises(ImportError, match="bar"):
         _chain_import('foo', 'bar')
--- a/openpathsampling/tests/test_integration_tools.py
+++ b/openpathsampling/tests/test_integration_tools.py
@@ -1,0 +1,16 @@
+import pytest
+
+from openpathsampling.integration_tools import _chain_import
+
+@pytest.mark.parametrize('modules', [
+    ('foo', 'os.path'),
+    ('os.path', 'foo')
+])
+def test_chain_import(modules):
+    import os.path
+    mod = _chain_import(*modules)
+    assert mod is os.path
+
+def test_chain_import_error():
+    with pytest.raises(ImportError):
+        _chain_import('foo', 'bar')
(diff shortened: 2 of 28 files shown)


dwhswenson:
This is ready for review and comment. I will leave it open for at least 24 hours, merging no earlier than Sun 22 Aug 23:00 GMT (19:00 my local).

Changing the way we import `reducePeriodicBoxVectors` in `engines.openmm.tools` is what fixes the errors we were getting in the RC. However, most of the changes were to drop the `simtk` name everywhere that I could. Additionally, notebook tests required ensuring that we didn't import from `simtk` in a way that caused the warning message to be emitted.

Right now, we're silencing the import warning that OpenMM emits when we import MDTraj. That's not ideal, because it actually silences all possible times the OpenMM error would be emitted (warning emitted when source is imported, which only happens once).  I have a PR already merged into MDTraj that should fix the issue there, so we can relax the logger silencing next time MDTraj releases.

I also changed the frequency that we check for OpenMM RCs to daily, instead of every other day. The only argument to do it less frequently is that it adds more noise in the main Actions tab. I think being able to provide rapid feedback on the RC outweighs that.

I turned the tests of the RC off in https://github.com/openpathsampling/openpathsampling/pull/1051/commits/ef6cf33ea1c0cf939d6e10a6ea406c7795989665. See tests for https://github.com/openpathsampling/openpathsampling/pull/1051/commits/e53fd6ee8842f6bb6f54de54fe435d5001903456 for the passing tests against the RC.

It looks like all changes merged in since 1.5.1 were things that can go into a patch release, so I think I'll go ahead and cut 1.5.2 from `master` after this is merged in, so that OPS is ready for OpenMM 7.6 when it comes out.

openpathsampling/engines/openmm/__init__.py
```diff
@@ -1,10 +1,9 @@
...
     raise RuntimeError("Install OpenMM to use this feature")
 
-try:
-    import simtk.openmm
-    import simtk.openmm.app
-except ImportError:
+from openpathsampling.integration_tools import HAS_OPENMM
+
+if not HAS_OPENMM:
     HAS_OPENMM = False
```
sroet, line 7: the extra HAS_OPENMM is redundant
```suggestion
from openpathsampling.integration_tools import HAS_OPENMM

if not HAS_OPENMM:
```

openpathsampling/tests/test_integration_tools.py
```diff
@@ -0,0 +1,16 @@
...
+    ('foo', 'os.path'),
+    ('os.path', 'foo')
+])
+def test_chain_import(modules):
+    import os.path
+    mod = _chain_import(*modules)
+    assert mod is os.path
+
+def test_chain_import_error():
+    with pytest.raises(ImportError):
```
sroet, line 15: Do you mind testing "last error" behavior? 
```suggestion
    with pytest.raises(ImportError, match="bar"):
```

sroet (changes requested):
LGTM,

2 small nitpicks:
- one unnecessary line of code
- make sure we test against enforcing raising the "last" import error

Don't hang this PR on these issues

sroet:
As a side comment, can we make sure these tests only run on openpathsampling/openpathsampling? I am getting failure emails with:
```
Run benc-uk/workflow-dispatch@v1
  with:
    workflow: Test OpenMM Release Candidate
Error: Parameter token or opts.auth is required
```

dwhswenson:
> As a side comment, can we make sure these tests only run on openpathsampling/openpathsampling? 

Yes. This should be fixed by https://github.com/openpathsampling/openpathsampling/pull/1051/commits/2f6af6a79c30e2067fb9c6701468be55fbe75fd6.  I'm still very confused by some aspects of how GitHub actions deals with forks. First, I don't know why they made the decision to enable actions in forks by default (discussion [here](https://github.community/t/stop-github-actions-running-on-a-fork/17965/3) illustrates reasons this is confusing, including that it is different from about every other CI platform). Second, I have no idea why I am not getting those error messages. One of the reasons I develop from a personal fork is to make my developer experience the same as any other contributor. (So that I would get annoyed by these messages as quickly as anyone else.) Maybe at some point I changed a preference so I wouldn't get doubled email notifications? But I can't find that preference now....

sroet (approved):
perfect, thanks. LGTM!

PR #1052, opened 2021-08-23
Prevent non-string names in StorableNamedObject

dwhswenson:
Objects in OPS can be named. These names should always be strings. This PR causes a `TypeError` to be raised if a non-string is assigned as the `name` of a `StorableNamedObject`.

Some context in discussion of https://github.com/openpathsampling/openpathsampling/pull/1050.


sroet (approved):
LGTM!

Issue #1061, opened 2021-08-28
Nightly tests failing on forks


bdice:
I have been getting emails about CI failure every day since I forked this repository. I think I finally figured out that it was because my fork's `stable` branch wasn't up to date, but I'm still not sure (I tried updating the tags on my fork earlier, but that didn't fix it). Regardless of the cause, I would advocate for the nightly CI checks to be skipped on all forks except `openpathsampling/openpathsampling` in the interest of conserving energy (the Python 2.7 tests take about 90 minutes from what I can tell) and reducing notification noise.

Example failure: https://github.com/bdice/openpathsampling/runs/3448877047?check_suite_focus=true

I'm going to open a PR that I hope will address the problem.

bdice:
This also affects releases -- I pushed to the `stable` branch of my fork in order to stop the nightly tests from failing. In the nightly tests, autorelease saw that my `stable` branch pointed to 1.3.0 and was out of date with the current 1.5.2 release. But after pushing to `stable`, that triggered a workflow for releases, which I also did not want. CI logs here: https://github.com/bdice/openpathsampling/actions/runs/1177536916

This is also addressed in PR #1062.

PR #1062, opened 2021-08-28
Limit nightly CI and release/deployment CI to the base repository (not forks).

bdice:
See issue #1061. This prevents forks from running nightly CI on the primary branches, which should only be needed in the base repository.

Maintainers: feel free to push to this branch directly, or make a copy of this branch on the base repo `openpathsampling/openpathsampling` if needed for testing purposes. It is hard to test scheduled CI triggers but I think this should work since I based it on other GitHub workflows.

### DIFF
--- a/.github/workflows/autorelease-deploy.yml
+++ b/.github/workflows/autorelease-deploy.yml
@@ -5,6 +5,7 @@ on:
 
 jobs:
   deploy_pypi:
+    if: ${{ github.repository == 'openpathsampling/openpathsampling' }}
     runs-on: ubuntu-latest
     name: "Deploy to PyPI"
     steps:
--- a/.github/workflows/autorelease-gh-rel.yml
+++ b/.github/workflows/autorelease-gh-rel.yml
@@ -6,6 +6,7 @@ on:
 
 jobs:
   release-gh:
+    if: ${{ github.repository == 'openpathsampling/openpathsampling' }}
     runs-on: ubuntu-latest
     name: "Cut release"
     steps:
--- a/.github/workflows/autorelease-prep.yml
+++ b/.github/workflows/autorelease-prep.yml
@@ -12,6 +12,7 @@ defaults:
 
 jobs:
   deploy_testpypi:
+    if: ${{ github.repository == 'openpathsampling/openpathsampling' }}
     runs-on: ubuntu-latest
     name: "Deployment test"
     steps:
@@ -47,6 +48,7 @@ jobs:
           repository_url: https://test.pypi.org/legacy/
         name: "Deploy to testpypi"
   test_testpypi:
+    if: ${{ github.repository == 'openpathsampling/openpathsampling' }}
     runs-on: ubuntu-latest
     name: "Test deployed"
     needs: deploy_testpypi
--- a/.github/workflows/deploy_docs.yml
+++ b/.github/workflows/deploy_docs.yml
@@ -13,6 +13,7 @@ defaults:
 
 jobs:
   docs:
+    if: ${{ github.repository == 'openpathsampling/openpathsampling' }}
     runs-on: ubuntu-latest
     name: "docs"
     steps:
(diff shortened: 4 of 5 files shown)


dwhswenson:
Sorry that this has been a problem, and thanks for letting us know. I'm not sure why you're getting this (or maybe why others aren't?) @sroet did have similar issues with the OpenMM RC check (see https://github.com/openpathsampling/openpathsampling/pull/1051#issuecomment-903638134 and my response), but doesn't seem to have had this one.

For the specifics of this PR, I want to get @sroet's comments on this before merging.  I think that running the nightly builds in his fork has been part of his workflow; if so, this exact solution won't work. In that case, a fix might be to move the if-statement in the main tests workflow to the step for the Autorelease check. That wouldn't stop the whole workflow from running, but should at least prevent you from getting the error messages.

The changes in the Autorelease workflows should actually be made in Autorelease (those files are generated by the `autorelease vendor actions` command, and not really supposed to be edited locally). I've already made the PR to fix those (https://github.com/dwhswenson/autorelease/pull/100) and will release a new version of Autorelease ASAP.

sroet:
> @sroet did have similar issues with the OpenMM RC check (see #1051 (comment) and my response), but doesn't seem to have had this one.

That is due to the fact that I forked this repo before the `stable` branch existed and never had a reason to push a local version of it.

> For the specifics of this PR, I want to get @sroet's comments on this before merging. I think that running the nightly builds in his fork has been part of his workflow; if so, this exact solution won't work. In that case, a fix might be to move the if-statement in the main tests workflow to the step for the Autorelease check. That wouldn't stop the whole workflow from running, but should at least prevent you from getting the error messages.

I do use nightlies on `master` to be notified of urgent failures. However, as I have since become a maintainer on the main repo I might actually be notified as well if it happens on the main repo, but I have not seen it fail since that change happened (I think)

I also run weekly local builds with cutting edge packages, as this was the most efficient way I found to notice deprecation warnings.

dwhswenson:
> That is due to the fact that I forked this repo before the stable branch existed and never had a reason to push a local version of it.

There are two test failures. One is from pushing to stable, so this is why you haven't seen that one. But the other [is a nightly build on master](https://github.com/bdice/openpathsampling/runs/3448877047?check_suite_focus=true). It comes from autorelease comparing the current version to what can be deduced from the tags. Updating tags in his fork should have fixed it for @bdice, but that's not a burden we should be putting on contributors.

My fork [has no tags](https://github.com/dwhswenson/openpathsampling/tags) (I never updated them); this may be why autorelease doesn't complain for me. @sroet, you [have pre-1.0 tags](https://github.com/sroet/openpathsampling/tags). I'm pretty sure that Autorelease *should* complain there, but maybe there's a bug in the logic so that it complains if you skip patch or minor versions, but not if you skip major *and* minor/patch versions?

> However, as I have since become a maintainer on the main repo I might actually be notified as well if it happens on the main repo, but I have not seen it fail since that change happened (I think)

I think it may be that the notification goes to the person who pushed the commit (merged the PR), which is usually me (except for one I apparently forgot was ready to merge? sorry!) We *have* had a few failing nightlies recently, I believe due to Dropbox being flaky at serving files for the analysis test notebooks. I found an email notification from Aug 27 about such a failure. (I just automatically re-run when I see that; if it doesn't fail again, I don't worry.)

dwhswenson (changes requested):
@bdice: Given that @sroet does make use of running nightlies on his fork, could you try moving the if-statement to the step named "Autorelease check"? I think that will prevent the test failures. It also keeps behavior similar to GitHub defaults. As to ***why*** GitHub defaults to running workflows in forks, I don't know, and I'm [not the only one](https://github.community/t/stop-github-actions-running-on-a-fork/17965) who thinks it was a bad choice. But generally, I try to keep things near the platform defaults so contributors get expected behavior.

bdice:
Closing as stale.

PR #1065, opened 2021-09-09
improvements to unpatch after using SimStore

dwhswenson:
There's an `unpatch` method in the `monkey_patches` to undo the patches that tell OPS that you'll be using SimStore. This generally isn't used by users (who tend to be either fully in SimStore or fully not, at least within a given session), but it is important for unit testing.

This PR adds some `importlib.reload` commands that are required to fully undo the patch.

### DIFF
--- a/openpathsampling/experimental/storage/monkey_patches.py
+++ b/openpathsampling/experimental/storage/monkey_patches.py
@@ -57,6 +57,13 @@ _PREPATCH = {
     ]
 }
 
+_UNPATCH_MODULES = [
+    paths.netcdfplus,
+    paths.collectivevariable,
+    paths.collectivevariables,
+    paths
+]
+
 def monkey_patch_saving(paths):
     paths.netcdfplus.FunctionPseudoAttribute.to_dict = \
             function_pseudo_attribute_to_dict
@@ -93,8 +100,11 @@ def unpatch(paths):
     for cls, old in _PREPATCH.items():
         cls.to_dict = old['to']
         cls.from_dict = old['from']
-    importlib.reload(paths.netcdfplus)
-    importlib.reload(paths.collectivevariable)
-    importlib.reload(paths.collectivevariables)
-    importlib.reload(paths)
+
+    # we loop over the modules twice as quick-and-dirty solution to avoid
+    # figuring the DAG topological order for imports (should actually in
+    # order leaf to root)
+    for module in _UNPATCH_MODULES * 2:
+        importlib.reload(module)
+
     return paths
--- a/openpathsampling/experimental/storage/monkey_patches.py
+++ b/openpathsampling/experimental/storage/monkey_patches.py
@@ -93,4 +93,8 @@ def unpatch(paths):
     for cls, old in _PREPATCH.items():
         cls.to_dict = old['to']
         cls.from_dict = old['from']
+    importlib.reload(paths.netcdfplus)
+    importlib.reload(paths.collectivevariable)
+    importlib.reload(paths.collectivevariables)
+    importlib.reload(paths)
     return paths


openpathsampling/experimental/storage/monkey_patches.py
```diff
@@ -93,4 +93,8 @@ def unpatch(paths):
     for cls, old in _PREPATCH.items():
         cls.to_dict = old['to']
         cls.from_dict = old['from']
+    importlib.reload(paths.netcdfplus)
+    importlib.reload(paths.collectivevariable)
+    importlib.reload(paths.collectivevariables)
+    importlib.reload(paths)
```
sroet, line 99: wouldn't only running the last line be sufficient? ([that is my impression when looking at the description](https://docs.python.org/3/library/importlib.html#importlib.reload))

```suggestion
    importlib.reload(paths)
```

otherwise I would like to see a list of modules that needs to be reloaded

sroet (changes requested):

openpathsampling/experimental/storage/monkey_patches.py
dwhswenson: > wouldn't only running the last line be sufficient?

Nope. Without all 4 lines, the tests in the CLI fail. (This is taken from the CLI tests, which have a few places that require unpatching -- I found myself duplicating some existing testing code for openpathsampling/openpathsampling-cli#43, so it made more sense to move it upstream). Note that `reload` reloads a *module*, not a package. [Here's an SO link aiming for a more general solution](https://stackoverflow.com/questions/15506971/recursive-version-of-reload).

One option would be to use a recursive reload as given on SO, but (1) explicitly doing the needed reloads is less code; and (2) I want to discourage adding more things to the list of modules to reload.

> otherwise I would like to see a list of modules that needs to be reloaded

Not entirely clear on what you mean here -- would it be something like:

```python
# global in monkey_patches
_MODULES_MODIFIED = [
    paths.netcdfplus,
    paths.collectivevariable,
    paths.collectivevariables,
    paths
]
```

and then

```python
# in unpatch
for module in _MODULES_MODIFIED:
    importlib.reload(module)
```

The major concern there is that order can matter, and although a list will preserve order, it tends to de-emphasize the importance of order. This is why `paths` comes last -- reloading `paths.collectivevariable` makes it so that `paths.collectivevariable.MDTrajFunctionCV` is unpatched, but `paths.MDTrajFunctionCV` would not be unpatched. In fact, even switching the order of importing `collectivevariable` (a module) and `collectivevariables` (a package) breaks this, because `collectivevariables` imports from `collectivevariable`, and then `openpathsampling` imports from `collectivevariables`. Essentially, if you make a DAG of dependencies, you need to re-import in DAG topological order (leaf to root). In the SO link, you'll see references about running the function twice: that's a less efficient way to get the same result, but easy to implement for the general case.

Anyway, I'll make that change if you think it is makes the code clearer, but these were the reasons behind my thinking. Let me know!

sroet: I did indeed mean that.

However, I am still confused where in this code `collectivevariable` is monkeypatched (and therfor why it needs unpatching), I can not find any reference to it  in this file before this reload...

dwhswenson: > However, I am still confused where in this code collectivevariable is monkeypatched (and therfor why it needs unpatching), I can not find any reference to it in this file before this reload...

It makes use of `netcdfplus.ObjectJSON.callable_to_dict`, which gets patched.

Probably the cleaner thing is to have the whole `unpatch` function replaced by a recursive reload, which should get us back to the unpatched state. But that's beyond the scope of what I want to do right now (and all of this deals with edge case uses and is temporary code until 2.0, so making it perfect is not a high priority.)

> make a list and reload twice (and please add a comment why we loop over the list twice)

In 3f0244f!

sroet (approved):
LGTM

PR #1066, opened 2021-09-09
Clear old ExternalMDSnapshots during `iter_generate`

dwhswenson:
This clears cached details from snapshots that have a `clear_cache` method (namely, `ExternalMDSnapshot`s) after the partial trajectory has been yielded to back to the `generate` function.

OPS currently requires that the entire trajectory fit in memory; this relaxes that requirement for users of the Gromacs engine. This is in response to a user-reported problem that might have been caused by this.


dwhswenson:
The coverage failure here is because we don't actually install Gromacs for testing on CI. There is a new test added that checks the desired behavior (except the exception -- but the exception is a do-nothing, and we don't currently have an engine that would trigger it.)

Relevant coverage from a local run (where gmx is installed):

```
$ pytest openpathsampling/tests/test_gromacs_engine.py --cov=openpathsampling --cov-report html
```
image

My inclination is to leave as-is, and merge with the slight drop in project coverage.

dwhswenson:
After improved tests in recent commits, we still have basically the coverage as discussed in https://github.com/openpathsampling/openpathsampling/pull/1066#issuecomment-916866183. Location of the missing lines shifted a little in those changes, but the principle is the same

@sroet : Asking for a review on this, despite the patch coverage being low. Most of the missing lines are because we don't test Gromacs integration in CI. Local tests with Gromacs installed show missing coverage only for the `AttributeError` safety check that isn't relevant for any existing engines.

openpathsampling/engines/dynamics_engine.py
```diff
@@ -656,10 +669,21 @@ def iter_generate(self, initial, running=None, direction=+1,
...
         yield trajectory
+        self._clear_snapshot_cache(snapshot)
+
+    def _clear_snapshot_cache(self, snapshot):
+        if self.clear_snapshot_cache and snapshot is not None:
+            try:
+                # isolate the AttributeError to just getting this method
+                clear_cache = snapshot.clear_cache
+            except AttributeError:
+                clear_cache = lambda: None
```
sroet, line 685: This can be solved slightly better (in my opinion)
```suggestion
            clear_cache = getattr(snapshot, 'clear_cache', lambda: None)
```

openpathsampling/engines/external_snapshots/snapshot.py
```diff
@@ -106,10 +106,9 @@ def set_details(self, xyz, velocities, box_vectors):
             pass
         else:
             raise RuntimeError("Can't set details if frame already exists.")
-        finally:
-            self._xyz = xyz
-            self._velocities = velocities
-            self._box_vectors = box_vectors
+        self._xyz = xyz
+        self._velocities = velocities
+        self._box_vectors = box_vectors
```
sroet, line 111: This is probably better to be put in the `except` clause (cant suggest becuase this involves unchanged lines) (and the except should probably catch a *very* specific error

openpathsampling/engines/dynamics_engine.py
```diff
@@ -377,16 +378,24 @@ def stop_conditions(self, trajectory, continue_conditions=None,
...
+            # Current thought: No short circuit means that CVs may be
+            # calculated as part of stop conditions here, so keep as is.
+            # (NB: ensembles/volumes use short-circuit logic, so there isn't
+            # really a promise that CVs get calculated.)
+            # if stop:
+            #     return True
 
         return stop
 
+
```
sroet, line 398: this triggers my linter
```suggestion
```

sroet (changes requested):
I think there is a bug in the stop logic, please update and add a test. + some code formatting issues

openpathsampling/engines/dynamics_engine.py
dwhswenson: Yup. And that also fixes the uncovered line in the local tests 😉

openpathsampling/engines/dynamics_engine.py
```diff
@@ -377,16 +378,24 @@ def stop_conditions(self, trajectory, continue_conditions=None,
...
-                for condition in continue_conditions:
-                    can_continue = condition(trajectory, trusted)
-                    stop = stop or not can_continue
-            else:
-                stop = not continue_conditions(trajectory, trusted)
+        if callable(continue_conditions):
+            continue_conditions = [continue_conditions]
+
+        for condition in continue_conditions:
+            stop = not condition(trajectory, trusted)
```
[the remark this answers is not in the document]
dwhswenson: Problem confirmed and test added in 301fc1d

openpathsampling/engines/external_snapshots/snapshot.py
dwhswenson: Moved this into the `except` clause in ebb21e9. I didn't specify the exception because I'm not entirely sure that we can predict what the exception should be (depends on the implementation of `engine.read_frame_data`).

dwhswenson:
@sroet This should be ready for another round of review. Again, patch coverage issues are because we don't test Gromacs integration in CI, but tests are in place if you have gmx installed locally.

sroet (changes requested):
1 redundant call left, LGTM otherwise

sroet (approved):
LGTM

PR #1068, opened 2021-09-10
New analysis testing utilities

dwhswenson:
Writing tests for analysis code is hard. I always tell people that the input to most analysis methods should be a list of steps. But how do you mock up a sequence of steps?

This PR adds some utilities in `tests/analysis/utils/` to help create mock simulation results. These utilities have their own test in that same directory.

This was developed because the headache of writing such tests has been blocking both #1026 and #1048 for months. I decided to make a separate task out of creating tools to solve that generic problem.

The approach used here is to use as much of the internal OPS machinery as possible, instead of just reproducing the data it generates (as was done in the [OPSPiggybacker](https://github.com/dwhswenson/OPSPiggybacker/)). Here, we make use of `unittest.patch` to inject the data that OPS would have created via simulation. This means that if internal structure of movers (or of `Details` they return) changes, this code should automatically use that new behavior.

The scope of this PR is only aiming to support one-way shooting moves, path reversal moves, and repex moves, as well as wrapping as with an `OrganizeByMoveGroup` global strategy. Support for other movers will be added as needed in the future.


dwhswenson:
This is ready for review and comment. I will leave it open for at least 48 hours, merging no earlier than Tue 14 Sep 20:00 GMT (16:00 my local).

Coverage doesn't show, since the `utils` code in inside the `tests` directory. However, local tests showed coverage for everything except the `ImportError` that preserves Py27 compatibility.

For what it's worth, I'm already playing with this in a branch where I've mixed this stuff and #1026. Writing tests for #1026 is ***so*** easy now.

openpathsampling/tests/analysis/utils.py
```diff
@@ -0,0 +1,500 @@
...
+import random
+try:
+    from unittest.mock import Mock, patch
+except ImportError:
+    from mock import Mock, patch
+
+from openpathsampling.tests.test_helpers import make_1d_traj
+import openpathsampling as paths
+
+def _get_only(iterable, condition, error_msg):
```
sroet, line 27: pep8: 2 blank expected, found 1
```suggestion


def _get_only(iterable, condition, error_msg):
```

openpathsampling/tests/analysis/utils.py
```diff
@@ -0,0 +1,500 @@
...
+    possibilities = [item for item in iterable if condition(item)]
+    if len(possibilities) != 1:
+        msg = "expected 1 {error_msg}; found {nfound}: found".format(
+            error_msg=error_msg,
+            nfound=len(possibilities),
+            found=possibilities
+        )
```
sroet, line 34: pep8: `found` from format not used in str
```suggestion
        msg = "expected 1 {error_msg}; found {nfound}: {found}".format(
            error_msg=error_msg,
            nfound=len(possibilities),
            found=possibilities
        )
```

openpathsampling/tests/analysis/utils.py
```diff
@@ -0,0 +1,500 @@
...
+
+
+class AnalysisTestSetupError(Exception):
+    """Raised for when an internal error occurs during test setup.
+
+    These usually indicate a problem with test suite, not with the code
+    itself.
+    """
+
+def make_trajectory(lower, upper):
```
sroet, line 46: expected 2 blank lines
```suggestion


def make_trajectory(lower, upper):
```

openpathsampling/tests/analysis/utils.py
```diff
@@ -0,0 +1,500 @@
...
+def make_tis_trajectory(cv_max, lower_bound=-0.1):
+    """Make a TIS trajectory with a given maximum x value.
+
+    """
+    increasing = make_trajectory(lower_bound, cv_max)
+    if cv_max >= 1.0:
+        return increasing
+    else:
+        decreasing = increasing.reversed[1:]
+        return increasing + decreasing
```
sroet, line 80: The cutoff between switching from increasing to both should probably be user configurable, also 2 blank lines expected and only 1 found
```suggestion


def make_tis_trajectory(cv_max, lower_bound=-0.1, final_state=1.0):
    """Make a TIS trajectory with a given maximum x value.

    """
    increasing = make_trajectory(lower_bound, cv_max)
    if cv_max >= final_state:
        return increasing
    else:
        decreasing = increasing.reversed[1:]
        return increasing + decreasing
```

openpathsampling/tests/analysis/utils.py
```diff
@@ -0,0 +1,500 @@
...
+
+def _select_by_input_ensembles(movers, ensembles):
```
sroet, line 82: 2 blank lines expected
```suggestion


def _select_by_input_ensembles(movers, ensembles):
```

openpathsampling/tests/analysis/utils.py
```diff
@@ -0,0 +1,500 @@
...
+
+        # make a move change for the inner step (selecting which mover
+        # within the move type)
+        inner_mock = MockRandomChoiceMover(group_selector, change)
+        inner_change = inner_mock(inputs)
+
+        # make a move change for the outer step (selecting which move type
+        # to do from the root_mover)
+        outer_mock = MockRandomChoiceMover(root_mover, inner_change)
+        outer_change=  outer_mock(inputs)
```
sroet, line 183: whitespace complaint
```suggestion
        outer_change = outer_mock(inputs)
```

openpathsampling/tests/analysis/utils.py
```diff
@@ -0,0 +1,500 @@
...
+                )
+            else:
+                raise AnalysisTestSetupError(
+                    "Step tried to force acceptance on a trial that can "
+                    "not be accepted."
+                )
+
+
+
+    def safety_checks(self, change, inputs):
```
sroet, line 335: to many blank lines
```suggestion

    def safety_checks(self, change, inputs):
```

openpathsampling/tests/analysis/utils.py
```diff
@@ -0,0 +1,500 @@
...
+                                               group_name=group_name)
+
+
+def _do_single_step(init_conds, move, org_by_group):
+    change = move(init_conds)
+    if org_by_group:
+        change = move.wrap_org_by_group(change, init_conds)
+    return change
+
+def run_moves(init_conds, moves, org_by_group=True):
```
sroet, line 465: expected 2 blank lines
```suggestion


def run_moves(init_conds, moves, org_by_group=True):
```

sroet (changes requested):
Looks great,

2 critical points:
- I am not convinced "make_trajectory" does what it is supposed to, 
- I would like to have access to the `final_state` in the `make_tis_trajectory`

Also pointed out the issues my development environment was complaining about, feel free to ignore those

dwhswenson:
Sorry that this has been on pause for so long. I'll go ahead and take the PEP8 improvements. However, some of the other comments from review made it clear to me that my code was not communicating its purpose well enough, because that purpose was not obvious to @sroet. I've been trying to figure out what the best way to make the purpose more clear would be -- but rather than try to predict was others will think, it's better to just ask!

Fundamental ideas that I don't think have been communicated clearly enough here are mainly about the `make_trajectory` method:

1. This is a specific set of fixtures intended for analysis, and they are intended to be used together. Specifically, the `make_trajectory` method here is *only* intended to be used with the specific network fixture created here. Some of @sroet's concerns deal with the issue that someone might use `make_trajectory` for a different network.
2. The trajectories created are designed to fit a style I've tended to use, which discretizes at intervals of 0.1. A number of @sroet's concerns deal with my implicit assumption that the inputs are at intervals of 0.1.

Here are a few options I've considered on how to make these things clearer to others reading the code. If you have options, please suggest!

Regarding point 1:

* document it in docstrings
* convert these separate but related fixtures to a single instance of some class, where, e.g., `network` and `scheme` might be properties, and `make_trajectory` is a method specific to the fixture

Regarding point 2:

* docstrings
* convert `make_trajectory` into something that takes integers in the range [-1, 10], which makes the discretization more explicit. This might also involve changing the network itself to actually use the state borders at 0.0, 10.0 instead of 0.0, 1.0.

These are the things I thought of; again, I'm very open to other options. The code in this PR is very much intended for re-use by contributors (please let me make it easier for you to write tests!), so it is extremely important that it is easily understood.

I'm somewhat leaning toward the more complicated solution in both cases (especially point 1, where it enables me to encapsulate the `make_trajectory` method with the fixture)

openpathsampling/tests/analysis/conftest.py
```diff
@@ -0,0 +1,21 @@
+import pytest
+import openpathsampling as paths
+
+@pytest.fixture
+def tis_network():
+    paths.InterfaceSet._reset()
+    cv = paths.FunctionCV("x", lambda s: s.xyz[0][0])
```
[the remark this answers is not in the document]
dwhswenson: The CVs used here are not compatible with SimStore, so this should not be used if the intent is to also test SimStore storage.

Of course, that will change in 2.0 when `paths.FunctionCV` *is* `paths.experimental.storage.CollectiveVariable`.

dwhswenson:
I ended up doing some pretty significant changes since the last review, so this is probably worth a complete re-review, as opposed to just reviewing changes since last review. I re-organized according to the discussion in https://github.com/openpathsampling/openpathsampling/pull/1068#issuecomment-962610214 and following.

Some of the things that are currently in `tests/analysis/utils` might be moved to `tests/utils` in the future, because I think they might be useful in testing more broadly (using a consistent network for 2-state TPS and 2-state TIS setups instead of custom setups for every test class).

Here's what's now included in this PR:	

* `tests/analysis/utils/fixture_classes.py`: This includes the core fixture classes and tools to create fixtures within a `conftest.py`. These now group the `make_trajectory`/`make_tis_trajectory` in an object with the `network` and `scheme`, so it should be clear that these are all related.
* `tests/analysis/utils/test_fixture_classes.py`: Tests for the stuff in `fixture_classes`. Note that this actually tests by way of the specific setups used in `conftest.py`, so technically is also testing that.
* `tests/analysis/utils/mock_movers.py`: This is the core of mocking out pathmovers for analysis. Code here is largely unchanged from the previous, but moved to a separate and more clearly-named module.
* `tests/analysis/utils/test_mock_movers.py`: Tests for above. This is largely unchanged from previous, except in that the range (-0.1, 1.0) was changed to (-1, 10) and trajectory inputs are now all integers.
* `tests/analysis/conftest.py`: Fixtures that will be the most common setups.

openpathsampling/tests/analysis/conftest.py
```diff
@@ -0,0 +1,51 @@
+# TODO: Currently this is located in tests/analysis/conftest.py. However,
+# this might be more suitable for a higher-level conftest.py, or perhaps it
+# should be moved into `fixtures` subdirectory of tests with the necessary
+# objects imported into the main tests/conftest.py, for use across the test
+# suite.
+from functools import partial
+
+import pytest
```
sroet, line 8: my linter says these are not used here
```suggestion
```

openpathsampling/tests/analysis/utils/mock_movers.py
```diff
@@ -0,0 +1,464 @@
...
+            nfound=len(possibilities),
+            found=possibilities
+        )
+        raise AnalysisTestSetupError(msg)
+    return possibilities[0]
+
+
+class AnalysisTestSetupError(Exception):
+    """Raised for when an internal error occurs during test setup.
+
```
sroet, line 40: hard to read grammar
```suggestion
    """Raised when an internal error occurs during test setup.

```

openpathsampling/tests/analysis/utils/mock_movers.py
```diff
@@ -0,0 +1,464 @@
...
+    try:
+        signature = tuple(ensembles)
+    except TypeError:
+        signature = tuple([ensembles])
+
+    mover = _get_only(
+        iterable=movers,
+        condition=lambda m: set(m.ensemble_signature[0]) == set(signature),
+        error_msg="mover matching signature{sig}".format(sig=signature)
```
sroet, line 63: maybe a space here? (feel free to ignore)
```suggestion
        error_msg="mover matching signature {sig}".format(sig=signature)
```

sroet (approved):
Some extra imports, and some possible corrections on the docs. LGTM otherwise

sroet (approved):
LGTM, feel free to merge on green

PR #1071, opened 2021-09-19
Switch CI install to mamba

dwhswenson:
Py27 builds have taken too long for a while, and now they're so bad that they're failing CI. (20+ minutes was bad, but then it jumped to 2 hours -- and now it takes more than 6 hours?!?!?!) I don't know why this is getting worse, but this is insane.

In this PR, I'm going to try to switch to [mamba](https://github.com/mamba-org/mamba). I've personally been using mamba for a while now. I tested locally and could build a Py27 environment for OPS in a few minutes, using the same command that hangs in CI. (The conda version of the same command is *still* trying to run -- the problem isn't with GitHub actions, since it affects me, too.)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -57,6 +57,7 @@ jobs:
         run: |
           if [ -z "$MINIMAL" ] ; then
             source devtools/conda_install_reqs.sh
+            python -m pip install nose pytest pytest-cov coveralls nbval
           else
             python -m pip install -r devtools/minimal.txt \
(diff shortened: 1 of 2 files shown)


dwhswenson:
Okay, "Install requirements" step for 2.7 came in at 1m2s (and around 3min for the Py3 builds). I'd been thinking about doing this since shortly after I personally switched to mamba... WTF took me so long?!

However, this seems to be giving a segfault at the end of the unit testing. That looks like an issue with Pytest, since the tests complete correctly, then it seems to hang and then die. I'll investigate.

PR #1072, opened 2021-09-21
Switch CI install to mamba

dwhswenson:
Second try at #1071. From that PR:

> Py27 builds have taken too long for a while, and now they're so bad that they're failing CI. (20+ minutes was bad, but then it jumped to 2 hours -- and now it takes more than 6 hours?!?!?!) I don't know why this is getting worse, but this is insane.
> 
> In this PR, I'm going to try to switch to [mamba](https://github.com/mamba-org/mamba). I've personally been using mamba for a while now. I tested locally and could build a Py27 environment for OPS in a few minutes, using the same command that hangs in CI. (The conda version of the same command is _still_ trying to run -- the problem isn't with GitHub actions, since it affects me, too.)

Then:

> Okay, "Install requirements" step for 2.7 came in at 1m2s (and around 3min for the Py3 builds). I'd been thinking about doing this since shortly after I personally switched to mamba... WTF took me so long?!
>
> However, this seems to be giving a segfault at the end of the unit testing. That looks like an issue with Pytest, since the tests complete correctly, then it seems to hang and then die. I'll investigate.

This PR is going to try to track down why pytest is giving a core dump.


dwhswenson:
Okay... the problem is netcdf4. If this is installed by `mamba`, tests fail. If it is installed by `pip`, tests pass.

dwhswenson:
Update after helpful discussion in mamba-org/mamba#1175 : If I restructure our build to ensure that everything is coming from `conda-forge` (nothing from the `defaults`/`main` channel), then I get the same error in `conda`:

| Commit | Pass | Install reqs | Install testing |  Install `netcdfplus` |
|-------|-----|----------|------------|------|
| efcb126| ❌  | `conda` | `pip` | `conda` |
| ddefb77 | ❌ | `mamba` | `pip` | `mamba` |
| 02c56e4 | ❌ | `mamba` | `mamba` | `mamba` |
| fad30aa | ✅ | `mamba` | `mamba` | `pip` |

.github/workflows/tests.yml
```diff
@@ -69,6 +70,8 @@ jobs:
         run: conda list
       - name: "Autorelease check"
         run: python devtools/autorelease_check.py
+      #- name: "DEBUG: enable SSH login"
+        #uses: mxschmitt/action-tmate@v3
```
sroet, line 74: leftover debug code?

please remove or describe when this is needed
```suggestion

```

dwhswenson:
Okay... it's fixed! And now the 2.7 build is done running *before* the full builds of 3.x, instead of waiting an extra 2 hours.

The segfault problem appears to have been that, in the reactive flux analysis tests, storage was removed without closing it (or letting it go out of scope, which would implicitly close it). I still don't know why this is only a problem when we install from `conda-forge` (possibly still a bug report to file somewhere), or why it technically required two tests to run in order to show up, but we're passing now.

To get the kind of commit history that will actually be useful for this, I'm going to close this PR, cherry pick a few commits into a new PR, and then add other commits that reproduce anything else worth keeping in here. I don't want to squash multiple ideas into 1 commit, and I don't want to merge 30 debugging commits.

BTW, new tool I learned about during this: [mxschmitt/action-tmate](https://github.com/marketplace/actions/debugging-with-tmate). Makes it so you can SSH into your runner and work inside the CI running interactively to debug. ***So*** useful when isolating the test that was causing the error (since I couldn't reproduce the error locally).

devtools/conda_install_reqs.sh
```diff
@@ -9,12 +9,23 @@
...
+fi
+
+INSTALL_CMD="$EXE install -y -q -c conda-forge -c omnia --override-channels"
+
+# TODO: is this preinstall needed? We're certainly no longer using pyyaml,
+# and it looks like we should be, but aren't, using future in setup_cfg_reqs
 if [ ! -z "$OPS_ENV" ]
 then
-    conda create -q -y --name $OPS_ENV conda future pyyaml python=$CONDA_PY
+    $INSTALL_CMD --name $OPS_ENV conda future pyyaml python=$CONDA_PY
```
sroet, line 25: does conda needs to be installed here?
```suggestion
    $INSTALL_CMD --name $OPS_ENV future pyyaml python=$CONDA_PY
```

sroet (commented):
This seems to pass and pretty quick. I would like this PR to be squashed down (as the final changes are quite minor).

Only minor comments:
I had one question about debug code, and about installing `conda`. 
feel free to ignore all of this and merge when ready if it is too much of a hassle

otherwise LGTM

PR #1073, opened 2021-09-22
Switch CI install to mamba

dwhswenson:
Third (and, I hope, final!) version of this. See also #1072 and #1071 for the history of trying to debug this problem.

Main points, quoting from previous:

> Py27 builds have taken too long for a while, and now they're so bad that they're failing CI. (20+ minutes was bad, but then it jumped to 2 hours -- and now it takes more than 6 hours?!?!?!) I don't know why this is getting worse, but this is insane.
> 
> In this PR, I'm going to try to switch to [mamba](https://github.com/mamba-org/mamba). I've personally been using mamba for a while now. I tested locally and could build a Py27 environment for OPS in a few minutes, using the same command that hangs in CI. (The conda version of the same command is _still_ trying to run -- the problem isn't with GitHub actions, since it affects me, too.)

However, that change resulted in a segfault that appears to have been because a test didn't close a storage before removing the file (which is fixed in this PR).

In addition to fixing that segfault and switching to `mamba` for installation, this PR:

* Adds a commented debugging step (allowing one to SSH into the worker) to the main CI workflow.
* Improves the requirements installation in several ways, including giving a way to override/replace conda installs with pip installs, and using a more consistent (and purely conda-forge) installation procedure.

### DIFF
--- a/devtools/conda_install_reqs.sh
+++ b/devtools/conda_install_reqs.sh
@@ -9,19 +9,23 @@
 
 DEVTOOLS_DIR=`dirname "${BASH_SOURCE[0]}"`
 
-if [ ! command -v mamba ]
+if ! command -v mamba &> /dev/null
 then
     EXE="conda"
 else
     EXE="mamba"
 fi
 
+INSTALL_CMD="$EXE install -y -q -c conda-forge -c omnia --override-channels"
+
+# TODO: is this preinstall needed? We're certainly no longer using pyyaml,
+# and it looks like we should be, but aren't, using future in setup_cfg_reqs
 if [ ! -z "$OPS_ENV" ]
 then
-    $EXE create -q -y --name $OPS_ENV conda future pyyaml python=$CONDA_PY
+    $INSTALL_CMD --name $OPS_ENV conda future pyyaml python=$CONDA_PY
     source activate $OPS_ENV
 else
-    $EXE install -y -q future pyyaml  # ensure that these are available
+    $INSTALL_CMD future pyyaml  # ensure that these are available
 fi
 
 # for some reason, these approaches to pinning don't always work (but conda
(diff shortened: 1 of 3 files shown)


devtools/conda_install_reqs.sh
```diff
@@ -9,12 +9,23 @@
...
+fi
+
+INSTALL_CMD="$EXE install -y -q -c conda-forge -c omnia --override-channels"
+
+# TODO: is this preinstall needed? We're certainly no longer using pyyaml,
+# and it looks like we should be, but aren't, using future in setup_cfg_reqs
 if [ ! -z "$OPS_ENV" ]
 then
-    conda create -q -y --name $OPS_ENV conda future pyyaml python=$CONDA_PY
+    $INSTALL_CMD --name $OPS_ENV conda future pyyaml python=$CONDA_PY
```
sroet, line 25: only leftover comment from the other PR:

"Is the conda install here necessary?
```suggestion
    $INSTALL_CMD --name $OPS_ENV future pyyaml python=$CONDA_PY
```

sroet (approved):
LGTM, one leftover question. Feel free to merge without ;)

dwhswenson:
I restructured the part in question a little more thoroughly than just the recommended change. That `if` statement was from a very old way of generating our list of requirements; we no longer need `pyyaml` (nor `future`, since even in the Py2 tests, this script is only run from Py3 -- the Py2 environment would be created *by* this script.)

The script *is* used by the `conda_ops_dev_install` script, though, so a few pieces of functionality need to be left in (like creating a new environment).

@sroet : If you have a chance to take another look, please do. If not, I'm pretty sure the changes are fine, and I'll aim to merge this before the next overnight scheduled tests run.

sroet (approved):
LGTM

dwhswenson:
Whoops. forgot to merge this before I went to bed last night!

PR #1075, opened 2021-10-05
Account for modifier acceptance in engine movers

### LINKED ISSUE #794: snapshot modification: where is it taken into account for the acceptance?
I tried to find my way through the snapshot modifiers, in particular the RandomVelocities. If I am not completely mistaken in general the acceptance criterion for modified shooting snapshots should include the ratio of the stationary distributions at the snapshots. See for example the "Trajectory based rare event methods" review by Bolhuis and Dellago Equations 77 and 93/94. I am aware that the configurational part of the densities cancel when only drawing new momenta, but the kinetic part of the density will not in general. So I would expect that a \rho(x_{SP}) / \rho(x`_{SP}) is multiplied with the sample.bias somewhere. 
Where is that happening? (Or why is it not needed?)
I looked in the PathMover code and as far as I can tell all EngineMovers only take into account the bias of the selector.
Would be great if someone can clarify. (If it is indeed needed and missing I will try to fix it)

sroet:
Closes #794 for all OPS classes
However, a new issue should be opened to discuss a refactor of `EngineMover` to take this and the other bias options [described in this comment](https://github.com/openpathsampling/openpathsampling/issues/794#issuecomment-923830025) into account for OPS 2.0.

Custom implementations of both `EngineMover` and `SnapshotModifier` should tyake care to implement this correctly on their end.

Big assumption (that might break API): **A key that starts with "bias" should not be present in `run_details` if it is not intended to be used for the bias calculation of a sample** 

todo:

- [x] add functionality
- [x]  add tests
- [x]  update docstrings
- [x]  make touched files pep8 compliant
- [x]  remove nose from touched tests (partial #756)

commit descriptions:
4f017dd adds probability_ratio to snapshot modifiers
ad7fad0 allows the usage of this probability ratio to be used as a bias in EngineMovers
0e42bb9 reverts an oversight in my thought pattern about the acceptance bias of velocities that are directly sampled from the correct boltzman distribution
ad33488 and b351417 add tests for the added functionality
abf0d99 alters the docstrings that were already present
276e3b3 fixes pep8 complaints from my env
f591ca7 drops nose from the tests
8a6057e Fixes an py2.7 issue in the tests where you can't initiate a class without overwriting *all* abstract methods

[Edit]: added a link to the mentioned comment

### DIFF
--- a/openpathsampling/tests/test_shooting.py
+++ b/openpathsampling/tests/test_shooting.py
@@ -232,3 +232,17 @@ class TestConstrainedSelector(SelectorTest):
         mod_traj += mytraj[idx+1:]
         prob = self.sel.probability_ratio(frame, mytraj, mod_traj, mod_frame)
         assert prob == expected
+
+
+class TestDeprecations(object):
+    @pytest.mark.parametrize('selector_class', [ShootingPointSelector,
+                                                FirstFrameSelector,
+                                                FinalFrameSelector])
+    def test_new_snapshot(self, selector_class):
+        selector = selector_class()
+        mytraj = make_1d_traj(coordinates=[-0.5, -0.4, -0.3, -0.1,
+                                           0.1, 0.2, 0.3, 0.5])
+        with pytest.deprecated_call(match='new_snapshot'):
+            _ = selector.probability_ratio(mytraj[1], mytraj, mytraj)
+        # Reset the warning for the next class
+        NEW_SNAPSHOT_SELECTOR.has_warned = False
--- a/openpathsampling/tests/test_spring_shooting.py
+++ b/openpathsampling/tests/test_spring_shooting.py
@@ -109,10 +109,14 @@ class TestSpringShootingSelector(SelectorTest):
         sel = SpringShootingSelector(delta_max=1,
                                      k_spring=1)
         sel._acceptable_snapshot = True
-        ratio = sel.probability_ratio(None, None, None)
+        # TODO OPS 2.0:  Last value here is not None to silence deprecation
+        # warnings, should be set to None after the completion
+        ratio = sel.probability_ratio(None, None, None, 1)
         assert ratio == 1.0
         sel._acceptable_snapshot = False
-        ratio = sel.probability_ratio(None, None, None)
+        # TODO OPS 2.0:  Last value here is not None to silence deprecation
+        # warnings, should be set to None after the completion
+        ratio = sel.probability_ratio(None, None, None, 1)
         assert ratio == 0.0
 
     def test_sanity_breaking_fw_pick(self):
--- a/openpathsampling/tests/test_spring_shooting.py
+++ b/openpathsampling/tests/test_spring_shooting.py
@@ -297,7 +297,7 @@ class TestSpringShootingSelector(SelectorTest):
         sel.restart_from_step(step)
         assert sel.trial_snapshot == 10
 
-    def test_deprectation(self):
+    def test_deprecation(self):
         sel = self.default_selector
         # Override to mimick a good pick
         sel._acceptable_snapshot = True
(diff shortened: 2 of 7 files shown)


sroet:
@dwhswenson This is ready for a review, all tests pass. 

Codecov issues are in code that was only altered to deal with pep8 (276e3b3). Do you have any tips on what kinds of tests should be added for that, or should I just revert the changes?

I would advice to have a look at the diff between abf0d99 and master and then the changes of 276e3b3 (should only be style, no altered code/functionality) and f591ca (only changes the tests)



Also, we should add an API 2.0 issue to at least align the init call of `AbstractTwoWayShootingMover` with `EngineMover` but preferably also refactor EngineMover to introduce the [proposed stages](https://github.com/openpathsampling/openpathsampling/issues/794#issuecomment-923830025) with their associated bias calculations

dwhswenson (commented):
Before I go too much into detail, the main question is whether you're intending to follow this up with a the refactor for 2.0.

* **Implementation**: There are a couple of things I'd suggest changing in the implementation here, mainly to make a more complete refactor more straightforward. If you're going to have a PR for 2.0 coming soon, then we can deal most of those issues there (here, they'd require adding more compatibility code); if not, I'd rather make it easier to refactor later by getting the structure closer here. Specifically, issues are:

  1. Rather than taking the `bias*` namespace in details, I think it would make sense to directly call each of the `probability_ratio` methods in the same place. It's a little odd in the current setup that we need to look different places for the code that calculates different parts of the bias.
  2. If we're adding an implicit shooting point modifier for all engine movers, then I think should actually use it (even if it is often no modification, i.e., `new is old`), which would involve moving modification into the `EngineMover.__call__` method (just like shooting point selection).

  I'll go into more detail if needed, but those are the main issues.

* **Coverage**: It looks like most of this is related to stuff from `pathmover_inout.py`. Don't worry about it -- we can merge with failing patch coverage in this case. I'm personally in favor of completely removing `pathmover_inout.py` in OPS 2.0. It's untested and generally problematic. See discussion in [#557](https://github.com/openpathsampling/openpathsampling/pull/557).

dwhswenson:
More thoughts coming later, but here's a really quick one:

> even with NoModification I am pretty sure `new == old` but not `new is old` (only a copy is returned).

Looking at the code, you're correct. In usage here, I think we want the default modifier (for one-way shooting, e.g.) to return `new` such that `new is old`. My suggestion (pretty easy change to the current code):

```python
class NoModification(SnapshotModifier):
    def __init__(self, as_copy=True):
        # masking change to only certain atoms is nonsense for no modification
        super(NoModification, self).__init__(subset_mask=None)
        self.as_copy = as_copy

    def __call__(self, snapshot):
        return snapshot.copy() if self.as_copy else snapshot
```

Then use `NoModification(as_copy=False)` as the default in the changes here.

openpathsampling/tests/test_spring_shooting.py
```diff
@@ -292,6 +297,15 @@ def test_correct_loading():
         sel.restart_from_step(step)
         assert sel.trial_snapshot == 10
 
+    def test_deprectation(self):
```
dwhswenson, line 300:
```suggestion
    def test_deprecation(self):
```

(spelling)

(And why test the deprecation instead of making to not issue the deprecation?)

openpathsampling/tests/test_shooting.py
```diff
@@ -232,3 +232,17 @@ def test_probability_ratio_modified_coordinates(self, new_coord, expected):
         mod_traj += mytraj[idx+1:]
         prob = self.sel.probability_ratio(frame, mytraj, mod_traj, mod_frame)
         assert prob == expected
+
+
+class TestDeprecations(object):
+    @pytest.mark.parametrize('selector_class', [ShootingPointSelector,
+                                                FirstFrameSelector,
+                                                FinalFrameSelector])
```
dwhswenson, line 240: Any reason not to just update these now (and make a quick subclass/instance of the base class if it isn't abstract) to test the deprecation call?

dwhswenson (changes requested):
Pretty good. One typo and questions/suggestions about updating classes within our code rather than leaving them warning of deprecations.

openpathsampling/tests/test_shooting.py
dwhswenson: Got it. Thanks for the clarification!

PR #1079, opened 2021-10-06
add no_copy option to NoModifier and Deprecate 'subset_mask'

sroet:
This is a cherry picked version of https://github.com/openpathsampling/openpathsampling/pull/1075/commits/79c961ee94b03dbbd05cabc20c52d95a798e1682 from #1075 in order to seperate out the comments

Copies of relevant comments:
> @dwhswenson wrote:
> More thoughts coming later, but here's a really quick one:
>
>    even with NoModification I am pretty sure new == old but not new is old (only a copy is returned).
>
>Looking at the code, you're correct. In usage here, I think we want the default modifier (for one-way shooting, e.g.) to return `new`  such that new is old. My suggestion (pretty easy change to the current code):
>```
>class NoModification(SnapshotModifier):
>   def __init__(self, as_copy=True):
>     # masking change to only certain atoms is nonsense for no modification
>       super(NoModification, self).__init__(subset_mask=None)
>       self.as_copy = as_copy
>
>    def __call__(self, snapshot):
>        return snapshot.copy() if self.as_copy else snapshot
>```
> Then use NoModification(as_copy=False) as the default in the changes here.

> @sroet wrote:
> That was way more nasty than expected, as we actually test for subset_mask on NoModification (as a proxy for subclassing SnapshotModifier) Also, apparently old.copy() != old for toy snapshots, I did not really investigate this, but I am afraid we are losing precision on the copy when seeing all the assert_almost_equal in the tests for this modifier...

> @dwhswenson wrote:
> > I can cherry pick over that commit into a separate PR if that is preferred
> 
> Yeah, if it's going to be a headache, let's move it into a separate PR to keep discussion a little cleaner.
> 
> > That was way more nasty than expected, as we actually test for subset_mask on NoModification (as a proxy for subclassing `SnapshotModifier`)
> 
> That part is easily fixed by subclassing in tests. I was lazy before; sorry! man_shrugging
> 
> > Also, apparently `old.copy() != old` for toy snapshots, I did not really investigate this, but I am afraid we are losing precision on the copy when seeing all the `assert_almost_equal` in the tests for this modifier...
> 
> I'm not surprised by this. There's enough bouncing between C and Python representations that I wouldn't be surprised if these are only `almost_equal`.

### DIFF
--- a/openpathsampling/tests/test_snapshot_modifier.py
+++ b/openpathsampling/tests/test_snapshot_modifier.py
@@ -70,7 +70,12 @@ class TestNoModification(object):
                                                      [2.5, 2.6, 2.7]]))
 
     def test_apply_to_subset(self):
-        mod = NoModification(subset_mask=[1,2])
+        # TODO OPS 2.0: this test should be testing SnapshotModifier instead of
+        # NoModification, but python 2.7 does not allow for the initialisation
+        # without overwriting the abstract __call__ for now this raises a
+        # DeprecationWarning
+        mod = NoModification(subset_mask=[1, 2])
+
         copy_1Dx = self.snapshot_1D.coordinates.copy()
         new_1Dx = mod.apply_to_subset(copy_1Dx, np.array([-1.0, -2.0]))
         assert_array_almost_equal(new_1Dx, np.array([0.0, -1.0, -2.0, 3.0]))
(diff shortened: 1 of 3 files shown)


sroet:
> I think that was mislabeled as tested functionality. Since `subset_mask` has no functionality in `NoModification`, it can't actually do anything there. It was just me being lazy and not implementing a subclass within the tests, and not intended functionality. If anyone is affected, I'll just point to the old adage that ["Every change breaks someone's workflow."](https://xkcd.com/1172/)

Perfect, just wanted to make sure that is what you intended. Implemented it, please have another look

openpathsampling/tests/test_snapshot_modifier.py
```diff
@@ -106,10 +118,20 @@ def test_call(self):
...
-        assert_true(self.snapshot_1D.velocities is not new_1D.velocities)
-        assert_true(self.snapshot_3D.coordinates is not new_3D.coordinates)
-        assert_true(self.snapshot_3D.velocities is not new_3D.velocities)
+        assert self.snapshot_1D.coordinates is not new_1D.coordinates
+        assert self.snapshot_1D.velocities is not new_1D.velocities
+        assert self.snapshot_3D.coordinates is not new_3D.coordinates
+        assert self.snapshot_3D.velocities is not new_3D.velocities
+        # TODO OPS 2.0: the following tests should probabily work
+        # assert new_1D == self.snapshot_1D
+        # assert new_3D == self.snapshot_3D
```
dwhswenson, line 127: Actually, those probably won't work if we use UUID for snapshot hash/equality as under discussion in #1080. I think we need helper functions (probably only in testing) to check (near) equality of snapshots. In practice, the only reason I can think of to make a copy of snapshot is for testing CV behavior (new UUID means that caching results by UUID doesn't work; forces you to recalculate instead of using cached values.)

dwhswenson (changes requested):
Trivial change: One comment to either remove or at least take into consideration with regards to discussion in #1080.

openpathsampling/tests/test_snapshot_modifier.py
sroet: removed them, I do agree that this might not be wanted behavior

dwhswenson (approved):
LGTM

PR #1086, opened 2021-10-29
SimStore: Fix issue with internalized snapshot box vectors

dwhswenson:
This fixes a bug when using `engines.features.box_vectors` in SimStore. Previously, these were not saved.

As far as I can tell, this bug only occurs if you are using the internalized Gromacs snapshots, introduced in #933, and storing them with the new SimStore storage system. (However, it could be an issue for engines that aren't included in core OPS.) The OpenMM engine uses a box vectors feature in `engines.features.statics` and the Gromacs engine uses a box vectors feature in `engines.external_snapshots.features.box_vectors`, so neither of these are affected. I only found this bug when I was going to suggest using SimStore + internalized snapshots as part of a [user question](https://mattermodeling.stackexchange.com/questions/6967/) that came up today. It seems that I overlooked this feature when adding SimStore storage support.

New tests fail without the (simple) fix in the feature file and pass with the fix.

### DIFF
--- a/openpathsampling/engines/features/box_vectors.py
+++ b/openpathsampling/engines/features/box_vectors.py
@@ -3,6 +3,10 @@ numpy = ['box_vectors']
 
 dimensions = ['n_spatial']
 
+schema_entries = [
+    ('box_vectors', 'ndarray.float32({n_spatial},{n_spatial})')
+]
+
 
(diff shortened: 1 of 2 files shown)


sroet (approved):
LGTM

PR #1087, opened 2021-11-01
DOCS: Add GitHub support to `copyfiles`, remove online notebook downloads

dwhswenson:
In order to move notebooks into the right locations, a while ago I added a `copyfiles` script to our docs directory. This is specifically used to bring our docs from examples into the docs, where Sphinx can find them.

Today, I discovered that it was using GitLab to download (very outdated) copies of our tutorial notebooks. So I updated the script to allow us to use GitHub instead.

Then I realized that I we're not even *using* those downloaded notebooks in our docs. 🤷‍♂️ It definitely makes more sense to just *link* to the repositories, especially given the option of launching the tutorial with Binder.

In any case, I left the addition of support for GitHub, and removed the downloads of the tutorial notebooks from online (this also would have been an annoyance to anyone trying to build our docs without internet connectivity). I left previous usage commented as an example for any future users.

This should be immediately ready for review.

### DIFF
--- a/docs/copyfiles/copyfiles.py
+++ b/docs/copyfiles/copyfiles.py
@@ -33,9 +33,20 @@ def copy_gitlab(source, dest, files):
         print(url)
         urllib.request.urlretrieve(url, dest / f)
 
+def copy_github(source, dest, files):
+    domain = "http://raw.githubusercontent.com/"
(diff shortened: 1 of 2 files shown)


sroet (approved):
LGTM

PR #1088, opened 2021-11-07
SimStore: Fix storable function serialization

dwhswenson:
This fixes a bug in SimStore storable functions serialization. When support for function periodicity was added in #990, it looks like I failed to add the periodicity information to the `StorableFunction.to_dict`. This fixes that, and includes a test to prevent regression.

Currently, function periodicity is only used by the (as yet unreleased) command line tools for setting up a simulation, where it is only used at object creation time to determine whether a volume should be periodic or not. This means that the only use case where this bug could be a problem would be if the user is using the unreleased CLI `wizard` to create *new* volumes for periodic CVs *after* loading that CV from a file. This is an unlikely use case, so I don't think this is a critical concern, but it does push for an OPS release before the CLI wizard is released.

### DIFF
--- a/openpathsampling/experimental/simstore/test_storable_function.py
+++ b/openpathsampling/experimental/simstore/test_storable_function.py
@@ -375,7 +375,7 @@ class TestStorableFunction(object):
             return obj
 
         func = StorableFunction(func=identity, period_min=-2, period_max=2,
-                               store_source=True)
+                                store_source=True)
         dct = func.to_dict()
         assert 'identity' in dct['source']
         assert dct['period_max'] == 2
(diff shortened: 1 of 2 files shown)


dwhswenson:
This is ready for review and comment. I will leave it open for at least 24 hours, merging no earlier than Mon 08 Nov 14:00 GMT (09:00 my local).

openpathsampling/experimental/simstore/test_storable_function.py
```diff
@@ -374,12 +371,21 @@ def test_validate(self, is_valid):
 
 
     def test_to_dict_from_dict_cycle(self):
-        pytest.skip()
-        pass
+        def identity(obj):
+            return obj
+
+        func = StorableFunction(func=identity, period_min=-2, period_max=2,
+                               store_source=True)
```
sroet, line 378: indent is 1 off
```suggestion
                                store_source=True)
```

sroet (approved):
1 indentation suggestion, feel free to ignore. LGTM otherwise

PR #1090, opened 2021-11-09
Docs: New Installation docs

dwhswenson:
This PR is a rewrite of the installation page of the docs. The old installation page was originally written with the expectation that we'd need to onboard new contributors who had very little Python experience, and that we'd want to give them multiple options for how to get OPS installed.

However, the install docs are one of the first pages *users* come to. So I've re-written them to be more focused on user concerns. I kept a brief mention of installing by pip/setuptools, but basically assume that if you're doing that, you know what you're doing.

The new docs include also discuss various tools within the molecular simulation ecosystem that OPS integrates with, and they emphasize that all you need to do to get OPS to work with those tools is install them on your system.

Docs-only; this should be immediately ready for review. I will leave it open for at least 24 hours, merging no earlier than Wed 10 Nov 20:00 GMT (15:00 my local).

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -35,7 +35,7 @@ With that, you should be ready to use OPS!
 Installation with ``pip`` or ``setuptools``
 -------------------------------------------
 
-OPS can also be installed with ``pip`` via the command ``pip install
+OPS can also be installed with ``pip`` via the command ``python -m pip install
 openpathsampling``, or the source can be downloaded and installed using
 ``setuptools`` and the command ``python setup.py install`` (run from the
 root directory of the source repository). We only recommend these
@@ -45,7 +45,7 @@ Developer install
 -----------------
 
 OpenPathSampling can be installed as an "editable" (developer) install
-through standard mechanisms (``pip install -e .`` or ``python setup.py
+through standard mechanisms (``python -m pip install -e .`` or ``python setup.py
 develop``). 
 
 Additionally, we provide a script that downloads the OPS source repository,
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -1,103 +1,134 @@
-.. _install:
-
-************
+============
 Installation
-************
+============
 
-OpenPathSampling currently only works on Mac and Linux. It is tested against
-Python 2.7, 3.6, and 3.7, although there may be some corners of the code
-that aren't Python 3-compatible yet.
+This page has instructions on installing OPS, installing software packages
+that integrate with OPS, and testing that your OPS installation is working.
 
-.. note:: As of OpenPathSampling 1.1, OpenMM will no longer be automatically
-          installed when you install OPS. However, the OpenMM engine will be
-          immediately available if you install OpenMM yourself. See the
-          `OpenMM installation instructions
-          <http://docs.openmm.org/latest/userguide/application.html#installing-openmm>`_
-          for a detailed guide, but ``conda install -c conda-forge -c omnia
-          openmm`` will work for most people. (COMING SOON: details on OPS
-          integrations with other tools.)
+OpenPathSampling currently only works on POSIX systems (macOS and
+UNIX/Linux). It is tested against Python 2.7, 3.7, 3.8, and 3.9.
+
+Installing OpenPathSampling
+===========================
 
 .. _install-with-conda:
 
-Standard Install with Conda
-===========================
+Installation with ``conda`` (recommended)
+-----------------------------------------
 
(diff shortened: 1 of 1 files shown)


docs/install.rst
```diff
@@ -1,103 +1,134 @@
...
   $ conda install -c conda-forge openpathsampling
 
 With that, you should be ready to use OPS!
 
-.. _developer-install-conda:
+Installation with ``pip`` or ``setuptools``
+-------------------------------------------
+
+OPS can also be installed with ``pip`` via the command ``pip install
+openpathsampling``, or the source can be downloaded and installed using
```
sroet, line 39: future proofing this to the recommended method (`pip install openpathsampling` in python 3.10 installs it in a python3.1 subdirectory...)
```suggestion
OPS can also be installed with ``pip`` via the command ``python -m pip install
openpathsampling``, or the source can be downloaded and installed using
```

docs/install.rst
```diff
@@ -1,103 +1,134 @@
...
+``setuptools`` and the command ``python setup.py install`` (run from the
+root directory of the source repository). We only recommend these
+installation mechanisms for users who are familiar with these tools.
+
+Developer install
+-----------------
+
+OpenPathSampling can be installed as an "editable" (developer) install
+through standard mechanisms (``pip install -e .`` or ``python setup.py
```
sroet, line 48:
```suggestion
through standard mechanisms (``python -m pip install -e .`` or ``python setup.py
```

docs/install.rst
```diff
@@ -1,103 +1,134 @@
...
+In addition to the optional packages listed below, OPS ships with MDTraj
+built-in.
 
-Then, in the directory containing the source code, you can install it with ::
++------------------------+-------------------------------------------------+
+| | **PyEMMA**           |                                                 |
+| | |pyemma-conda|_      |                                                 |
++------------------------+-------------------------------------------------+
+| | **py-PLUMED**        |                                                 |
+| | |plumed-conda|_      |                                                 |
```
sroet, line 106: is this just a python wrapper or does it include an actual install of plumed as well (I trul

sroet (changes requested):
please update the pip commands to the current recommended `python -m pip`, and 1 question about py-plumed. LGTM otherwise

docs/install.rst
dwhswenson: py-plumed [recipe requires plumed](https://github.com/conda-forge/py-plumed-feedstock/blob/278ece433668b8128a9db0d73832daad2264d02a/recipe/meta.yaml#L30) (which has a separate [feedstock](https://github.com/conda-forge/plumed-feedstock/)), so installing from conda brings them both. Other installations of the Python wrapper may require separate install of PLUMED (for example, if you get it from [PyPI](https://pypi.org/project/plumed/), where the requirement is documented pretty clearly.)

Since the Python wrappers allow you to provide the PLUMED kernel at runtime there's no reason not to install with conda -- if you have a custom build of PLUMED, you can still use the conda Python wrappers.

docs/install.rst
dwhswenson: Taking the suggestions, but do you have a link for the dangers of straight `pip install`? I knew that `python -m` was preferred (and should have known to write that here). But I always thought it was a "make sure that `pip` is associated with the `python` you want." I hadn't heard of Python 3.10 issues.

Personally, I switched to to `python -m pip` because sometimes conda doesn't seem to `hash -r` correctly. I will often `which python` to check that it is correct in a new env, but don't think to `which pip`, and after completely borking my conda setup when installing code for a software paper I was reviewing..... Never again!

sroet: I normally refer to [this blog post](https://snarky.ca/why-you-should-use-python-m-pip/). It is mainly making sure you are installing into the expected python (which I have seen very experienced users F up on HPC systems due to `module load` shenanigans) on top of that there is the benefit:

> And if you're on Windows there is an added benefit to using python -m pip as it lets pip update itself. Basically because pip.exe is considered running when you do pip install --upgrade pip, Windows won't let you overwrite pip.exe. But if you do python -m pip install --upgrade pip you avoid that issue as it's python.exe that's running, not pip.exe.

the 3.10 -> 3.1 issue is just something I noticed when trying to build mdtraj on 3.10 (which is possible :smile: ). I think that one got fixed recently in both pip and conda (conda also used to complain about not finding any packages for python 3.1 if you try to conda update from a python 3.10 env :laughing: )

sroet (approved):
LGTM

PR #1091, opened 2021-12-06
Guard against openmm import warnings (including those from OpenMMTools)

sroet:
For a while, the notebook tests have been failing due to the deprecation warning about importing `simtk.openmm` instead of `openmm`. 

~I would like to use the correct import, but openmm 7.6 did not build for python 2.7, so instead I just have to ignore the warning.~

The PR guards the import with a try except that falls back to the old behaviour


dwhswenson (commented):
Thanks for the fix (and sorry for not getting to it earlier; I was out of internet range last week and had been preparing for life without internet for the week before that, as well as preparing for life in general).

A couple of thoughts:

1. If this is OpenMM 7.6-related, why didn't this start happening a *lot* earlier?
2. It seemed to me that this error coincided with the most recent MDTraj release. What happens if the MDTraj import is in a different cell?
2. Does this work with a try/except, e.g.:
  ```python
  try:
      import openmm as omm
      import openmm.unit as u
  except ImportError:
      import simtk.openmm as omm
      import simtk.unit as u
  ```

I'd prefer something that doesn't involve an ignore if we can help it (don't want to fall into the ignore-everything issue we had with PyEMMA integration testing), but as long as it is only the imports, I suppose it wouldn't be a problem.

(Aside: this is another set of integration tests that really should be in the main pytest run. Not asking that the change be part of this PR; just mentioning that if it ever starts giving more headaches, that's a completely valid solution.)

sroet:
>1. If this is OpenMM 7.6-related, why didn't this start happening a lot earlier?

No clue, it might be due to caching, or if it is pulling numpy 1.21 (as opemmtools currently can't be imported with that numpy version)

> 2. It seemed to me that this error coincided with the most recent MDTraj release. What happens if the MDTraj import is in a different cell?


Screenshot from 2021-12-07 09-59-24


> 3. Does this work with a try/except, e.g.:

That seems reasonable, if it is not due to numpy 1.21, this should be fine (implemented in f30ed0d)

sroet:
> (and sorry for not getting to it earlier; I was out of internet range last week and had been preparing for life without internet for the week before that, as well as preparing for life in general).

No worries, the nice thing about opensource is that I can fix it if I get annoyed enough by it :laughing:

sroet:
Alright, this is probably not due to the import warnings.  One error I saw pop up in testing was due to numba not playing nice with numpy 1.21 for openmmtools...

sroet:
This is clearly comming from openmmtools, not too much I can do about that after https://github.com/choderalab/openmmtools/pull/528 (I forgot I had apparently already fixed this :facepalm: ) 
@dwhswenson how do we want to proceed from here?
- split out openmmtools import and ignoring that one? 
- keep the tests failing until openmmtools does a release (unknown timeline)?

dwhswenson:
> * split out openmmtools import and ignoring that one?

This is definitely the right choice. Please also add a comment that says something like `# ignore until next OpenMMTools release, see openpathsampling/openpathsampling#1091` so that it's easy to figure out when that is no longer necessary (and maybe we can even completely undo this PR once it is no longer needed).

If I understand correctly, it sounds like it should be possible to undo this once our CI is picking up `openmmtools > 0.20.3`. (Writing that explicitly now so I don't need to compare date of this PR to OpenMMTools release dates in the future. 😉)

sroet:
> This is definitely the right choice. Please also add a comment that says something like # ignore until next OpenMMTools release, see openpathsampling/openpathsampling#1091 so that it's easy to figure out when that is no longer necessary (and maybe we can even completely undo this PR once it is no longer needed).

Done, and all green. Please have another look

dwhswenson (approved):
LGTM! Thanks!

PR #1092, opened 2021-12-10
Actually check for CVDefinedVolume.lambda_max == "inf"

sroet:
while looking at the correctness of the docstring in #1068 I noticed a tiny bug in CVDefinedVolume.

fixed bug in 8ea8d38
added test in 11c4119
dropped nose from test_volume.py in 41ffb9b

### DIFF
--- a/openpathsampling/tests/test_volume.py
+++ b/openpathsampling/tests/test_volume.py
@@ -91,10 +91,20 @@ class TestCVDefinedVolume(object):
         assert_equal(volA(0.50), False)
         assert_equal(volA(0.51), False)
 
+    def test_inf_upper_boundary(self):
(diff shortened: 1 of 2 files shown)


dwhswenson (approved):
Nice catch. Will merge on green.

By the way, do you have a script for making the nose -> pytest change, or is that all manual effort? I'm about to start the same on the tests in #787 so we can merge that in. Manual updates aren't *that* bad, but a script would be easier!

sroet:
> By the way, do you have a script for making the nose -> pytest change, or is that all manual effort? I'm about to start the same on the tests in #787 so we can merge that in. Manual updates aren't that bad, but a script would be easier!

Nope, just doing it by hand while I think about some programming/scientific issues (it is brainless enough to be almost like meditating for me xD)

PR #1093, opened 2021-12-14
Fix build for OpenMM RCs

dwhswenson:
Our automated check against new OpenMM release candidates picked up a new RC last night, and the build [is failing](https://github.com/openpathsampling/openpathsampling/runs/4517804647?check_suite_focus=true) with a conda `UnsatifiableError`. I'm going to try to fix this on our side. Among other things, it seems like there's a weird combination trying to pull in `python=3.10` while also pulling in `pypy3.6`, which seems like an obvious source of conflict. We don't currently test against 3.10, so I'm guessing that's the issue (this fails before we try to install the OpenMM RC, so I suspect the problem due to changes in the GitHub Actions environment, not due to the RC itself).

I'll do a little on this now, but if my first tries don't solve, I'll come back to it later in the day (after 19:00 GMT).

The offending conda command in [line 17 in the "Install requirements" stage](https://github.com/openpathsampling/openpathsampling/runs/4517804647?check_suite_focus=true#step:5:17):

```
conda install -y -q -c conda-forge -c omnia --override-channels python=3.10   future psutil numpy scipy pandas netcdf4 svgwrite networkx matplotlib ujson!=2.*,!=3.*,!=4.0.0,!=4.0.1 mdtraj jupyter openmm!=7.5.1 openmmtools py-plumed pyemma   sqlalchemy!=1.4.0 dill   nose pytest pytest-cov coveralls nbval
```

### DIFF
--- a/.github/workflows/test-openmm-rc.yml
+++ b/.github/workflows/test-openmm-rc.yml
@@ -4,8 +4,8 @@ name: "Test OpenMM Release Candidate"
 on:
   workflow_dispatch:
   # use this for debugging
-  #pull_request:
-    #branch: master
+  pull_request:
+    branch: master
 
 defaults:
   run:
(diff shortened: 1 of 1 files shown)


dwhswenson:
Yep, it was just a matter of needing to pin Python to 3.9, as can be seen in 6648835. Figuring out the issue with Python 3.10 will be future work; it is probably in an (optional) integration requirement, since conda-forge [opened a PR to release OPS against Python 3.10](https://github.com/conda-forge/openpathsampling-feedstock/pull/20) in our feedstock. In additional good news, the tests against the RC pass, so it looks like we have no problems with the RC this time.

This is ready for review and comment. I will leave it open for at least 8 hours, merging no earlier than Wed 15 Dec 00:00 GMT (Tue 14 Dec 19:00 my local).

sroet:
LGTM, do you mind making an issue to remember to undo this?

sroet (approved):

PR #1094, opened 2021-12-14
enable py310 tests

sroet:
should be the only thing required to test https://github.com/conda-forge/openpathsampling-feedstock/pull/20

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -49,12 +49,12 @@ jobs:
       - uses: conda-incubator/setup-miniconda@v2
         with: 
           auto-update-conda: true
-          python-version: ${{ matrix.CONDA_PY }}
+          python-version: "${{ matrix.CONDA_PY }}"
           miniforge-variant: Mambaforge
       - name: "Install requirements"
         env:
           MINIMAL: ${{ matrix.MINIMAL }}
-          CONDA_PY: ${{ matrix.CONDA_PY }}
+          CONDA_PY: "${{ matrix.CONDA_PY }}"
         run: |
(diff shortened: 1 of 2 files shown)


dwhswenson:
Related: #1095 (written at same time! 😆)

I suspect the issues will be with the various optional requirements. These can be manually managed based on the value of Python version `CONDA_PY` by modifying `devtools/conda_install_reqs.sh`.

sroet:
alright, so the first things required for this to work is openmm needs to have a python 3.10 build (conda search openmm comes up with nothing) 
do we want to add an issue on their tracker?

dwhswenson:
> do we want to add an issue on their tracker?

No need -- they won't release one anyway until the 7.7 release (in RC now). That will be followed quickly by an OpenMMTools release, too (on mobile now, but there's an issue on their feedstock, IIRC).

sroet:
this is waiting on:  https://github.com/conda-forge/openmm-feedstock/pull/68 or some equivalent

dwhswenson:
Tried re-running this today. Looks like there still isn't a Python 3.10 build on OpenMM on conda-forge. See https://github.com/conda-forge/openmm-feedstock/pull/68. Currently looks like there are problems with pypy on aarch64  CPython 3.10 on win.

dwhswenson:
(And apparently we both found that within about 10 seconds of each other 😆)

sroet:
retriggering CI as openmm is now available for python 3.10

sroet:
This is now waitng for a pyemma release for python 3.10

dwhswenson:
Superseded by #1133.

Issue #945, opened 2020-11-17
Deprecation warnings due to invalid escape sequences in Python 3.8


tirkarthi:
Deprecation warnings are raised due to invalid escape sequences. This can be fixed by using raw strings or escaping the literals. pyupgrade also helps in automatic conversion : https://github.com/asottile/pyupgrade/ . 

```
find . -iname '*.py' -not -path "wandb/vendor/*" | xargs -P4 -I{} python3 -Wall -m py_compile {}  
./openpathsampling/tests/test_external_engine.py:48: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if line is '':
./openpathsampling/netcdfplus/attribute.py:202: DeprecationWarning: invalid escape sequence \*
  """
./openpathsampling/netcdfplus/attribute.py:408: DeprecationWarning: invalid escape sequence \*
  """Turn a callable class or function generating a callable object into a CV
./openpathsampling/range_logic.py:1: DeprecationWarning: invalid escape sequence \ 
  '''
./openpathsampling/engines/features/numpydoctools.py:207: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  if section is not 'HEAD':
./openpathsampling/experimental/simstore/my_types.py:4: DeprecationWarning: invalid escape sequence \.
  "ndarray\.(?P<dtype>[a-z0-9]+)(?P<shape>\([0-9\,\ ]+\))"
./openpathsampling/experimental/simstore/serialization_helpers.py:50: DeprecationWarning: invalid escape sequence \(
  encoded_uuid_re = re.compile("UUID\((?P<uuid>[\-]?[0-9]+)\)")
./openpathsampling/numerics/histogram.py:596: DeprecationWarning: invalid escape sequence \_
  """Set up both x-axis and y-axis for plotting.
./openpathsampling/treelogic.py:278: DeprecationWarning: invalid escape sequence \*
  """
./openpathsampling/treelogic.py:335: DeprecationWarning: invalid escape sequence \*
  """
./openpathsampling/treelogic.py:366: DeprecationWarning: invalid escape sequence \*
  """
./openpathsampling/treelogic.py:402: DeprecationWarning: invalid escape sequence \*
  """
./openpathsampling/treelogic.py:433: DeprecationWarning: invalid escape sequence \*
  """
./docs/premake.py:14: DeprecationWarning: invalid escape sequence \_
  fname = re.sub('\_', '-', fname)
[...]

PR #1096, opened 2021-12-20
make a missed string into a rawstring

sroet:
This fixes a DeprecationWarning that popped up in the running tests after merging #787 
```
openpathsampling/engines/toy/pes.py:374
/home/sroet/github_files/openpathsampling/openpathsampling/engines/toy/pes.py:374: DeprecationWarning: invalid escape sequence \s
"""Simple double-well potential. Independent in each degree of freedom.
```


dwhswenson:
LGTM, thanks. We may have a few other of these that need treatment, too; see #945.

PR #1097, opened 2022-01-03
update copyright to 2022

sroet:
Happy new year!

This updates the copyright I found with `grep -rl "2021" ./ --exclude=*.{pdb}`

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2014-2019 the contributors to OpenPathSampling
+Copyright (c) 2014-2022 the contributors to OpenPathSampling
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -107,7 +107,7 @@ master_doc = 'index'
 
 # General information about the project.
 project = u'OpenPathSampling'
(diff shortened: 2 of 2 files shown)


dwhswenson (commented):
Looks mostly good to me (after I added a commit to update LICENSE as well). One question is whether to change the name list in the docs copyright notice. That text shows up in the footer of every page in the docs, and I think it should be considered as primarily relating to the docs, as opposed to the code, which is covered in LICENSE. (Though, obviously, the code is used to generate the docs.)

I think this should acknowledge the existence of contributors beyond the original "gang of 4" that it currently lists. There are a few options here:

* `the contributors to OpenPathSampling`: this is the text used in the license, which puts no emphasis on any particular person
* `David W.H. Swenson, Jan-Hendrik Prinz, John Chodera, Peter Bolhuis, and other contributors to OpenPathSampling`: This is a bit wordy, but keeps everyone listed who is currently listed
* `David W.H. Swenson and other contributors to OpenPathSampling`: here I'm taking credit for the fact that nearly all the text in the docs directory was written by me (which means that if any explanations on the website are poorly written, that's probably also my fault)

Any thoughts?

sroet:
> Any thoughts?

Depends on your commitment to keeping them up to date :wink:.  If you are committed to that, I would go for option 3 (as you said, you are responsible for most of the text), otherwise I would go for option 1. (no one particular claims responsibility). However, I would at least ping the other contributors if you plan on dropping their names.

dwhswenson:
Not worrying about changing who the copyright credits right now. Let's merge this instead of waiting until 2023. 🤣

PR #1102, opened 2022-01-27
Add FutureWarning for default filesetter external engines

sroet:
this adds a `FutureWarning` to let users know that in OPS 2.0 the default `FileSetter` will become random string

This was part of the discussion on #1101

PR #1103, opened 2022-01-27
Update standard filenamesetter for OPS 2.0

sroet:
This does a couple things:
- cherry-picks in #1102
- the default `filename_setter` for `ExternalEngines` are now `RandomStringFilenames` instead of `FilenameSetter`
- `FilenameSetter` is now a subclass of `RandomStringFilenames` (instead of the other way around) (this was needed to still be able to use Filename setter after changing the default options)
- `FilenameSetter now returns a string that is padded up to `self.length` with zeroes
- `GromacsEngine.set_filenames` only resets the `filename_setter` if the `filename_setter` has a `reset` attribute
- updated the tests for this behaviout
- solved pep8 complaints on all touched files
- dropped `nose` from all touched tests

### DIFF
--- a/openpathsampling/deprecations.py
+++ b/openpathsampling/deprecations.py
@@ -1,7 +1,6 @@
 from __future__ import print_function
 import sys
 import warnings
-from collections import namedtuple
 from functools import wraps
 from inspect import isclass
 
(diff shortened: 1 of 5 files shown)


sroet:
failing CI is due know issue with importing openmm (or openmmtools) in notebooks is raising a warning: 

```
Traceback:
Unexpected output fields from running code: {'stderr'}
```

sroet:
@dwhswenson all green, this is ready for review

Issue #1101, opened 2022-01-27
Continuing a Path Sampling Simulation with GROMACS


jjfsteuer:
First of all a big thanks to the developers for the great software, documentation and continuous development. 

I am currently trying to continue a TPS simulation using OPS with the GROMACS Engine. 

My initial run consists of 100 steps, and to append to that simulation storage I use the following code:

```
storage = paths.Storage(filename, "a")
tps_calc = storage.pathsimulators[0]
tps_calc.restart_at_step(storage.steps[-1]) #
tps_calc.run(200)
storage.close()
```

While the OPS package appears to do what I expect

```
Working on Monte Carlo cycle number 101
Starting simulation...
Working on first step
```
the executed GROMACS command does not adopt the corresponding index

`gmx -nobackup mdrun -s topol.tpr -o ././prod_trr/0000001.trr -e ././prod_edr/0000001.edr -g ././prod_log/0000001.log`

and the trajectory files from the initial run are overwritten.

I was able to reproduce the same problem for the alanine dipeptide example for GROMACS (trying to continue from the simulation in AD_tps_2a_run_flex.ipynb). 
Again, the appending run appears to continue with the TPS where it is supposed to, but the GROMACS filenames are restarted at 0000001.

Any advice or suggestion what might be going wrong is greatly appreciated!
Best, 
Jakob

dwhswenson:
> We can add a restart_at_step() for the external engine (like we have doen for SpringShooting) and add this to the examples/docs

I think this is the best solution here. The best way for OPS 2.0 might be for simulation objects to implement a `restart_at_step` method, then use the serialization tools in SimStore to identify all simulation objects, and for `simulation.restart_at_step` to call that for all objects that implement it (this also solves it for `SpringShooting`).

@jjfsteuer : As @sroet mentions in his suggestion for 2.0, 2.0 will default to using random filenames instead of sequential numbers. You can already use this functionality, and I would recommend it. It should solve the problem you're having. See #933 for details and some other reasons that it was originally implemented. Of course, this doesn't change the data you've already created, but for future data, it's easier.

 (Note to self -- not sure that the 2.0 branch actually *has* that default yet, but it should be possible to do.)

sroet:
> (Note to self -- not sure that the 2.0 branch actually has that default yet, but it should be possible to do.)

It does not, hence my proposal for a deprecation warning + 2.0 update

jjfsteuer:
Thank you so much for the quick response!

[...]

PR #1105, opened 2022-02-03
Add `restart_at_step` for `ExternalEngine` (for numeric filenames)

sroet:
with #1103 and #1102, I believe this implements the last thing discussed in  #1101:

It adds a `restart_at_step` for `ExternalEngine`s, which (if `self.filename_setter` is `FilenameSetter`) set's it's count to the last written trajectory + 1

this will error out if it tries to read filenames that don't end in a number, for now a `ValueError: invalid literal for int() with base 10: ''`, but this can be updated to make it more clear.

Also updated the AD gromacs example to include the required restart code.


dwhswenson:
@sroet : I seem to be having trouble updating this PR to the new `master`: can you pull in the upstream changes and push to this branch? Just want to get a fresh CI run to make sure everything is still fine.

I'm also trying to find anything else that should be put into a 1.7 release; this seemed to be the only PR that was really needed. (I think 1.8 is going to focus a lot of storage stuff, so I'll put storage handlers into that.)

dwhswenson (approved):
Code looks fine, and now tested against recent updates!

Issue #1108, opened 2022-03-22
Experimental Storage fails at reloading


gianmarcolazzeri:
Hello! When restoring a "db" storage, function `tuple_keys_from_dict` in `openpathsampling.experimental.dict_searization_helpers` doesn't treat properly an input dictionary in which the key it attempts to tuple is a string. I don't know if it is because the input dictionary shouldn't be like this, but found an easy fix by replacing:
```
    def tuple_keys_dict_from_dict(name, dct):
        keys = dct[name + "_tuple_keys"]
        values = dct[name + "_values"]
        return {tuple(k): v for k, v in zip(keys, values)}
```
with
```
    def tuple_keys_dict_from_dict(name, dct):
        keys = dct[name + "_tuple_keys"]
        values = dct[name + "_values"]
        return {(tuple(k) if type(k) is not str else k): v for k, v in zip(keys, values)}
```

dwhswenson:
Hi @gianmarcolazzeri, thanks for the report!

I'm curious about the situation where this is happening -- `tuple_keys_dict_from_dict` is a workaround for a few cases where OPS stores a dictionary that uses tuples for keys, so it shouldn't be used on dicts with strings as keys. I have a couple questions:

1. Is this happening somewhere within normal usage of the core OPS library, or is this something that's popping up in custom code you've written?
2. If it happens in the core library, then there's a bug somewhere that I need to fix; could you give more details on how it happens? If it is in custom code, this may mean that there's an easier way to accomplish what you're trying to do -- I'm happy to give advice on that if you'd like.

[...]

PR #1109, opened 2022-03-24
Prevent double monkey-patching experimental storage

dwhswenson:
It was possible to double monkey-patch for the experimental storage. This caused many headaches downstream (e.g., OPS CLI) because that meant that something downstream had to track state of whether OPS had been monkey-patched. It also might be the problem behind #1108. This PR makes it so that monkey-patching is idempotent.

Marked WIP because it still needs tests (which will be slightly a pain), but this is the implementation I'm planning on.

### DIFF
--- a/openpathsampling/experimental/storage/monkey_patches.py
+++ b/openpathsampling/experimental/storage/monkey_patches.py
@@ -120,4 +120,8 @@ def unpatch(paths):
     for module in _UNPATCH_MODULES * 2:
         importlib.reload(module)
 
+    global _IS_PATCHED_SAVING
+    global _IS_PATCHED_LOADING
+    _IS_PATCHED_SAVING = False
+    _IS_PATCHED_LOADING = False
     return paths
--- a/openpathsampling/experimental/storage/monkey_patches.py
+++ b/openpathsampling/experimental/storage/monkey_patches.py
@@ -64,7 +64,14 @@ _UNPATCH_MODULES = [
     paths
 ]
 
+_IS_PATCHED_SAVING = False
+_IS_PATCHED_LOADING = False
+
 def monkey_patch_saving(paths):
+    global _IS_PATCHED_SAVING
+    if _IS_PATCHED_SAVING:
+        return paths
+
     paths.netcdfplus.FunctionPseudoAttribute.to_dict = \
             function_pseudo_attribute_to_dict
     paths.TPSNetwork.to_dict = tuple_keys_to_dict(
@@ -73,9 +80,14 @@ def monkey_patch_saving(paths):
     paths.MISTISNetwork.to_dict = tuple_keys_to_dict(
         paths.MISTISNetwork.to_dict, 'input_transitions'
     )
+    _IS_PATCHED_SAVING = True
     return paths
 
 def monkey_patch_loading(paths):
+    global _IS_PATCHED_LOADING
+    if _IS_PATCHED_LOADING:
+        return paths
+
     paths.CallableCV.from_dict = classmethod(callable_cv_from_dict)
     paths.netcdfplus.FunctionPseudoAttribute.from_dict = \
             classmethod(from_dict_attr_to_class(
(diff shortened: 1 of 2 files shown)


dwhswenson:
Tests weren't quite as bad as I feared, once I sat down to write them.

This is ready for review and comment. I will leave it open for at least 48 hours, merging no earlier than Thu 28 Jul 00:00 GMT (Wed 27 Jul 19:00 my local).

openpathsampling/experimental/storage/monkey_patches.py
```diff
@@ -64,7 +64,14 @@ def inner(cls, dct):
     paths
 ]
 
+_IS_PATCHED_SAVING = False
+_IS_PATCHED_LOADING = False
```
sroet, line 68: For global variables, I would like to have an indication where they come from (and slightly more protection against a separate package doing the same thing), do you mind changing these (and all the other ones) to
```suggestion
_OPS_IS_PATCHED_SAVING = False
_OPS_IS_PATCHED_LOADING = False
```

openpathsampling/experimental/storage/test_monkey_patches.py
```diff
@@ -0,0 +1,43 @@
...
+    paths = monkey_patch_all(paths)
+    try:
+        patched_dict = patching_tps_network.to_dict()
+        paths = monkey_patch_all(paths)
+        double_patched_dict = patching_tps_network.to_dict()
+    finally:
+        unpatch(paths)
+
+    assert orig_dict != patched_dict
+    assert patched_dict == double_patched_dict
```
sroet, line 29: I would probably rewrite this so that the whole test is caught in the try (including the asserts) as now 
`double_patched_dict` is not guaranteed to be defined if the `monkey_path_all` fails.

I would probably rewrite like:
```suggestion
    try:
        patched_dict = patching_tps_network.to_dict()
        paths = monkey_patch_all(paths)
        double_patched_dict = patching_tps_network.to_dict()
        assert orig_dict != patched_dict
        assert patched_dict == double_patched_dict
    except:
        unpatch(paths)
        raise
    else:
        unpatch(paths)
```

openpathsampling/experimental/storage/test_monkey_patches.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+def test_unpatch(patching_tps_network):
+    orig_dict = patching_tps_network.to_dict()
+    global paths
+    paths = monkey_patch_all(paths)
+    try:
+        patched_dict = patching_tps_network.to_dict()
+    finally:
+        paths = unpatch(paths)
```
sroet, line 39: same as above
```suggestion
    try:
        patched_dict = patching_tps_network.to_dict()
    except:
        unpatch(paths)
        raise
    else:
        paths = unpatch(paths)
```

sroet (changes requested):
A couple comments, nothing mayor, so feel free to ignore if you have a different opinion

openpathsampling/experimental/storage/test_monkey_patches.py
dwhswenson: I think you meant to add the rest of the test to the try part of the suggestion?  Is the suggestion, as is, any different in practice from the current code? I thought that just using `finally` would be the same as `except: do_something(); raise` followed by `else: do_something()`?

openpathsampling/experimental/storage/test_monkey_patches.py
dwhswenson: I'm not concerned by this -- the `finally` clause doesn't prevent the error from raising. So if there's an error, the asserts should not be reached.

openpathsampling/experimental/storage/monkey_patches.py
dwhswenson: globals aren't true globals -- they're still module scope (even if the module they're scoped to is `__main__`). [The `global` keyword is necessary to set module-level globals from within a function.](https://stackoverflow.com/a/7060755) So I don't think this is a problem.

sroet: you are right, sorry about my confusion.

openpathsampling/experimental/storage/test_monkey_patches.py
sroet: Sorry, I misremembered how a `try` without an `except` worked. I somehow was convinced it would catch `all` (don't know why I thought that. :facepalm:  )

you code is better than my suggested change.

sroet:
All my comments have been handled (all of them where due to me how python does things). This LGTM, feel free to merge

PR #1114, opened 2022-05-30
(Possible temporary) Fix dill in testing

### LINKED ISSUE #1112: Dill 0.3.5 causes problems in test suite
Tests for the experimental storage have been failing since dill 0.3.5 was released. I think this is actually a problem related to how our tests create a function that mimics a typical user-created function, as opposed to a fundamental problem in SimStore. That is, the incompatibility is in the code for the tests, not the main code itself. (Users don't need to worry, except that this makes it harder for us to ensure correct behavior is continuing.)

I've opened an issue on dill to see if they can fix it on their end or give us advice on how to fix it on our end. https://github.com/uqfoundation/dill/issues/482

dwhswenson:
Resolves #1112.

This should at least get tests working. I'm keeping an eye on https://github.com/uqfoundation/dill/issues/482 / https://github.com/uqfoundation/dill/pull/486 for possible better long-term fixes, although it might make sense to just leave this as-is. Serialization of callables shouldn't be a performance bottleneck for us, so as long as this works, I'm not worried if this might be a little slower.

### DIFF
--- a/openpathsampling/experimental/simstore/callable_codec.py
+++ b/openpathsampling/experimental/simstore/callable_codec.py
@@ -127,7 +127,7 @@ class CallableCodec(object):
 
             return {
                 '__callable_name__': obj.__name__,
-                '_dilled': dill.dumps(obj)
+                '_dilled': dill.dumps(obj, recurse=True)
             }
         return obj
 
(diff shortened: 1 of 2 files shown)


sroet (approved):
LGTM, feel free to merge on green

Issue #1117, opened 2022-08-08
Upon doing Alanine Dipeptide tutorial given on openpathsampling website, I have faced a error in the CommittorSimulation


skhabib388:
Upon doing the Alanine Dipeptide tutorial given on openpathsampling website using GROMACS, all the steps I have done like what it is showing in the tutorial. All the steps up to committor analysis were fine. But I have faced one error while doing CommittorSimulation. I am giving the line of codes and errors below.

Code lines:
output_storage = paths.Storage("committor_simulation.nc", "w")
simulation = paths.CommittorSimulation(storage=output_storage,
                                       engine=engine,
                                       states=[C_7eq, alpha_R],
                                       randomizer=randomizer,
                                       initial_snapshots=snapshots_to_run)
 simulation.run(n_per_snapshot=10)

Error:

Working on snapshot 1 / 1000; shot 1 / 10
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Input In [22], in <module>
----> 1 simulation.run(n_per_snapshot=10)

File ~/Downloads/data_analysis_water_exchange/AlaDipeptide/openpathsampling/pathsimulators/shoot_snapshots.py:145, in ShootFromSnapshotsSimulation.run(self, n_per_snapshot, as_chain)
    143     start_snap = self.randomizer(start_snap)
    144 else:
--> 145     start_snap = self.randomizer(snapshot)
    147 sample_set = paths.SampleSet([
    148     paths.Sample(replica=0,
    149                  trajectory=paths.Trajectory([start_snap]),
    150                  ensemble=self.starting_ensemble)
    151 ])
    152 sample_set.sanity_check()

File ~/Downloads/data_analysis_water_exchange/AlaDipeptide/openpathsampling/snapshot_modifier.py:213, in RandomVelocities.__call__(self, snapshot)
    210     except AttributeError:
    211         pass  # use default
--> 213 new_snap = make_snapshot(snapshot)
    214 return new_snap

File ~/Downloads/data_analysis_water_exchange/AlaDipeptide/openpathsampling/snapshot_modifier.py:160, in RandomVelocities._default_random_velocities(self, snapshot, beta, subset)
    157 vel_subset = self.extract_subset(velocities, subset)
    159 # raises AttributeError if snapshot doesn't support masses feature
--> 160 all_masses = snapshot.masses
    161 masses = self.extract_subset(all_masses, subset)
    163 n_spatial = len(vel_subset[0])

[...]

PR #1119, opened 2022-08-09
[WIP] engine_masses feature; use in Gromacs engine

dwhswenson:
Apparently Gromacs snapshots didn't have masses associated? See #1117. I was pretty sure we've used masses here before, but can't find an example of it.

Note that the better long-term solution for the needs of #1117 will be an `engine.randomize_velocities` -- we already have the machinery to *use* this, but haven't actually implemented any examples yet.

### DIFF
--- a/openpathsampling/engines/external_snapshots/snapshot.py
+++ b/openpathsampling/engines/external_snapshots/snapshot.py
@@ -13,6 +13,7 @@ logger = logging.getLogger(__name__)
     features.coordinates,
     features.velocities,
     features.box_vectors,
+    features.engine_masses,
     ext_features.file_info
 ])
 class InternalizedMDSnapshot(BaseSnapshot):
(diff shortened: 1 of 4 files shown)


sroet:
Leaving a comment here to keep the issue clean, but about
> We'll need to be able to correctly randomize velocities, which requires both creating random velocities and then applying constraints. I don't think we have a way to apply constraints to a Gromacs snapshot yet. These are things that would be good to add, but I probably can't get to this in the very near future.

I am pretty sure that was fixed in PyRETIS by letting gromacs run a `0` step simulation, wouldn't a similar solution work for OPS?

sroet:
> (Goal for the next two weeks is to get 1.6 released before my ACS talk on Aug 24; I'm just glad to finally be back to working on OPS!) 

Good luck with your talk! Just ping me if you want my eyes on something.

PR #1123, opened 2022-11-13
update tests to use setup_method instead of setup, same for teardown

sroet:
I noticed today that a bunch of 
```
openpathsampling/tests/test_pathmover.py::TestTwoWayShootingMover::test_to_dict_from_dict
  /usr/share/miniconda3/envs/test/lib/python3.7/site-packages/_pytest/fixtures.py:900: PytestRemovedIn8Warning: Support for nose tests is deprecated and will be removed in a future release.
  openpathsampling/tests/test_pathmover.py::TestTwoWayShootingMover::test_to_dict_from_dict is using nose-specific method: `setup(self)`
  To remove this warning, rename it to `setup_method(self)`
  See docs: https://docs.pytest.org/en/stable/deprecations.html#support-for-tests-written-for-nose
    fixture_result = next(generator)
```
came form our test suite. This solves them (together with a similar warning for `teardown`)

### DIFF
--- a/openpathsampling/tests/pathsimulators/test_hooks.py
+++ b/openpathsampling/tests/pathsimulators/test_hooks.py
@@ -170,7 +170,7 @@ class TestPathSimulatorHook(object):
 
 
 class TestSelfOrSimProperty(object):
-    def setup(self):
+    def setup_method(self):
         self.attr_name = "test_attr"
         self.simulation = MagicMock(test_attr="sim_attr")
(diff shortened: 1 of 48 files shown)


dwhswenson (approved):
Thanks! This issue had also come up at $DAYJOB, and I was *not* looking forward to figuring out how to get this change right without it being tedious. I'm guessing the key was to search for `def setup(self):`, as opposed to just looking for `setup`?

Merging!

PR #1128, opened 2023-01-07
Fix Dropbox download of precalculated results

dwhswenson:
Something recently broke in the download of results that are used in the notebook tests.

TBH, I should probably move that into a git repo or something, instead of keeping them in my personal Dropbox. The use of my personal Dropbox for this probably predates the existence of an OpenPathSampling organization on GitHub, and (except few times I was stupid enough to move those files), it's never really caused us problems. But still. Not best practice.

### DIFF
--- a/examples/ipynbtests.sh
+++ b/examples/ipynbtests.sh
@@ -33,8 +33,8 @@ case $PYTHON_VERSION in
         echo "Unsupported Python version: $PYTHON_VERSION"
 esac
 
-curl -OLk $mstis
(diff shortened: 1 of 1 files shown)


dwhswenson:
Apparently I needed to specify HTTP 1.1?

This is ready for review and comment. I will leave it open for at least 24 hours, merging no earlier than Mon 09 Jan 00:00 GMT (Sun 08 Jan 18:00 my local).

PR #1129, opened 2023-01-15
Fix testing against OpenMM RCs

dwhswenson:
Some change has caused the token for workflow dispatch (the mechanism we used to conditionally launch a workflow) to fail. However, in the meantime, GitHub has added a native approach for reusing workflows. So this is intended to switch to using that.

And we're probably going to have some things to change thanks to the OpenMM 8.0 release candidate....

### DIFF
--- a/.github/workflows/check-openmm-rc.yml
+++ b/.github/workflows/check-openmm-rc.yml
@@ -29,6 +29,6 @@ jobs:
 
   test_rc:
     needs: check_rc
-    uses: ./.github/workflow/test-openmm-rc.yml
+    uses: ./.github/workflows/test-openmm-rc.yml
     secrets: inherit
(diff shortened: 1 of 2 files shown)


dwhswenson:
> And we're probably going to have some things to change thanks to the OpenMM 8.0 release candidate....

Apparently not! Tests against the RC ran as part of [f8a0335](https://github.com/openpathsampling/openpathsampling/pull/1129/commits/f8a0335ebd5b228b3a3c0582ceb39c5ce945519c), and [passed](https://github.com/openpathsampling/openpathsampling/actions/runs/3924103322/jobs/6708135547)!

This is ready for review and comment. I will leave it open for at least 12 hours, merging no earlier than Mon 16 Jan 04:00 GMT (Sun 15 Jan 22:00 my local). (Short window for minor changes and fact that I kind of don't want another overnight test failure in my inbox.)

PR #1132, opened 2023-09-15
Get tests working again

dwhswenson:
Taking a few days to do some OPS upkeep; after far too much neglect. This PR will get our tests back in working order.

Importantly, as of this PR (and the next release), we're (finally) dropping Python 2.7 support. It is too much of a maintenance burden, and everyone else dropped it over 3 years ago. There should be no active projects going with Python 2.7.

- [x] Drop Python 2 support (finally)
- [x] Fix up Gromacs tests for newer grompp warnings

### DIFF
--- a/openpathsampling/engines/external_engine.py
+++ b/openpathsampling/engines/external_engine.py
@@ -13,6 +13,8 @@ import signal
 import shlex
 import time
 
+from subprocess import PIPE
+
 import sys
 if sys.version_info > (3, ):
     long = int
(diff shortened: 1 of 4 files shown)


dwhswenson:
I'll leave this where it is; looks like we might need some more work to get things working on Python 3.10/3.11.

This is ready for review and comment. Without review, I will merge this sometime after Sat 16 Sep 15:00 GMT (10:00 my local).

openpathsampling/engines/external_engine.py
```diff
@@ -257,9 +259,10 @@ def start(self, snapshot=None):
         self.start_time = time.time()
         try:
             logger.info(self.engine_command())
-            # TODO: add the ability to have handlers for stdin and stdout
             self.proc = psutil.Popen(shlex.split(self.engine_command()),
-                                     preexec_fn=os.setsid)
+                                     preexec_fn=os.setsid,
+                                     stdout=PIPE,
+                                     stderr=PIPE)
```
sroet, line 265: Is this actually needed? I don't see these pipes used anywhere?

sroet (approved):
1 question, LGTM otherwise. Feel free to merge on green.

openpathsampling/engines/external_engine.py
dwhswenson: This is for the `_communicate()` method. It makes it so that you can get stdout/stderr -- this was somewhat useful in debugging (told me that I wasn't generating the tpr, so I knew to look at the `grompp` command). But it seemed like `communicate` is blocking (so we just run the full length trajectory, never checking if we should kill it). So `PIPE`/`communicate` are probably *only* useful when debugging.

PR #1133, opened 2023-09-15
Support Python 3.10; Python 3.11

dwhswenson:
This is a follow-up on #1132, and should only be merged after that one.

This PR will bring OPS tests in line with the [NEP29](https://numpy.org/neps/nep-0029-deprecation_policy.html) guidance on Python version support: namely, we will drop anything before 3.9, and support everything after 3.9 that is also supported by released versions of our major dependencies (typically involved waiting on support in the OpenMM ecosystem to filter through).

To-do:

- [x] Test on Python 3.10
- [x] Test on Python 3.11

Due to major changes in opcodes in Python 3.11, we're using SimStore's `CallableCodec` to serialize callables in Python 3.11, even in netcdfplus storage. This has the side effect that we needed to change the `CallableCodec`'s output to be base64-encoded so that it was serializable by `ujson` (SimStore's JSON codec ecosystem knows how to store raw bytes; `ujson` does not.) This means that SimStore storages will change, although we should still be able to load the old objects. It also means that `dill` is now a strict requirement for OPS (technically only on Python 3.11, but we'll include it in packaging everywhere).

A few extra to-do's as a result:

* ~Add warning that on Python 3.11 you should probably switch to SimStore~ [I decided against this. It has always been the case that you could use different code paths for serialization in different Python versions. This is just a *much* different code path. I'll probably add `FutureWarning` on `Storage` after the next release.]
- [x] Add warning when loading old SimStore storages that the serialization has changed
- [x] Add test to ensure we can load old SimStore storages

### DIFF
--- a/examples/ipynbtests.sh
+++ b/examples/ipynbtests.sh
@@ -37,7 +37,6 @@ case $PYTHON_VERSION in
     "3.11")
         mstis="$dropbox_base_url2/c4idtymcaqvtftigce49c/toy_mstis_1k_OPS1_py311.nc?rlkey=soa6ba7okx66439egjyuscc5l&dl=1"
         mistis="$dropbox_base_url2/s0o7br93s60q0vez88phr/toy_mistis_1k_OPS1_py311.nc?rlkey=22vdvjovt25bj4a6gh3m4vrvj&dl=1"
-        unzip="true"
         ;;
     *)
         echo "Unsupported Python version: $PYTHON_VERSION"
--- a/examples/ipynbtests.sh
+++ b/examples/ipynbtests.sh
@@ -33,6 +33,10 @@ case $PYTHON_VERSION in
         mstis=$dropbox_base_url/8rr0tt25xlm47cs/toy_mstis_1k_OPS1_py38.nc
         mistis=$dropbox_base_url/r3d5s5txbnpste0/toy_mistis_1k_OPS1_py38.nc
         ;;
+    "3.11")
+        mstis=$dropbox_base_url/8rr0tt25xlm47cs/toy_mstis_1k_OPS1_py38.nc
+        mistis=$dropbox_base_url/r3d5s5txbnpste0/toy_mistis_1k_OPS1_py38.nc
+        ;;
     *)
         echo "Unsupported Python version: $PYTHON_VERSION"
 esac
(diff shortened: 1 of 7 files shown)


dwhswenson:
This is ready for review and comment. Without review, I will merge this after at least 24 hours, no sooner than Mon 18 Sep 20:00 GMT (15:00 my local).

examples/ipynbtests.sh
```diff
@@ -29,6 +30,14 @@ case $PYTHON_VERSION in
         mstis=$dropbox_base_url/8rr0tt25xlm47cs/toy_mstis_1k_OPS1_py38.nc
         mistis=$dropbox_base_url/r3d5s5txbnpste0/toy_mistis_1k_OPS1_py38.nc
         ;;
+    "3.10")
+        mstis=$dropbox_base_url/8rr0tt25xlm47cs/toy_mstis_1k_OPS1_py38.nc
+        mistis=$dropbox_base_url/r3d5s5txbnpste0/toy_mistis_1k_OPS1_py38.nc
+        ;;
+    "3.11")
+        mstis="$dropbox_base_url2/c4idtymcaqvtftigce49c/toy_mstis_1k_OPS1_py311.nc?rlkey=soa6ba7okx66439egjyuscc5l&dl=1"
+        mistis="$dropbox_base_url2/s0o7br93s60q0vez88phr/toy_mistis_1k_OPS1_py311.nc?rlkey=22vdvjovt25bj4a6gh3m4vrvj&dl=1"
```
sroet, line 39: Why are these strings while the previous ones aren't?

sroet (approved):
These changes looks reasonable, couple questions. I feel unsure about checking the new encode/decode paths, so will have to trust you and the tests for that. LGTM otherwise

setup.cfg
```diff
@@ -27,6 +26,7 @@ classifiers =
 
 [options]
 include_package_data = True
+python_requires = >3.7
```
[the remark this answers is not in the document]
dwhswenson: Probably should be `>=3.7`, actually. The thought here is that we haven't actually dropped Python 3.7 support quite yet, although we're not testing on it. Since there's an opcode change between 3.7 and 3.8, I want to stretch the window just a little longer for anyone who started on 3.7. (Even after opcode changes, you can reload old data with safemode, but that's a hassle.) This is especially important because we had sometimes been slow to officially support new versions of Python.

I'm hoping to get to a 1.6 release in the next few weeks, and with that off my shoulders, more frequent releases should be easier. I'd like this to be `>=3.9` by the end of the year, and then to actually follow the NEP29 drop schedule.

examples/ipynbtests.sh
dwhswenson: Dropbox changed the link format. It used to be this nice simple URL with no GET parameters, and could be treated as barewords. Now we have GET parameters, which include `=`. This confuses `bash`, which tries to treat it as assignment. Unfortunately, the `rlkey` parameter is absolutely required for access.

PR #1136, opened 2023-10-17
Pin nbval != 0.10.0

dwhswenson:
I'm not sure why we're just now picking up nbval 0.10.0 (released in January), but it seems to be causing tests to fail.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -52,7 +52,7 @@ test =
     pytest
     pytest-cov
     coveralls
-    nbval
+    nbval!=0.10.0
 simstore =
     sqlalchemy>=1.4.1
 


dwhswenson:
Launched the PR; forgot to check if it fixed tests. 😂

This is ready for review and comment. Without review, I will merge this after at least 18 hours, no sooner than Wed 18 Oct 22:00 GMT (17:00 my local).

sroet (approved):
Change seems to work and test pass. LGTM feel free to merge

dwhswenson:
For reference, I think the relevant nbval issue is https://github.com/computationalmodelling/nbval/issues/194. It looks like an nbval dev doesn't think this is an intentional change, so I'm assuming it will get fixed someday; this is why I'm pinning instead of changing the notebook.

PR #1137, opened 2024-01-02
Fix engine props for OpenMM 8.1

dwhswenson:
In OpenMM 8.1, it seems that we must pass `None` for engine properties. The empty dict is no longer treated the same.


dwhswenson:
This seems to fix the failing tests. This is ready for review and comment. Without review, I will merge this after at least 18 hours, no sooner than Wed 03 Jan 17:00 GMT (11:00 my local).

PR #1142, opened 2024-01-28
Fix a few more setup -> setup_method tests

dwhswenson:
In #1123, most of our setup/teardown methods were fixed to use pytest's preferred (and now required) `setup_method`/`teardown_method` nomenclature. This fixes a few more.

---

**UPDATE**: This fixes 3 problems caused by upstream breaking changes:

1. As mentioned above, pytest now requires `setup_method` / `teardown_method` instead of `setup` and `teardown`.
2. Newer versions of pytest only work with nbval 0.10.0 or later. nbval 0.10.0 introduced what I consider to be a bug, which is that return values that you hide in-notebook with a `;` are seen by nbval as output. This required using `NBVAL_IGNORE` in some notebooks.
3. pytest also broke XUnit-style tests that include staticmethods as tests (I consider this to also be a bug). The fix was to just make them instance methods.

### DIFF
--- a/examples/tests/test_netcdfplus.ipynb
+++ b/examples/tests/test_netcdfplus.ipynb
@@ -235,6 +235,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "# NBVAL_IGNORE_OUTPUT\n",
     "st.nodes.save(Node(10));"
    ]
   },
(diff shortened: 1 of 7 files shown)


dwhswenson:
Current failure here is because the newer pytest requires nbval 0.10.0, which we pin to avoid against because it broke notebook tests (outputs changed). This will take a little work to resolve.

dwhswenson:
Finally getting CI to green again! See top post for updated contents of this PR. This is ready for review and comment. Without review, I will merge this after at least 18 hours, no sooner than Tue 23 Apr 19:00 GMT (14:00 my local).

sroet (approved):
I assume the ipynb changes outside of `"# NBVAL_IGNORE_OUTPUT` are just some other update diffs? LGTM otherwise

dwhswenson:
Yeah, the other notebook changes were just "rerun the notebook with updated software and maybe that will fix it?" before actually having to look in detail at the problem.

PR #1145, opened 2024-04-26
Test against Python 3.12

dwhswenson:
Let's see how much of Python 3.12 we can support... we might need to only support the `MINIMAL` build in 3.12, because I think some optional dependencies (like OpenMMTools) still don't have a 3.12 build. Currently, this tries to bring us up to SPEC0/NEP29 Python versions (3.10 and above), although I may keep 3.9 around if we can't build on 3.12.

This also cleans up a bunch of old devtools/CI things. I had the habit of leaving around old approaches so I wouldn't have to dig through git to switch back if I found my new approach didn't work. A few things this removes:

* `after_success.sh`: Replaced by an actions workflow that builds docs when we merge to `master`.
* `pytests.sh`, `.travis.yml.disabled`: specific to Travis; there's no way we're going back there
* `miniconda_install.sh`: We now use actions maintained by others to install conda/mamba
* `minimal.txt`, `minimal_testing.txt`, `experimental_reqs.txt`: These were requirements.txt-style files to get our requirements. They've been replaced by a script that extracts requirements from the setup.cfg, so that setup.cfg can be a single source of truth (although `experimental_reqs` was used until this PR).

I'm a little concerned about removing `miniconda_install.sh`, just because I'm afraid I may have, at some point, recommended using this repo as a single source of truth on that file. There may be CI somewhere that is still using it. If so, hopefully anyone who needs it can vendor an old copy of it into their repos, or, as we have, update to better approaches provisioning conda/mamba.


dwhswenson:
Huh, looks like:

1. We *were* using the `minimal*.txt` files in CI. Fixed that so that we *do* have a proper single source of truth.
2. PyEMMA has been abandoned, and was the only thing that didn't build on 3.12. I removed it from our test matrix, but (as with MSMBuilder) will leave the integration in place for users who might be using it. (OpenMMTools apparently runs on 3.12, but I know they can't test there, since they still have some `nose` in their tests!)
3. Also, apparently MDTraj builds on Linux and Py3.12, but not yet Mac. Which made it a bit of a headache for me to generate the new test data files needed for Py3.12 changes in serialization.

Anyway, now this is done! I plan to follow it up pretty quickly with a 1.6.1 release.

This is ready for review and comment. Without review, I will merge this after at least 24 hours, no sooner than Mon 29 Apr 21:00 GMT (16:00 my local).

devtools/conda_install_reqs.sh
```diff
@@ -36,7 +36,7 @@ WORKAROUNDS=""
 REQUIREMENTS=`python ${DEVTOOLS_DIR}/setup_cfg_reqs.py`
 TESTING=`python ${DEVTOOLS_DIR}/setup_cfg_reqs.py --extra test`
 INTEGRATIONS=`cat ${DEVTOOLS_DIR}/tested_integrations.txt | tr "\n" " "`
-EXPERIMENTAL=`cat ${DEVTOOLS_DIR}/experimental_reqs.txt | tr "\n" " "`
+EXPERIMENTAL=$(python ${DEVTOOLS_DIR}/setup_cfg_reqs.py --extra simstore)
```
sroet, line 39: Hardcoding the name of the extra in this file feels wrong.
For the current state of the code I understand that this would work, but for future maintainability it might be nice to rename the `extra` `simstore` to `experimental` and don't give users the options to choose which of the experimental feature dependencies they install, just make it all or nothing

Or we could make a catch all `extra` using [this SO answer](https://stackoverflow.com/a/70983864)

PR #1141, opened 2024-01-14
Cut off `nose`

dwhswenson:
Resolves #1140, and will allow us to enable Python 3.12 testing (at least in minimal testing; unsure whether some upstreams are caught up yet).

This is a decent bit of tedious work (which is why it hasn't been done yet.) I can facilitate it with some vim macros, but there will be edge cases, so results from macros need to be manually validated. My guess is that I'll do it in the following steps:

- [x] Replace `assert_equal(a, b)` with `assert a == b`
- [x] Replace `assert_not_equal(a, b)` with `assert a != b`
- [x] Replace `raise SkipTest()` with `pytest.skip()`
- [x] Replace `@raises(...)` with `with pytest.raises(...)`
- [x] Replace `assert_almost_equal`, probably with things from `numpy.testing`
- [x] Replace various methods in `test_helpers` that are built on `nose`
- [x] Clean up any remaining bits that aren't included above
- [x] Remove `nose` from the build process

Adding Python 3.12 to the CI matrix will come in a later PR.

PR #1146, opened 2024-04-29
OpenPathSampling 1.6.1

dwhswenson:
OpenPathSampling 1.6.1 is a minor update, primarily focused on allowing OPS to run/be tested on Python 3.12. In practice, that means removing dependency on tools that prevented Python 3.12 support. In particular, this removes `nose` from our testing requirements, and deprecates our integration with PyEMMA, which is no longer maintained.

**DEPRECATION:** As of OPS 1.6.1, we no longer recommend the use of `PyEMMAFeaturizerCV`. PyEMMA is no longer maintained. Please do not create collective variables based on PyEMMA.

## Miscellaneous improvements

* Test against Python 3.12; deprecate PyEMMA ([#1145](https://github.com/openpathsampling/openpathsampling/pull/1145) #Deprecations)
* Cut off `nose` ([#1141](https://github.com/openpathsampling/openpathsampling/pull/1141) #upstream_changed)
* Fix a few more setup -> setup_method tests ([#1142](https://github.com/openpathsampling/openpathsampling/pull/1142) #upstream_changed)


setup.cfg
```diff
@@ -55,6 +55,3 @@ test =
     nbval>=0.10.0
 simstore =
     # simstore requirements have been moved to main
-
-[bdist_wheel]
-universal = 1
```
sroet, line 61: Why was this removed?

sroet (approved):
(I only checked the last 4 commits: 878620a - 3584801)
One question, not a blocker. LGTM otherwise

setup.cfg
dwhswenson: Turns out I'd misremembered what `bdist_wheel.universal` does.  It isn't the equivalent of noarch; it is noarch + compatible with Python 2 and 3. We're no longer compatible with Python 2. With `universal = 1`, wheels claim to be Python 2 compatible, even if `python_requires` is >3 (e.g., https://pypi.org/project/openpathsampling/1.6.0/#files)

I probably knew that when I first added that to a project, but kinda forgot.

PR #1148, opened 2024-04-29
Update 2.0 after release of 1.6.1

dwhswenson:
Update `dev-2.0` with current status of `master`.

I was hoping there would be a way to make this a reviewable diff, but no. Even the merge commit, https://github.com/openpathsampling/openpathsampling/commit/35327d3749e0cdfa4b15b707e53cd7e514664867,  where I fixed merge conflicts, is probably too much to review. I will mention that while fixing merge conflicts, I went ahead and removed the `PyEMMAFeaturizerCV` (which is deprecated and slated for removal to 2.0).


dwhswenson:
This is ready for review and comment. Without review, I will merge this after at least 24 hours, no sooner than Wed 01 May 02:00 GMT (Tue 30 Apr 21:00 my local).

@sroet, review of the last 3 commits isn't too tedious. I hoped that the merge commit would have been possible, but 🤷‍♂️. The change on CI scripts was that I ran `diff` against the normal CI script and kept the stuff I considered (potentially) relevant.

sroet (approved):
Looking at the final 3 commits ( 88a8c0d - bbf0cab ); those look reasonable. 
I will trust the rest of the lines (and the passing of the tests). Feel free to merge

PR #1149, opened 2024-07-04
Update Python version in test-openmm-rc

dwhswenson:
Our script to test OpenMM RCs was pinned to Python 3.9, which we no longer support. Looks like there's another OpenMM RC out, so it is time to update this!!


dwhswenson:
This is ready for review and comment. Without review, I will merge this after at least 8 hours, no sooner than Thu 04 Jul 22:00 GMT (18:00 my local).

(Short merge window because this is a trivial change and I'd like to get this in so that the next regular RC test runs in time to catch any issues.)

sroet (approved):
LGTM!

dwhswenson:
Thanks for the quick review! Merging.

sroet:
> Thanks for the quick review! Merging.

You caught me when I didn't have anything to do on a train

PR #1150, opened 2024-07-17
Fix histogram xvals

dwhswenson:
Turns out that a line of old uncovered code hadn't been updated for Python 3! Fixed the line, added test coverage.


dwhswenson:
This is ready for review and comment. Without review, I will merge this after at least 18 hours, no sooner than Thu 18 Jul 17:00 GMT (12:00 my local).

PR #1151, opened 2024-07-20
Volume shooting point selector

dwhswenson:
This adds a volume-based shooting point selector. It will only select points from a given volume.

(I thought we already had this? I thought this was implemented as part of "Shooting Range" by @hejung in one of our ESDWs? Guess that code never made it into main. Not too hard to recreate this much of it, though.)

### DIFF
--- a/openpathsampling/shooting.py
+++ b/openpathsampling/shooting.py
@@ -271,3 +271,12 @@ class FirstFrameSelector(ShootingPointSelector):
         if new_snapshot is None:
             NEW_SNAPSHOT_SELECTOR.warn(stacklevel=3)
         return 1.0
+
+
+
+class VolumeShootingSelector(ShootingPointSelector):
+    def __init__(self, volume):
+        self.volume = volume
+
+    def f(self, frame, trajectory):
+        return float(self.volume(frame))
--- a/openpathsampling/shooting.py
+++ b/openpathsampling/shooting.py
@@ -273,7 +273,6 @@ class FirstFrameSelector(ShootingPointSelector):
         return 1.0
 
 
-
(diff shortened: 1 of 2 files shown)


dwhswenson:
This is ready for review and comment. Without review, I will merge this after at least 24 hours, no sooner than Mon 22 Jul 20:00 GMT (15:00 my local).

openpathsampling/shooting.py
```diff
@@ -271,3 +271,11 @@ def probability_ratio(self, snapshot, old_trajectory,
...
             NEW_SNAPSHOT_SELECTOR.warn(stacklevel=3)
         return 1.0
+
+
+class VolumeSelector(ShootingPointSelector):
+    def __init__(self, volume):
+        self.volume = volume
+
+    def f(self, frame, trajectory):
+        return float(self.volume(frame))
```
sroet, line 281: any particular reason you went for `float` instead of `int` casting?
(I understand it makes sense from a probablility point of view, but not really from the point of view that the output of volume should be `bool`)

sroet (approved):
one question about your reasoning, looks good either way

openpathsampling/shooting.py
dwhswenson: I'm thinking of it as a special case of a shooting point biasing function, like `BiasedSelector`. In that case you bias toward certain shooting points using a CV, which returns floats. Really, the actual value that comes out here doesn't matter, as long as it is either 0 or some constant $k$ (I took $k=1$).

FWIW, even `UniformSelector` technically returns `1.0` for its `f`. So this also keeps the typing consistent across different selectors.

PR #1154, opened 2024-08-12
Custom `PathDensityHistogram._add_ops_trajectory`

dwhswenson:
This adds a touch point for path density histogram subclasses to change what occurs when adding an OPS trajectory to a PathDensityHistogram.

I needed this when playing with the idea of having a custom histogram that would also track all trajectories that passed through a given histogram bin.

This doesn't break the API, and doesn't even need changes in the tests.


dwhswenson:
This is ready for review and comment. Without review, I will merge this after at least 24 hours, no sooner than Wed 14 Aug 03:00 GMT (Tue 13 Aug 23:00 my local).

Issue #1080, opened 2021-10-06
Hashing, Equality, and UUIDs in OPS 2.0


dwhswenson:
One of the major changes in 2.0 should be in the way objects in OPS are hashed. Although this is the sort of thing that seems like a subtle point that only programmers care about, we've seen situations where it causes headaches for real users.

NOTE: This supercedes issue #701. See some initial thoughts there; we didn't make changes in 1.x for backward compatibility, but this is a significant issue in real use that we should address in 2.0.

### Equality and hashing

Just as a reminder, we have 3 types of "equality" that we can use in Python:

1. `is`-equality: Same object in memory, `obj1 is obj2`. No way to override.
2. `hash`-equality: `hash(obj1) == hash(obj2)`, defined by `__hash__`, and used in for performance in dicts, etc. Must be immutable over object lifetime.
3. `==`-equality: `obj1 == obj2`, defined by `__eq__`

These are in decreasing order of strictness, i.e.

```
(obj1 is obj2) ==> (obj1 == obj2) ==> (hash(obj1) == hash(obj2))
```

where `==>` is "implies" and hash is only relevant is the objects are hashable. Two objects that compare equal must have the same hash. (Obviously, a specific object in memory must have the same hash as itself and must be equal to itself.) However, two objects can compare equal or hash to the same value without being the same object in memory. More subtle, two objects can hash to the same value without comparing equal.

### How hashes are used

Hashes are used to speed up `dict`/`set` lookup in Python, but after a candidate is found, a further check with `==` is also done. This means that two objects with the same `hash` can simultaneously be keys in a `dict` if and only if they are not `==`.

The `__hash__` method should return an integer, usually accomplished by wrapping other hashables in `hash`. Importantly, the implementation of `hash` is an implementation detail, and may not be preserved across different Python implementations or versions of Python.

Further reading: [this gist of mine](https://gist.github.com/dwhswenson/50e435ed67c4b3cd6baee75a8ececdfb) and [this article](https://hynek.me/articles/hashes-and-equality/).

### We violate the Python contract on hashes and equality

In OPS, we sometimes override the `__eq__` operator while continuing to use the UUID as the `hash`. This means that we have objects that compare equal but give different hashes. This becomes a problem if you have objects that are `==` and that you want to put into a hashed container such as a `set` or a `dict`.

### Hashing by UUID causes pain for analysis

[...]

PR #1155, opened 2024-08-13
Let UUID determine equality for ensembles

dwhswenson:
Long ago, we tried to use the string representations of ensembles as the comparison for equality. This was a decent idea, although there are some issues with the ways we handle equality and hashing.

I recently came across a weird edge case where a periodic volume with an edge exactly at -pi was serialized in 2 different ways: one with the edge at -pi and one with the edge at +pi. I'd still like to look into what happened to make that happen, but this would fix the most obvious effect (which is that the resulting ensembles weren't equal). In any case, this is a good shortcut: the same UUID indicates the same object, assuming we can trust the meaning of UUID.

### DIFF
--- a/openpathsampling/ensemble.py
+++ b/openpathsampling/ensemble.py
@@ -253,7 +253,10 @@ class Ensemble(with_metaclass(abc.ABCMeta, StorableNamedObject)):
     def __eq__(self, other):
         if self is other:
             return True
-        if self.__uuid__ == other.__uuid__:
+        if (
+            self.__class__ == other.__class__
+            and self.__uuid__ == other.__uuid__
+        ):
             return True
 
         return str(self) == str(other)
--- a/openpathsampling/ensemble.py
+++ b/openpathsampling/ensemble.py
@@ -253,6 +253,9 @@ class Ensemble(with_metaclass(abc.ABCMeta, StorableNamedObject)):
     def __eq__(self, other):
         if self is other:
             return True
+        if self.__uuid__ == other.__uuid__:
+            return True
+
         return str(self) == str(other)
 
     def __ne__(self, other):


dwhswenson:
This is ready for review and comment. Without review, I will merge this after at least 24 hours, no sooner than Wed 14 Aug 05:00 GMT (01:00 my local).

openpathsampling/ensemble.py
```diff
@@ -253,6 +253,12 @@ def __init__(self):
     def __eq__(self, other):
         if self is other:
             return True
+        if (
+            self.__class__ == other.__class__
+            and self.__uuid__ == other.__uuid__
+        ):
+            return True
+
         return str(self) == str(other)
```
sroet, line 262: Do you actually still want to check the string representation if the UUID fails?

sroet (changes requested):
1 question, LGTM otherwise

openpathsampling/ensemble.py
dwhswenson: Backward compatibility.

This is something that needs to change in OPS 2.0. See https://github.com/openpathsampling/openpathsampling/issues/1080. (Proposal 2 in that issue is similar to what we did with [`gufe`](https://github.com/OpenFreeEnergy/gufe), more docs [here](https://gufe.readthedocs.io/en/latest/guide/serialization.html). I've been keeping an eye on that to see what headaches it causes; there's some conceptual weirdness in the flyweight model, so I'm waiting to see if users hate it.)

dwhswenson: Also, there's this, which would break:

```python
In [1]: import openpathsampling as paths

In [2]: e1 = paths.EmptyEnsemble()

In [3]: e2 = paths.EmptyEnsemble()

In [4]: e1 == e2  # should be True!
Out[4]: True

In [5]: e1.__uuid__ == e2.__uuid__
Out[5]: False
```

sroet (approved):

PR #1157, opened 2024-08-30
SimStore performance improvements and fixes

dwhswenson:
This includes several performance improvements and minor fixes for SimStore. Quick manifest:

* Added and started using `is_storage_string` and `has_storage_uuid`, which use the fast `ClassIsSomething` model to cache classes with given attributes. The normal `is_string` and `has_uuid` methods were taking a larger fraction of time than they should.
* Added default support for SQL index on UUID (which should have been there all along).
* Fix an edge case where old-style CVs were no reloaded correctly in SimStore, because the new JSON serialization was already making a callable and the code to make a callable from the marshal data was no longer needed.

This also helps clear up some of the external testing in ops-storage-notebooks which has been off for a while.

### DIFF
--- a/openpathsampling/experimental/simstore/serialization_helpers.py
+++ b/openpathsampling/experimental/simstore/serialization_helpers.py
@@ -14,7 +14,7 @@ from .class_lookup import (
 )
 from .proxy import GenericLazyLoader
 from .uuids import (
-    has_uuid, get_uuid, set_uuid, encode_uuid, decode_uuid, encoded_uuid_re,
+    get_uuid, set_uuid, encode_uuid, decode_uuid, encoded_uuid_re,
     is_uuid_string
 )
 
--- a/openpathsampling/experimental/simstore/serialization_helpers.py
+++ b/openpathsampling/experimental/simstore/serialization_helpers.py
@@ -8,7 +8,10 @@ import collections
 from .tools import flatten_all, nested_update, group_by_function
 from .tools import is_iterable, is_mappable, is_numpy_iterable
 from . import tools
-from .class_lookup import is_storage_iterable, is_storage_mappable
+from .class_lookup import (
+    is_storage_iterable, is_storage_mappable, is_storage_string,
+    has_storage_uuid,
+)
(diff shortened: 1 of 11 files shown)


openpathsampling/experimental/simstore/serialization_helpers.py
```diff
@@ -8,7 +8,10 @@
...
 from .tools import is_iterable, is_mappable, is_numpy_iterable
 from . import tools
-from .class_lookup import is_storage_iterable, is_storage_mappable
+from .class_lookup import (
+    is_storage_iterable, is_storage_mappable, is_storage_string,
+    has_storage_uuid,
+)
 from .proxy import GenericLazyLoader
 from .uuids import (
     has_uuid, get_uuid, set_uuid, encode_uuid, decode_uuid, encoded_uuid_re,
```
sroet, line 17: shouldn't `has_uuid` be dropped from the imports? (only reference is in commented out code)
Bit surprised linting did not catch this

openpathsampling/experimental/simstore/sql_backend.py
```diff
@@ -35,7 +35,7 @@
 }
 
 universal_sql_meta = {
-    'uuid': {'uuid': {'primary_key': True}},
+    'uuid': {'uuid': {'primary_key': True, 'index': True}},
```
sroet, line 38: :partying_face:

sroet (approved):
One import can be dropped, feel free to ignore. LGTM otherwise

openpathsampling/experimental/simstore/serialization_helpers.py
dwhswenson: Turns out this was trickier than it appeared. In an early version of the file, `has_uuid` was defined in `serialization_helpers.py`. Then I split it off into `uuids.py`. But I allowed the transitive import.

Fixed that in [8bf1f31](https://github.com/openpathsampling/openpathsampling/pull/1157/commits/8bf1f31493719f0d55537bdea987796cfa282f23).

PR #1158, opened 2024-09-05
Storage Interfaces

dwhswenson:
These are intended to abstract out filesystem storage to be an arbitrary key-value store. As examples, I've implemented storage on the filesystem and in memory. Object storage (e.g., S3) can be implemented separately.

I'm making this a separate PR so that it can be used in future PRs.

I'd also appreciate suggestions on a better name! `StorageHandler` is kind of awful. [UPDATE: Changed to `StorageInterface`]


dwhswenson:
This is ready for review and comment. Without review, I will merge this after at least 24 hours, no sooner than Fri 06 Sep 15:00 GMT (17:00 my local).

openpathsampling/utils/storage_handlers.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def load(self, storage_label, target_path):
+        target_path.parent.mkdir(parents=True, exist_ok=True)
+        local_path = self.root / storage_label
+        _logger.debug("Copying file from {str(local_path)} "
+                      f"to {str(target_path)}")
+        shutil.copyfile(local_path, target_path)
+
+    def delete(self, storage_label):
+        os.remove(self.root / storage_label)
```
sroet, line 69: `self.transfer` can deal with directories, this does not, which functionality is intended?

openpathsampling/utils/storage_handlers.py
```diff
@@ -0,0 +1,108 @@
...
+        local_path = self.root / storage_label
+        local_path.parent.mkdir(parents=True, exist_ok=True)
+        shutil.copyfile(source_path, self.root / storage_label)
+
+    def load(self, storage_label, target_path):
+        target_path.parent.mkdir(parents=True, exist_ok=True)
+        local_path = self.root / storage_label
+        _logger.debug("Copying file from {str(local_path)} "
+                      f"to {str(target_path)}")
+        shutil.copyfile(local_path, target_path)
```
sroet, line 66: this could use docs, specifically of the difference between load and store (load seems to do something I would not expect from the name)

openpathsampling/utils/storage_handlers.py
```diff
@@ -0,0 +1,108 @@
...
+        self._data = {}
+
+    def store(self, storage_label, source_path):
+        with open(source_path, mode='rb') as f:
+            self._data[storage_label] = f.read()
+
+    def load(self, storage_label, target_path):
+        with open(target_path, mode='wb') as f:
+            f.write(self._data[storage_label])
+        return self._data[storage_label]
```
sroet, line 94: this could use docs, specifically on the difference between load and store

openpathsampling/utils/storage_handlers.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def transfer(self, storage_label, source_path):
+        """Transfer a file to the storage label from the source path.
+
+        In some cases, this can be made faster than store followed by
+        os.remove, so this method can be overridden. (Example: moving ona
+        file system is faster than copying.)
+        """
+        self.store(storage_label, source_path)
+        os.remove(source_path)
```
sroet, line 42: this can't deal with directories

sroet (changes requested):
There are:
-  missing tests (mainly `__contains__` and `list_directory` calls)
- a typo
- Some unnecessary TemporaryDirectory magic
- Some missing docs on what `store` and `load` are intended to do (especially `load` behavior was unexpected)
- Strange layout of the base test class vs inheritance
- Some inconsistency if directories are handled or not

> I'd also appreciate suggestions on a better name! StorageHandler is kind of awful.

ChatGPT came up with (no strong preference from me):
> StorageManager
StorageHandler
AbstractStorage
StorageController
StorageService
DataStorageInterface
StorageProvider
StorageFacade
StorageGateway
StorageAbstraction

dwhswenson (commented):
Thanks for the useful review! I made a few significant changes:

* Now nothing should work with directories. That's a little harder for wrapping an object store/key-value model, which is the goal here.
* Decided that I like `StorageInterface` more; changed some names
* Overall, better structure of ABCs with significantly better docs, although I count on inheritance when the subclasses have nothing to add to the doctoring.

Please take another look when you can!

sroet (approved):
One possible reuse of a variable, feel free to ignore.
LGTM otherwise

PR #1161, opened 2024-10-04
Bump 1.7.1.dev0

dwhswenson:
Bump to master after release.


dwhswenson:
Version bump so CI doesn't fail (expecting local version definition to be bumped from release). IMO, no review necessary. Merging, and then let's move toward OPS 1.8!

PR #1163, opened 2024-10-15
Update to use setup-miniconda v3; use miniforge

dwhswenson:
Mambaforge has been deprecated. See https://conda-forge.org/news/2024/07/29/sunsetting-mambaforge/

Now miniforge is effectively identical to the old mambaforge. At the same time, I'm updating the version action we're using.


dwhswenson:
This passes, and affects essentially no code, so I'm merging this without review. (It will enable development on other branches.)

PR #1169, opened 2024-12-23
Update for SPEC0 as of Dec 2024

dwhswenson:


dwhswenson:
OpenMM seems to exist on PyPI with Python 3.13 support, but not on conda-forge. We'll either have to change our install to take OpenMM from PyPI, or wait a bit longer on this.

Until users need Python 3.13, I'll just wait.

sroet:
> OpenMM seems to exist on PyPI with Python 3.13 support, but not on conda-forge. We'll either have to change our install to take OpenMM from PyPI, or wait a bit longer on this.

That is strange as it should have been up since 12-12-2024:
Screenshot from 2025-01-06 17-00-03

sroet:
On a tangental note, since conda `23.10`, conda is using the `mamba` solver under the hood by default: https://conda.github.io/conda-libmamba-solver/user-guide/ so you can probably drop back to a default conda install for the tests

dwhswenson:
> That is strange as it should have been up since 12-12-2024:

That's MDTraj. 😉 Check for OpenMM; it is still missing 3.13.

sroet:
> > That is strange as it should have been up since 12-12-2024:
> 
> That's MDTraj. 😉 Check for OpenMM; it is still missing 3.13.

:sweat_smile: I see it is missing now. Brain is apparently still in xmas holiday mode.

sroet:
Openmm 8.2.0 is now available on conda-forge for python 3.13 😄

dwhswenson:
Looks like now we're waiting on numba (transitive dependency from openmmtools). They're actively working on that feedstock (as in, recent commit still running builds... 2+ hour build process on Windows!)

dwhswenson:
close-reopen cycle on this to get tests to run again

dwhswenson:
This is now failing with the same issue as #1173 (I was hoping maybe it was a Python 3.10-specific issue?). But that means it got through the installation phase. So; good news: we can probably merge this soon. Bad news: I have to figure out #1173 first!

PR #1170, opened 2024-12-23
Trajectory Exports: Core and integration in `DynamicsEngine`

dwhswenson:
One of the major things that came out of the path sampling software meeting in Leiden in September 2024 was a desire for OPS to export trajectories more easily into formats that people are more familiar with.

In this PR, I'm starting to add functionality for external trajectory export. The eventual goal is that you'll be able to export a trajectory using `engine.export_trajectory(trajectory, filename)`. That method will take an optional `writer` parameter that takes a callable to perform the actual output. This allows a single unified interface from the user perspective, but allows individual callable (as, e.g., instances of a class) to handle details of implementation; for example, allowing lossy output while defaulting to lossless. Different engines can also have different default trajectory writers.

This PR will add support for this in the base `DynamicsEngine` class, where the `_default_trajectory_writer` will return a `SimStoreTrajectoryWriter`, since all engines should be able to export in that format. A future PR will add better support for OpenMM and Gromacs engines, as well as lossy output using MDTraj.

- [x] `SimStoreTrajectoryWriter`
- [x] Tests for `SimStoreTrajectoryWriter`
- [x] `DynamicsEngine.export_trajectory`
- [x] Tests for `DynamicsEngine.export_trajectory`
- [x] Add to developer docs to explain this to engine developers.

Overall plan:
1. **This PR**: Add support for core functionality and defaults for SimStore exports.
2. Upcoming PR: Add support for TRR export for Gromacs and OpenMM engines (as defaults for those engines). Add support for MDTraj-based export for lossy exports.
3. Later PR: Add support to export a list of `MCStep`s, using a given `TrajectoryWriter` and some definition of where to store step information.
4. Later PR: Add new type of storage (based on SimStore) that uses lossless external trajectory storage for snapshot data storage.

### DIFF
--- a/docs/developers/engines.rst
+++ b/docs/developers/engines.rst
@@ -29,4 +29,5 @@ implement engines that do not only deal with standard molecular dynamics.
     snapshot_features
     engines_direct_api
     engines_indirect_api
+    engines_trajectory_export
     advanced_engines
--- a/docs/developers/engines_trajectory_export.rst
+++ b/docs/developers/engines_trajectory_export.rst
@@ -1,0 +1,24 @@
+Trajectory Export from Engines
+==============================
+
+Users frequently want trajectories to be exported from Python into format
+that they can analyze with other tools. Since the expected trajectory format
+can differ between engines, a developer can set a default trajectory writer
+using the :meth:`._default_trajectory_writer` method. This method should
+return a callable that takes a trajectory, a filename, and the boolean flag
+``force``, where the ``force`` flag forces overwrite of an existing file (a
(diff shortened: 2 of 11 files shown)


dwhswenson:
This is ready for review and comment. Without review, I will merge this after at least 48 hours, no sooner than Fri 27 Dec 14:00 GMT (09:00 my local).

Note to reviewers: There are also bunch of changes here that involve either loading `FunctionCV` from `paths` at a later time, or calling `super()` instead of `super(ClassName, self)`. These are both related to some weird aspects of unpatching as part of the unit tests, but I'd also argue that both are better practices anyway.

PR #1172, opened 2025-01-04
Trajectory Exports: MDTraj and TRR formats; Gromacs and OpenMM support

dwhswenson:
Add support in OpenMM and Gromacs engines to export trajectories. This includes lossless export (as TRR) and lossy export (via MDTraj, where the export format will be determined from the filename). We use TRR format as the default lossless format for both OpenMM and Gromacs.

- [x] Lossless: TRRTrajectoryWriter
- [x] Tests for TRRTrajectoryWriter
- [x] Lossy: MDTrajTrajectoryWriter
- [x] Tests for MDTrajTrajectoryWriter
- [x] OpenMMEngine._default_trajectory_writer and tests for OpenMMEngine.export_trajectories
- [x] GromacsEngine._default_trajectory_writer and tests for GromacsEngine.export_trajectories

Overall plan:

* Add support for core functionality and defaults for SimStore exports. (#1170)
* **This PR**: Add support for TRR export for Gromacs and OpenMM engines (as defaults for those engines). Add support for MDTraj-based export for lossy exports.
* Later PR: Add support to export a list of MCSteps, using a given TrajectoryWriter and some definition of where to store step information.
* Later PR: Add new type of storage (based on SimStore) that uses lossless external trajectory storage for snapshot data storage.

### DIFF
--- a/openpathsampling/exports/trajectories/mdtrajtrajectorywriter.py
+++ b/openpathsampling/exports/trajectories/mdtrajtrajectorywriter.py
@@ -14,22 +14,23 @@ class MDTrajTrajectoryWriter(TrajectoryWriter):
             raise ImportError("MDTraj is not available")
 
         self.mdtraj_selection = mdtraj_selection
+        self._sel = None  # the actual atom indices of selection
 
     def _write(self, trajectory, filename):
         mdt = trajectory.to_mdtraj()
 
-        sel = None
-        if isinstance(self.mdtraj_selection, str):
-            sel = mdt.topology.select(self.mdtraj_selection)
-        elif isinstance(self.mdtraj_selection, Iterable):
-            sel = list(self.mdtraj_selection)
-        elif self.mdtraj_selection is None:
-            pass
-        else:
-            raise TypeError("mdtraj_selection must be a string or an "
-                            f"iterable; got '{self.mdtraj_selection}'")
+        if self._sel is None:
+            if self.mdtraj_selection is None:
+                pass
+            elif isinstance(self.mdtraj_selection, str):
+                self._sel = mdt.topology.select(self.mdtraj_selection)
+            elif isinstance(self.mdtraj_selection, Iterable):
+                self._sel = list(self.mdtraj_selection)
+            else:
+                raise TypeError("mdtraj_selection must be a string or an "
+                                f"iterable; got '{self.mdtraj_selection}'")
 
-        if sel is not None:
-            mdt = mdt.atom_slice(sel)
+        if self._sel is not None:
+            mdt = mdt.atom_slice(self._sel)
 
         mdt.save(filename)
--- a/openpathsampling/exports/trajectories/mdtrajtrajectorywriter.py
+++ b/openpathsampling/exports/trajectories/mdtrajtrajectorywriter.py
@@ -14,23 +14,22 @@ class MDTrajTrajectoryWriter(TrajectoryWriter):
             raise ImportError("MDTraj is not available")
 
         self.mdtraj_selection = mdtraj_selection
-        self._sel = None  # the actual atom indices of selection
 
     def _write(self, trajectory, filename):
         mdt = trajectory.to_mdtraj()
 
-        if self._sel is None:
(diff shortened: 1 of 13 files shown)


dwhswenson:
This is ready for review and comment. Without review, I will merge this after at least 24 hours, no sooner than Sun 05 Jan 01:00 GMT (Sat 04 Jan 19:00 my local).

The relevant test for Gromacs is currently skipped in CI (no Gromacs install), but works locally. That's the 1 line missing in the patch from CodeCov. I'm not going to set that to ignore, because the lack of Gromacs testing in CI is the result of some historical issues that have since been resolved. The better solution is to add Gromacs into the test environment in CI.

That's also in the context of me considering a major change to our CI setup in response to the development of better tooling for handling multiple conda environments for a project (`pixi`). Leaving that for future PRs, though.

openpathsampling/exports/trajectories/mdtrajtrajectorywriter.py
```diff
@@ -0,0 +1,35 @@
...
+            raise ImportError("MDTraj is not available")
+
+        self.mdtraj_selection = mdtraj_selection
+
+    def _write(self, trajectory, filename):
+        mdt = trajectory.to_mdtraj()
+
+        sel = None
+        if isinstance(self.mdtraj_selection, str):
+            sel = mdt.topology.select(self.mdtraj_selection)
```
sroet, line 23: just for my info: How expensive is this call and should we cache the result?

openpathsampling/exports/trajectories/trrtrajectorywriter.py
```diff
@@ -0,0 +1,26 @@
...
+from .core import TrajectoryWriter
+
+class TRRTrajectoryWriter(TrajectoryWriter):
+    """Write a trajectory to a Gromacs TRR file.
+
+    This inludes velocities, so it can be used for restarts.
+    """
+    def __init__(self):
+        if not HAS_MDTRAJ:  # -no-cov-
+            raise ImportError("MDTraj is not available")
```
sroet, line 12: I can guess your reason, but would like for it to be mentioned explicitly:
Any reason why you don't allow this writer to write out selections? (in case people want velocities, but not necessary all waters for analysis after the fact?)
(If this proposed functionality is in-scope, I would probably make it explicit by writing a separate class for it that emits significant warnings)

sroet (approved):
LGTM, 2 implementation questions, feel free to ignore

openpathsampling/exports/trajectories/mdtrajtrajectorywriter.py
dwhswenson: Moderately expensive, but I suspect we're still I/O bound here? Following this comment, I actually added caching ([f2e4897](https://github.com/openpathsampling/openpathsampling/pull/1172/commits/f2e4897c3b492dcd697a04d85393768949cc4a77)), and I realized I didn't like the result and reverted ([84c98d1](https://github.com/openpathsampling/openpathsampling/pull/1172/commits/84c98d11d6195ce664d49553570123d561076bdf)).

Let's say you create a writer with an MDTraj selection rule like `"not water and not ion"`. I can imagine that, as part of your analysis, you might apply that same writer to both the wild-type and mutant versions of a protein, and be very surprised that the selection criteria is totally wrong. That's a serious foot-gun.

Checking to prevent that involves a full check of MDTraj topologies, which I suspect is more expensive than using the MDTraj DSL.

openpathsampling/exports/trajectories/trrtrajectorywriter.py
dwhswenson: Yeah, this is supposed to be a lossless format. You could easily create a different subclass to allow MDTraj selections, too (or, now that MDAnalysis is LGPL, you could use it!)

In general, I think I'm going to aim to make lossless writers separate (maybe we'll even add a flag to them?) because we can use these as the "database source of truth" to replace storing everything in a single SimStore/NetCDFPlus file. Lossy formats should not be used for that, but do have potential value as trajectory exporters for external analysis.

PR #1173, opened 2025-05-11
Use `autorelease check` instead of custom script

dwhswenson:
We recently bumped up to setuptools 80 in our default builds. This is causing problems because of major changes in [how `setuptools` treats the `develop` command](https://setuptools.pypa.io/en/stable/history.html#v80-0-0).

In any case, instead of having a custom script that uses autorelease, it is better to use the `autorelease check` command, available since autorelease 0.3.0 (2021). That's what this PR will do.

- [x] See if using `autorelease check` fixes (otherwise, fix might be needed upstream in autorelease)
- [x] Remove old `devtools/autorelease_check.py`
- [x] Update autorelease for publishing, too


dwhswenson:
This is ready for review and comment. Without review, I will merge this after at least 24 hours, no sooner than Mon 12 May 21:00 GMT (16:00 my local).

sroet (approved):
LGTM!

PR #1180, opened 2026-03-02
OpenMMEngine: Persist `platform`; allow `openmm_properties` override in `initialize`

dwhswenson:
This allows OpenMM engines to persist their platform (if given to `__init__`). It also allow override of `openmm_properties` (and `platform`) in the `initialization()` method.

Resolves part of #1177, following the approach proposed at https://github.com/openpathsampling/openpathsampling/issues/1177#issuecomment-3942235556.

### DIFF
--- a/openpathsampling/engines/openmm/engine.py
+++ b/openpathsampling/engines/openmm/engine.py
@@ -242,19 +242,19 @@ class OpenMMEngine(DynamicsEngine):
 
     @staticmethod
     def _normalize_platform(platform):
-        if platform is None:
-            return None
-
-        if isinstance(platform, str):
-            return platform
-
-        try:
-            return platform.getName()
-        except (AttributeError, TypeError) as err:
-            raise TypeError(
-                "platform must be None, a platform name string, or an "
-                "openmm.Platform instance"
-            ) from err
+        match platform:
+            case None:
+                return None
+            case str() as name:
+                return name
+            case candidate if isinstance(candidate, openmm.Platform):
+                return candidate.getName()
+            case _:
+                raise TypeError(
+                    "platform must be None, a platform name string, or an "
+                    "openmm.Platform instance. "
+                    f"Got {type(platform)} instead"
+                )
 
     @property
     def simulation(self):
--- a/openpathsampling/engines/openmm/engine.py
+++ b/openpathsampling/engines/openmm/engine.py
@@ -79,7 +79,8 @@ class OpenMMEngine(DynamicsEngine):
             system,
             integrator,
             openmm_properties=None,
-            options=None):
+            options=None,
+            platform=None):
         """
         Parameters
         ----------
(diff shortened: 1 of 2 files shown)


dwhswenson:
Leaving this open for a final 48: Without review, I will merge this after at least 48 hours, no sooner than Thu 26 Mar 08:00 GMT (03:00 my local). If I understand, this resolves the primary concerns from @Bernadette-Mohr in #1177. Adding @sroet for review, if he has time.

openpathsampling/engines/openmm/engine.py
```diff
@@ -220,14 +227,35 @@ def from_new_options(
...
+        if isinstance(platform, str):
+            return platform
+
+        try:
+            return platform.getName()
+        except (AttributeError, TypeError) as err:
+            raise TypeError(
+                "platform must be None, a platform name string, or an "
+                "openmm.Platform instance"
+            ) from err
```
sroet: Shouldn't this be an actual test against `isinstance`? As there are a couple other python objects that implement a `getName` which would absolutely wreck this ([`threading.thread.getName()`](https://docs.python.org/3/library/threading.html#threading.Thread.getName) comes to mind, even if it is Deprecated)


```suggestion
        if isinstance(platform, openmm.Platform):
            return platform.getName()
        
        raise TypeError(
            "platform must be None, a platform name string, or an "
            "openmm.Platform instance. "
            f"Got {type(platform)} instead"
        )
```

This whole function might be nice for a `match case` rewrite, depending on minimal supported  python version

sroet (changes requested):
One question about the normalize_platform function, LGTM otherwise

openpathsampling/engines/openmm/engine.py
dwhswenson: Good call on the match-case. I tended to avoid that (prefer dispatch dict over match-case for the most common ways people use cases in C/bash), but we are >=3.10, so this is allowed -- and in a situation like this, looks pretty good.

Implemented in [2e318c6](https://github.com/openpathsampling/openpathsampling/pull/1180/commits/2e318c64a559ea22c4d5f313af804ada83bc7089).

openpathsampling/tests/test_openmm_engine.py
```diff
@@ -296,3 +310,89 @@ def test_export_trajectory(self, tmp_path):
...
+                return 'CPU'
+
+        integrator = mm.LangevinIntegrator(
+            300*u.kelvin,
+            old_div(1.0, u.picoseconds),
+            2.0*u.femtoseconds
+        )
+        integrator.setConstraintTolerance(0.00001)
+
+        with pytest.raises(TypeError, match='platform must be None'):
```
sroet, line 354: Maybe nice to also make sure the bad class is mentioned in the error 


```suggestion
        with pytest.raises(TypeError, match=r'platform must be None.*PlatformLikeObject'):
```

sroet (approved):
Thanks for the rewrite and updated test. Small nitpick on the test, but not blocking. LGTM

PR #1181, opened 2026-03-22
Fix SimStore problems with scalarize singletons in numpy 2

dwhswenson:
This is similar to one of the things we fixed in #1178 for array scalar conversion -- not sure why the tests for this one only started failing recently, but this will solve it.


sroet (approved):
Small change and thoroughly tested. LGTM!

PR #1183, opened 2026-05-31
Fix OpenMM RC test workflow

dwhswenson:
Looks like this is failing for us due to some issues with an action no longer providing pip. This is intended to bring the OpenMM RC workflow more into line with our main testing workflow.

Assisted-by: Codex.app:GPT-5.4

### DIFF
--- a/.github/workflows/check-openmm-rc.yml
+++ b/.github/workflows/check-openmm-rc.yml
@@ -7,8 +7,8 @@ on:
   schedule:
     - cron: "0 9 * * *"
   # use this for debugging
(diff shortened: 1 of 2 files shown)


dwhswenson:
This is ready for review and comment. Without review, I will merge this after at least 24 hours, no sooner than Mon 01 Jun 18:00 GMT (13:00 my local).

Note that commit c22b84a temporarily added running on PR to the OpenMM RC check workflows, and that they ran and passed on that commit. That should mean that this PR works for what was intended.

